PR #2, opened 2016-09-14
Add jshint linter

fijshion:


ale_linters/javascript/jshint.vim
```diff
@@ -0,0 +1,50 @@
...
+        \   'nr': -1,
+        \})
+    endfor
+
+    return output
+endfunction
+
+call ALEAddLinter('javascript', {
+\   'executable': 'jshint',
+\   'command': 'jshint --reporter unix --config ' . g:ale_jshint_config_loc . ' -',
```
w0rp: You should use a default for the config location or make the config option in some way. I now offer a `command_callback` option which lets you generate the command string which will be run in a job. The function takes a buffer number as its only argument. You can look at the shell linter I just added around an hour ago for an example of its use.

Thank you for adding more tools! You contribution is much appreciated. I need to add an option for selecting which linters you want, which is part of why I now require a `'name'` option in the dictionary here for identifying which linter it is for which filetype. I'll add that once my RSI clears up. I've been having some issues recently.

w0rp:
One thing I'm wondering about is that I might make all of the linters change directory through some means to the directory of the file before running, to make sure they search up parent directories to find a configuration file. A good default for JSHint might be to not specify any configuration files at all in the command, and let JSHint look up parent directories to find one. That's what I used to do with JSHint before I switched to eslint. I think it works in Vim for me at the moment because I separately change directories automatically when I can.

ale_linters/javascript/jshint.vim
fijshion: Sorry - I forgot to include plugin/ale/aaflags.vim in this commit. Force pushed an update. Let me know if you like it.

Thank you for writing this! I've been using it all day.. it's very useful and speedy.

plugin/ale/aaflags.vim
```diff
@@ -40,3 +40,8 @@ endif
 if !exists('g:ale_echo_cursor')
     let g:ale_echo_cursor = 1
 endif
+
+" Set this to the location of the jshint configuration file
+if !exists('g:ale_jshint_config_loc')
+    let g:ale_jshint_config_loc = '.jshintrc'
+endif
```
w0rp: Oh sorry, I only just saw this now. This conflicts with some recent changes I did. Check my comments about reading from parent directories as  default. I would maybe move this flag into the linter file itself, and keep only the options which apply to all linters in this file.

fijshion:
Looking up a config recursively sounds cool. JSHint is a little annoying because it doesn't look for a default configuration file when in stdin mode. That's why I added the flag. I'll clean it up more tonight.

ale_linters/javascript/jshint.vim
w0rp: No problem. I'll keep adding more to it especially after I am able to recover from some recent RSI issues I have been experiencing.

w0rp:
Ah, I see. That sounds unfortunate. `eslint` does do that with the stdin option. I wonder if it might be related to a need to switch directories, if it's just always based off of the filename you pass in. `eslint` also has a `filename` option for a hint. Maybe there is something there.

I might look into always somehow changing directories when running jobs, maybe something like `cd `dirname %` && ...` for commands or similar.

w0rp:
You could also write the "search up parent directories" part yourself, as you can now write callbacks for the command to run which will let you do that, if you look at what `ALEAddLinter` will now accept on `master`. I have been thinking I'll probably need a utility function at some point for doing that. Maybe something like `ale#util#FindConfig(buffer, filename_callback)`.

fijshion:
I force pushed the change to move the jshint_config_loc flag to the jshint.vim file.

w0rp:
I'll take this as it is for now. If you or others want to improve it, they can go ahead and do that. I'll write that function for finding files in parent directories at some point soon.

PR #3, opened 2016-09-15
Add rubocop linter

ynonp:
For ruby projects add support for rubocop linter using "ruby" file type

### DIFF
--- a/ale_linters/ruby/rubocop.vim
+++ b/ale_linters/ruby/rubocop.vim
@@ -7,9 +7,8 @@ let g:loaded_ale_linters_ruby_rubocop = 1
 function! ale_linters#ruby#rubocop#Handle(buffer, lines)
     " Matches patterns line the following:
     "
-    " <text>:47:14: Missing trailing comma. [Warning/comma-dangle]
-    " <text>:56:41: Missing semicolon. [Error/semi]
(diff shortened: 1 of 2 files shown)


ale_linters/ruby/rubocop.vim
```diff
@@ -0,0 +1,45 @@
...
+        \   'type': type ==# 'C' ? 'E' : 'W',
+        \   'nr': -1,
+        \})
+    endfor
+
+    return output
+endfunction
+
+call ALEAddLinter('ruby', {
+\   'executable': 'rubocop',
```
w0rp: With changes I have made, I'll need to add a `name` key here now, which is now required. I'll merge this and add it in. Could you test this again after I make that change?

ale_linters/ruby/rubocop.vim
```diff
@@ -0,0 +1,45 @@
+if exists('g:loaded_ale_linters_ruby_rubocop')
```
w0rp: I'll add Author comments to the top of this file and others to credit you. Let me know if there is a preferred name you would like in the files via email or just comments on GitHub.

ynonp: Cool! just use my github username ynonp

PR #13, opened 2016-09-26
Option to keep sign column always open

vmarovic:
Add option to keep sign column open event when there is no signs to show.

I found this works better for me since It stops sign columns from constantly pushing text left and right.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Each directory within corresponds to a particular filetype in Vim, and each file
 in each directory corresponds to the name of a particular linter.
 
 ### Always showing gutter
-To prevent left gutter to disappear when there is no errors set
+You can keep the sign gutter open at all times by setting the g:ale_sign_column_always to 1
 
 ```vim
 let g:ale_sign_column_always = 1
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -82,6 +82,14 @@ This plugin will look for linters in the [`ale_linters`](ale_linters) directory.
 Each directory within corresponds to a particular filetype in Vim, and each file
 in each directory corresponds to the name of a particular linter.
 
+### Always showing gutter
+To prevent left gutter to disappear when there is no errors set
```
w0rp: The grammar isn't the best here. You could write something like, "You can keep the sign gutter open at all times by setting the `g:ale_sign_column_always` to `1`."

Maybe we should add an "Options" section which explains all of the options, even though there's a link to the .vim file.

vmarovic: Fixed readme. Options section would probably be useful. On the note of keeping track if all errors are gone perhaps we should look into adding some info into status line. Could be nice to have it as option.

w0rp:
Hmm, I'll think about it.

PR #15, opened 2016-09-29
Fix typo ALErrorSign -> ALEErrorSign

CookiesBestDinner:
Error sign was showing up with default colours


w0rp:
Nice spot! Thanks!

PR #26, opened 2016-10-03
Add support for javascript.jsx filetype

gavocanov:
As first thnx a lot for the plugin, works really well!!

When working with [React](https://facebook.github.io/react/) it almost a standard to use the [vim-jsx](https://github.com/mxw/vim-jsx) plugin which adds a `javascript.jsx` filetype for `javascript`.

This was not handled by `ale` by default, with this PR it works.


w0rp:
Hmm, I'll take this. If performing mostly the same setup twice for the JS linters becomes an issue in future, I may do something like automatically define any JS linter also for JSX.

PR #29, opened 2016-10-03
Add csslint for css

KabbAmine:
Using the stdin wrapper script


w0rp:
Aha! I didn't see this pop up. I was just working on this right this minute myself. Have a look at my new commits, and let me know what you think. Tell me if you find any issues.

KabbAmine:
No issues so far, LGTM

I was working on sass-lint implementation (for sass & scss) that use the same error format as csslint, so if you're not on it I'll push a PR.

w0rp:
You could take the function that's there now, move it into the file with
generic handlers, and apply it to those too. That should be nice and easy.
Stick your name at the top of the files if you add them. :+1:

PR #30, opened 2016-10-03
Add sass/scss support with sass-lint

KabbAmine:


w0rp:
Looks pretty good. I'll move that handler into one file all three can use, and change the letter case a bit. I'd write "SCSS" and "SASS" in the README file instead of "Scss" and "Sass", and I'd write the linter names as `sasslint`, etc.

w0rp:
I just merged this now via the command line after fixing the conflicts and what I mentioned above. Thanks!

PR #32, opened 2016-10-03
Use gcc format for vint output

KabbAmine:
Vint allows to use a custom output format so I made it gcc compatible.

The advantages:

- Less code (We are lazy, everybody knows that)
- The warnings and errors are differentiated now

Also, even if the errors were reported correctly the callback function was wrong (`ale_linters#haskell#ghc#Handle` instead of `ale_linters#vim#vint#Handle`)

### DIFF
--- a/ale_linters/vim/vint.vim
+++ b/ale_linters/vim/vint.vim
@@ -1,4 +1,5 @@
 " Author: w0rp <devw0rp@gmail.com>
+" Modified by: KabbAmine <amine.kabb@gmail.com>
 " Description: This file adds support for checking Vim code with Vint.
 
 if exists('g:loaded_ale_linters_vim_vint')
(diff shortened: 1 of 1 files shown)


ale_linters/vim/vint.vim
```diff
@@ -7,40 +8,11 @@ endif
...
-        \   'col': l:match[2] + 0,
-        \   'text': text,
-        \   'type': 'W',
-        \   'nr': -1,
-        \})
-    endfor
-
-    return output
-endfunction
+let s:format = '-f "{file_path}:{line_number}:{column_number}: {severity}: {description} (see {reference})'
```
KabbAmine: You'll notice that I didn't close the double quote. Looks like `stdin_wrapper` or something else adds an extra double quote.
Try by yourself and you'll understand.

w0rp: Hmm, that sounds like a bug. I'll look into that.

ale_linters/vim/vint.vim
```diff
@@ -1,4 +1,5 @@
 " Author: w0rp <devw0rp@gmail.com>
+" Modified by: KabbAmine <amine.kabb@gmail.com>
```
w0rp: I figure I'll just add a comma in the one above and put your name there.

KabbAmine: Sure

w0rp:
Just so you know, my only hold up on this is needing to look into that double quote bug. I haven't managed to get the time to repeat it and fix it yet.

KabbAmine:
No problem, take your time.

I will also look into it if I have time.

w0rp:
I have merged this now via the commandline after adjusting the Author line and the error handler a little. I don't know what's going on with the double quote there, but it seems to work, so okay then.

PR #36, opened 2016-10-03
Add linter yamllint for yaml

KabbAmine:
A yaml file can have `yaml` or `yml` extension, so we should either rethink how the stdin wrapper script proceed (Maybe pass 2 arguments to the script?) or find a better way in `ALEAddLinter()`.


w0rp:
The file extension might not matter for most linters. It will matter for some.

KabbAmine:
> The file extension might not matter for most linters. It will matter for some.

I was talking about how the extension is hard coded in the command (`g:ale#util#stdin_wrapper . ' .yml yamllint -f parsable'`), but in this example yamllint can make the difference and lint both `.yml` and `.yaml`, so no problem here.

PR #41, opened 2016-10-04
Add tidy support for html

KabbAmine:
As the title said.

I've added 2 options:

- `g:ale_html_tidy_executable` to define the executable to use (In my config for example I'm using `tidy5`)
- `g:ale_html_tidy_args` for the CLI flags.

Generally speaking, it would be great to make those kind of options in all the linters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ name. That seems to be the fairest way to arrange this table.
 | D | [dmd](https://dlang.org/dmd-linux.html)^ |
 | Fortran | [gcc](https://gcc.gnu.org/) |
 | Haskell | [ghc](https://www.haskell.org/ghc/)^ |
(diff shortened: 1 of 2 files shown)


w0rp:
Yeah, more of the linters could let you configure the executable paths. Some of them do already, I think.

I just documented the flags available for everything in the README, and I think the flags specific to each linter should eventually go in there too.

w0rp:
Looks good to me. Cheers! :+1: :beers:

KabbAmine:
> I just documented the flags available for everything in the README, and I think the flags specific to each linter should eventually go in there too.

For my plugins, I document everything in the README then copy/convert it to the `doc/plugin.txt` file.

w0rp:
Writing the `:help` files is on the list of things to do. It'll happen eventually.

PR #51, opened 2016-10-06
correctly parse eslint warning type

jwilliams108:
The eslint linter is not properly identifying warnings as the eslint output isn't correctly parsed for the type. The marker string needs to be split first.


w0rp:
Yep, looks good. Thanks!

PR #53, opened 2016-10-06
GHC Linter for Haskell Stack

dahlia:
[According to the official website of Haskell][1], [Stack][2] is one of widely used ways to install Haskell.  It is also a modern frontend to Cabal, the package manager for Haskell.

Because Stack manages GHC installations by itself, it requires to use `stack ghc` command instead of vanilla `ghc` command.  So I added a `stach-ghc` linter for Haskell.

[1]: https://www.haskell.org/downloads
[2]: https://haskellstack.org/


w0rp:
I'll take this. We might have to make a note about how to turn off both linter types in future. They are kind of the same type of linter, so I'm not sure this needs to go into another file.

PR #55, opened 2016-10-06
Add support for env in bash scripts

kalmiz:
Bash scripts are often using the `env` command instead of an actual command. This PR adds support for it but keeps the original functionality.

### DIFF
--- a/ale_linters/sh/shell.vim
+++ b/ale_linters/sh/shell.vim
@@ -30,25 +30,25 @@ function! ale_linters#sh#shell#GetExecutable(buffer)
     if len(banglines) == 1
         if strpart(banglines[0], 0, 2) == '#!'
             let parts = filter(split(strpart(banglines[0], 2), ' '), "v:val != ''")
-			if len(parts) > 0 
-				if strpart(parts[0], len(parts[0]) - 3) == 'env'
-				   let parts = parts[1:]
-				endif
-				if len(parts) > 0
-					let tmp = ale_linters#sh#shell#ParseShebang(parts)
-				else
-					let tmp = ''
-				endif
-				if strpart(tmp, 0, 1) != '/' && tmp != ''
-					let res = systemlist('which ' . tmp)
-					if len(res) > 0
-						let tmp = res[0]
-					endif
-				endif
-				if tmp != ''
-					let shell = tmp
-				endif
-			endif
+            if len(parts) > 0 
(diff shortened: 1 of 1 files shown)


ale_linters/sh/shell.vim
```diff
@@ -13,17 +13,42 @@ if !exists('g:ale_linters_sh_shell_default_shell')
...
+                if strpart(parts[0], len(parts[0]) - 3) == 'env'
+                   let parts = parts[1:]
+                endif
+                if len(parts) > 0
+                    let tmp = ale_linters#sh#shell#ParseShebang(parts)
+                else
+                    let tmp = ''
+                endif
+                if strpart(tmp, 0, 1) != '/' && tmp != ''
+                    let res = systemlist('which ' . tmp)
```
w0rp: Do we need to run a system call to evaluate shebang lines?

Are there many shells which won't be installed to /bin or /usr/bin? I suppose maybe `zsh` on Mac OSX might end up in some strange place.

kalmiz: The issue here is that `env` command looking for the executable in your `$PATH`. For example, if you have the `~/bin` directory in your `$PATH` and you have a `bash` binary in it then `env bash` will use that. In that case, if you hardcode the path then vim will pick up a different binary for lint then the bash for run. I understand your concern and however the `system` won't run if your are using a full path without the `env`, we could add a cache for it: store the shebang line and only run the parser if the cache is empty - first time - or the shebang line is not equal with the cache.

w0rp: Would regex not be easier then a recursive function call for looking for `env`? I find the `strpart` calls hard to follow.

kalmiz: I'm using recursive function because according to the `env` manual it can have options:
`env [-iv] [-P altpath] [-S string] [-u name] [name=value ...] [utility [argument ...]]`
My recursive function is try to find the first string which doesn't start with a `-`.
(I just realized that it will fail if it contains custom environment variables  - the `[name=value ...]` part. I will fix that)
In my opinion `strpart` in the recursive function is easy to follow but if you are talking about the other `strpart` calls then I can try to rewrite them using regex but I think that would be more confusing. Sorry, I forgot to comment that recursive function, will do.

w0rp:
Check out issue #69 and the branch `shell-detection`. Does that address your needs? It will never be a _perfect_ implementation without a system call, but it can be _good enough_.

PR #56, opened 2016-10-06
Add basic scala support using scalac

kalmiz:


ale_linters/scala/scalac.vim
```diff
@@ -0,0 +1,61 @@
...
+            continue
+        endif
+
+        let text = l:match[3]
+        let type = 'W'
+        if l:match[2] == 'error'
+            let type = 'E'
+        endif
+        let col = 0
+        if exists('a:lines[ln+1]')
```
w0rp: Can you not check `if ln + 1 < len(a:lines)` instead? It looks like you're doing a bounds check. If the string is empty for some reason, then `stridx` below will just return `-1` anyway.

ale_linters/scala/scalac.vim
```diff
@@ -0,0 +1,61 @@
...
+
+        if len(l:match) == 0
+            continue
+        endif
+
+        let text = l:match[3]
+        let type = 'W'
+        if l:match[2] == 'error'
+            let type = 'E'
+        endif
```
w0rp: You could have used a ternary expression for this, but this will do.

ale_linters/scala/scalac.vim
```diff
@@ -0,0 +1,61 @@
...
+        let col = 0
+        if exists('a:lines[ln+1]')
+            let col = stridx(a:lines[ln+1], '^')
```
w0rp: You should add spaces around the binary operator `+` here.

ale_linters/scala/scalac.vim
kalmiz: I converted it to a ternary expression.

ale_linters/scala/scalac.vim
kalmiz: You are right it's a bounds check. I rewrote it.

ale_linters/scala/scalac.vim
kalmiz: I added the spaces.

kalmiz:
Thanks for your comments! I have implemented your suggestions and added the documentation too.

PR #57, opened 2016-10-07
updating config option format and adding to doc

jwilliams108:
I noticed you added some docs around linter specific options, so I added a section for phpcs. I also updated the option variable name to better match similar options in other linters.


w0rp:
Looks good to me.

PR #58, opened 2016-10-07
Add an initial getStatuslineStr function with customizable output

KabbAmine:
You can now get a formatted string to use with the statusline with `ale#sign#GetStatuslineStr()`
The default output looks like that:

```vim
2 error(s) 1 warning(s)
```

e_a_w

Note that the string will not print errors or warnings if they're not present.

```vim
" Only errors
1 error(s)
" Or warnings
3 warnings
```

e
w

Also, return an empty string if no issues at all.

-----------------------

You can specify the format to use using a list in `g:ale_statusline_format` where:

- Both elements are strings
- The 1st one is for errors and the 2nd for warnings

The default value is `['%d error(s)', '%d warning(s)']`

For example with:

```vim
let g:ale_statusline_format = ['⨉ %d', '⬥ %d']
```

ale

-----------------------

If you like my implementation, I'll update the documentation and force push. 
Also, I'm not sure if the `sign.vim` file is the proper place for this function.

*__N.B:__ The red color is not a part of the PR, its proper to my config*

**UPDATE** Correct some typos


plugin/ale/sign.vim
```diff
@@ -40,6 +40,34 @@ execute 'sign define ALEWarningSign text=' . g:ale_sign_warning
 \   . ' texthl=ALEWarningSign'
 sign define ALEDummySign
 
+function! ale#sign#GetStatuslineStr()
```
w0rp: Could you move this into a `statusline.vim` file? I think this belongs in its own file. The `Str` in the name is redundant. I wonder if the "public" functions for ALE should all be named in PascalCase without the paths. Something like `ALEGetStatusLine`.

plugin/ale/sign.vim
```diff
@@ -40,6 +40,34 @@ execute 'sign define ALEWarningSign text=' . g:ale_sign_warning
...
+	redir END
+
+	let [errors, warnings] = [0, 0]
+	for s in split(signs, "\n")
+		if s =~# 'ALEErrorSign$'
+			let errors += 1
+		elseif s =~# 'ALEWarningSign$'
+			let warnings += 1
+		endif
+	endfor
```
w0rp: Instead of parsing the output of `sign place`, you should use the `g:ale_buffer_loclist_map` variable. It maps the buffer to the `loclist` items set up by ALE. That's less likely to lead to bugs later on, and should be easier to work with. The List is all in the standard loclist format used by Vim.

plugin/ale/sign.vim
```diff
@@ -40,6 +40,34 @@ execute 'sign define ALEWarningSign text=' . g:ale_sign_warning
...
 
+function! ale#sign#GetStatuslineStr()
+	" Returns a formatted string that can be integrated in the 
+	" statusline
+
+	redir => signs
+	silent exec 'sign place buffer=' . bufnr('%')
+	redir END
+
+	let [errors, warnings] = [0, 0]
```
w0rp: I say just put each variable declaration on its own line. They don't need to be combined into one line like this.

plugin/ale/sign.vim
```diff
@@ -40,6 +40,34 @@ execute 'sign define ALEWarningSign text=' . g:ale_sign_warning
...
+			let errors += 1
+		elseif s =~# 'ALEWarningSign$'
+			let warnings += 1
+		endif
+	endfor
+
+	let errors = errors ?
+				\	printf(g:ale_statusline_format[0], errors) : ''
+	let warnings = warnings ?
+				\	printf(g:ale_statusline_format[1], warnings) : ''
```
w0rp: If these expressions are going off to the right margin anyway, you might as write only one line each here.

plugin/ale/sign.vim
```diff
@@ -40,6 +40,34 @@ execute 'sign define ALEWarningSign text=' . g:ale_sign_warning
...
+
+	" Different formats if no errors or no warnings
+	return !empty(errors) && !empty(warnings) ?
+				\	printf('%s %s', errors, warnings) :
+				\	(empty(errors) ? warnings : errors)
```
w0rp: I think if statements would be better here. If you are going to continue A line, I'd prefer a hanging indent, like so.

```vim
let x = 2
\   + 3
```

Then the following lines don't get shuffled around when the first line changes in length.

plugin/ale/aaflags.vim
```diff
@@ -39,3 +39,11 @@ let g:ale_warn_about_trailing_whitespace =
...
 " This flag can be set to 1 to keep sign gutter always open
 let g:ale_sign_column_always = get(g:, 'ale_sign_column_always', 0)
+
+" String format for statusline
+" Its a list where:
+" * The 1st element is for errors
+" * The 2nd element is for warnings
+let g:ale_statusline_format = get(
+			\	g:, 'ale_statusline_format', ['%d error(s)', '%d warning(s)']
+			\ )
```
w0rp: I'd use a hanging indent here, as mentioned below. You could probably leave the first two arguments to `get` on the first line, and then up-indent the value on the second line, leaving the closing parenthesis on its own line. Like so:

```vim
let g:ale_statusline_format = get(g:, 'ale_statusline_format', 
\   ['%d error(s)', '%d warning(s)']
\)
```

plugin/ale/sign.vim
KabbAmine: I agree for `ALEGetStatusLine`, it should follow `ALELint`, `ALEGetLinter`...
I've used the `plugin#file#fun` notation to follow your conventions, in fact it is useful only when the function exists in an autoload file, otherwise there is need to use it.

A specific file for the statusline is what I was thinking, so yes I'll do it.

plugin/ale/sign.vim
KabbAmine: Ok

plugin/ale/sign.vim
KabbAmine: Its just a quicker way to write it, but as you like, I'll change it.

plugin/ale/sign.vim
KabbAmine: Ok, they were longer the 1st time, and I forgot to join them when I made changes.

plugin/ale/sign.vim
KabbAmine: Ok

plugin/ale/aaflags.vim
KabbAmine: Ok

KabbAmine:
Changes made and force pushed.

------------

**BONUS**

In my config and for all the syntax checkers that I've used, I always configure the status line to show a special character for when there are no errors.

Check the following:

ale

If you want, I can add an option for that in `g:ale_statusline_format`.

e.g.
```vim
let g:ale_statusline_format = ['%d error(s)', '%d warning(s)', 'ok']
```

For example there is my own custom statusline function:
```vim
function! SLAle(mode) abort " {{{1
	if !exists('*ALEGetStatusLine')
		return ''
	endif
	if empty(ALEGetLinters(&ft))
		return ''
	endif
	let l:str = ALEGetStatusLine()
	return a:mode ?
				\	(!empty(l:str) ? l:str : ''):
				\	(!empty(l:str) ? '' : '')
endfunction
```

FredrikAugust:
Is there anyway you could get this to work with vim-airline?

KabbAmine:
> Is there anyway you could get this to work with vim-airline?

@FredrikAugust I've quickly looked on the API and you can do something like that:

```vim
let g:airline_section_error = '%{ALEGetStatusLine()}'

" Or
call airline#parts#define_function('ALE', 'ALEGetStatusLine')
let g:airline_section_error = airline#section#create_right(['ALE'])
```

Note that from what I understand, this replace the whole airline `error` section, so it may be a better way.

When this PR will be merged, I'll add (maybe in the README or in gist) a way to make it work with airline (I've stopped using those kind of plugins long time ago).

FredrikAugust:
Sounds great, I'm also tempted to move away from them, but atm I'm too lazy to migrate the setup to default status bar. Some documentation on that would be great!

Speaking of migrating, do you think you could share the setup for your statusbar there? Looks really nice :tada:

KabbAmine:
@FredrikAugust All what you need is `:h statusline`
I've just pushed my ale configuration, check my [vim repo](https://github.com/kabbamine/myvimfiles), in `config/statusline.vim`

w0rp:
@KabbAmine I like settings for this. We should consider putting something like that in the README. Maybe a section about how to configure this and also the signs to make them prettier for systems which have decent Unicode support, etc.

w0rp:
This looks pretty great. Do you want to add in the "ok" option too? I like the idea.

Could you add information about the option and the feature to the help file, and maybe some information to the README?

KabbAmine:
Let me first add the *no errors* option (OK), then I'll update the documentation files and force push again.

For the extra settings, it should be in a commit a part, maybe in some kind of FAQ in the doc file (or README?)

KabbAmine:
@w0rp Please check if everything is ok (Maybe rewording or typos fixing).

doc/ale.txt
```diff
@@ -319,6 +329,12 @@ ALEGetLinters(filetype)                                       *ALEGetLinters()*
   Return all of linters configured for a given filetype as a |List| of
   |Dictionary| values in the format specified by |ALEAddLinter()|.
 
+ALEGetStatusLine()                                           *ALEGetStatusLine()*
+  Return a formatted string that can be added to the statusline.
+	The output's format is defined in |`g:ale_statusline_format`|.
+  To enable it, the following should be present in your |statusline| settings: >
+  %{ALEGetStatusLine()}
+
```
w0rp: Do you not need to end the code block here with `<`? I'm not sure myself.

KabbAmine: Not necessary, the code block is stopped when no more indentation (2 spaces or more)

README.md
```diff
@@ -123,6 +124,31 @@ let g:ale_sign_error = '>>'
...
+```vim
+%{ALEGetStatusLine()}
+```
+
+When errors are detected a string showing the number of errors will be shown.  
+You can customize the output format using the global list `g:ale_statusline_format` where:
+
+- The 1st element is for errors
+- The 2nd element is for warnings
+- The 3rd element is for when no erros are detected
```
w0rp: There is a typo I just noticed here, but I'll just add the 'r' on master.

doc/ale.txt
```diff
@@ -219,6 +219,16 @@ you can disable these warnings for some linters by setting this option to `0`.
...
 
+g:ale_statusline_format                               *g:ale_statusline_format*
+
+Type: |List|
+Default: `['%d error(s)', '%d warning(s)', 'OK']`
+
+This variable defines the format of |`ALEGetStatusLine()`| output.
+- The 1st element is for errors
+- The 2nd element is for warnings
+- The 3rd element is for when no erros are detected
```
w0rp: The typo was copy and pasted here.

PR #60, opened 2016-10-07
Add custom color support for sign column (different color without err…

theanimal666:
With this patch the sign bar changes it's background color when no errors were found to provide instant visual feedback that file a is linting even when sign column is configured to stay open all the time.

Please note: these are my first few lines of vim code, it seems to work but i have real clue if this won't break anything else i did not notice.

Possible improvement: 
Get current configured color of the SignColumn as default instead of hardcoded value

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,10 @@ vimrc file for all given linters is as follows:
 | `g:ale_sign_offset` | an offset for sign ids | `1000000` |
 | `g:ale_echo_cursor` | echo errors when the cursor is over them | `1` |
 | `g:ale_warn_about_trailing_whitespace` | enable trailing whitespace warnings for some linters | `1` |
+| `g:ale_sign_term_color` | color of the sign bar in term | `lightgrey` |
+| `g:ale_sign_term_color_no_error` | color of the sign bar in term when no errors are found (only visible with g:ale_sign_column_always=1) | `green` |
+| `g:ale_sign_gui_color` | color of the sign bar in GUI | `lightgrey` |
+| `g:ale_sign_gui_color_no_error` | color of the sign bar in GUI when no errors are found (only visible with g:ale_sign_column_always=1) | `green` |
 
 ### Selecting Particular Linters
 
--- a/plugin/ale/sign.vim
+++ b/plugin/ale/sign.vim
@@ -33,6 +33,12 @@ let g:ale_sign_warning = get(g:, 'ale_sign_warning', '--')
 " The dummy sign will use the ID exactly equal to the offset.
 let g:ale_sign_offset = get(g:, 'ale_sign_offset', 1000000)
 
+" sign bar color
+let g:ale_sign_term_color = get(g:, 'ale_sign_term_color', 'lightgrey')
+let g:ale_sign_term_color_no_error = get(g:, 'ale_sign_term_color_no_error', 'green')
+let g:ale_sign_gui_color = get(g:, 'ale_sign_gui_color', 'lightgrey')
+let g:ale_sign_gui_color_no_error = get(g:, 'ale_sign_gui_color_no_error', 'green')
+
 " Signs show up on the left for error markers.
 execute 'sign define ALEErrorSign text=' . g:ale_sign_error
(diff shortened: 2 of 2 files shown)


KabbAmine:
Even if its not documented, the sign colors can be configured via the following groups:

```
ALEError
ALEWarning
ALEErrorSign
ALEWarningSign
```

And you can link them or apply your own colors easily:

```vim
" To have the same color as Directory group
hi! link ALEError Directory

" Use your own colors
hi ALEError ctermfg=red ctermbg=blue......
```

theanimal666:
My patch is settings the background color of the SignColumn to the background of the signs themselves. 
Here are screenshots for what i did to make it clear, what i did.

screen shot 2016-10-07 at 15 57 44
screen shot 2016-10-07 at 15 57 56

KabbAmine:
My bad, I misunderstood you.

Anyway, IMHO its not a good idea, or at least not a thing that should be enabled by default.
Why? Because its:

- Quite distracting (and disturbing)
- May conflict with other plugins using signs (e.g gitgutter)
- Make other signs difficult to see

See by yourself (folds + gitgutter + ale)

s

Note that I'm not the plugin's author, so wait for @w0rp.

theanimal666:
oh, yes i see - not good at all ;)
thats what i meant with i have no idea what else this may break.

KabbAmine:
No need to close the PR. You may get some interesting points from other people than me.

theanimal666:
ok - reopened lets see what others think about

w0rp:
I think the highlight groups need to be added to the documentation. I planned to do that eventually.

plugin/ale/sign.vim
```diff
@@ -134,6 +147,7 @@ function! ale#sign#SetSigns(buffer, loclist)
     endfor
 
     if !g:ale_sign_column_always && len(signlist) > 0
+    " if len(signlist) > 0
```
w0rp: There's a bit of commented out code left here.

README.md
```diff
@@ -83,6 +83,10 @@ vimrc file for all given linters is as follows:
 | `g:ale_sign_offset` | an offset for sign ids | `1000000` |
 | `g:ale_echo_cursor` | echo errors when the cursor is over them | `1` |
 | `g:ale_warn_about_trailing_whitespace` | enable trailing whitespace warnings for some linters | `1` |
+| `g:ale_sign_term_color` | color of the sign bar in term | `lightgrey` |
+| `g:ale_sign_term_color_no_error` | color of the sign bar in term when no errors are found (only visible with g:ale_sign_column_always=1) | `green` |
+| `g:ale_sign_gui_color` | color of the sign bar in GUI | `lightgrey` |
+| `g:ale_sign_gui_color_no_error` | color of the sign bar in GUI when no errors are found (only visible with g:ale_sign_column_always=1) | `green` |
```
w0rp: Don't forget to add new options to the `doc` file too, and make sure the table doesn't get too large.

plugin/ale/sign.vim
```diff
@@ -120,6 +126,13 @@ function! ale#sign#SetSigns(buffer, loclist)
         exec 'sign unplace ' . current_id . ' buffer=' . a:buffer
     endfor
 
+	" set colot of SignColumn
+	if len(signlist) > 0
+		execute ':highlight SignColumn ctermbg=' . g:ale_sign_term_color . ' guibg=' .g:ale_sign_gui_color
+	else
+		execute ':highlight SignColumn ctermbg=' . g:ale_sign_gui_color_no_error . ' guibg=' . g:ale_sign_gui_color_no_error
+	endif
```
w0rp: The indentation looks a bit screwy here. You should `:set expandtab` for editing these files.

theanimal666:
As pointed out by KabbAmine it breaks things in combintation with other plugins, so not sure if it won't be better to just close this. 
With https://github.com/w0rp/ale/pull/58 merged the same thing can be done e.g. is a far better solution to have this in the statusbar.

w0rp:
I'll close this one. I suppose if you want some short information on whether or not there are any errors, the statusline function would be a better fit.

PR #64, opened 2016-10-08
Add cpp linter

Geam:
It's mainly an copy/paste from the c one with few modification for cpp.


doc/ale.txt
```diff
@@ -273,6 +275,16 @@ Default: `'-Wall'`
...
 
+-------------------------------------------------------------------------------
+4.4. cpp-gcc                                       *ale-linter-options-cpp-gcc*
+
+g:ale_cpp_gcc_options                                   *g:ale_cpp_gcc_options*
+
+Type: |String|
+Default: `'-Wall'`
+
+This variable can be change to modify flags given to gcc.
```
w0rp: There's a typo here. They should be "changed" not "change."

w0rp:
I guess those typos can be sorted out afterwards. :+1: to documenting the Fortran options. That's still undocumented. Both can be done in another PR.

Geam:
Ok for C++ instead of Cpp in the doc and README. I've put cpp cause `set: filetype` give me cpp and I couldn't use ++ in the variable name (vim doesn't like it) but yes, it's wrong in the doc. 

For the parentheses, you mean in the README ? If yes, I think we should add a linter specific section under Options cause, if there is many linter for a language it can get messy. I will try something and if it's good enough, I will submit it.

w0rp:
I was wondering if it could be written `C++ (filetype cpp)`, or maybe people will just know that anyway? Something like that.

w0rp:
Yep, that'll do it. Cheers! :beers:

Geam:
I've modify it to match the cython one

PR #65, opened 2016-10-08
Support verilator for verilog

mshr-h:
I've added [verilator](http://www.veripool.org/projects/verilator/wiki/Intro) for verilog.


w0rp:
I see no problems with this. Cheers! :beers:

PR #66, opened 2016-10-08
Add missing options to the doc and correct some typo

Geam:
This is not ready to merge, I'm still writing options for jshint, shellcheck and shell but I need somewhere to ask questions so, I made the PR (don't know if that the good way to do it, first time contributing to a project).

My concerns are : 
- won't the doc be too fussy especially the table of content ? (First time writing a doc for vim)
- option entry naming convention in the table of content : 
The first two options were for eslint and phpcs so the naming was pretty strait forward.
I've write the c, c++ and fortran one and I couldn't write just gcc as it would have been confusing. So I prepend with the language name and a dash.
Should I keep this convention for those were the linter is not use for more than one language ? Or just put the checker name ?

Cause I'm not great in english (trying my best) I will correct all the typo in the last commit.

Oh, would you prefere if I make a PR for each option or as I've start, all the correction in one PR ?


w0rp:
The disambiguation of the linter by placing 'language-` before it looks fine to me. There probably won't be that many linters which apply to more than one language. GCC is kind of a one-off.

doc/ale.txt
```diff
@@ -286,6 +288,45 @@ Default: `'-Wall'`
...
+This variable can be changed to change the path to tidy.
+
+
+g:ale_html_tidy_args                                     *g:ale_html_tidy_args*
+
+Type: |String|
+Default: '-q -e -language en'
+
+This variable can be changed to change the arguments provide to the
+executable.
```
w0rp: This sentence should read "provided", as it's past tense.

doc/ale.txt
```diff
@@ -286,6 +288,45 @@ Default: `'-Wall'`
...
+
+
+File encoding
+The linter will auto-detect the file encoding and provide the good option to
+the executable. If it can't detect the file encoding, it will fall back to
+utf8.
```
w0rp: I'd remove line 321 here with the words "File encoding," as there isn't any sub-subsection markup. The following sentences don't make the best use of English, but the ideas are there. I have an alternative suggestion.

"ALE will attempt to automatically detect the appropriate file encoding to provide to html-tidy, and fall back to UTF-8 when encoding detection fails."

doc/ale.txt
```diff
@@ -286,6 +288,45 @@ Default: `'-Wall'`
...
+
+The recognize file encoding are the following :
```
w0rp: You can change this to "The recognized file encodings are as follows:"

w0rp:
I have merged this with some amendments now.

PR #67, opened 2016-10-08
Add linter HTMLHint for html files

KabbAmine:
Better than tidy IMO :smile:


w0rp:

PR #73, opened 2016-10-09
Modify shell's pattern to handle locals other than english, fixes #69

### LINKED ISSUE #69: How to specify bash for sh files
I want to use only `bash` to lint sh files (I still have `shellcheck` installed and don't want to use for now), so which one of the following should I use?

```vim
" It should be after the linter name, which is 'shell'
let g:ale_linters = {'sh': ['shell']}

let g:ale_linters = {'sh': ['bash']}
```

Note that none of the above works.

KabbAmine:
As the title said


w0rp:
Yep, makes sense. Cheers! :beers:

PR #77, opened 2016-10-10
Minor fixes:

KabbAmine:
- Ensure that php linter pattern does not include spaces:
    PHP can return errors with extra spaces like the following:
    `PHP Parse error:  syntax error, unexpected end of file in t.php on line 4`

- Set option locally to buffer

- Rename `noErrors` variable according to the project's naming convention

- Make the jsonlint pattern a little better

### DIFF
--- a/plugin/ale/cursor.vim
+++ b/plugin/ale/cursor.vim
@@ -68,14 +68,14 @@ function! ale#cursor#TruncatedEcho(message)
 
     " We need to turn T for truncated messages on for shortmess,
     " and then then we need to reset the option back to what it was.
-    let shortmess_options = &shortmess
+    let shortmess_options = getbufvar('%', '&shortmess')
 
     try
         " Echo the message truncated to fit without creating a prompt.
-        set shortmess+=T
+        setlocal shortmess+=T
         exec "norm :echomsg message\n"
     finally
-        let &shortmess = shortmess_options
+        call setbufvar('%', '&shortmess', shortmess_options)
     endtry
 endfunction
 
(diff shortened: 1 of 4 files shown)


plugin/ale/cursor.vim
```diff
@@ -72,7 +72,7 @@ function! ale#cursor#TruncatedEcho(message)
 
     try
         " Echo the message truncated to fit without creating a prompt.
-        set shortmess+=T
+        setl shortmess+=T
```
w0rp: `setlocal` would be better here. do you need to also change the `let` below, or the value of the variable above?

You should probably write the full name of `setlocal` here. It might be easy to miss the `l`. Though we do use `exec` a lot instead of `execute`, etc.

w0rp: I wonder if we can get a `Vint` configuration setting to check for that, or something.

KabbAmine: Ok after re-checking, the `ProhibitAbbreviationOption` is only for vim options, so no, there is no way to check this kind of error with `vint`.

w0rp: I'll add coding standards to `CONTRIBUTING.md` later.

plugin/ale/cursor.vim
```diff
@@ -72,7 +72,7 @@ function! ale#cursor#TruncatedEcho(message)
...
         exec "norm :echomsg message\n"
```
KabbAmine: @w0rp Why did you use exec and norm here?

KabbAmine: And `echomsg` instead of a simple `echo`? It will pollute the messages history.

KabbAmine:
Should be ok now.

w0rp:
Cheers! :beers:

PR #80, opened 2016-10-10
First pass at optimizing ale to autoload

### LINKED ISSUE #59: Refactoring
Actually, the plugin is getting bigger and bigger, so IMO we should make some refactoring for the next versions.

- Create a directory structure more *vimish*, e.g:
```sh
├── ale_linters/
├── lib(scripts?)/
│   ├── stdin-wrapper
│   └── dmd-wrapper
├── doc/
│   ├── ale.txt
├── plugin/
│   └── ale.vim (content of zmain and maybe aaflags.vim)
├── autoload/
│   ├── ale.vim
│   └── ale/
│       ├── aaflags.vim
│       ├── cursor.vim
│       ├── handlers.vim
│       ├── sign.vim
│       ├── util.vim
├── example.gif
├── LICENSE
├── CONTRIBUTING.md
└── README.md
```
- Use local variables & functions when we can.
- Document everything.
- Test cases, using `Vader` or the vim 8 testing framework (#7)
- Use an OOP approach (Not mandatory but make the code more readable)
- Follow the [google Vimscript Style Guide](https://google.github.io/styleguide/vimscriptguide.xml) (Vint is following this guide).

What do you think @w0rp?

neersighted:
First off, the structure/function names should be revised a bit,
but I will wait for @w0rp's input before unifying the naming style.
Second off, the docs probably need some more work, I just did some
simple find-and-replace work.

With that said, this pull brings major performance gains for ale. On my
slowest system, fully loading ale and all its code takes around 150ms.

I have moved all of ale's autoload-able code to autoload/, and in
addition, implemented lazy-loading of linters. This brings load time on
that same system down to 5ms.

The only downside of lazy loading is that `g:ale_linters` cannot be
changed at runtime; however, it also speeds up performance at runtime by
simplfying the logic greatly.

Please let me know what you think!

Closes #59


neersighted:
I would like to unify function naming and simplify the `engine.vim` code if possible before merging -- some advice on how to proceed here would be great!

autoload/ale/linter.vim
```diff
@@ -0,0 +1,52 @@
...
+    endif
+
+    if has_key(g:ale_linters, a:filetype)
+        " Filter loaded linters according to list of linters specified in option.
+        for linter in g:ale_linters[a:filetype]
+            execute 'runtime! ale_linters/'.a:filetype.'/'.linter.'.vim'
+        endfor
+    else
+        execute 'runtime! ale_linters/'.a:filetype.'/*.vim'
+    endif
```
w0rp: I like this trick. I think you should put spaces around every binary operator, though, including `.`.

autoload/ale/linter.vim
```diff
@@ -0,0 +1,52 @@
...
+
+    return [] " Yes, we miss the first run, but it avoids some code duplication.
```
w0rp: Will this prevent linters from being run when the files are opened if we were to remove the `BufEnter` event from the `autocmd` line? Could you not check `s:linters` again before returning here and then return that?

What happens if you are editing a filetype which has no linters defined for it? Will it keep running the `runtime` command again and again, trying to find some linters? Maybe you should set `[]` for the filetype when the runtime command fails to find any linters for a given filetype.

autoload/ale/engine.vim
```diff
@@ -0,0 +1,218 @@
...
+        \   'output': [],
+        \}
+
+        if has('nvim')
+            " In NeoVim, we have to send the buffer lines ourselves.
+            let input = join(getbufline(a:buffer, 1, '$'), "\n") . "\n"
+
+            call jobsend(job, input)
+            call jobclose(job, 'stdin')
+        elseif has('win32')
```
w0rp: You'll have to rebase again and incorporate the changes on master, although they may be temporary. There is an open issue with sending buffer data to jobs in MacVim GUI, but I believe the issue might be down to old versions of MacVim, so I might apply this only for the Windows version of Vim here in the end anyway.

autoload/ale/engine.vim
```diff
@@ -0,0 +1,218 @@
...
+        elseif a:linter.output_stream ==# 'both'
+            let job = jobstart(command, {
+            \   'on_stdout': 'ale#engine#GatherOutputNeoVim',
+            \   'on_stderr': 'ale#engine#GatherOutputNeoVim',
+            \   'on_exit': 'ale#engine#HandleExitNeoVim',
+            \})
+        else
+            let job = jobstart(command, {
+            \   'on_stdout': 'ale#engine#GatherOutputNeoVim',
+            \   'on_exit': 'ale#engine#HandleExitNeoVim',
```
w0rp: Could these not still be in script functions, so they cannot be accessed from the outside? If so, that would provide better encapsulation.

autoload/ale/engine.vim
```diff
@@ -0,0 +1,218 @@
+" Stores information for each job including:
+"
+" linter: The linter dictionary for the job.
+" buffer: The buffer number for the job.
+" output: The array of lines for the output of the job.
+let s:job_info_map = {}
+
+function! ale#engine#GetJobID(job) abort
```
w0rp: Most of these functions should probably be script functions with `s:`, as they aren't intended to be visible anywhere else, unless there's some kind of performance reason to name them otherwise.

autoload/ale.vim
```diff
@@ -0,0 +1,49 @@
...
+
+        if !executable(l:executable)
+            " The linter's program cannot be executed, so skip it.
+            continue
+        endif
+
+        call ale#engine#ApplyLinter(buffer, linter)
+    endfor
+endfunction
+
```
w0rp: There's an extra blank line here.

autoload/ale/util.vim
```diff
@@ -0,0 +1,81 @@
...
+
+    for item in a:loclist
+        if item.lnum == 0
+            " When errors appear at line 0, put them at line 1 instead.
+            let item.lnum = 1
+        elseif item.lnum > last_line_number
+            let item.lnum = last_line_number
+        endif
+    endfor
+endfunction
```
neersighted: fair enough

autoload/ale/sign.vim
```diff
@@ -98,7 +93,7 @@ function! ale#sign#CombineSigns(loclist)
 endfunction
 
 " This function will set the signs which show up on the left.
-function! ale#sign#SetSigns(buffer, loclist)
+function! ale#sign#SetSigns(buffer, loclist) abort
```
neersighted: vint threw up, sooo

doc/ale.txt
```diff
@@ -481,21 +500,21 @@ ALEAddLinter(filetype, linter)                                 *ALEAddLinter()*
...
-  |Dictionary| values in the format specified by |ALEAddLinter()|.
+  |Dictionary| values in the format specified by |ale#linter#define()|.
 
 
-ALEGetStatusLine()                                         *ALEGetStatusLine()*
+ale#statusline#status()                               *ale#statusline#status()*
   Return a formatted string that can be added to the statusline.
   The output's format is defined in |`g:ale_statusline_format`|.
   To enable it, the following should be present in your |statusline| settings: >
-  %{ALEGetStatusLine()}
```
neersighted: got it

autoload/ale/engine.vim
neersighted: yeah, in hindsight, no point making them public

autoload/ale/engine.vim
neersighted: done

autoload/ale/linter.vim
neersighted: Didn't think of the `BufEnter` problem; you're right about just checking `s:linters` as well. In trying to keep this simple I missed some obvious issues.

autoload/ale/engine.vim
w0rp: Sorry to be a bother, but I'm going to revert those changes now. They didn't fix the issues one guy is having with MacVim, and the code as it was before didn't cause problems in MacVim in my tests.

w0rp: You can change it in this PR. Sorry for the bother.

neersighted: done!

neersighted:
If everything looks good to you, this is merge-ready. Still not quite happy with the mix of capitalization everywhere, but thats a niggle for another day.

autoload/ale/linter.vim
```diff
@@ -34,19 +34,31 @@ function! ale#linter#define(filetype, linter) abort
...
+        execute 'runtime! ale_linters/' . a:filetype . '/*.vim'
     endif
 
-    return [] " Yes, we miss the first run, but it avoids some code duplication.
+    if has_key(s:linters, a:filetype)
+        " If we found a linter, return it now.
+        return s:linters[a:filetype]
+    else
+        " If we couldn't load a linter, blacklist it in the future.
+        let g:ale_linters[a:filetype] = []
```
w0rp: Did you mean to assign to `s:linters` here?

w0rp:
Can you point to some examples of the mix of capitalisation? Do you mean for variable names? I could fix those issues, probably. All of the functions ought to be written in `PascalCase`, and all of the variables ought to be written `with_underscores`.

autoload/ale/linter.vim
neersighted: \*facedesk\*

neersighted:
In hindsight, it was I who introduced the mix last night working on this. Would you like all the one-word functions (such as `ale#engine#invoke`, etc) capitalized?

w0rp:
Yeah, if you're adding the lowercase function names, you can change them to PascalCase. After that I'll merge this. These are some great changes.

Once this is in, we can see if we can get some Vader tests in too. Then we'll really be cooking with gas.

neersighted:
On another note, I'm still getting no output from vint. It works fine when I run shell commands, and fine when I run the command specified by `linter#Get()`... Any ideas?

w0rp:
Do other linters work, or is it only `vint` which isn't working? Do any other linters using  `g:ale#util#stdin_wrapper ` work for you? Does your version of `vint` output errors to stderr instead of stdout?

w0rp:
Vint is the only linter which uses a special format with the wrapper script with no closing quote. Maybe you could try reverting it back to using its own function for handling the errors again, and see if that helps. There may be some issue there perhaps. (Take away `s:format` and see if you can parse the warnings/errors like other linters do.)

w0rp:
I'll merge this as it is now, and have a look at the Vint problems myself.

neersighted:
Sweet!

w0rp:
Merging that introduced some cursor bugs I had to fix, but that was quickly remedied. The `autocmd` line for the cursor had to be run in the main plugin file, and the loop for setting `linter_name` on the items was missing. That should be working now.

neersighted:
Got it, thanks!

PR #83, opened 2016-10-10
Fixup stdin-wrapper

neersighted:
* Use a more universal shebang.
* Use the template feature of mktemp to avoid silly gymnastics.

### DIFF
--- a/stdin-wrapper
+++ b/stdin-wrapper
@@ -1,4 +1,4 @@
-#!/bin/bash -eu
+#!/usr/bin/env bash
 
 # Author: w0rp <devw0rp@gmail.com>
 # Description: This script implements a wrapper for any program which does not accept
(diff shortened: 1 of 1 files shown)


stdin-wrapper
```diff
@@ -10,11 +10,8 @@
 file_extension="$1"
 shift
 
-temp_file=`mktemp`
-mv "$temp_file" "$temp_file$file_extension"
-temp_file="$temp_file$file_extension"
-
-trap "rm $temp_file" EXIT
+temp_file=$(mktemp --tmpdir "ale-XXX$file_extension")
```
w0rp: Should more Xs be used here? Maybe as many as `mktemp` uses by default.

neersighted: It's mixed-case alphanumeric, so I don't think its really an issue.

w0rp: Might as well err on the safe side. It couldn't hurt.

PR #85, opened 2016-10-10
add shellcheck option to the doc and sort options by theme

Geam:
I've sort the options cause I find it quite messy to have option for same language not one after one other. If you don't want it, I can reset to commit where I add the shellcheck option.

Sorry I was not available as much as I wanted this week-end for PR #66.

### DIFF
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -439,7 +439,7 @@ g:ale_linters_sh_shellckeck_exclusions *g:ale_linters_sh_shellckeck_exclusions*
   Type: |String|
   Default: `''`
 
-  Set this variable to exclude test for shellcheck (-e option);
+  Set this variable to exclude test(s) for shellcheck (-e/--exclude option).
 
 
 ===============================================================================
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -11,13 +11,13 @@ CONTENTS                                                         *ale-contents*
   3. Global Options.............................|ale-options|
   4. Linter Specific Options....................|ale-linter-options|
     4.1. eslint.................................|ale-linter-options-eslint|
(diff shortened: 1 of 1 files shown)


doc/ale.txt
```diff
@@ -393,41 +407,40 @@ g:ale_fortran_gcc_options                           *g:ale_fortran_gcc_options*
...
+4.9. shellchek                                  *ale-linter-options-shellcheck*
 
-g:ale_javascript_jshint_executable         *g:ale_javascript_jshint_executable*
+g:ale_linters_sh_shellckeck_exclusions *g:ale_linters_sh_shellckeck_exclusions*
 
   Type: |String|
-  Default: `'jshint'`
+  Default: `''`
+
+  Set this variable to exclude test for shellcheck (-e option);
```
w0rp: Should that read "exclude tests?" It looks like that's intended to be plural.

w0rp:
I saw this kind of sorting coming. The headings might get sorted again in future, possibly some different way. I'm pretty happy with what you've done here.

doc/ale.txt
w0rp: Also, it looks like there's a random semicolon on the end of the line.

Geam: Woopsie the semi-column, too much c/cpp latetly.
For the -e as I'm not a shellcheck user I've report to the doc and I don't think it's suppose to be plural : https://github.com/koalaman/shellcheck/wiki/Ignore
I'll dig through that a bit more.

Geam: Ok it's suppose to be plural : https://github.com/koalaman/shellcheck/issues/14

w0rp:
Cheers! :beers:

PR #91, opened 2016-10-11
Support Vint style issues, and enabling/disabling them

w0rp:
This pull request will add the `-s` flag for the Vint linter, and an option to turn style checks on and off. I decided to enable more warnings by default, and you can turn them off if they annoy you.

@neersighted Take a look, and let me know what you think.

### DIFF
--- a/ale_linters/vim/vint.vim
+++ b/ale_linters/vim/vint.vim
@@ -7,11 +7,20 @@ endif
 
 let g:loaded_ale_linters_vim_vint = 1
 
+" This flag can be used to change enable/disable style issues.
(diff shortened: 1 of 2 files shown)


neersighted:
Looks good to me, but how are you parsing the `style_problem` severity in your handler code?

neersighted:
It defaults to `warning` if not recognized as `error`, doesn't it...

w0rp:
@neersighted Yeah. If the severity isn't and error, it's considered a warning.

neersighted:
:+1:, looks good then.

PR #92, opened 2016-10-11
Remove many redundant handlers

neersighted:
Some of these are debatable, as there is some special behavior we want to save. All of these work on my system, but some more testing might be required.

To preserve the special cases, and to simplify some of the Unix-compatible handlers I left untouched, some method for extending handlers or inheriting from handlers is desirable.

This is just a rough draft, let me know what you think.

### DIFF
--- a/ale_linters/haskell/ghc.vim
+++ b/ale_linters/haskell/ghc.vim
@@ -7,63 +7,12 @@ endif
 
 let g:loaded_ale_linters_haskell_ghc = 1
 
-function! ale_linters#haskell#ghc#Handle(buffer, lines)
-    " Look for lines like the following.
-    "
-    " /dev/stdin:28:26: Not in scope: `>>>>>'
-    let l:pattern = '^[^:]\+:\(\d\+\):\(\d\+\): \(.\+\)$'
-    let l:output = []
-
-    " For some reason the output coming out of the GHC through the wrapper
-    " script breaks the lines up in strange ways. So we have to join some
-    " lines back together again.
-    let l:corrected_lines = []
-
-    for l:line in a:lines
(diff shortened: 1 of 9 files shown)


ale_linters/haskell/ghc.vim
```diff
@@ -7,69 +7,18 @@ endif
 
 let g:loaded_ale_linters_haskell_ghc = 1
 
-function! ale_linters#haskell#ghc#Handle(buffer, lines)
```
w0rp: Test the GHC handler again. It did some funky things. I might have fixed this now with changes I have made since I first wrote it to the `stdin-wrapper` script. It might have been an IFS bug.

neersighted: It works on my machine (tm). I tested on the git-annex source code.

neersighted:
There are possible some other handlers that could be similarly merged, but I think this is a good start.

ale_linters/verilog/iverilog.vim
```diff
@@ -25,7 +25,7 @@ function! ale_linters#verilog#iverilog#Handle(buffer, lines)
         endif
 
         let l:line = l:match[1] + 0
-        let l:type = l:match[2] ==# 'warning' ? 'W' : 'E'
```
mshr-h: In iverilog, there are 2 types of error representation, `error` and `syntax error`.
So, I recommend not to change this statement.

neersighted: How's that?

mshr-h: OK! :thumbsup:

w0rp:
Cheers! :beers:

PR #93, opened 2016-10-11
Clean and reorganize flags/preferences

neersighted:
Just some housekeeping work, centralizing all the preferences and removing redundant variable declarations

### DIFF
--- a/plugin/ale.vim
+++ b/plugin/ale.vim
@@ -2,6 +2,8 @@
 " Description: Main entry point for the plugin: sets up prefs and autocommands
 "   Preferences can be set in vimrc files and so on to configure ale
 
+" Sanity Checks
+
 if exists('g:loaded_ale')
     finish
 endif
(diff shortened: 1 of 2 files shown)


plugin/ale.vim
```diff
@@ -2,41 +2,66 @@
...
-let g:ale_lint_on_text_changed = get(g:, 'ale_lint_on_text_changed', 1)
-
 " This flag can be set with a number of milliseconds for delaying the
 " execution of a linter when text is changed. The timeout will be set and
 " cleared each time text is changed, so repeated edits won't trigger the
 " jobs for linting until enough time has passed after editing is done.
 let g:ale_lint_delay = get(g:, 'ale_lint_delay', 200)
 
+" This flag can be set to 0 to disable linting when text is changed.
+if get(g:, 'ale_lint_on_text_changed', 1)
```
w0rp: These changes won't set the global variables to a default value. That can be a problem if you want to read them in other files. I'd keep setting these flags to their default value.

neersighted: They weren't used anywhere else, so I figured this was more elegant. But if you prefer, it is trivial to keep them as they were.

neersighted: Rebased

w0rp:
Cheers! :beers:

PR #94, opened 2016-10-11
Add system to alias linter folders, and register multiple filetypes

### LINKED ISSUE #87: Alias system required
With my lazy-loading work (and indeed, before it, in the case of the `sh` linter), `shell` linter, linters which support multiple file-types are not actually used.

This is presently the case for `jsx` and `zsh`. Two solutions are obvious:

* Create symlinks (easiest and simplest, in my opinion)
* Create an alias table

I think the symlink approach is more elegant, as it avoids referencing any linters in the core, and can be trivially and transparently managed, eg `ln -s ale_linters/sh/shell.vim ale_linters/zsh/shell.vim`.

Thoughts?

neersighted:
Closes #87

This change is backwards compatible, as ale#linter#Define() takes
variable arguments, with the linter being the last. It still needs to be
documented, however. Not sure how to describe vargs in vimscript terms.

### DIFF
--- a/ale_linters/javascript/eslint.vim
+++ b/ale_linters/javascript/eslint.vim
@@ -49,14 +49,7 @@ function! ale_linters#javascript#eslint#Handle(buffer, lines)
     return l:output
 endfunction
 
-call ale#linter#Define('javascript', {
-\   'name': 'eslint',
-\   'executable': g:ale_javascript_eslint_executable,
-\   'command': g:ale_javascript_eslint_executable . ' -f unix --stdin --stdin-filename %s',
-\   'callback': 'ale_linters#javascript#eslint#Handle',
-\})
-
-call ale#linter#Define('javascript.jsx', {
+call ale#linter#Define('javascript', 'javascript.jsx', {
 \   'name': 'eslint',
 \   'executable': g:ale_javascript_eslint_executable,
 \   'command': g:ale_javascript_eslint_executable . ' -f unix --stdin --stdin-filename %s',
--- a/ale_linters/javascript/jscs.vim
+++ b/ale_linters/javascript/jscs.vim
@@ -43,14 +43,7 @@ function! ale_linters#javascript#jscs#Handle(buffer, lines)
     return l:output
 endfunction
 
-call ale#linter#Define('javascript', {
-\   'name': 'jscs',
-\   'executable': 'jscs',
-\   'command': 'jscs -r unix -n -',
-\   'callback': 'ale_linters#javascript#jscs#Handle',
-\})
-
-call ale#linter#Define('javascript.jsx', {
+call ale#linter#Define('javascript', 'javascript.jsx', {
 \   'name': 'jscs',
 \   'executable': 'jscs',
 \   'command': 'jscs -r unix -n -',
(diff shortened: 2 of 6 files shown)


w0rp:
I meant `a:filetype` on the RHS above. I looks like I can't edit my own review comments.

neersighted:
That was what I did initially, but that falls down in some cases. What if not all javascript linters cope with jsx? In fact, I ran into the issue with shellcheck, which does not support zsh. In short, linters need to be able to indicate what they support, blanket aliasing is not a good idea.

neersighted:
Also, no point on making my implementation here user configurable, as it depends on linter support.

w0rp:
Users may want to use linters for new filetypes they define and then detect with their own detection rules. Maybe we could use a mapping to a mapping of linters to run? `{'javascript.jsx': {'javascript': ['eslint', ...]}, ...}`

w0rp:
Actually, we don't need to limit the linters in that way for one variable, as we can combine two of them.

```vim
let g:ale_linter_aliases = {'javascript.jsx', 'javascript'}
let g:ale_linters = {'javascript.jsx', ['eslint']}
```

Then we read the original filetype, not the mapped one, to limit which linters will be run.

neersighted:
I'll make it configurable sure, but I think that syntax is a kludge at best. It still won't work unless a linter registers support.

Think of it this way, I have two linters in the `sh` folder, and both register for `sh`.
Only one support zsh. if `#Get()` returns the list for `sh`, an incorrect linter will be run.

So then, how do we know? The linter either has to notify ale that it support zsh, or ale has to have it hardcoded that `sh/shell.vim` supports zsh despite it not claiming such. The vargs syntax is simply to avoid duplicate/copy-pasted `#Define()` blocks.

neersighted:
That can work as a second feature, sure, but I don't like that you have to configure jsx support, and instead of it working out of the box. Why not combine both approaches for maximum flexibility? The user preference can be just that, but the core is aware of the linters and vice-versa, so it can do things in a streamlined fashion.

w0rp:
It can work out of the box if you set up defaults. Set up default aliases, and merge user defined aliases with the defaults. Set up default arrays to limit which linters are run for filetypes like `zsh`. Then it will all work.

neersighted:
See, as in my earlier comment, now the core has to hardcode 'sh/shell.vim' support for 'zsh', which is rather inelegant in my opinion. I prefer this approach, where the linter is the one source of truth for filetype support. The only reason we even need to have a `s:folder_aliases` list is for lazy-loaded, and we can switch to symlinks if we decide that is more elegant.

w0rp:
I'd prefer to set up some aliases in the main file, and make the aliases configurable.

neersighted:
How about this approach as a middle ground?

w0rp:
That still won't quite work, because it won't allow you to define an entirely new filetype, and run exiting linters against that new filetype.

w0rp:
https://github.com/w0rp/ale/compare/master...w0rp:filetype-aliasing?expand=1 This is the kind of thing I am talking about. That will allow users to configure the one-to-one filetype mappings and the linters to run for each new filetype however they like.

I made it just load every linter for a given filetype there, as there's no good way to limit the selections there, and it shouldn't penalising the loading times massively. There will only ever be maybe 3 .vim files per filetype anyway.

Take a look.

neersighted:
Fair enough, I'm happier with that approach then I thought I would be. Let's go with that!

w0rp:
Cool, I'll add some documentation and merge it later on.

PR #97, opened 2016-10-11
Implement a more efficient statusbar

neersighted:
The statusbar now keeps its state in a separate variable, in order to
avoid excess iterations. The engine now updates said variable on run,
and a new function is made available for external statusbars to call (to
avoid dependencies on internal implementation details of ale).

To keep things light, the status bar code is not loaded unless invoked
by the user or an external plugin. On the first load it will update
itself from the global loclist, after that, the engine will handle all
updates.

The external integration function, `ale#statusline#Count()`, will return
a tuple in the format `[E, W]` (where E is errors, W is warnings), unless
no data exists (ie, the plugin doesn't have a linter for a file or has
not run yet), in which case it returns 0/false.

In addition, I fixed a few errors (`ale#cleanup#Buffer` never ran correctly, echocursor was being set incorrectly, etc) I came across as I worked.

I will make use of this in my airline pull request to reduce coupling to ale internals.

### DIFF
--- a/autoload/ale/statusline.vim
+++ b/autoload/ale/statusline.vim
@@ -1,20 +1,12 @@
 " Author: KabbAmine <amine.kabb@gmail.com>
 " Description: Statusline related function(s)
 
-function! ale#statusline#Status() abort
-    " Returns a formatted string that can be integrated in the
-    " statusline
-
-    let l:buffer = bufnr('%')
-    let l:loclist = g:ale_buffer_loclist_map
-
-    if !has_key(l:loclist, l:buffer)
-        return ''
-    endif
-
+" Update the buffer error/warning count with data from loclist.
+function! ale#statusline#Update(buffer, loclist) abort
     let l:errors = 0
     let l:warnings = 0
-    for l:entry in l:loclist[l:buffer]
+
+    for l:entry in a:loclist
         if l:entry.type ==# 'E'
             let l:errors += 1
         else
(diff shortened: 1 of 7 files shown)


w0rp:
I'll look through this eventually. I think first I want some Vader tests to be written for the statusline. I've been creating regressions without tests to check for them already, and I'd like to stop that.

neersighted:
I think this makes the statusline more testable, if anything. I can work on some tests for the status line, but I would prefer to start with this merged, versus the current implementation. Up to you ultimately, but this is a lot easier to mock than loclists.

w0rp:
I don't think we should do any kind of refactoring without tests now we have the means to write them. It will help a lot to have the tests running on Travis CI, which will happen soon enough.

autoload/ale/statusline.vim
```diff
@@ -1,29 +1,54 @@
...
+" Returns a tuple of errors and warnings (or false if no data exists)
+" for use in third-party integrations.
+function! ale#statusline#Count(buffer) abort
+    if !has_key(g:ale_buffer_count_map, a:buffer)
+        if has_key(g:ale_buffer_loclist_map, a:buffer)
+            call ale#statusline#Update(a:buffer, g:ale_buffer_loclist_map[a:buffer])
+            return ale#statusline#Count(a:buffer)
+        else
+            return 0
+        endif
```
w0rp: You could do this instead.

```vim
if !has_key(g:ale_buffer_count_map, a:buffer)
    call ale#statusline#Update(a:buffer, get(g:ale_buffer_loclist_map, a:buffer, []))
endif

return g:ale_buffer_count_map[a:buffer]
```

The data will be replaced when there are actually errors or warnings to display, and you will get `[0, 0]` when there is no data, so there won't be any difference between the initial load and when ALE doesn't find problems later on.

autoload/ale/statusline.vim
```diff
@@ -1,29 +1,54 @@
...
+    let l:buffer = bufnr('%')
+
+    if !has_key(g:ale_buffer_count_map, l:buffer)
+        if has_key(g:ale_buffer_loclist_map, l:buffer)
+            call ale#statusline#Update(l:buffer, g:ale_buffer_loclist_map[l:buffer])
+            return ale#statusline#Status()
+        else
+            return ''
+        endif
+    endif
```
w0rp: Similar to the above, you can write...

```vim
if !has_key(g:ale_buffer_count_map, l:buffer)
    call ale#statusline#Update(a:buffer, get(g:ale_buffer_loclist_map, a:buffer, []))
endif
```

w0rp: Then the return value here will always be consistent between when ALE is first loaded, and when it reports no errors.

plugin/ale.vim
```diff
@@ -120,10 +122,10 @@ let g:ale_warn_about_trailing_whitespace =
 
 " Housekeeping
 
-augroup ALECleanup
+augroup ALECleanupGroup
     autocmd!
     " Clean up buffers automatically when they are unloaded.
-    autocmd BufUnload * call ale#cleanup#Buffer('<abuf>')
+    autocmd BufUnload * call ale#cleanup#Buffer(expand('<abuf>'))
```
w0rp: :+1: I missed that when I first added it. We definitely need to add that test for cleanup.

autoload/ale/statusline.vim
neersighted: I am trying to differentiate between `[0,0]` == OK, and `[0, 0]` == ale has not run. Any ideas on that front?

s-ol: `[errors, warnings, status_code]`?

w0rp:
Yep, looks good. I think there are two stages here. 1. Improve statusbar performance. 2. Implement some new behaviour for a different initial status when no errors have been reported, if we determine that's desirable.

Cheers! :beers:

PR #108, opened 2016-10-13
Add autocommand test

neersighted:


neersighted:
Crap, running into some problems here. It works on my machine, but my machine is much faster than Travis' workers. And I'd hate to just increase the `sleep` here... Any ideas?

w0rp:
Maybe there is some way you can call a function to stop all of the jobs, so they will exit earlier.

neersighted:
I'm not familiar enough with the async API to implement that. If you have any bright ideas, go ahead, otherwise, might just have to `sleep 5` for this to work.

w0rp:
I'll see what I can dig up from the documentation later this evening.

w0rp:
After reading the documentation, I think the best two options would be to either increase the sleep time, which isn't fantastic, or to expose the script function so you can call it directly. If you dig through Google, there is some discussion of running script functions in Vader, but it involves parsing the source files and might not be the best.

neersighted:
How about a blocking version? We can test ale#Lint seperately by mocking the engine, and we can mock ale#Lint with a blocking version to test the engine.

w0rp:
The best bet is to probably call `HandleExit` directly and test that.

w0rp:
I have a possible solution for you now. I just added a test which checks that the loclist gets updated after linting is done, and I now use a call for a new function like so:

```vim
call ale#engine#WaitForJobs(2000)
```

That will wait up to 2 seconds for linters to finish their work, but possibly complete sooner than that if jobs are done. So far it seems to work. Adjustments might be needed later if some problems come up around how execution of handlers after jobs finish works.

neersighted:
That sound perfect! I'll go ahead and try to incorporate this now.

neersighted:
Looks like that code only works under Vim 8, not NeoVim. I'm getting an unknown function, `job_status`.

neersighted:
Regardless, the tests pass in regular Vim. I am having the loclist test fail on my machine though (in Docker).

```
  Starting Vader: /testplugin/test/test_linting_updates_loclist.vader
    (1/1) [  GIVEN] Some JavaScript with problems
    (1/1) [EXECUTE] The loclist should be updated after linting is done
    (1/1) [EXECUTE] (X) {} should be equal to {'3': [{'lnum': 1, 'bufnr': 3, 'vcol': 0, 'linter_name': 'eslint', 'nr': -1, 'type': 'W', 'col': 10, 'text': 'Infix operators must be spaced. (space-infix-ops)'}, {'lnum': 2, 'bufnr': 3, 'vcol': 0, 'linter_name': 'eslint', 'nr': -1, 'type': 'E', 'col': 10, 'text': 'Missing semicolon. (semi)'}]}
  Success/Total: 0/1
```

neersighted:
Ah, looks like that was a matter of pulling down the latest docker image. My bad.

w0rp:
It doesn't yet work in NeoVim or on non Unix machines. I'll see what I can do about that.

I think there's a problem with the Makefile where it doesn't know how to pull a more recent build of the Docker image if you already have some older image. I'm not sure how to fix that. Maybe an ID can be written into the Makefile to check against, so that pulling again with a new ID will cause the new image to be loaded.

w0rp:
There's one more thing I just noticed. Could you `unlet` the `g:success` variable in an `After` block so that variable isn't around for other tests? One unfortunate side effect of this is that the variable won't be available by the time the `Then` block is run, so you'll have to move the assertion into the `Execute` block. I don't know why Vader runs `After` before `Then`. Maybe there's a better way to handle this.

w0rp:
Thanks for adding the test! I'll mark this one off on the ongoing issue.

PR #120, opened 2016-10-18
Use `cat` instead of `read -r` to stream stdin to file

hauleth:


stdin-wrapper
```diff
@@ -6,15 +6,15 @@
...
 file_extension="$1"
 shift
 
 temp_file=$(mktemp --tmpdir "ale-XXXXXXXXX$file_extension")
 trap 'rm $temp_file' EXIT
 
-while read -r; do
-    echo "$REPLY" >> "$temp_file"
-done
+cat - > "$temp_file"
```
w0rp: Is there a reason for this?

hauleth: In theory it would be faster (mmaping and other fancy stuff). Also this allow me to to change it to named pipe soon.

w0rp: Makes sense.

What will the named pipe be used for?

w0rp:
I might see if I can get the wrapper script covered with a test first. I'll give it a go near my lunch break.

stdin-wrapper
hauleth: Instead of creating temporary file and writing data to it provide named pipe for the process to "fool it" that it is reading file, not stream. This is [reaison d'etre of named pipes](https://groups.google.com/d/msg/golang-dev/R9hbrKG18hQ/6RjEiOOG_hoJ).

w0rp:
I wasn't able to easily make the tests run `Vint`, even though that should be installed in the Docker image. I'll try later.

hauleth:
@w0rp now it should be fine with BSD tooling also. Old version couldn't be used with BSD `mktemp` and other small incompatibilities.

stdin-wrapper
```diff
@@ -1,20 +1,28 @@
...
 
+set -eu
+
 # All of the following arguments are read as command to run.
 file_extension="$1"
 shift
 
-temp_file=$(mktemp --tmpdir "ale-XXXXXXXXX$file_extension")
-trap 'rm $temp_file' EXIT
+temp_dir=$(mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
```
w0rp: Won't this always create the directory `/tmp/mytmpdir`? Don't you need some Xs to get a random name?

w0rp: Can `mktemp -d` fail? If it does, I don't think the part after `||` will do any better, which as I said above, doesn't work.

stdin-wrapper
```diff
@@ -1,20 +1,28 @@
...
+# at you) do not fit in line and require filename ending. Otherwise it would be
+# simple as
+#
+#     "$@" /dev/stdin
+#
+# without all that hackery with copying `/dev/stdin`
+cp /dev/stdin "$temp_file"
 
-"$@" "$temp_file"
+exec "$@" "$temp_file"
```
w0rp: Will the use of `exec` make the `trap` above execute too early, and delete the directory before the file has been read? I expect `exec` to end the process for this script, and that to trigger the trap.

stdin-wrapper
hauleth: `mktemp -d` will work with GNU and fail with BSD, that is why there is `mktemp -d -t 'mktemp -d -t mytempdir`.

w0rp: The use of temporary directories should help with supporting Java linting. I don't think there's any way to run `javac` where it doesn't output a `.class` file, so `rm -r` on a temporary directory should clean that up nicely.

stdin-wrapper
hauleth: Yeah, you were right. Fixed now.

stdin-wrapper
w0rp: Ah, so it's BSD only? On BSD, will it create just `/tmp/mytempdir`, or will it create a random directory name?

w0rp: If you add some X's, you might be able to use the same command for both GNU and BSD. Maybe...

stdin-wrapper
```diff
@@ -1,20 +1,28 @@
...
-    echo "$REPLY" >> "$temp_file"
-done
+# In perfect world it wouldn't be needed, but some tools (`go vet`, I am looking
+# at you) do not fit in line and require filename ending. Otherwise it would be
+# simple as
+#
+#     "$@" /dev/stdin
+#
+# without all that hackery with copying `/dev/stdin`
+cp /dev/stdin "$temp_file"
```
hauleth: Additional feature: predictable filename, now it is always `file.ext`, not some random string.

stdin-wrapper
hauleth: @w0rp I think that instead of using `javac` as a linter we should add support for Sonar or other external static analyzers. `javac` does poor job as a linter.

hauleth: However [Syntastic linter](https://github.com/scrooloose/syntastic/blob/master/syntax_checkers/java/javac.vim) looks nice, so I believe that we could port that one to Ale. Unfortunately I am not Java dev so I don't have time and resources to port that.

stdin-wrapper
w0rp: We could probably even change the wrapper script at some point to write out the exact same basename of the file that you were originally given. That would help a lot with situations where the linter expects the filename to be a certain way.

w0rp:
This all looks good to me. You can leave this how it is, and I'll try and get the test for the wrapper script working before I merge this.

w0rp:
I haven't been able to get a test to work inside the Docker container, although strangely tests which use the stdin-wrapper script work outside of the container. I'm not sure what the problem is there yet. `vint` runs just find inside the container. I tried the `stdin-wrapper` script on my machine with Vint, and it seems to work fine.

There is a problem with the changes to the `dmd-wrapper` script, however. The line which gets the import paths results in that variable being set to the value `'No valid root package found - aborting.'` if the linting is performed on a file which doesn't have a DUB package. I think you'll have to handle that failure better and set the string to be empty in that case.

hauleth:
I have fixed that.

w0rp:
Cool, that should do it. Cheers! :beers:

PR #123, opened 2016-10-19
Added support for hlint

jparoz:


w0rp:
I see no problems here. Cheers! :beers:

PR #126, opened 2016-10-20
Add support for cppcheck

bartlibert:
I'm using a new handling function because the cppcheck "gcc-compatible" output mode does not include column numbers, so it does not work with the already existing gcc handler.


autoload/ale/handlers.vim
```diff
@@ -76,6 +75,35 @@ function! ale#handlers#HandleGCCFormat(buffer, lines) abort
     return l:output
 endfunction
 
+function! ale#handlers#HandleCppCheckFormat(buffer, lines)
```
w0rp: Could you add a vader unit test for this function? The other generic handler functions have tests for them now.

bartlibert:
I have added a test (sorry, thought about it, but then forgot)

PR #129, opened 2016-10-20
Add ansible-lint support and factor out PEP8 handler

### LINKED ISSUE #128: Feature request: add ansible-lint support
Please add support of [`ansible-lint`](https://github.com/willthames/ansible-lint) when `ft=ansible`. Again, ideally it should work in conjunction with `yamllint` when `ft=ansible.yaml`.

neersighted:
Still need to implement aliasing. @w0rp, what would be the most elegant way to set this up? (ansible-lint for ansible, yamllint for yaml, both for ansible.yaml).

Closes #128

### DIFF
--- a/autoload/ale/handlers.vim
+++ b/autoload/ale/handlers.vim
@@ -103,6 +103,43 @@ function! ale#handlers#HandleCppCheckFormat(buffer, lines) abort
 
 endfunction
 
+function! ale#handlers#HandlePEP8Format(buffer, lines)
+    " Matches patterns line the following:
+    "
+    " Matches patterns line the following:
+    "
+    " stdin:6:6: E111 indentation is not a multiple of four
+    " test.yml:35: [EANSIBLE0002] Trailing whitespace
+    let l:pattern = '^' . s:path_pattern . ':\(\d\+\):\?\(\d\+\)\?: \[\?\(\([[:alpha:]]\)[[:alnum:]]\+\)\]\? \(.*\)$'
+    let l:output = []
+
+    for l:line in a:lines
+        let l:match = matchlist(l:line, l:pattern)
+
+        if len(l:match) == 0
+            continue
+        endif
+
+        let l:code = l:match[3]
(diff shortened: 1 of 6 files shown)


w0rp:
You can add another entry to the existing directory to map the ansible filetype to the YAML filetype. We might need to implement something different for this though, like aliasing filetypes while still keeping linters for that new file type. Maybe for x mapping to y, we could load the linters for x as well as y for x.

autoload/ale/handlers.vim
```diff
@@ -103,6 +103,43 @@ function! ale#handlers#HandleCppCheckFormat(buffer, lines) abort
...
 
+function! ale#handlers#HandlePEP8Format(buffer, lines)
+    " Matches patterns line the following:
+    "
+    " Matches patterns line the following:
+    "
+    " stdin:6:6: E111 indentation is not a multiple of four
+    " test.yml:35: [EANSIBLE0002] Trailing whitespace
+    let l:pattern = '^' . s:path_pattern . ':\(\d\+\):\?\(\d\+\)\?: \[\?\(\([[:alpha:]]\)[[:alnum:]]\+\)\]\? \(.*\)$'
+    let l:output = []
```
w0rp: Could you add a test for this in the Vader tests, so all of the handlers in the handlers files are covered?

neersighted:
Actually, we should handle dot-seperated filetypes. http://stackoverflow.com/questions/8413781/automatically-set-multiple-file-types-in-filetype-if-a-file-has-multiple-exten

Vim allows dot separated filetypes to apply multiple filetypes. With this support, there would be no need for javascript.jsx to be explicitly handler for instance, we just run the checkers for both JavaScript and JSX.

w0rp:
Hmm, makes sense. Create the issue for it if you haven't already. I guess we'd have to change the loading of linters to run through a list of filetypes, rather than just the one.

neersighted:
I would merge this now, and implement dot-seperated support separately.

w0rp:
Sure. I'll take this. We can sort out the dot-separated filetypes in another pull request.

PR #133, opened 2016-10-21
Added c++14 standard support and POSIX API compile time linker flags and ISO c++14 atomic and thread library support…

ehzShelter:
…for avoiding annoying errors and warnings..!

### DIFF
--- a/ale_linters/cpp/gcc.vim
+++ b/ale_linters/cpp/gcc.vim
@@ -9,13 +9,15 @@ let g:loaded_ale_linters_cpp_gcc = 1
 
 " Set this option to change the GCC options for warnings for C.
 if !exists('g:ale_cpp_gcc_options')
-    let g:ale_cpp_gcc_options = '-Wall'
+  " added c++14 standard support
+  " POSIX thread and standard c++ thread and atomic library Linker
+    let g:ale_cpp_gcc_options = '-std=c++14 -ldl -lpthread -lm -latomic -Wall -Wextra -Wshadow -pedantic'
 endif
 
 call ale#linter#Define('cpp', {
-\   'name': 'gcc',
+\   'name': 'g++',
 \   'output_stream': 'stderr',
-\   'executable': 'gcc',
+\   'executable': 'g++',
 \   'command': 'gcc -S -x c++ -fsyntax-only '
 \       . g:ale_cpp_gcc_options
 \       . ' -',
--- a/ale_linters/cpp/gcc.vim
+++ b/ale_linters/cpp/gcc.vim
@@ -11,7 +11,14 @@ let g:loaded_ale_linters_cpp_gcc = 1
 if !exists('g:ale_cpp_gcc_options')
   " added c++14 standard support
   " POSIX thread and standard c++ thread and atomic library Linker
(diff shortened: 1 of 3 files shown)


ale_linters/cpp/gcc.vim
```diff
@@ -9,13 +9,15 @@ let g:loaded_ale_linters_cpp_gcc = 1
 
 " Set this option to change the GCC options for warnings for C.
 if !exists('g:ale_cpp_gcc_options')
-    let g:ale_cpp_gcc_options = '-Wall'
+  " added c++14 standard support
+  " POSIX thread and standard c++ thread and atomic library Linker
+    let g:ale_cpp_gcc_options = '-std=c++14 -ldl -lpthread -lm -latomic -Wall -Wextra -Wshadow -pedantic'
```
w0rp: Could any of these options cause problems for other users? I do like making the warnings and errors as nagging as possible by default.

ehzShelter: No...  Otherwise... 
```cpp
#include <iostream>

int main(void)
{
      auto myStr = "Hi Vim ale";
      std::cout << myStr << std::endl;
}

```

### Otherwise It throws error message..

ehzShelter: auto c++11 features :)

ehzShelter: -std=c++14 backward compatible with previous version of c++..
Ref:

http://en.cppreference.com/w/

ehzShelter: Another thing...! for c++14, In operating system, gcc aka g++ version must be 4.9.0 or greater.

Reference : https://isocpp.org/blog/2014/04/gcc-4.9.0

w0rp: It's a couple of years old now, so there probably aren't that many users where this wouldn't work for them, and they can always configure the option however they want.

w0rp:
I think I'll go with this. Could you update the help file to document the same default string?

ehzShelter:
```
vimrc
-let g:syntastic_c_compiler = 'clang'		
 -let g:syntastic_c_compiler_options = ' -std=c11 -latomic -Werror -pedantic-errors -Wno-error=unused-parameter -Wno-error=unused-command-line-argument'		
 -		
 -" Syntastic Plugin C++14 support for Clang Compiler		
 -" Make sure clang have system path		
 -let g:syntastic_cpp_compiler = 'clang++'		
 -" flags more specific to clang		
 -let g:syntastic_cpp_compiler_options = ' -std=c++14 -stdlib=libc++ -lc++abi -ldl -latomic -Werror -Wself-assign -Wno-error=unused-command-line-argument -Wno-error=unused-parameter -pedantic-errors -Wextra-tokens -Wambiguous-member-template -Wbind-to-temporary-copy '
```

Make it interface-able through .vimrc 
In Syntastic Plugin, I can change it according to my choice...

Then people can meet their compilation requirements, Because some people like more strict check whereas some people don't

ehzShelter:
I have updated the docs default string and Now it is more minimalist !

w0rp:

PR #135, opened 2016-10-21
Add ale_lint_in_normal_mode option

renyard:
This PR implements an option to lint when exiting insert mode and making changes in normal mode.

(I found continuous linting in insert mode a little off putting and this seems like a happy medium without having to wait until a save.)

### DIFF
--- a/plugin/ale.vim
+++ b/plugin/ale.vim
@@ -70,6 +70,20 @@ if g:ale_lint_on_save
     augroup END
 endif
 
+" This flag can be set to 1 to enable linting when leaving insert mode.
+let g:ale_lint_in_normal_mode = get(g:, 'ale_lint_in_normal_mode', 0)
+if g:ale_lint_in_normal_mode
+    augroup ALERunOnInsertLeaveGroup
+        autocmd!
+        autocmd InsertLeave * call ale#Queue(0)
+    augroup END
+
+    augroup ALERunOnTextChangedNormalGroup
+        autocmd!
+        autocmd TextChanged * call ale#Queue(g:ale_lint_delay)
+    augroup END
+endif
+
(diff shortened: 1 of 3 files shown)


plugin/ale.vim
```diff
@@ -70,6 +70,20 @@ if g:ale_lint_on_save
...
+if g:ale_lint_in_normal_mode
+    augroup ALERunOnInsertLeaveGroup
+        autocmd!
+        autocmd InsertLeave * call ale#Queue(0)
+    augroup END
+
+    augroup ALERunOnTextChangedNormalGroup
+        autocmd!
+        autocmd TextChanged * call ale#Queue(g:ale_lint_delay)
+    augroup END
```
w0rp: I wonder if it will be confusing that you can set the option for linting when text is changed to being off, but then have it so it's still linting when text is changed. I wonder if this `autocmd` should be removed.

Another option would be to split the option into three options, for toggling linting in normal mode, insert mode, and when leaving insert mode, with the old option kept for turning the first two off easily.

renyard:
Thanks for the comments.

I think you're right that the multiple `autocmd` under one option could be confusing, but also think splitting it out into more options could be just as bad. I wonder if documenting this as a configuration example in the readme, demonstrating the configurability of the plugin could be good?

Let me know your thoughts and I'd be happy to move this out into my vimrc and write it up for the docs or readme, if you think it will be useful.

Thanks for a great plugin!

w0rp:
I think documenting that you can do something like this would be best. You could write something for the README file if you want. I will eventually re-arrange the README so there's an FAQ section and put something like that in there.

w0rp:
I'm closing this, in favour of adding something like this to the FAQ instead.

Issue #45, opened 2016-10-05
Add quickfix support


w0rp:
I addeed loclist support on day 1, but not quickfix list support. Some people will probably like to use the quickfix list. Some support should be added for populating the quickfix list, and for automatically opening the list when there are errors.

w0rp:
This has been resolved already.

PR #137, opened 2016-10-22
Factor out list support, and implement opening/closing

### LINKED ISSUE #136: Request: add option to open loclist when there are errors
Hi, thanks for ale, I'm finding it to be a great way to run linting without any UI lag. One option that I miss from neomake is `g:neomake_open_list`, which will automatically open the location list if there are errors (the option to preserve focus in the buffer you're editing is also nice). Any chance of something like this being added to ale?

The neomake description of this option is here:
https://github.com/neomake/neomake/blob/master/doc/neomake.txt#L242

neersighted:
This PR factors out the loclist logic to be more generic and support the
quickfix list (as well as moving it to its own file). It also implements
the opening and closing of said lists, with logic to ensure we don't
close the list when another plugin is using it.

I am not sure what the best way to address preferences is, as I wanted
to maintain backward compaibility. I think the implementation here is
acceptable, but @w0rp's opinion may differ.

The settings still need to be documented, but I will follow up on that
after the settings interface is finalized.

Closes #136, #45

### DIFF
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -105,16 +105,16 @@ function! s:HandleExit(job) abort
     " for efficient lookup of the messages in the cursor handler.
     call sort(g:ale_buffer_loclist_map[l:buffer], 'ale#util#LocItemCompare')
 
-    if g:ale_set_loclist
-        call setloclist(0, g:ale_buffer_loclist_map[l:buffer])
+    if g:ale_set_quickfix || g:ale_set_loclist
+        call ale#list#SetLists(g:ale_buffer_loclist_map[l:buffer])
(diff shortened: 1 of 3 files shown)


w0rp:
@neersighted If you can add some tests for this, I'll take this. It might be best to check out my pull request for stopping jobs when buffers exit first, which puts everything in the one global now.

neersighted:
@w0rp Yeah, been a bit busy and haven't had time to finish the tests. I think I'll wait for that to merge first to make things easy.

w0rp:
This pull request has been superseded by https://github.com/w0rp/ale/pull/266.

PR #138, opened 2016-10-22
Add support for dot-seperate linters, improve linter tests

### LINKED ISSUE #132: Support dot-seperated filetypes
Eg, http://stackoverflow.com/questions/8413781/automatically-set-multiple-file-types-in-filetype-if-a-file-has-multiple-exten

This would simplify support for things like `javascript.jsx`, which would be considered both `jsx` and `javascript`, or `ansible.yaml`, which would be both `ansible` and `yaml`. Syntastic supports this, as does the syntax highlighting engine.

This also means that some linters can run only for jsx, and some apply to both jsx and javascript, making this finer-grained than the existing alias system (which can stay).

The change is relatively simple, and just requires some parsing in `linter.vim`.

neersighted:
This PR first and formost implements support for dot-seperated filetypes,
a very trivial change.

This closes #132

But more importantly, this PR vastly improves the test quality for
`ale#linter#Get`. It enables us to reset the state of ale's internal
linter cache, to facilitate better testing, as well as making use of
mocked linters instead of depending on linters on disk (which may
change). In addition, a dummy linter is defined to test the autoloading
behavior.

Header guards were removed from all linters as:

* A: ale won't try and load linters if they already exist in memory
* B: we can't reset state for testing if they can't be loaded again

### DIFF
--- a/ale_linters/ansible/ansible-lint.vim
+++ b/ale_linters/ansible/ansible-lint.vim
@@ -1,12 +1,6 @@
 " Author: Bjorn Neergaard <bjorn@neersighted.com>
 " Description: ansible-lint for ansible-yaml files
 
-if exists('g:loaded_ale_linters_ansible_ansiblelint')
-    finish
-endif
-
-let g:loaded_ale_linters_ansible_ansiblelint = 1
-
 call ale#linter#Define('ansible', {
 \   'name': 'ansible',
 \   'executable': 'ansible',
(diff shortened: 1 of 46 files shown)


neersighted:
I would probably implement dot-seperated support in linter limits for that, to make it easy. As far as documentation goes, on it!

neersighted:
As far as header guards go, we explicitly load our linters and we do not load them again if they exist in memory (the `s:linters` variable). I don't think its an issue here, as with them we can't reset state to mock out linters.

w0rp:
That's good enough for me. This is pretty good stuff, cheers! :beers:

PR #139, opened 2016-10-22
Added support for mdl, a markdown linter

sbdchd:
Took me a bit [searching](https://github.com/w0rp/ale/issues/16) to realize that this plugin doesn't rely on `&errorformat`.

`let &errorformat = '%f: %l: %m'` --> `let pattern = ':\(\d*\): \(.*\)$'`

https://github.com/mivok/markdownlint

### DIFF
--- a/ale_linters/markdown/mdl.vim
+++ b/ale_linters/markdown/mdl.vim
@@ -1,0 +1,35 @@
+" Author: Steve Dignam <steve@dignam.xyz>
+" Description: Support for mdl, a markdown linter
+
+function! ale_linters#markdown#mdl#Handle(buffer, lines) abort
+    " matches: '(stdin):173: MD004 Unordered list style'
+    let l:pattern = ':\(\d*\): \(.*\)$'
+    let l:output = []
+
+    for l:line in a:lines
+        let l:match = matchlist(l:line, l:pattern)
+
+        if len(l:match) == 0
+            continue
+        endif
+
+        call add(l:output, {
+        \   'bufnr': a:buffer,
+        \   'lnum': l:match[1] + 0,
+        \   'vcol': 0,
+        \   'col': -1,
+        \   'text': l:match[2],
+        \   'type': 'W',
+        \   'nr': -1,
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+call ale#linter#Define('markdown', {
+\   'name': 'mdl',
+\   'executable': 'mdl',
+\   'command': 'mdl',
+\   'callback': 'ale_linters#markdown#mdl#Handle'
+\})
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ name. That seems to be the fairest way to arrange this table.
 | JavaScript | [eslint](http://eslint.org/), [jscs](http://jscs.info/), [jshint](http://jshint.com/) |
 | JSON | [jsonlint](http://zaa.ch/jsonlint/) |
 | Lua | [luacheck](https://github.com/mpeterv/luacheck) |
+| Markdown | [mdl](https://github.com/mivok/markdownlint) |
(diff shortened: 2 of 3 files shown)


ale_linters/markdown/mdl.vim
```diff
@@ -0,0 +1,41 @@
...
+
+if exists('g:loaded_ale_linters_markdown_mdl')
+    finish
+endif
+
+let g:loaded_ale_linters_markdown_mdl = 1
+
+function! ale_linters#markdown#mdl#Handle(buffer, lines) abort
+    " matches: '(stdin):173: MD004 Unordered list style'
+    let pattern = ':\(\d*\): \(.*\)$'
```
mshr-h: I think it's better to be explicitly scoped variable like this.
```vim
let l:pattern = ':\(\d*\): \(.*\)$'
```

ale_linters/markdown/mdl.vim
```diff
@@ -0,0 +1,41 @@
...
+    let output = []
```
mshr-h: Same as the above.

ale_linters/markdown/mdl.vim
```diff
@@ -0,0 +1,41 @@
...
+
+    for line in a:lines
+        let match = matchlist(line, pattern)
```
mshr-h: Same as the above.

neersighted:
This format is close enough to the Unix format that you can use an existing handler, such as `ale#handler#HandleUnixFormatAsError`.

ale_linters/markdown/mdl.vim
w0rp: I found the problem with the build. Piping the errors through `sed` to fix the paths was messing up the error codes. I switched the Makefile to using Bash, and used the `pipefail` option to fix that. If the pull request was rebased again, it should fail the build now.

w0rp:
The function `ale#handler#HandleUnixFormatAsWarning` could be used here I think, as it looks like you want to output warnings.

I haven't used this linter myself, but I might start using it. I could use something to check my Markdown syntax. It's cool that you're adding support for it.

sbdchd:
@mshr-h by default, variables in functions are [locally scoped](https://neovim.io/doc/user/eval.html#local-variable).

Also, I tried using `ale#handler#HandleUnixFormatAsWarning` as the callback, but ale didn't return any errors.

neersighted:
@sbdchd It is more a matter of style. In this repo we use locally scoped variables, as suggested by `vint -s`. I think coding style hasn't made it into the `CONTRIBUTING` file yet.

ale_linters/markdown/mdl.vim
```diff
@@ -0,0 +1,41 @@
...
+
+        if len(l:match) == 0
+            continue
+        endif
+
+        call add(l:output, {
+        \   'bufnr': a:buffer,
+        \   'lnum': l:match[1] + 0,
+        \   'vcol': 0,
+        \   'col': -1,
```
w0rp: I just noticed that the column number here is -1. Could you set that to 0 or 1 instead? I don't think the plugin expect negative column numbers anywhere.

ale_linters/markdown/mdl.vim
```diff
@@ -0,0 +1,41 @@
+" Author: Steve Dignam <steve@dignam.xyz>
+" Description: Support for mdl, a markdown linter
+
+if exists('g:loaded_ale_linters_markdown_mdl')
```
neersighted: One other note, we don't use header guards in linters anymore. See #138

w0rp:
I have merged this via the command line, after changing that `col` value to `0`.

neersighted:
@w0rp You can actually edit the PR yourself now by default.

PR #140, opened 2016-10-22
Clean up some test filenames and extraneous code

neersighted:


neersighted:
Yeah, state is still cleaned up. Vader has includes, so we could make cleanup generic and include it in each test.

w0rp:
Yeah, the includes might be the way to go. If the state is still cleaned up, then this is fine with me. :+1:

Issue #40, opened 2016-10-04
What about java?


ehzawad:
Is there any way integrate java's built in javac ?

w0rp:
Yeah, I think there is. I think that can be supported. If anyone else wants to give it a go, go for it. Otherwise, I'll probably get to it eventually.

w0rp:
I had a quick look at this. The best bet to support checking Java files with `javac` is to modify the Bash wrapper script so it creates temporary directories, and create a .java file with the same name in the directory, and then write some support for automatically adding some directories to the classpath. It will be a bit tricky, but probably doable.

ehzShelter:
Bit tricky.. waiting for stable full support anyway.. :+1:

ehzShelter:
I am confused about $CLASSPATH finding of javac... Because in Syntastic plugin that w as an issue.. Javac cannot load user define java package properly...

neersighted:
CLASSPATH is required because ale does not run against files on disk, it runs on buffers in Vim's memory.

w0rp:
I'd wait until the open pull request for rewriting most of the stdin-wrapper script is done. After that is over, and the Windows script is re-written in PowerShell, I can go one step further and mirror the filename of the .java file exactly in the temporary directory which is created, and let it dump the .class file in there, which will be deleted. Then it will be just a matter of detecting the CLASSPATH and running `javac`.

ehzShelter:
After installing this plugin, I have already removed old caboose Vim-Syntastic Plugin. I am noob in vimscript, Otherwise I would try to contribute..Anyway, Waiting for javac support...:)

w0rp:
Very basic `javac` support has been added now. You can configure the `CLASSPATH` with some options, but ALE will currently not do anything clever to figure out how to set up the `CLASSPATH` automatically. I'll let a few people try it out and raise some issues for suggestions about how to automatically determine paths for Java.

notaduck:
@w0rp, I'm using the algs4 package from princeton and compiles the document with javac-algs4, is there any way to use this instead of the standard javac?

w0rp:
At the moment, no, but support for that isn't too hard to add. I'll open an issue for that.

w0rp:
#1476 There's an issue for tracking that.

notaduck:
Sry! I just noticed I could add the .jar file to the classpath

w0rp:
Okay, sounds good.

PR #141, opened 2016-10-23
Added support for javac (with eclipse classpath support for now)

Farenjihn:
This is my first real try at VimL so I am unaware of the quality of this changeset.

I had to do a few things to get javac to work with ALE, mainly:
1. javac doesn't support compiling from stdin so I clone the current buffer into /tmp/java_ale and tell javac to generate *.class files there by setting the appropriate classpath.
2. I clean the files left in /tmp/ every time a buffer with the ".java" extension is closed. If there is a better behaviour to have about this please tell me so I can change it.
3. For now I only check for a ".classpath" file in the current working directory and parse it using a few lines of python to create the final javac command. I can try and add maven/other tools support later in.
4. I parse the classpath file every time a buffer with the ".java" extension is opened. This is because it might have changed in between if the project the user is working on is not stable. If this seems inappropriate tell me what it should do.

All in all this is me trying to contribute to this awesome plugin. Hope it can help!

### DIFF
--- a/ale_linters/java/javac.vim
+++ b/ale_linters/java/javac.vim
@@ -9,6 +9,7 @@ let g:loaded_ale_linters_java_javac = 1
 let g:ale_java_javac_classpath = ''
 
 let s:eclipse_classpath = ''
+let s:tmppath = '/tmp/java_ale/'
 
 function! ale_linters#java#javac#Handle(buffer, lines) abort
     " Look for lines like the following.
@@ -69,7 +70,7 @@ endif
 endfunction
 
 function! ale_linters#java#javac#GetCommand(buffer)
-    let l:path = '/tmp/java_ale/' . expand('%:p:h')
+    let l:path = s:tmppath . expand('%:p:h')
     let l:file = expand('%:t')
     let l:buf = getline(1, '$')
 
@@ -85,7 +86,7 @@ function! ale_linters#java#javac#GetCommand(buffer)
 endfunction
 
 function! ale_linters#java#javac#CleanupTmp()
-    call delete('/tmp/java_ale/', 'rf')
+    call delete(s:tmppath, 'rf')
 endfunction
 
 autocmd! BufEnter *.java call ale_linters#java#javac#CheckEclipseClasspath()
--- a/ale_linters/java/javac.vim
+++ b/ale_linters/java/javac.vim
@@ -1,0 +1,101 @@
+" Author: farenjihn <farenjihn@gmail.com>
+" Description: Lints java files using javac
+
+if exists('g:loaded_ale_linters_java_javac')
+    finish
+endif
+
+let g:loaded_ale_linters_java_javac = 1
+let g:ale_java_javac_classpath = ''
+
+let s:eclipse_classpath = ''
+
+function! ale_linters#java#javac#Handle(buffer, lines) abort
+    " Look for lines like the following.
+    "
+    " Main.java:13: warning: [deprecation] donaught() in Testclass has been deprecated
+    " Main.java:16: error: ';' expected
+
+    let l:pattern = '^.*\:\(\d\+\):\ \(.*\):\(.*\)$'
+    let l:output = []
+
+    for l:line in a:lines
+        let l:match = matchlist(l:line, l:pattern)
+
+        if len(l:match) == 0
+            continue
+        endif
+
+        call add(l:output, {
+            \   'bufnr': a:buffer,
+            \   'lnum': l:match[1] + 0,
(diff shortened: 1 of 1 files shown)


ale_linters/java/javac.vim
```diff
@@ -0,0 +1,106 @@
+" Author: farenjihn <farenjihn@gmail.com>
+" Description: Lints java files using javac
+
+if exists('g:loaded_ale_linters_java_javac')
```
neersighted: We don't use header guards any more, as of a recent PR.

ale_linters/java/javac.vim
```diff
@@ -0,0 +1,106 @@
...
+            \   'type': l:match[2] ==# 'error' ? 'E' : 'W',
+            \   'nr': -1,
+            \})
+    endfor
+
+    return l:output
+endfunction
+
+function! ale_linters#java#javac#ParseEclipseClasspath()
+let l:eclipse_classpath = ''
```
neersighted: Indentation here would be good.

ale_linters/java/javac.vim
```diff
@@ -0,0 +1,106 @@
...
+
+if exists('g:loaded_ale_linters_java_javac')
+    finish
+endif
+
+let g:loaded_ale_linters_java_javac = 1
+let g:ale_java_javac_classpath = ''
+
+let s:eclipse_classpath = ''
+let s:tmppath = '/tmp/java_ale/'
```
neersighted: We need to be compatible with windows as well.

I think instead of doing your file magic, we should rewrite the stdin-wrapper to use the same filename as the current filename.

neersighted:
Looks good, especially for a first pass. Most of the issues I see are just formatting. However, I think a more elegant way to fix this would be making the stdin-wrapper use the original filename. I am working on a PR to that affect now.

neersighted:
Just linking to #40

Farenjihn:
The wrapper script mentioned in https://github.com/w0rp/ale/issues/40 does indeed sound better. I'll wait for your PR before changing this one to fit with the new wrapper.

w0rp:
We should first get using the original filename in the temporary directory in, and an updated script for wrapping commands for Windows.

The auotcmd for the classpath won't quite work out when you're editing multiple files, and there'll be too much work for Vim to do there. I think we should first implement the command chaining I have described in a recent issue, and use that.

w0rp:
You just merged the master branch for your pull request. You should use `git rebase` for pull requests instead, and never `git merge`. If you use `git reset --hard <commit_sha_before_merged_commits>` then `git rebase upstream/master`, then `git push -f` you can delete the merged commits, rebase, and replace your remote copy of the branch to fix the issue.

Farenjihn:
There we go. I appreciate your support and patience, really helps out. Thank you.

ale_linters/java/javac.vim
hauleth: @neersighted this shouldn't be hard with current new `stdin-wrapper`.

neersighted: Yeah, the current one will work. I still want to rewrite it to use the correct filename but `anything.class` should work for now.

w0rp: That's the plan. We need to first write a PowerShell script to replace the executable for Windows, so it writes to a temporary directory too, and then replace the file extension arguments with just the filename itself. I'd suggest replacing all of the concatenation with `ale#util#stdin_wrapper` with a `wrap_stdin` argument for the linters.

gabro:
Hey there, is there any news about this?

w0rp:
The Windows and Unix stdin-wrapper scripts behave largely the same now. Now they both need to be modified to use the whole filename instead of just the file extension. After that, this linter might work.

w0rp:
I never forgot about this one. Now the mechanisms are finally in place to make this work. I'll merge this to another branch now and finish it off.

w0rp:
I took away the Eclipse classpath support for now.

I'm not sure how useful it is to parse an Eclipse `.classpath` file if you're editing Java files in Vim. Eclipse will be needed to update the `.classpath` file.

My plan now is to just to get `javac` to run with no detection of the class path, let someone who starts using it get annoyed that it doesn't detect a classpath, open a GitHub issue proposing some kind of solution, and then go from there.

dhinus:
> My plan now is to just to get javac to run with no detection of the class path, let someone who starts using it get annoyed that it doesn't detect a classpath, open a GitHub issue proposing some kind of solution, and then go from there.

🖐 /me raises hand and opens https://github.com/w0rp/ale/issues/361

PR #142, opened 2016-10-23
Add node_modules support for JSHint, and use the global config as a fallback

w0rp:
This pull request will apply the same basic logic I just applied to `eslint` to `jshint` as well, for using a local `node_modules` version. In addition, it will now use the global JSHint configuration path only as a default when `.jshintrc` cannot be found, rather than always explicitly overriding the local configuration file.

I'll leave this up for others to take a look at.

@bartlibert Does this look good to you? I believe you are using ALE with a JSHint configuration file.


w0rp:
I feel pretty good about this, so I'll merge it.

bartlibert:
Just to let you know, it seems to work as expected with this integrated.

PR #143, opened 2016-10-23
Stop jobs when buffers close

### LINKED ISSUE #107: Jobs should be tracked seperately
Getting an error on the latest master:
```
Error detected while processing function <SNR>105_HandleExitNeoVim[1]..<SNR>105_HandleExit:
line   29:
E716: Key not present in Dictionary: 1
E15: Invalid expression: g:ale_buffer_should_reset_map[l:buffer]
```

w0rp:
This pull request closes #107.

I rewrote how jobs and other information are stored so that buffers can be mapped back to still-running jobs, so those jobs can be cancelled when a buffer is closed. There are now only two global variables for tracking the work ALE does, `g:ale_buffer_loclist_map`, which holds the loclist data, and is documented for public use, and `g:ale_buffer_info` which holds everything else. Some job information is still private to the `engine.vim` file, as it doesn't need to be visible elsewhere.

I also added a new test for checking that running a linter will set the signs appropriately. Tests are the only way to go for preventing regressions. I got rid of some `Then` blocks from some tests, because it's just much easier to write cleanup for tests without `Then`, as `After` runs before `Then`. I'm considering just politely banning the `Then` blocks. They look nice, but just don't work as well as putting your assertions on the end of `Execute` blocks.

@neersighted Take a look at this. I'll use this branch at work to test this myself, and leave this PR open for a little bit.

### DIFF
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -26,6 +26,7 @@ function! ale#engine#InitBufferInfo(buffer) abort
         \   'job_list': [],
         \   'should_reset': 1,
         \   'dummy_sign_set': 0,
+        \   'loclist': [],
         \}
     endif
 endfunction
@@ -121,28 +122,28 @@ function! s:HandleExit(job) abort
         " Set the flag for resetting the loclist to 0 again, so we won't
         " empty the list later.
         let g:ale_buffer_info[l:buffer].should_reset = 0
-        let g:ale_buffer_loclist_map[l:buffer] = []
+        let g:ale_buffer_info[l:buffer].loclist = []
     endif
 
     " Add the loclist items from the linter.
-    call extend(g:ale_buffer_loclist_map[l:buffer], l:linter_loclist)
+    call extend(g:ale_buffer_info[l:buffer].loclist, l:linter_loclist)
 
     " Sort the loclist again.
     " We need a sorted list so we can run a binary search against it
     " for efficient lookup of the messages in the cursor handler.
-    call sort(g:ale_buffer_loclist_map[l:buffer], 'ale#util#LocItemCompare')
+    call sort(g:ale_buffer_info[l:buffer].loclist, 'ale#util#LocItemCompare')
 
     if g:ale_set_loclist
-        call setloclist(0, g:ale_buffer_loclist_map[l:buffer])
+        call setloclist(0, g:ale_buffer_info[l:buffer].loclist)
     endif
 
     if g:ale_set_signs
-        call ale#sign#SetSigns(l:buffer, g:ale_buffer_loclist_map[l:buffer])
+        call ale#sign#SetSigns(l:buffer, g:ale_buffer_info[l:buffer].loclist)
     endif
 
     if exists('*ale#statusline#Update')
         " Don't load/run if not already loaded.
-        call ale#statusline#Update(l:buffer, g:ale_buffer_loclist_map[l:buffer])
+        call ale#statusline#Update(l:buffer, g:ale_buffer_info[l:buffer].loclist)
     endif
 
     " Call user autocommands. This allows users to hook into ALE's lint cycle.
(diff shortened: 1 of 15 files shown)


w0rp:
I should also mention, one of the effects of this is that it should be possible to run two processes for a single linter across two buffers at the same time now.

neersighted:
As the author of Vader has expressed support for moving after to be truely after everything, how about a PR against Vader? https://github.com/junegunn/vader.vim/blob/master/autoload/vader.vim#L284

autoload/ale/engine.vim
```diff
@@ -71,17 +100,11 @@ function! s:HandleExit(job) abort
...
-
     let l:linter = l:job_info.linter
     let l:output = l:job_info.output
     let l:buffer = l:job_info.buffer
 
-    if !has_key(g:ale_buffer_should_reset_map, l:buffer)
-        " A job ended for a buffer which has been closed, so stop here.
-        return
-    endif
+    call s:StopPreviousJobs(l:buffer, l:linter)
```
neersighted: A comment here might be helpful. Why do we stop jobs here?

autoload/ale/engine.vim
```diff
@@ -270,8 +294,8 @@ function! ale#engine#WaitForJobs(deadline) abort
 
     let l:job_list = []
 
-    for l:job_id in keys(s:job_info_map)
-        call add(l:job_list, s:job_info_map[l:job_id].linter.job)
+    for l:info in values(g:ale_buffer_info)
```
neersighted: Comments here as well could make this more obvious.

autoload/ale/statusline.vim
```diff
@@ -14,34 +14,48 @@ function! ale#statusline#Update(buffer, loclist) abort
...
 
-    let g:ale_buffer_count_map[a:buffer] = [l:errors, l:warnings]
+    let g:ale_buffer_info[a:buffer].count = [l:errors, l:warnings]
 endfunction
 
-" Returns a tuple of errors and warnings for use in third-party integrations.
-function! ale#statusline#Count(buffer) abort
+" Set the error and warning counts, calling for an update only if needed.
+" If counts cannot be set, return 0.
+function! s:SetupCount(buffer) abort
```
neersighted: I like this simplification a lot.

autoload/ale/statusline.vim
```diff
@@ -14,34 +14,48 @@ function! ale#statusline#Update(buffer, loclist) abort
...
+    if !has_key(g:ale_buffer_info, a:buffer)
+        " Linters have not been run for the buffer yet, so stop here.
+        return 0
+    endif
+
     " Cache is cold, so manually ask for an update.
-    if !has_key(g:ale_buffer_count_map, a:buffer)
+    if !has_key(g:ale_buffer_info[a:buffer], 'count')
         call ale#statusline#Update(a:buffer, get(g:ale_buffer_loclist_map, a:buffer, []))
```
neersighted: I like the simplification in this function a lot. This is somewhat unrelated, but is there anyway we can decouple from g:ale_buffer_loclist_map in this file?

neersighted:
junegunn/vader.vim#97

autoload/ale/statusline.vim
w0rp: I'm considering removing that global entirely. I'm in two minds about it. It's kind of an easy format to read in case anyone wants to use it for something or other. Then again, maybe it's only useful in this plugin itself.

autoload/ale/engine.vim
w0rp: Sure.

w0rp:
Maybe we could get rid of the buffer loclist global like you say. I could roll that into the new global map, then offer a function for getting just that list from before out again. I might do that tomorrow.

w0rp:
@neersighted Check this out. I have merged everything into `g:ale_buffer_info` now and I have added a function for getting the loclist for other plugins to use, if they want.

w0rp:
I'm happy enough with my testing now to put this into `master`.

PR #145, opened 2016-10-24
Add support for mlint, a MATLAB linter

awlayton:
I had to redirect stderr to stdout, so this probably doesn't work in Windows. Let  me know if I should do it a different way. I looked some and didn't see current support for any other linters that output to stderr.

### DIFF
--- a/ale_linters/matlab/mlint.vim
+++ b/ale_linters/matlab/mlint.vim
@@ -1,8 +1,6 @@
 " Author: awlayton <alex@layton.in>
 " Description: mlint for MATLAB files
 
-let g:loaded_ale_linters_matlab_mlint = 1
-
 let g:ale_matlab_mlint_executable =
 \   get(g:, 'ale_matlab_mlint_executable', 'mlint')
 
--- a/ale_linters/matlab/mlint.vim
+++ b/ale_linters/matlab/mlint.vim
@@ -1,0 +1,59 @@
+" Author: awlayton <alex@layton.in>
+" Description: mlint for MATLAB files
+
+if exists('g:loaded_ale_linters_matlab_mlint')
+    finish
+endif
+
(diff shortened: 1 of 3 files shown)


ale_linters/matlab/mlint.vim
```diff
@@ -0,0 +1,59 @@
...
+            continue
+        endif
+
+        let l:lnum = l:match[1] + 0
+        let l:col = l:match[2] + 0
+        let l:code = l:match[3]
+        let l:text = l:match[4]
+
+        " Suppress erroneous waring about file name
+        if l:code ==# 'FNDEF'
```
w0rp: We might be able to turn this on in future. There is some support for copying the same filename to the temporary directory to come eventually.

ale_linters/matlab/mlint.vim
```diff
@@ -0,0 +1,59 @@
+" Author: awlayton <alex@layton.in>
+" Description: mlint for MATLAB files
+
+if exists('g:loaded_ale_linters_matlab_mlint')
```
w0rp: We don't need the `exists` guards any more. They are being removed to aid some testing, and we should only be loading these files at most once per filetype anyway, I believe.

awlayton:
Thanks for pointing out the right way to handle stderr. I think I addressed all your changes @w0rp.

ale_linters/matlab/mlint.vim
```diff
@@ -0,0 +1,57 @@
+" Author: awlayton <alex@layton.in>
+" Description: mlint for MATLAB files
+
+let g:loaded_ale_linters_matlab_mlint = 1
```
w0rp: You left the "loaded" flag lying around.

w0rp:

PR #150, opened 2016-10-25
Fix ALELint autocommand example

neersighted:
`User` is required for User autocommands.


neersighted:
I see now that you updated the example, but it is still incorrect. The correct line is `autocmd User ALELint call myfunction()`. Mind manually editing that in so I can close this?

w0rp:
Sure, I'll do that.

PR #151, opened 2016-10-27
Add pylint formatter

keith:
This customizes the output format of pylint in order to use the same
pep8 parser

One thing to note here, pylint likes to output lots of comments or warnings, both of which use `--` in the gutter. There's a possibility we'd prefer the warnings to be displayed with `>>`, but I'll leave that up for debate here.

### DIFF
--- a/ale_linters/python/pylint.vim
+++ b/ale_linters/python/pylint.vim
@@ -1,0 +1,9 @@
+" Author: keith <k@keith.so>
+" Description: pylint for python files
+
+call ale#linter#Define('python', {
+\   'name': 'pylint',
+\   'executable': 'pylint',
+\   'command': g:ale#util#stdin_wrapper . ' .py pylint --output-format text --msg-template="{path}:{line}:{column}: {msg_id} {msg}" --reports n',
+\   'callback': 'ale#handlers#HandlePEP8Format',
+\})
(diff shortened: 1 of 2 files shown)


ale_linters/python/pylint.vim
```diff
@@ -0,0 +1,9 @@
+" Author: keith <k@keith.so>
+" Description: pylint for python files
+
+call ale#linter#Define('python', {
+\   'name': 'pylint',
+\   'executable': 'pylint',
+\   'command': 'pylint --output-format text --msg-template="{path}:{line}:{column}: {msg_id} {msg}" --reports n %s',
```
w0rp: I think there's a problem here. I think you're formatting in the filename, which means that the linter is running on the version of the file on disk, but not on the contents of the current buffer. You should try using `ale#util#stdin_wrapper`, which is used in some other linters.

keith: Ah ok, I was wondering why many linters were using stdin. I will update this. I don't _think_ we should need the wrapper here

keith: Ah yes we do, looking into this now.

keith:
I've updated this to use the stdin wrapper. One problem I had while doing that was temp files with `_` in their path, I had to add that to the regex for this to work. Also it seemed like `-` was consuming `:` characters until it was escaped as well.

w0rp:
Cool, I'll take this and add it to the README and help file after I look at some other pull requests.

keith:
Awesome, thank you!

dbrodie:
Would it be possible to add an args callback like exists for flake8 to allow for disabling certain warnings, etc...?

w0rp:
Yeah, you can open an issue for that and I or someone else will look at it eventually.

PR #153, opened 2016-10-27
Choose shell dialect based on vim syntax

keith:
Shellcheck is smart enough to check the shebang in a given file to
determine which dialect to use. Unfortunately this doesn't work for
files without shebangs, even if it might be apparent what dialect should
be used, such as "bashrc" or "foo.bash". Luckily `filetype.vim` defines
specific vars based on which shell dialect is being used based on a huge
list of conditions. With this change we take those into account for all
the types shellcheck supports, otherwise we fallback to letting it try
and decide.

For more context I also made this change to syntastic a while ago https://github.com/scrooloose/syntastic/pull/1766

### DIFF
--- a/ale_linters/sh/shellcheck.vim
+++ b/ale_linters/sh/shellcheck.vim
@@ -16,9 +16,25 @@ else
     let s:exclude_option = ''
 endif
 
+function! s:GetDialectArgument()
+    if exists('b:is_bash') && b:is_bash
+        return '-s bash'
+    elseif exists('b:is_sh') && b:is_sh
+        return '-s sh'
+    elseif exists('b:is_kornshell') && b:is_kornshell
+        return '-s ksh'
+    endif
+
+    return ''
+endfunction
+
+function! ale_linters#sh#shellcheck#GetCommand(buffer)
+  return 'shellcheck ' . s:exclude_option . ' ' . s:GetDialectArgument() . ' -f gcc -'
+endfunction
+
 call ale#linter#Define('sh', {
 \   'name': 'shellcheck',
 \   'executable': 'shellcheck',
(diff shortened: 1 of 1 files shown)


keith:
There were some concerns raised on the other PR that I'd be happy to fix if you're interested in going in this direction at all!

ale_linters/sh/shellcheck.vim
```diff
@@ -16,9 +16,21 @@ else
...
 endif
 
+function! s:GetDialectArgument()
+    if exists('b:is_bash') && b:is_bash
+        return '-s bash'
+    elseif exists('b:is_sh') && b:is_sh
+        return '-s sh'
+    elseif exists('b:is_kornshell') && b:is_kornshell
+        return '-s ksh'
+    endif
```
w0rp: You could probably include zsh here too. 

I didn't know about these special buffer variables.

Could you use `command_callback` instead below with a function which returns the command? Then you could re-evaluate the command each time ALE runs the linter. This here will only be evaluated once.

keith: I've updated this to evaluate multiple times. I haven't added zsh here since it seems like right now this isn't run for zsh files since they have the `zsh` `ft` and not `sh`

keith: Ah interesting. Well there isn't an internal var like this for zsh since it's expected to be a different filetype. I'll check the ft first instead

w0rp:
Just so you know, I've been seeing emails come through for this and other pull requests. I've been busy this weekend, but I will get to them eventually. I wish GitHub had a new post feature on it, or something.

keith:
No rush at all! Thanks!

w0rp:
Looks good to me. I'll take this. Cheers! :beers:

keith:
Thanks!!

PR #155, opened 2016-10-27
Added options variable for htmlhint

deathmaz:
Currently htmlhint has a bug which prevents htmlhint from using .htmlhintrc file
But it works if I add it as a param to command:
```viml
let g:ale_html_htmlhint_options = '--config ~/.htmlhintrc --format=unix stdin'
```

### DIFF
--- a/ale_linters/html/htmlhint.vim
+++ b/ale_linters/html/htmlhint.vim
@@ -2,11 +2,11 @@
 " Description: HTMLHint for checking html files
 
 " CLI options
-let g:ale_html_htmlhint_options = get(g:, 'ale_html_htmlhint_options', '--format=unix stdin')
+let g:ale_html_htmlhint_options = get(g:, 'ale_html_htmlhint_options', '--format=unix')
 
 call ale#linter#Define('html', {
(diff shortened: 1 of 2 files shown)


ale_linters/html/htmlhint.vim
```diff
@@ -1,9 +1,12 @@
...
 " Description: HTMLHint for checking html files
 
+" CLI options
+let g:ale_html_htmlhint_options = get(g:, 'ale_html_htmlhint_options', '--format=unix stdin')
+
 call ale#linter#Define('html', {
 \   'name': 'htmlhint',
 \   'executable': 'htmlhint',
-\   'command': 'htmlhint --format=unix stdin',
+\   'command': 'htmlhint ' . g:ale_html_htmlhint_options,
```
w0rp: Could you leave the `stdin` option here? The linter can't work without that.

deathmaz: done

deathmaz:
If you don't mind i'll try to prepare another PR to make it work similar to `jshint`

PR #157, opened 2016-10-30
Add FlowType support

zperrault:
Adds support for FlowType.

I still need to test that this is graceful when a project doesn't have a `.flowconfig` or `flow-bin` isn't installed.
# 152

### DIFF
--- a/ale_linters/javascript/flow.vim
+++ b/ale_linters/javascript/flow.vim
@@ -9,7 +9,7 @@ let g:ale_javascript_flow_use_global =
 
 function! ale_linters#javascript#flow#GetExecutable(buffer) abort
   if g:ale_javascript_flow_use_global
-    return g:ale_javascript_flow_use_global
+    return g:ale_javascript_flow_executable
   endif
 
   return ale#util#ResolveLocalPath(
@@ -27,7 +27,7 @@ endfunction
 function! ale_linters#javascript#flow#Handle(buffer, lines)
   let l:flow_output = json_decode(join(a:lines, ''))
 
-  if l:flow_output
+  if has_key(l:flow_output, 'errors')
     let l:output = []
 
     for l:error in l:flow_output.errors
--- a/ale_linters/javascript/flow.vim
+++ b/ale_linters/javascript/flow.vim
@@ -1,0 +1,62 @@
+" Author: Zach Perrault -- @zperrault
+" Description: FlowType checking for JavaScript files
+
+let g:ale_javascript_flow_executable =
+\   get(g:, 'ale_javascript_flow_executable', 'flow')
+
+let g:ale_javascript_flow_use_global =
+\   get(g:, 'ale_javascript_flow_use_global', 0)
+
+function! ale_linters#javascript#flow#GetExecutable(buffer) abort
+  if g:ale_javascript_flow_use_global
+    return g:ale_javascript_flow_use_global
+  endif
+
+  return ale#util#ResolveLocalPath(
+  \   a:buffer,
+  \   'node_modules/.bin/flow',
+  \   g:ale_javascript_flow_executable
+  \)
+endfunction
+
+function! ale_linters#javascript#flow#GetCommand(buffer) abort
+  return ale_linters#javascript#flow#GetExecutable(a:buffer)
+  \   . ' check-contents --json --from ale'
+endfunction
+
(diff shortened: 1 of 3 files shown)


doc/ale.txt
```diff
@@ -11,19 +11,20 @@ CONTENTS                                                         *ale-contents*
...
-    4.6.  cpp-gcc...............................|ale-linter-options-cpp-gcc|
-    4.7.  fortran-gcc...........................|ale-linter-options-fortran-gcc|
-    4.8.  shell.................................|ale-linter-options-shell|
-    4.9.  shellcheck............................|ale-linter-options-shellcheck|
-    4.10. vint..................................|ale-linter-options-vint|
-    4.11. luacheck..............................|ale-linter-options-luacheck|
-    4.12. c-cppcheck............................|ale-linter-options-c-cppcheck|
-    4.13. cpp-cppcheck..........................|ale-linter-options-cpp-cppcheck|
-    4.14. htmlhint..............................|ale-linter-options-htmlhint|
+    4.2.  flow..................................ale-linter-options-flow
```
neersighted: You need pipes here to create a link

ale_linters/javascript/flow.vim
```diff
@@ -0,0 +1,66 @@
...
+
+let g:ale_javascript_flow_executable =
+\   get(g:, 'ale_javascript_flow_executable', 'flow')
+
+let g:ale_javascript_flow_use_global =
+\   get(g:, 'ale_javascript_flow_use_global', 0)
+
+function! ale_linters#javascript#flow#GetExecutable(buffer) abort
+  if g:ale_javascript_flow_use_global
+    return g:ale_javascript_flow_use_global
```
w0rp: You're returning the wrong variable here by mistake.

ale_linters/javascript/flow.vim
```diff
@@ -0,0 +1,66 @@
...
+      " Each error is broken up into parts
+      let l:text = ''
+      let l:line = 0
+      for l:message in l:error.message
+        " Comments have no line of column information
+        if l:message.line + 0
+          let l:line = l:message.line + 0
+        endif
+        let l:text = l:text . ' ' . l:message.descr
+      endfor
```
w0rp: Are there multiple errors in this loop, or only one error split into an array for some reason? I wonder if you should maybe call `add(l:output, ...)` inside of this inner for loop with each error message.

zperrault: Each error has multiple messages. Some of those messages have the proper line numbers and others don't 😕 . I was creating a single message per error. Here's an example of the output I'm parsing.

```json
{
  "flowVersion":"0.33.0",
  "errors":[
    {
      "kind":"infer",
      "level":"error",
      "message":[
        {
          "context":"foo.forEach(() => null)",
          "descr":"property `forEach`",
          "type":"Blame",
          "loc":{
            "source":"-",
            "type":"SourceFile",
            "start":{"line":12,"column":5,"offset":96},
            "end":{"line":12,"column":11,"offset":103}
          },
          "path":"-",
          "line":12,
          "endline":12,
          "start":5,
          "end":11
        },
        {
          "context":null,
          "descr":"Property not found in",
          "type":"Comment",
          "path":"",
          "line":0,
          "endline":0,
          "start":1,
          "end":0
        },
        {
          "context":"foo.forEach(() => null)",
          "descr":"String",
          "type":"Blame",
          "loc":{
            "source":"-",
            "type":"SourceFile",
            "start":{"line":12,"column":1,"offset":92},
            "end":{"line":12,"column":3,"offset":95}
          },
          "path":"-",
          "line":12,
          "endline":12,
          "start":1,
          "end":3
        }
      ]
    }
  ],
  "passed":false
}
```

ale_linters/javascript/flow.vim
zperrault: Good catch.

ale_linters/javascript/flow.vim
w0rp: I see. I guess you get different parts of the same error? I wonder if you could maybe take the column number information from there.

zperrault: @w0rp Yeah I was playing with that. The tricky part is that the `Comment` type messages don't have any loc information. 😕

w0rp:
I think this should do for now. Maybe we can include column numbers in the future and we won't know what people think of the messages until we just get people to try it. Cheers! :beers:

PR #158, opened 2016-10-31
Add support for clang for C language

mshr-h:
As the title said, I've added [clang](http://clang.llvm.org/) for C language.


w0rp:
I see no problems here. Cheers! :beers:

PR #164, opened 2016-11-03
flow linting should only happen on flow typed files.

delianides:
flow linting lights up on files that don't have the `@flow` comment. Using the `respect-pragma` option forces flow to only check files with the `@flow` heading.

The `--respect-pragma` was introduction with Flow 0.30 (current is 0.34) so I believe you'd have to have that version for linting to work.


w0rp:
I'll take this. We might have to add a global for letting people configure the options. People may want different things.

PR #165, opened 2016-11-05
changed while read -r loop to cat > for portability and simplicity

purpleP:
Very simple change that I've tested with ghc linter.

### DIFF
--- a/stdin-wrapper
+++ b/stdin-wrapper
@@ -16,6 +16,6 @@ temp_dir=$(mktemp -d 2>/dev/null || mktemp -d -t 'ale_linter')
 temp_file="$temp_dir/file$file_extension"
 trap 'rm -r "$temp_dir"' EXIT
 
-cat >> "$temp_file"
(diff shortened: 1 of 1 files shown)


stdin-wrapper
```diff
@@ -16,8 +16,6 @@ temp_dir=$(mktemp -d 2>/dev/null || mktemp -d -t 'ale_linter')
 temp_file="$temp_dir/file$file_extension"
 trap 'rm -r "$temp_dir"' EXIT
 
-while read -r; do
-  echo "$REPLY" >> "$temp_file"
-done
+cat >> "$temp_file"
```
purpleP: Yes, one is enough. As to why this is more portable - http://unix.stackexchange.com/questions/321303/how-to-use-namedpipe-as-temporary-file
I'm just trusting this guy, because he obviously knows better than me.

w0rp:
Looks good to me. :+1:

Issue #20, opened 2016-10-02
Add checkers for TypeScript


prashcr:
Specifically, [tsc](https://github.com/Microsoft/TypeScript) (or maybe [tsuquyomi](https://github.com/Quramy/tsuquyomi)) and [tslint](https://github.com/palantir/tslint)
### Status
- [x] TSLint
- [ ] Tsuquyomi/tsc

w0rp:
I'm totally in favour of this. I think I'm generally in favour of adding more support for linting everything. I'll probably start using TypeScript personally soon enough.

prashcr:
Great!

### tslint
Style checking
Needs stdin wrapper.
Out of curiosity, what's the rationale behind requiring linters to accept stdin?

### tsc
Syntax and type checking
tsc is the official typescript compiler, however it's really slow (takes 2 seconds for a hello world).

TypeScript also comes with a tool called TSServer which uses some caching magic to run the same checks way faster, at the cost of a few hundred ms at startup. Aside from checking, it also adds intellisense autocomplete. This is what's used in Sublime/VS Code/WebStorm.

# tsuquyomi
This plugin wraps TSServer and makes it really easy to use.
However, this uses [vimproc.vim](https://github.com/Shougo/vimproc.vim), an async task runner, to populate the quickfix list.

Syntastic's approach to this problem was to add a syntastic checker in the tsuquyomi repo itself and add an option to disable tsuquyomi's use of the quickfix list.

I'm not sure how to approach this.
Do we just use tsuquyomi and accept that people need to install an async task runner when they're already using vim 8? And do we try and integrate changes into their repo so that we can feed its output directly into ALE?
On the other hand, directly wrapping TSServer also seems like too much work that shouldn't belong here.
Maybe we should fork tsuquyomi to use vim8 async?

w0rp:
I suppose we'd need to rewrite what `tsuquyomi` does to use the new Vim 8 functionality. It probably used vimproc because the new async functions weren't available at the time.

The linters need to accept stdin input because we need to be able to send the text for files which are open without first saving the files back to disk. The linting is applied by sending the contents of the text buffer directly to a given program. I might write a Bash wrapper script will could be used for sending input to most programs. Then eventually I might add a _Batch_ wrapper script for Windows, which will be... interesting.

w0rp:
I just remembered that I actually already wrote such a Bash wrapper script. I forgot that I did that.

prashcr:
[...]

PR #170, opened 2016-11-10
Add typecheck linter for typescript

mkusher:
Solves #20 

- Linters name: `typecheck`
- Required npm package: `ts-type-checker`(https://www.npmjs.com/package/ts-type-checker)

It uses my simple wrapper around typescript compiler . `npm i -g ts-type-checker` would be enough.
typecheck.vim is almost a copy of tslint.vim.

### DIFF
--- a/ale_linters/typescript/typecheck.vim
+++ b/ale_linters/typescript/typecheck.vim
@@ -12,8 +12,6 @@ function! ale_linters#typescript#typecheck#Handle(buffer, lines)
 
     for l:line in a:lines
         let l:match = matchlist(l:line, l:pattern)
-        echom len(l:match)
-        echom join(l:match, ", ")
 
         if len(l:match) == 0
             continue
--- a/ale_linters/typescript/typecheck.vim
+++ b/ale_linters/typescript/typecheck.vim
@@ -1,0 +1,47 @@
+" Author: Prashanth Chandra https://github.com/prashcr, Aleh Kashnikau https://github.com/mkusher
+" Description: type checker for TypeScript files
+
+function! ale_linters#typescript#typecheck#Handle(buffer, lines)
+    " Matches patterns like the following:
+    "
+    " hello.ts[7, 41]: Property 'a' does not exist on type 'A'
+    " hello.ts[16, 7]: Type 'A' is not assignable to type 'B'
+    "
+    let l:pattern = '.\+.ts\[\(\d\+\), \(\d\+\)\]: \(.\+\)'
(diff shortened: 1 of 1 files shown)


ale_linters/typescript/typecheck.vim
```diff
@@ -0,0 +1,47 @@
...
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+call ale#linter#Define('typescript', {
+\   'name': 'typecheck',
+\   'executable': 'typecheck',
+\   'command': 'typecheck %s',
```
w0rp: That will be running the command on the file as it is on disk, not the text you are editing as you type. You'll either need to use an option to use `stdin` data for the file, or use the stdin_wrapper file, probably with an argument to include the directory you are editing the file in or something similar.

ale_linters/typescript/typecheck.vim
```diff
@@ -0,0 +1,47 @@
...
+    " hello.ts[7, 41]: Property 'a' does not exist on type 'A'
+    " hello.ts[16, 7]: Type 'A' is not assignable to type 'B'
+    "
+    let l:pattern = '.\+.ts\[\(\d\+\), \(\d\+\)\]: \(.\+\)'
+    let l:output = []
+
+    for l:line in a:lines
+        let l:match = matchlist(l:line, l:pattern)
+        echom len(l:match)
+        echom join(l:match, ", ")
```
w0rp: You left some echo lines here.

mkusher: oh, my fault

ale_linters/typescript/typecheck.vim
mkusher: no, typecheck uses stdin by default

w0rp: Does it still use stdin if a filename is given? The `'%s'` here will be replaced with the filename.

mkusher: yes, typecheck needs file name to find tsconfig.json, but it will read file contents from stdin

w0rp: Ah, sounds good to me, then. Sounds similar to the `--stdin-filename` option for `eslint`, but as a positional argument.

w0rp:
I think this is good. I'll let Travis CI look at this. We might find that it could be a bit slow, but it's certainly better than nothing. If it delays other checks, you can always turn it off. If someone can implement something using HTTP and the test server, I'd also accept that.

PR #174, opened 2016-11-14
Add g:ale_python_flake8_executable

### LINKED ISSUE #172: Option to specify flake8 command line
AFAICT currently ALE hardcodes the executable as `flake8`.  I would like to sometimes use `python3 -m flake8` instead.

This is because `flake8` uses Python's standard library `ast` module, which only supports the same major Python version that you used to run `flake8`.   I use different Python versions for different projects, and I would like to specify which one I want with a vim autocommand.

mgedmin:
Closes #172.

### DIFF
--- a/ale_linters/python/flake8.vim
+++ b/ale_linters/python/flake8.vim
@@ -1,9 +1,21 @@
 " Author: w0rp <devw0rp@gmail.com>
 " Description: flake8 for python files
 
+let g:ale_python_flake8_executable =
+\   get(g:, 'ale_python_flake8_executable', 'flake8')
+
+function! ale_linters#python#flake8#GetExecutable(buffer) abort
+    return g:ale_python_flake8_executable
+endfunction
+
+function! ale_linters#python#flake8#GetCommand(buffer) abort
+    return ale_linters#python#flake8#GetExecutable(a:buffer)
+    \   . ' -'
+endfunction
+
 call ale#linter#Define('python', {
 \   'name': 'flake8',
-\   'executable': 'flake8',
-\   'command': 'flake8 -',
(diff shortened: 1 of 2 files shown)


mgedmin:
Actually, I'm testing this and I cannot get it to work -- I assume ALE looks for `python3 -m flake8` as an executable in $PATH, instead of looking just for `python3` :(

ale_linters/python/flake8.vim
```diff
@@ -1,9 +1,21 @@
...
+function! ale_linters#python#flake8#GetCommand(buffer) abort
+    return ale_linters#python#flake8#GetExecutable(a:buffer)
+    \   . ' -'
+endfunction
+
 call ale#linter#Define('python', {
 \   'name': 'flake8',
-\   'executable': 'flake8',
-\   'command': 'flake8 -',
+\   'executable_callback': 'ale_linters#python#flake8#GetExecutable',
```
w0rp: This will fail when the value for the executable is `'python3 -m flake8'`, as that's not a path to an executable file. I don't know a good answer for that at the moment.

mgedmin: Maybe `g:ale_python_flake8_executable="python"` and `g:ale_python_flake8_args="-m flake8 -"`?

w0rp:
Aha, you just spotted the same bug I did. I don't know a good answer at the moment. You could try parsing the first argument from the path, and quoting and spaces can be a bitch there.

mgedmin:
This version works for me.

w0rp:
I wonder if the the `-` argument should always be there, and not in the options. It is required for the plugin to work.

w0rp:
The tests passed, and I'm happy with this. Cheers! :beers:

Issue #173, opened 2016-11-14
Run linters in the directory of the file being linted


mgedmin:
`flake8` looks for project-specific settings in `setup.cfg` or `tox.ini`, found in the closest surrounding parent directory of the command line argument.  (If there are multiple arguments, `flake8` starts looking in the common shared parent directory.)

This logic does not work when ale runs `flake8 -` instead of `flake8 path/to/file/in/question.py`.  In practical terms, this is only painful when the `setup.cfg` you want to be used resides in a subdirectory of the current working directory.  Example:

```
- (current working directory)
    - package1/
        - setup.cfg
        - src/
            - file.py
    - package2/
        - setup.cfg
        - src/
            - another_file.py
```

When I'm editing `./package1/src/file.py`, I'd like `flake8` to pick up the project-specific settings from `./package1/setup.cfg`.

I think the simplest/most generic way of achieving this is to run the `flake8 -` process with the working directory set to the directory of the file I'm editing, i.e. `./package1/src/` in this case.

I don't see a way to ask vim's `job_start()` to run the job in a different working directory, so this would have to be done by changing the command itself to run `cd <directory> && <original-command>`.

I don't know if this should be done only for flake8, or if this is something that would be useful for other linters as well.  Maybe start cautiously and only do this for `flake8`?

Things to watch out for:
- the directory of the file might not exist (if I `:e new-subdir/file.py` and start typing)
- the linter command may be relative to $PWD (e.g. `./node_modules/.bin/eslint` or `./.venv/bin/flake8`), and blindly prepending `cd %:p:h && ` could break that.

w0rp:
The solution is easy. Find the `setup.cfg` file ourselves, and pass it via the command line. It's easier for `eslint`, which lets you set a filename in addition to taking input from stdin.

mgedmin:
The logic that flake8 uses to find the right config file is [a bit convoluted](https://gitlab.com/pycqa/flake8/blob/master/src/flake8/options/config.py) (AFAICT it looks for all `setup.cfg`, `tox.ini` and `.flake8` files in all parent directories, then merges the options from all of them, in a particular order).  I'm not sure whether it's possible to emulate that with a command line argument...

w0rp:
[...]

PR #176, opened 2016-11-20
Flow linter improvements

andreypopp:
Related #173

### DIFF
--- a/ale_linters/javascript/flow.vim
+++ b/ale_linters/javascript/flow.vim
@@ -25,7 +25,11 @@ function! ale_linters#javascript#flow#GetCommand(buffer) abort
 endfunction
 
 function! ale_linters#javascript#flow#Handle(buffer, lines)
-  let l:flow_output = json_decode(join(a:lines, ''))
+  let l:str = join(a:lines, '')
+  if l:str == ''
+    return []
+  endif
+  let l:flow_output = json_decode(l:str)
 
   if has_key(l:flow_output, 'errors')
     let l:output = []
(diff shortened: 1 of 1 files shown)


andreypopp:
Added some other improvements for Flow linter here.

ale_linters/javascript/flow.vim
```diff
@@ -21,11 +21,15 @@ endfunction
 
 function! ale_linters#javascript#flow#GetCommand(buffer) abort
   return ale_linters#javascript#flow#GetExecutable(a:buffer)
-  \   . ' check-contents --respect-pragma --json --from ale'
+  \   . ' check-contents --respect-pragma --json --from ale %s'
```
w0rp: Will this not run the linter on the file, instead of the buffer data? That would make it so you'll be getting the errors for the file as it is on disk, not as you are editing it.

andreypopp: This will still use stdin as input:
```
Usage: flow check-contents [OPTION]... [FILE]

Runs a flow check on the contents of stdin. If FILE is provided, then
check-contents pretends that the contents of stdin come from FILE

e.g. flow check-contents < foo.js
or   flow check-contents foo.js < foo.js
```

w0rp: Ahh, good to know! That makes matters much easier than. This behaviour is similar to the behaviour of `--stdin-filename` for `eslint`.

andreypopp: Right!

w0rp:
Given what you've said above, this looks pretty good to me. Thank you for fixing this bug. :+1:

PR #177, opened 2016-11-20
Add ocaml linter: merlin

andreypopp:


ale_linters/ocaml/merlin.vim
```diff
@@ -0,0 +1,15 @@
...
+
+function! ale_linters#ocaml#merlin#Handle(buffer, lines)
+  let l:errors = merlin#ErrorLocList()
+  return l:errors
+endfunction
+
+call ale#linter#Define('ocaml', {
+\   'name': 'merlin',
+\   'executable': 'ocamlmerlin',
+\   'command': 'true',
```
andreypopp: So in this case we read an error list directly from merlin (which keeps a connection to merlin server and uses it to fetch errors). This is why we use `true` command. Hacky... but works.

w0rp:
Are there any notes to make on which other plugin is required for this to work? You should probably only define this linter if the other plugin is installed. Then you can add documentation to inform users that they have to install another plugin for this to work, etc.

andreypopp:
@w0rp good point, I've added the check

ale_linters/ocaml/merlin.vim
```diff
@@ -0,0 +1,20 @@
+" Author: Andrey Popp -- @andreypopp
+" Description: Report errors in OCaml code with Merlin
+
+if !exists('g:merlin')
+  echo 'Error: merlin vim bindings are required for Ale merlin linter to work'
```
w0rp: That error might be a bit too annoying, as it will keep appearing until someone installs the other plugin, which they might not want. I'd put this in the documentation somewhere. Maybe something like `(see |ale-merlin-integration|)` in the Vim documentation and something similar in the README file, maybe just tell people to type `:help ale-merlin-integration` if you want to avoid writing the same text twice.

andreypopp: Am I right that you want me to keep `echo` statement but point users to ale's doc instead of the current message there?

andreypopp:
Updated pull request with docs and removed error message.

doc/ale.txt
```diff
@@ -26,10 +26,13 @@ CONTENTS                                                         *ale-contents*
     4.14. cpp-cppcheck..........................|ale-linter-options-cpp-cppcheck|
     4.15. htmlhint..............................|ale-linter-options-htmlhint|
     4.16. c-clang...............................|ale-linter-options-c-clang|
-  5. Commands/Keybinds..........................|ale-commands|
-  6. API........................................|ale-api|
-  7. Special Thanks.............................|ale-special-thanks|
-  8. Contact....................................|ale-contact|
+    4.17. python-flake8.........................|ale-linter-options-python-flake8|
```
andreypopp: Also this was missed from ToC.

andreypopp:
Friendly ping @w0rp, anything I can improve here?

w0rp:
The only thing I can think that's missing is an entry for the new linter with a documentation link to the new help section at the top of `doc/ale.txt`, like in the README. I'll merge it after you add that.

andreypopp:
Ok, done.

w0rp:
Looks good to me. :+1:

PR #180, opened 2016-11-24
add go build for build errors

dzhou121:
The current go vet, gofmt, go lint don't give the actual build errors. So go build is added to provide that.


andreypopp:
@dzhou121 awesome! please add info about it to [docs](https://github.com/w0rp/ale/blob/master/doc/ale.txt) and README.

w0rp:
I don't want to add another wrapper script to the repo. It won't work on Windows. Is there same way to run the commands from Vim instead?

renstrom:
How about using [`gotype`](https://godoc.org/golang.org/x/tools/cmd/gotype) instead of `go build`? It even supports stdin. However it might be even better to take it one step further and [`gometalinter`](https://github.com/alecthomas/gometalinter). Something like this:

```vim
let g:ale_go_gometalinter_args = get(g:, 'ale_go_gometalinter_args',
\   '--fast --disable=golint --disable=govet')

let s:format = '--format="{{.Path}}:{{.Line}}:{{if .Col}}{{.Col}}{{else}}0{{end}}: {{.Severity}}: {{.Message}} ({{.Linter}})"'

call ale#linter#Define('go', {
\   'name': 'gometalinter',
\   'output_stream': 'stdout',
\   'executable': 'gometalinter',
\   'command': 'gometalinter ' . s:format . ' ' . g:ale_go_gometalinter_args,
\   'callback': 'ale#handlers#HandleGCCFormat',
\})
```

w0rp:
@renstrom The `gotype` and `gometalinter` suggestions sound good to me. Do you want to open a pull request for those? I'll accept just about any pull request for linters if the code looks sane to me.

dzhou121:
@w0rp @renstrom 

gotype still work at package level. This is how Go works. If you have multiple files for one package, you have to provide all the files to gobuild or gotype for it to work properly.

E.g. if you defined a struct in file_1.go and used it in file_2.go, if you do 

$> go build file_2.go

It will report variable not defined. And you have to do

$> go build file_1.go file_2.go

renstrom:
Hmm I admit this isn't super trivial. But similar to Syntastic, can't we just `cd` to the project and run `go build` (ignoring `stdin`) and collect the output? This only works when saving though.

w0rp:
Can you specific the project path via an argument to `go build`? That would probably be easier. You might be able to use `cd` if you write `cd ... && ...` in the command.

renstrom:
Well, there's two ways to run `go build`. Either just `cd` to the project and run, or pass an import path. 

```bash
$ cd stuff/src/projects/github.com/w0rp/ale && go build
```

```bash
# This works relative to $GOPATH
$ go build github.com/w0rp/ale
```

The latter is a bit harder I guess, since it can be messy to figure out the import path. But perhaps more important, not all Go files are located such that they can be imported anyway.

dzhou121:
One important feature of ale is the ability to check on the fly. If you build the folder, you can only check after saving the file.

renstrom:
>One important feature of ale is the ability to check on the fly. If you build the folder, you can only check after saving the file.

Unfortunately I think that's just something we have to accept. I don't think it's feasible to copy an entire Go project to some temp path and substitute the current buffer with the correct file, setup `$GOPATH` etc and *then* run `go build`. Or is it...?

dzhou121:
You don't need to copy over the whole project. The files that belongs to the same package are enough.

dzhou121:
@w0rp I've reverted back to simple go build. Can this be merged now?

w0rp:
Yeah, that looks okay to me.

PR #183, opened 2016-11-24
Add puppet linters

### LINKED ISSUE #181: How about support for Puppet?
The two main validation tools for Puppet are `puppet-lint` and `puppet parser validate`, would be nice if ALE supported them too.

ace13:
Adds support for Puppet through `puppet parser validate` and `puppet-lint`.

This currently doesn't let `puppet-lint` do checks for auto-load layout, due to the way the stdin-wrapper works.

Closes #181


w0rp:
I see no issue here. Cheers! :beers:

PR #190, opened 2016-11-26
LaTeX Linters

abalmos:
Adds support for `chktex` and `lacheck`. 

`chktex`: 

Supports finding a project's `.chktexrc` file. It uses `ale#util#FindNearestFile` even though that is _not_ the search algorithm `chktex` uses. However, I find this to be the least surprising implementation for users that have a directory structure within their LaTeX project. For example, normal latex practice is to run `latex`, `lacheck`, etc. on the top most tex file in the root of the project. These programs then typically follow `\input{}` commands into the other files. As a result `chktex` expects `.chktexrc` to be in the current working directory. However, when editing a tex file from a sub-directory ale somewhat _incorrectly_ calls `chktex` (at least with respect to how `chktex` is usually called). Therefore, we look up the project tree for a `.chktexrc`.

`lacheck`:

`lacheck` has the same problem but it does not have a configuration file so I just throw away lint errors that are not for the file in the buffer -- see viml comments

### DIFF
--- a/ale_linters/tex/lacheck.vim
+++ b/ale_linters/tex/lacheck.vim
@@ -10,8 +10,7 @@ function! ale_linters#tex#lacheck#Handle(buffer, lines) abort
   " "book.tex", line 37: possible unwanted space at "{"
   " "book.tex", line 38: missing `\ ' after "etc."
 
-  " stdin-wrapper makes files like: /tmp/tmp.0hV8ww029I/file.tex
-  let l:pattern = '^".\+/file.tex", line \(\d\+\): \(.\+\)$'
+  let l:pattern = '^".\+", line \(\d\+\): \(.\+\)$'
   let l:output = []
 
   for l:line in a:lines
--- a/ale_linters/tex/lacheck.vim
+++ b/ale_linters/tex/lacheck.vim
@@ -1,0 +1,51 @@
+" Author: Andrew Balmos - <andrew@balmos.org>
+" Description: lacheck for LaTeX files
+
+let g:ale_tex_lacheck_executable =
+\   get(g:, 'ale_tex_lacheck_executable', 'lacheck')
+
+function! ale_linters#tex#lacheck#Handle(buffer, lines) abort
+  " Mattes lines like:
+  "
+  " "book.tex", line 37: possible unwanted space at "{"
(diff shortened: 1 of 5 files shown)


poohzrn:
@abalmos I have some huge tex projects and I have not managed to get this PR to find any errors
I have a structure like this: 
```
projectdir
----master.tex
----sectionsdir
    ----section1dir
         ----section1.tex
    ----section2dir
         ----section2.tex
```
If I run Ale on section1.tex I get no errors/ warnings

abalmos:
@poohzrn That's strange. A colleague of mine and I have been using this for a few weeks without issue for both `chktex` and `lacheck`.

Are you sure `chktex` and `lacheck` are on your path? If not, try setting `g:ale_tex_chktex_executable` and `g:ale_tex_lacheck_executable`.

I assume that you have lint errors for those files when you run `chktex` and `lacheck` by hand?

Is vim identifying the buffer filetype as `tex` or `plaintex`?

poohzrn:
Yeah fuck me, it works, I was just too tired.
I was on the wrong branch........

ale_linters/tex/lacheck.vim
```diff
@@ -0,0 +1,51 @@
...
+\   get(g:, 'ale_tex_lacheck_executable', 'lacheck')
+
+function! ale_linters#tex#lacheck#Handle(buffer, lines) abort
+  " Mattes lines like:
+  "
+  " "book.tex", line 37: possible unwanted space at "{"
+  " "book.tex", line 38: missing `\ ' after "etc."
+
+  " stdin-wrapper makes files like: /tmp/tmp.0hV8ww029I/file.tex
+  let l:pattern = '^".\+/file.tex", line \(\d\+\): \(.\+\)$'
```
w0rp: Could you use a bit more of a liberal regex here in terms of the filename? At some point in the future, I will make the basename of the files match the original filenames, so when I finally get that in, it will break this code here with this pattern.

w0rp:
Looks good to me! :+1:

w0rp:
Could you rebase and fix the conflict here?

abalmos:
@w0rp Pushed a rebase

w0rp:
Cheers! :beers:

PR #195, opened 2016-12-01
Add options variable for ruby-rubocop

mortonfox:
The reason for adding an options variable to Ruby Rubocop is I wanted to specify --lint when running Rubocop from ALE.

I find Rubocop's style cops too verbose and picky when I'm writing Ruby code, so I prefer to run it with only the lint cops enabled. I usually wait until I'm mostly done before turning on the Rubocop style cops.


w0rp:
Sure.

PR #198, opened 2016-12-04
Add support for tslint.json file detection in tslint

kyrisu:
By default tslint looks for the tslint.json file in the checked file directory. By providing option `-c` with the path to a tslint.json file we can override that.

The option is not set if the tslint.config is not found up the directory tree.

### DIFF
--- a/ale_linters/typescript/tslint.vim
+++ b/ale_linters/typescript/tslint.vim
@@ -37,12 +37,16 @@ function! ale_linters#typescript#tslint#Handle(buffer, lines)
     return l:output
 endfunction
 
-let s:tsconfig_path = ale#util#FindNearestFile(bufnr('%'), 'tslint.json')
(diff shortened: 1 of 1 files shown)


ale_linters/typescript/tslint.vim
```diff
@@ -37,9 +37,12 @@ function! ale_linters#typescript#tslint#Handle(buffer, lines)
     return l:output
 endfunction
 
+let s:tsconfig_path = ale#util#FindNearestFile(bufnr('%'), 'tslint.json')
+let s:tslint_options = empty(s:tsconfig_path) ? '' : '-c ' . s:tsconfig_path
```
w0rp: This runs only once when the file is loaded, so this won't work. You'll need to run this from within a `command_callback`, using the buffer number provided to the function.

w0rp:
Looks good! :+1: I'll let Travis look at it.

PR #201, opened 2016-12-05
Fix proper extension handling for tslint

kyrisu:
Right now tslint linter assumes .ts extension, unfortunately, that doesn't work when linting .tsx files (jsx in typescript).

I've added a change to take the file extension into account.


w0rp:
Looks fine to me. :+1:

PR #207, opened 2016-12-06
[new linter] add stylelint support

diartyz:
add stylelint support for
- css
- scss
- sass

### DIFF
--- a/ale_linters/css/stylelint.vim
+++ b/ale_linters/css/stylelint.vim
@@ -1,0 +1,8 @@
+" Author: diartyz <diartyz@gmail.com>
+
+call ale#linter#Define('css', {
+\   'name': 'stylelint',
+\   'executable': 'stylelint',
+\   'command': g:ale#util#stdin_wrapper . ' .css stylelint',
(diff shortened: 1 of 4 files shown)


w0rp:
Looks good to me. :+1:

aaronmcadam:
I'm having trouble getting `stylelint` to pick up my style violations. 

I've tried editing a file within a directory that has a `.stylelintrc` and a nested directory too, but I never see any violations in the gutter.

Is there something special I need to do @diartyz?

diartyz:
@aaronmcadam 
you doesn't need to specify a syntax, it will be automatically inferred by the file extension.
can you get the error when you using stylelint in shell ?
eg: `stylelint path/to/file`

aaronmcadam:
Thanks @diartyz, it turned out that my global `stylelint` was out of date and I was configuring a property that didn't exist with that version. It's working now :smile:

diartyz:
@aaronmcadam 
btw, you can use the local npm package now.
https://github.com/w0rp/ale/pull/226

PR #210, opened 2016-12-07
Catch g++ fatal error

apuignav:
In some cases I get "fatal error" in g++ (I don't know exactly why):

```
../Software/Toys/pdfs/RooKstMuMuFoldedP5.h:15:10: fatal error: 'RooAbsPdf.h' file not found
```

Catch these.


w0rp:
You broke the build there, and I thought it would. You should run `Vint` on your files. You'll need to add a `#` at the end of your regex there to make the match case-sensitive. Vint complains about this because operators like `==` change in behaviour depending on the user's configuration. (Thanks Vim...)

apuignav:
Damn, sorry for that, I thought Vint was installed. Fixed.

w0rp:
Cool, looks good to me.

Issue #117, opened 2016-10-17
Replace Windows stdin-wrapper executable with something better


hauleth:

w0rp:
I actually have another plan for this. I plan to eventually add support for chaining commands, so you can use the output of one command run asynchronously to run another job asynchronously. Then the need for special D script would go away. As far as the application of the idea goes, it would look like this:

```vim
function! GetDUBCommand(buffer)
    return 'dub describe'
endfunction

function! GetDMDCommand(buffer, dub_output_lines)
    " Build import paths arguments here
    return 'dmd ...'
endfunction

ale#linters#Define('d', {
\ ...
\   'command_chain' ['GetDUBCommand', 'GetDMDCommand'],
\})
``` 

The implementation is another story. That can be quite complicated.

bounceme:
What do you think about https://github.com/w0rp/ale/issues/16#issuecomment-254011781 ?
I can see about making a pr to replace the stdin wrappers ( that is what this issue is about right? ) with a vimscript one

w0rp:
This issue is about the DMD wrapper.  I don't think creating temporary files in the same directory as the source files is a good solution. Vim won't always manage to clean up the files after it is done, and you might not have permission to write the new files to the same directories.

bounceme:
They are made in $TMPDIR . The calls of `delete()` clean the files out

w0rp:
I see, you're copying the files to the temporary directory, and running the commands on that. That won't work when you need to check for files in a particular directory, say for when you import particular source files, etc.

bounceme:
True, the linters produce output based on the temps, and the output has the file names switched for the loclists. Your wrappers are a cleaner solution

hauleth:
I meant not `dmd-wrapper` but `stdin-wrapper.d`. I haven't used D in ages so `dmd-wrapper` isn't my area of interest.

w0rp:
Ah right, the D program for the Windows stdin-wrapper? It could potentially be replaced with a PowerShell script if it will work on most versions of Windows. I'll go with whatever actually works. The 32-bit executable should be working.

hauleth:
Depends on which Windows do you want to support. http://serverfault.com/questions/615275/in-which-version-of-windows-is-powershell-installed-by-default <http://serverfault.com/questions/615275/in-which-version-of-windows-is-powershell-installed-by-default>

TL;DR If we support only W7 and up, then we are ready to go.
--

Łukasz Niemier
lukasz@niemier.pl

[...]

PR #212, opened 2016-12-09
Add kotlinc checker for kotlin

francis36012:
This adds a checker (using kotlinc) for kotlin

### DIFF
--- a/ale_linters/kotlin/kotlinc.vim
+++ b/ale_linters/kotlin/kotlinc.vim
@@ -10,104 +10,107 @@ let g:ale_kotlin_kotlinc_use_module_file = get(g:, 'ale_kotlin_kotlinc_use_modul
 let g:ale_kotlin_kotlinc_module_filename = get(g:, 'ale_kotlin_kotlinc_module_filename', 'module.xml')
 
 function! ale_linters#kotlin#kotlinc#GetCommand(buffer)
-	let l:kotlinc_opts = g:ale_kotlin_kotlinc_options
-	let l:command = 'kotlinc '
+    let l:kotlinc_opts = g:ale_kotlin_kotlinc_options
+    let l:command = 'kotlinc '
 
-	" If the config file is enabled and readable, source it
-	if g:ale_kotlin_kotlinc_enable_config
-		if filereadable(expand(g:ale_kotlin_kotlinc_config_file, 1))
-			execute 'source ' . fnameescape(expand(g:ale_kotlin_kotlinc_config_file, 1))
-		endif
-	endif
+    " If the config file is enabled and readable, source it
+    if g:ale_kotlin_kotlinc_enable_config
+        if filereadable(expand(g:ale_kotlin_kotlinc_config_file, 1))
+            execute 'source ' . fnameescape(expand(g:ale_kotlin_kotlinc_config_file, 1))
+        endif
+    endif
 
-	" If use module and module file is readable use that and return
-	if g:ale_kotlin_kotlinc_use_module_file
-		let l:module_filename = fnameescape(expand(g:ale_kotlin_kotlinc_module_filename, 1))
-		if filereadable(l:module_filename)
-			let l:kotlinc_opts .= ' -module ' . l:module_filename
+    " If use module and module file is readable use that and return
+    if g:ale_kotlin_kotlinc_use_module_file
+        let l:module_filename = fnameescape(expand(g:ale_kotlin_kotlinc_module_filename, 1))
 
-			let l:command .= 'kotlinc ' . l:kotlinc_opts
-			return l:command
-		endif
-	endif
+        if filereadable(l:module_filename)
+            let l:kotlinc_opts .= ' -module ' . l:module_filename
+            let l:command .= 'kotlinc ' . l:kotlinc_opts
 
-	" We only get here if not using module or the module file not readable
-	if g:ale_kotlin_kotlinc_classpath !=# ''
-		let l:kotlinc_opts .= ' -cp ' . g:ale_kotlin_kotlinc_classpath
-	endif
+            return l:command
+        endif
+    endif
 
-	let l:fname = ''
-	if g:ale_kotlin_kotlinc_sourcepath !=# ''
-		let l:fname .= expand(g:ale_kotlin_kotlinc_sourcepath, 1) . ' '
-	endif
-	let l:fname .= shellescape(expand('%', 1))
+    " We only get here if not using module or the module file not readable
+    if g:ale_kotlin_kotlinc_classpath !=# ''
+        let l:kotlinc_opts .= ' -cp ' . g:ale_kotlin_kotlinc_classpath
+    endif
 
-	let l:command .= l:kotlinc_opts . ' ' . l:fname
-	return l:command
+    let l:fname = ''
+
+    if g:ale_kotlin_kotlinc_sourcepath !=# ''
+        let l:fname .= expand(g:ale_kotlin_kotlinc_sourcepath, 1) . ' '
+    endif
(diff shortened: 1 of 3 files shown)


ale_linters/kotlin/kotlinc.vim
```diff
@@ -0,0 +1,119 @@
...
+	" We only get here if not using module or the module file not readable
+	if g:ale_kotlin_kotlinc_classpath !=# ''
+		let l:kotlinc_opts .= ' -cp ' . g:ale_kotlin_kotlinc_classpath
+	endif
+
+	let l:fname = ''
+	if g:ale_kotlin_kotlinc_sourcepath !=# ''
+		let l:fname .= expand(g:ale_kotlin_kotlinc_sourcepath, 1) . ' '
+	endif
+	let l:fname .= shellescape(expand('%', 1))
```
w0rp: You might expand the wrong filename here. The callback might fire while another buffer is selected. You should take the filename from the buffer instead.

ale_linters/kotlin/kotlinc.vim
```diff
@@ -0,0 +1,119 @@
...
+
+	let l:command .= l:kotlinc_opts . ' ' . l:fname
```
w0rp: You probably need to use the `stdin_wrapper` here, and then you'll probably run into other issues. Are you running the linter on the file on disk, or does it read from Vim's buffer via stdin so you can check for warnings or errors as you type without having to save the file?

ale_linters/kotlin/kotlinc.vim
```diff
@@ -0,0 +1,119 @@
...
+		let l:type = l:match[4]
+		let l:text = l:match[5]
+
+		let l:bufnr = bufnr(l:file)
+		let l:curbufnr = bufnr('%')
+
+		" Skip if file is not loaded
+		if l:bufnr == -1 || l:bufnr != l:curbufnr
+			continue
+		endif
```
w0rp: Can you have a different path to the same file here and see no warnings or errors? It might be better to get an absolute path from what you parsed above, and the absolute path from the buffer's filename, and then compare the two.

francis36012: Great suggestion. I will be doing that.

ale_linters/kotlin/kotlinc.vim
```diff
@@ -0,0 +1,124 @@
...
+            continue
+        endif
+
+        let l:file = l:match[1]
+        let l:line = l:match[2] + 0
+        let l:column = l:match[3] + 0
+        let l:type = l:match[4]
+        let l:text = l:match[5]
+
+        let l:bufnr = bufnr(l:file)
```
w0rp: I think the problem is still here. What if the path from the linter is different from the path to the file used by Vim? Then you will probably fail to find the buffer number. You should use the path to the file taken from the buffer number given to this function, and compare it to the path you have parsed here, probably normalising both to absolute paths.

ale_linters/kotlin/kotlinc.vim
```diff
@@ -0,0 +1,124 @@
...
+        let l:kotlinc_opts .= ' -cp ' . g:ale_kotlin_kotlinc_classpath
+    endif
+
+    let l:fname = ''
+
+    if g:ale_kotlin_kotlinc_sourcepath !=# ''
+        let l:fname .= expand(g:ale_kotlin_kotlinc_sourcepath, 1) . ' '
+    endif
+    let l:fname .= shellescape(expand('#' . a:buffer . ':p'))
+    let l:command .= l:kotlinc_opts . ' ' . l:fname
```
w0rp: Will this compile on the file on disk, or on the text buffer in Vim? If it's the former, then you need to use the stdin_wrapper command so the Vim buffer is copied to a temporary file. Otherwise, you won't be able to check warnings or errors until the file is saved.

If this compiler is anything like `javac`, then you probably won't be able to implement this until after the Windows stdin-wrapper script has been written to save files in temporary directories, and both stdin-wrapper scripts copy the basename of files to the temporary directories verbatim. See issues #40 and #117.

ale_linters/kotlin/kotlinc.vim
francis36012: Thanks for the detailed feedback. Will the code below be enough then?

```vim
if fnamemodify(l:file, ':p') !=# expand('#' . a:buffer . ':p')
    continue
endif
```

w0rp: Yeah, I think that will work.

ale_linters/kotlin/kotlinc.vim
w0rp: Hmm, that will prevent you from correcting errors and warnings by typing then, or checking for errors and warnings before you save the file. I think this linter might have to wait until the issues I linked are resolved. Then it can use the stdin_wrapper script. You can keep The linter in your own fork for now and live without that behaviour. Hopefully I should get a little time to make this work soon.

w0rp:
There just the stdin_wrapper stuff I mentioned. I need to do a bit of work
to get that to work with compilers like these. You can keep it in your fork
for now. It will be a while till that's done.

ale_linters/kotlin/kotlinc.vim
francis36012: I don't think that will work, at least with the compiler being used here. If a buffer is copied into a temporary file and passed to the compiler, it mostly likely will result in duplicate definition errors because the original file may have already being specified in the linter's options (which will also be passed to the compiler).

w0rp:
@francis36012 Hello! It's been a while since I picked this up. Would you be interested in bringing this up to date? There is a new flag which will make linters run against files on disk, instead of Vim buffers. See #333. It should be perfect for this linter.

ale_linters/kotlin/kotlinc.vim
```diff
@@ -0,0 +1,122 @@
...
+
+    return l:output
+endfunction
+
+call ale#linter#Define('kotlin', {
+\   'name': 'kotlinc',
+\   'output_stream': 'stderr',
+\   'executable': 'kotlinc',
+\   'command_callback': 'ale_linters#kotlin#kotlinc#GetCommand',
+\   'callback': 'ale_linters#kotlin#kotlinc#Handle',
```
w0rp: For the comment I made just now, you pretty much need to just add `'lint_file': 1,` here.

ale_linters/kotlin/kotlinc.vim
```diff
@@ -0,0 +1,122 @@
...
+    if g:ale_kotlin_kotlinc_classpath !=# ''
+        let l:kotlinc_opts .= ' -cp ' . g:ale_kotlin_kotlinc_classpath
+    endif
+
+    let l:fname = ''
+
+    if g:ale_kotlin_kotlinc_sourcepath !=# ''
+        let l:fname .= expand(g:ale_kotlin_kotlinc_sourcepath, 1) . ' '
+    endif
+    let l:fname .= shellescape(expand('#' . a:buffer . ':p'))
```
w0rp: This should work fine Some changes have been made since you worked on this, and it might be possible for you to simply put `%s` in your command string now. ALE will replace that with the path to the file. What you have here already might work better for some reason, or maybe not as well? It's hard to tell.

francis36012:
Hello @w0rp, thanks for alerting me on the changes. I will be making the necessary changes to get the branch on track. It sounds like it won't be too much of a hassle.

francis36012:
Hey @w0rp, I added the flag, and brought the branch up to date. Can you merge now?

w0rp:
I just pushed my commit now for documentation restructuring, so feel free to open a pull request whenever.

PR #213, opened 2016-12-12
Add Elm linting via elm-make

BuffaloCoder:
Adds linting for Elm files using elm-make.

The process is similar to how [elm-vim](https://github.com/ElmCast/elm-vim) setup the [Syntastic](https://github.com/vim-syntastic/syntastic) linter.

### DIFF
--- a/ale_linters/elm/make.vim
+++ b/ale_linters/elm/make.vim
@@ -1,63 +1,64 @@
 " Author: buffalocoder - https://github.com/buffalocoder
 
 function! ale_linters#elm#make#Handle(buffer, lines)
-  let l:output = []
-  for l:line in a:lines
-    if l:line[0] ==# '['
-      let l:errors = json_decode(l:line)
+    let l:output = []
+    for l:line in a:lines
+        if l:line[0] ==# '['
+            let l:errors = json_decode(l:line)
 
-      for l:error in l:errors
-        call add(l:output, {
-        \   'bufnr': a:buffer,
-        \   'lnum': l:error.region.start.line,
-        \   'vcol': 0,
-        \   'col': l:error.region.start.column,
-        \   'type': (l:error.type ==? 'error') ? 'E' : 'W',
-        \   'text': l:error.overview,
-        \   'nr': -1,
-        \})
-      endfor
-    endif
-  endfor
+            for l:error in l:errors
+                call add(l:output, {
+                \    'bufnr': a:buffer,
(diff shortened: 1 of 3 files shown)


BuffaloCoder:
Looks like this should be changed to use the stdin wrapper. I'll look into that change tonight.

ale_linters/elm/make.vim
```diff
@@ -0,0 +1,64 @@
...
+
+" This function was copied from from https://github.com/ElmCast/elm-vim.
+" All credit goes to that project!
+" Returns the closest parent with an elm-package.json file.
+function! s:FindRootDirectory() abort
+    let l:elm_root = getbufvar('%', 'elmRoot')
+    if empty(l:elm_root)
+        let l:current_file = expand('%:p')
+        let l:dir_current_file = fnameescape(fnamemodify(l:current_file, ':h'))
+        let l:match = findfile('elm-package.json', l:dir_current_file . ';')
```
w0rp: This search can fail when the callback fires from a different buffer from the buffer the linter is being run for. You should try using the `FindFile` function in the util module. That will probably do the job.

ale_linters/elm/make.vim
```diff
@@ -0,0 +1,64 @@
...
+        endif
+    endif
+    return l:elm_root
+endfunction
+
+" Return the command to execute the linter in the projects directory.
+" If it doesn't, then this will fail when imports are needed.
+function! ale_linters#elm#make#GetCommand(buffer) abort
+    let l:root_dir = s:FindRootDirectory()
+    let l:dir_set_cmd = 'cd' . fnameescape(l:root_dir)
```
w0rp: Be careful here if the directory is an empty string. On UNIX machines, this might cause you to change directories to the home directory. It's probably best to check it the path is an empty string here and then only use `cd` if it's not empty.

ale_linters/elm/make.vim
```diff
@@ -0,0 +1,64 @@
...
+
+    return l:dir_set_cmd . '; elm-make --report=json %s --output='.shellescape(g:ale#util#nul_file)
```
w0rp: Does this run on the file on disk or on the text buffer? All linters for this project should be set up so you can check errors and warnings while you type, without having to save the file.

w0rp:
I haven't run it myself, but it looks good to me. :+1: Cheers! :beers:

PR #221, opened 2016-12-14
Improve formatting of ghc and hlint haskell linters

rob-b:
For ghc, it seemed that the conditional

```
if l:corrected_lines[-1] =~# ': error:$'
    let l:line = substitute(l:line, '\v^\s+', ' ', '')
endif
```

was never being reached. It's actually better to unconditionally
collapse whitespace anyway and so I simply removed the conditional
check.

For hlint I added more information about the error. This changes the
reported error from `Error: Avoid lambda` to something like:
` Error: Avoid lambda. Found: \ x -> foo x Why not: foo`

### DIFF
--- a/ale_linters/haskell/ghc.vim
+++ b/ale_linters/haskell/ghc.vim
@@ -29,9 +29,7 @@ function! ale_linters#haskell#ghc#Handle(buffer, lines)
             call add(l:corrected_lines, l:line)
         else
             if len(l:corrected_lines) > 0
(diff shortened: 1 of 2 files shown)


w0rp:
I see no problems here. :+1: If you want to go the extra mile, it would be good to have some Vader tests to cover these functions. The Vader tests aren't too hard to write once you get your head around them. Just avoid using `Then` unless you are using newer versions of Vader.

PR #223, opened 2016-12-14
Elm file filter & Windows bug fixes

BuffaloCoder:
Addresses https://github.com/w0rp/ale/issues/219. While I was at it I setup a Windows test environment and cleaned up issues related to that.

### DIFF
--- a/ale_linters/elm/make.vim
+++ b/ale_linters/elm/make.vim
@@ -3,20 +3,25 @@
 
 function! ale_linters#elm#make#Handle(buffer, lines)
     let l:output = []
+    let l:temp_dir = has('win32') ? $TMP : $TMPDIR
     for l:line in a:lines
         if l:line[0] ==# '['
             let l:errors = json_decode(l:line)
 
             for l:error in l:errors
-                call add(l:output, {
-                \    'bufnr': a:buffer,
-                \    'lnum': l:error.region.start.line,
-                \    'vcol': 0,
-                \    'col': l:error.region.start.column,
-                \    'type': (l:error.type ==? 'error') ? 'E' : 'W',
-                \    'text': l:error.overview,
-                \    'nr': -1,
-                \})
+                " Check if file is from the temp directory.
+                " Filters out any errors not related to the buffer.
+                if matchend(l:error.file, l:temp_dir) > 0
+                    call add(l:output, {
+                    \    'bufnr': a:buffer,
+                    \    'lnum': l:error.region.start.line,
+                    \    'vcol': 0,
+                    \    'col': l:error.region.start.column,
+                    \    'type': (l:error.type ==? 'error') ? 'E' : 'W',
+                    \    'text': l:error.overview,
+                    \    'nr': -1,
+                    \})
+                endif
             endfor
         endif
     endfor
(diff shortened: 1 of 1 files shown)


ale_linters/elm/make.vim
```diff
@@ -3,20 +3,25 @@
...
-                \    'lnum': l:error.region.start.line,
-                \    'vcol': 0,
-                \    'col': l:error.region.start.column,
-                \    'type': (l:error.type ==? 'error') ? 'E' : 'W',
-                \    'text': l:error.overview,
-                \    'nr': -1,
-                \})
+                " Check if file is from the temp directory.
+                " Filters out any errors not related to the buffer.
+                if l:error.file[0:len(l:temp_dir)-1] == l:temp_dir
```
w0rp: The linter didn't catch this one, but I wonder if you should use `==#` here or `==?`. One is case-sensitive, the other is case-insensitive. `==` is one or the other, depending on the user's Vim configuration.

It might make sense to run a case-sensitive path check for Unix systems, and a case-insensitive check for Windows computers.

I'd put some spaces around the binary `-` operator while you're at it.

ale_linters/elm/make.vim
```diff
@@ -32,10 +37,10 @@ function! ale_linters#elm#make#GetCommand(buffer) abort
         let l:dir_set_cmd = ''
     else
         let l:root_dir = fnamemodify(l:elm_package, ':p:h')
-        let l:dir_set_cmd = 'cd ' . fnameescape(l:root_dir) . '; '
+        let l:dir_set_cmd = 'cd ' . fnameescape(l:root_dir) . ' && '
     endif
 
-    let l:elm_cmd = 'elm-make --report=json --output='.shellescape(g:ale#util#nul_file)
+    let l:elm_cmd = 'elm-make --report=json --output='.shellescape('/dev/null')
```
w0rp: Will this break on Windows? I wonder if we should just remove the `shellescape` call where. We know the strings will be either `/dev/null` for Unix system or the special filename `nul` for Windows, which is the Windows equivalent.

ale_linters/elm/make.vim
BuffaloCoder: Will do!

ale_linters/elm/make.vim
BuffaloCoder: It actually fixes a Windows issue. Elm-make fails for "nul" because it lacks a file type of HTML or js. Instead, it uses "/dev/null" as a flag for not creating an output file.

I'll add a comment denoting as such when I apply the other changes.

w0rp: Huh, strange. Well, if it works, it works! Could you leave a comment explaining that above?

ale_linters/elm/make.vim
```diff
@@ -3,20 +3,31 @@
...
-                \    'nr': -1,
-                \})
+                " Check if file is from the temp directory.
+                " Filters out any errors not related to the buffer.
+                if l:is_windows
+                    let l:file_is_buffer = l:error.file[0:len(l:temp_dir)-1] ==? l:temp_dir
+                else
+                    let l:file_is_buffer = l:error.file[0:len(l:temp_dir)-1] ==# l:temp_dir
+                endif
+                if l:error.file[0:len(l:temp_dir)-1] == l:temp_dir
```
w0rp: I think you forgot to use the variable you created above here. Also, could you put a single blank line above if statements which don't start a new block?

BuffaloCoder: Corrected.  Don't know how I missed that one... Also, added the spaces around `-`

w0rp:
Looks good to me! :+1: I'll let Travis check it.

Issue #220, opened 2016-12-14
Error signs never unplaced


cympfh:
Once some signs placed, they never will be unplaced.
I can confirm this behavior on Vim8.0 (patch 1-133)/CentOS7.2.
But on the other platforms (Mac, Ubuntu), I cannot confirm this. 
So I am afraid that my vim is just wrong.

On my CentOS, I use ale with

```patch
--- a/autoload/ale/sign.vim
+++ b/autoload/ale/sign.vim
@@ -87,6 +87,7 @@ endfunction
 " This function will set the signs which show up on the left.
 function! ale#sign#SetSigns(buffer, loclist) abort
     let l:signlist = ale#sign#CombineSigns(a:loclist)
+    sign unplace *
```

w0rp:
Could you upload the file you were linting, and which linters you were running? It should be parsing the signs and removing the signs by their IDs. Could you also give me the output of `:sign place` when you have some signs showing that you can't get rid of? The format of that command is parsed to find signs to remove, and the locale configuration can change the output of that command in various ways.

cympfh:
Thanks. I thinks this is because of my locale (Japanese).

I checked with `rubocop` and the following ruby scripts.

```ruby
n = 0
```

```vim
:sign place
--- サイン ---
test.rb のサイン:
    行=1  識別子=1000001  名前=ALEWarningSign

:echo g:ale_buffer_info
{'1': {'job_list': [], 'echoed': 1, 'loclist': [{'lnum': 1, 'bufnr': 1, 'vcol': 0, 'linter_name': 'rubocop', 'nr': -1, 'type': 'W', 'col': 1, 'text': 'Useless assignment to variable - `n`.'}], 'new_loclist': []}}
```

Then

```ruby
n = 0
p n
```

```vim
:sign place
--- サイン ---
test.rb のサイン:
    行=1  識別子=1000001  名前=ALEWarningSign

:echo g:ale_buffer_info
{'1': {'job_list': [], 'loclist': [], 'new_loclist': []}}
```

cympfh:
ale works correctly with `LANG=en vim test.rb` 😄

w0rp:
Yeah, it's a weird issue I previously hit with a Russian user, and it looks like it's still broken for Japanese users. The pull request looks pretty good, but I'll require some more tests for this before I merge it.

w0rp:
I merged your pull request now and added test cases to cover this.

PR #224, opened 2016-12-15
Fix: sign pattern

cympfh:
#220


w0rp:
The bug fix looks good to me. I would like to add some tests for this. I think the code for reading the signs should be split into one function, and the code for reading the lines of output should be split into another. Then we could write unit tests with a bunch of examples for the latter, and another test against the former. You can give this a go, or else I'll do it later.

cympfh:
Okay. I'm waiting.

w0rp:
I'll merge this and add the tests now.

PR #226, opened 2016-12-16
[enhancement] add node_modules support for stylelint & htmlhint

diartyz:
@w0rp 

use local npm package frist instead of global package.
please review.


diartyz:
@w0rp 
ok, I update the doc.

PR #230, opened 2016-12-19
Add support for Rust using rustc and cargo

EinfachToll:
There are some things the user has to pay attention to when using this linter, so I added a few notes to the help file.

### DIFF
--- a/ale_linters/rust/rustc.vim
+++ b/ale_linters/rust/rustc.vim
@@ -1,0 +1,64 @@
+" Author: Daniel Schemala <istjanichtzufassen@gmail.com>
+" Description: rustc for rust files
+
+
+function! ale_linters#rust#rustc#handle_rustc_errors(buffer_number, errorlines)
+    let output = []
+    for errorline in a:errorlines
+        if errorline !~# '^{'
+            continue
+        endif
+        let error = json_decode(errorline)
+        for span in error.spans 
+            if span.is_primary
+                call add(output, {
+                            \ 'bufnr': a:buffer_number,
+                            \ 'lnum': span.line_start,
+                            \ 'vcol': 0,
+                            \ 'col': span.byte_start,
+                            \ 'nr': -1,
+                            \ 'text': error.message,
+                            \ 'type': toupper(error.level[0]),
+                            \ })
+            endif
+        endfor
+    endfor
+    return output
+endfunction
+
+
+function! ale_linters#rust#rustc#rustc_command(buffer_number)
+    " Try to guess the library search path. If the project is managed by cargo,
+    " it's usually <project root>/target/debug/deps/ or
+    " <project root>/target/release/deps/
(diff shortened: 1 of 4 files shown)


ale_linters/rust/rustc.vim
```diff
@@ -0,0 +1,94 @@
...
+                \ (l:span.file_name ==# l:file_name || l:span.file_name ==# '<anon>')
+                call add(l:output, {
+                            \ 'bufnr': a:buffer_number,
+                            \ 'lnum': l:span.line_start,
+                            \ 'vcol': 0,
+                            \ 'col': l:span.byte_start,
+                            \ 'nr': -1,
+                            \ 'text': l:error.message,
+                            \ 'type': toupper(l:error.level[0]),
+                            \ })
```
w0rp: Could you align the continued lines to a new level of indentation like in the rest of the codebase?

ale_linters/rust/rustc.vim
```diff
@@ -0,0 +1,94 @@
...
+
+if !exists('g:ale_rust_ignore_error_codes')
+    let g:ale_rust_ignore_error_codes = []
+endif
+
+
+function! ale_linters#rust#rustc#handle_rustc_errors(buffer_number, errorlines)
+    let l:file_name = fnamemodify(bufname(a:buffer_number), ':t')
+    let l:output = []
+    for l:errorline in a:errorlines
```
w0rp: Could you add a blank line before any loop or if statement which does not start a new block of indentation?

EinfachToll:
I made the requested stylistic changes.

> Is there any way to run Cargo on a temporary file, but set the path to include files in such a way that the other source files can be included? Maybe not.

I'll take a look.

> I remember hearing about some security issues with running rustc on files for Syntastic. That might be worth looking into. The solution there was that the checker was disabled by default.

I searched a but through discussions about syntastic and rust.vim, but found nothing about security issues. I'm also not sure how there can be any.

EinfachToll:
>> Is there any way to run Cargo on a temporary file, but set the path to include files in such a way that the other source files can be included? Maybe not.

>I'll take a look.

I took a look and unfortunately found no solution.

> Have a look at the conflicts that exist now.

I'm not sure what you mean here. I see that there is a conflict in doc/ale.txt, do you want me to resolve it or do you do this yourself?

w0rp:
@EinfachToll Yes, please fix the conflict in that file.

If there's no way to run the linter on temporary files, then maybe some linters should have an extra option for disabling running them as you type. There might be no other way around that. It might also make sense to apply the same to the `go build` linter.

w0rp:
You should rebase your branch instead of merging `master`, but I will just squash the commits anyway, so it's okay.

ivegotasthma:
Need any help with this, because it's a blocker for me.

w0rp:
I'll merge this. I might work on an option to disable running the linters while you type for specific linters, as it won't help here.

PR #231, opened 2016-12-23
Redirect go build output to /dev/null

fannheyward:
Named file file will be created after ALE, this can redirect the output to /dev/null if go build failed.


w0rp:
Looks fine to me. :+1:

PR #232, opened 2016-12-23
Added ability to reload compilation flags for c and c++

grzebiel:
I've noticed that `g:ale_c_gcc_options` and `g:ale_cpp_gcc_options` are set permamently to `linter.command`.

Here's a fix that alowes you to modify compilation flags dynamicly.


w0rp:
Yep, makes sense to me.

PR #235, opened 2016-12-28
Fix PHP lint generic error without line number

still-dreaming-1:
This makes php output more specific error messages. The format is the normal one ALE expects, but on some systems ALE does not work with PHP unless the display_errors=1 option is used. Without that option php will only output a generic message without a line number like "Errors parsing index.php".


still-dreaming-1:
Just to clarify, the reason I did this was to get Ale working on my system, as it does not notice PHP on my system without that option. I have tested it and it works for me.

w0rp:
Makes sense to me. Cheers! :beers:

PR #236, opened 2016-12-30
Pass file name to flake8

Chronial:
This is required for flake8 and its plugins to be able to apply file-specific rules.


w0rp:

PR #237, opened 2016-12-31
Add python mypy support

keith:
This adds support for [mypy](mypy-lang.org) the official optional python typechecker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ name. That seems to be the fairest way to arrange this table.
 | PHP | [php -l](https://secure.php.net/), [phpcs](https://github.com/squizlabs/PHP_CodeSniffer) |
 | Pug | [pug-lint](https://github.com/pugjs/pug-lint) |
 | Puppet | [puppet](https://puppet.com), [puppet-lint](https://puppet-lint.com) |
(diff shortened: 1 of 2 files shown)


w0rp:
Is there anything that must be done for this to get it to recognise imports correctly, say in combination with virtualenv?

keith:
Right now this definitely assumes that you're running vim in an environment where that's already handled. It feels to me like setting that up might be outside of the scope of what this plugin should handle. Thoughts?


--
Keith Smiley

renstrom:
>It feels to me like setting that up might be outside of the scope of what this plugin should handle. Thoughts?

I agree. Setting up a virtualenv (and activating it) is definitely something ALE shouldn't bother itself with. There are other plugins for that:

- https://github.com/lambdalisue/vim-pyenv
- https://github.com/jmcantrell/vim-virtualenv

w0rp:
I'll take this without the virtualenv detection. If enough people complain, it can be addressed later.

Could you change the variable name for the options from `_args` to `_options`? I've started insisting that every linter uses the same name convention for the options here: https://github.com/w0rp/ale/issues/257

w0rp:
I'll just merge this and do that myself. It should only take a few minutes.

keith:
Ah sorry, thanks for doing that! Thanks!


--
Keith Smiley

PR #239, opened 2017-01-01
Added hack linter for php

zefei:
This adds hack typechecker for hack (http://hacklang.org/) and php.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ name. That seems to be the fairest way to arrange this table.
 | MATLAB | [mlint](https://www.mathworks.com/help/matlab/ref/mlint.html) |
 | OCaml | [merlin](https://github.com/the-lambda-church/merlin) see `:help ale-integration-ocaml-merlin` for configuration instructions
 | Perl | [perl -c](https://perl.org/), [perl-critic](https://metacpan.org/pod/Perl::Critic) |
-| PHP | [php -l](https://secure.php.net/), [phpcs](https://github.com/squizlabs/PHP_CodeSniffer) |
+| PHP | [php -l](https://secure.php.net/), [phpcs](https://github.com/squizlabs/PHP_CodeSniffer), [hh_client](http://hacklang.org/) |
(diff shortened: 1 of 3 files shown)


ale_linters/php/hack.vim
```diff
@@ -0,0 +1,35 @@
...
+        call add(l:output, {
+        \   'bufnr': a:buffer,
+        \   'lnum': l:match[2] + 0,
+        \   'vcol': 0,
+        \   'col': l:match[3] + 0,
+        \   'text': l:match[5],
+        \   'type': 'E',
+        \})
+    endfor
+    return l:output
```
w0rp: I'd recommend putting a single blank line before any `return` statement, `if` statement, or any loop which doesn't start a new block.

README.md
```diff
@@ -72,7 +72,7 @@ name. That seems to be the fairest way to arrange this table.
 | MATLAB | [mlint](https://www.mathworks.com/help/matlab/ref/mlint.html) |
 | OCaml | [merlin](https://github.com/the-lambda-church/merlin) see `:help ale-integration-ocaml-merlin` for configuration instructions
 | Perl | [perl -c](https://perl.org/), [perl-critic](https://metacpan.org/pod/Perl::Critic) |
-| PHP | [php -l](https://secure.php.net/), [phpcs](https://github.com/squizlabs/PHP_CodeSniffer) |
+| PHP | [php -l](https://secure.php.net/), [phpcs](https://github.com/squizlabs/PHP_CodeSniffer), [hh_client](http://hacklang.org/) |
```
w0rp: Should we maybe name the linter itself `hh_client`, or maybe use the name `hack` here instead, and in the docs? It would be nice to make the name easier to guess for people who want to turn it on or off specifically. I'd also move it to the front of the list, as everything is supposed to be sorted by name here.

zefei:
Hey thanks for the review! I changed names to 'hack' and fixed styles.

w0rp:

PR #245, opened 2017-01-05
Add eslint option to pass in more options

zefei:
This adds an eslint option to customize the eslint command without affecting the executable callback, so that users can pass more options to eslint such as `'--rulesdir $ESLINT_RULES_DIR'` (this option cannot be set inside `.eslintrc`).


w0rp:
Looks fine to me. Merged! :+1:

Issue #248, opened 2017-01-08
Add linter for Erlang


evnu:
I'd like to add an Erlang linter based on running `erlc`. If you are interested, I will submit a pull request.

w0rp:
Yeah, go for it.

w0rp:
I just realised that this was merged ages ago.

PR #255, opened 2017-01-12
Add erlc lint for Erlang (#248)

evnu:
This pull-request adds support for linting Erlang by using `erlc` (See #248). `erlc` needs to be configured manually in order to find non-standard include files or `parse_transform` modules. For this, I added a new option `g:ale_erlang_erlc_flags`.

As the error format on `parse_transform` errors is different to the usual format, the error handling is split into two steps (check normal error/warning, check `parse_transform` error). I am currently unaware of a better way to achieve this.

### DIFF
--- a/ale_linters/erlang/erlc.vim
+++ b/ale_linters/erlang/erlc.vim
@@ -1,0 +1,74 @@
+" Author: Magnus Ottenklinger - https://github.com/evnu
+
+function! ale_linters#erlang#erlc#Handle(buffer, lines)
+  " Matches patterns like the following:
+  "
+  " error.erl:4: variable 'B' is unbound
+  " error.erl:3: Warning: function main/0 is unused
+  " error.erl:4: Warning: variable 'A' is unused
+  let l:pattern = '\v^([^:]+):(\d+): (Warning: )?(.+)$'
+
+  " parse_transforms are a special case. The error message does not indicate a location:
+  " error.erl: undefined parse transform 'some_parse_transform'
+  let l:pattern_parse_transform = '\v(undefined parse transform .*)$'
+  let l:output = []
+
+  for l:line in a:lines
+    let l:match = matchlist(l:line, l:pattern)
+    let l:match_parse_transform = matchlist(l:line, l:pattern_parse_transform)
+
+    " Determine if the output indicates an error. We distinguish between two cases:
+    "
+    " 1) normal errors match l:pattern
+    " 2) parse_transform errors match l:pattern_parse_transform
+    "
+    " If none of the patterns above match, the line can be ignored
+    if len(l:match) == 0 " not a 'normal' warning or error
+        if len(l:match_parse_transform) == 0 " also not a parse_transform error
+            continue
+        endif
+
+        let l:text = l:match_parse_transform[0]
+        call add(l:output, {
+                    \   'bufnr': a:buffer,
+                    \   'lnum': 0,
+                    \   'vcol': 0,
+                    \   'col': 0,
+                    \   'type': 'E',
+                    \   'text': l:text,
+                    \   'nr': -1,
+                    \})
+        continue
+    endif
+
+    let l:line = l:match[2]
+    let l:warning_or_text = l:match[3]
+    let l:text = l:match[4]
+
+    if !empty(l:warning_or_text)
+        let l:type = 'W'
+    else
+        let l:type = 'E'
+    endif
+
+    " vcol is Needed to indicate that the column is a character.
+    call add(l:output, {
+    \   'bufnr': a:buffer,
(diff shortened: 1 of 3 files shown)


ale_linters/erlang/erlc.vim
```diff
@@ -0,0 +1,88 @@
...
+    "
+    " If none of the patterns above match, the line can be ignored
+    if len(l:match) == 0 " not a 'normal' warning or error
+        if len(l:match_parse_transform) == 0 " also not a parse_transform error
+            continue
+        endif
+
+        let l:text = l:match_parse_transform[0]
+        call add(l:output, {
+                    \   'bufnr': a:buffer,
```
w0rp: Use a handling indent for continuing lines, like in the other source files. So you only increase the whitespace by one level in indentation.

ale_linters/erlang/erlc.vim
```diff
@@ -0,0 +1,88 @@
...
+  let l:output = []
+
+  let l:pattern_no_module_definition = '\v(no module definition)$'
+  let l:pattern_unused = '\v(.* is unused)$'
+
+  let l:is_hrl = expand('%:e') ==# 'hrl'
+
+  for l:line in a:lines
+    let l:match = matchlist(l:line, l:pattern)
+    let l:match_parse_transform = matchlist(l:line, l:pattern_parse_transform)
```
w0rp: You should only run this match if the first match fails. Otherwise you are wasting CPU cycles.

ale_linters/erlang/erlc.vim
```diff
@@ -0,0 +1,88 @@
...
+    " 1) normal errors match l:pattern
+    " 2) parse_transform errors match l:pattern_parse_transform
+    "
+    " If none of the patterns above match, the line can be ignored
+    if len(l:match) == 0 " not a 'normal' warning or error
+        if len(l:match_parse_transform) == 0 " also not a parse_transform error
+            continue
+        endif
+
+        let l:text = l:match_parse_transform[0]
```
w0rp: This variable is redundant.

ale_linters/erlang/erlc.vim
```diff
@@ -0,0 +1,88 @@
...
+
+  " parse_transforms are a special case. The error message does not indicate a location:
+  " error.erl: undefined parse transform 'some_parse_transform'
+  let l:pattern_parse_transform = '\v(undefined parse transform .*)$'
+  let l:output = []
+
+  let l:pattern_no_module_definition = '\v(no module definition)$'
+  let l:pattern_unused = '\v(.* is unused)$'
+
+  let l:is_hrl = expand('%:e') ==# 'hrl'
```
w0rp: This `expand` expression assumes that the buffer you are running this command on is the same buffer as the buffer you are checking. This might not be true. You should use the `a:buffer` variable to get the file extension for the buffer number.

ale_linters/erlang/erlc.vim
```diff
@@ -0,0 +1,88 @@
...
+                    \})
+        continue
+    endif
+
+    let l:line = l:match[2]
+    let l:warning_or_text = l:match[3]
+    let l:text = l:match[4]
+
+    " If this file is a header .hrl, ignore the following expected messages:
+    " - 'no module definition'
```
w0rp: This looks like a symptom of being unable to locate other files you are importing. Can `erlc` perform semantic analysis? It would be better to provide some way to determine where the modules live, or to let you set a module path.

evnu: `erlc` does not perform semantic analysis, or at least not to a large degree. Erlang basically distinguishes between two types of files, regular source files and header files. In a header file, the `-module` directive (from which this error comes) can be omitted, as a header file is not compiled by itself. A header is to be included in a regular source file, which cannot omit that directive.

Within `ale`, if we want to check if a `.hrl` file is properly formed, we need to handle it like a regular `.erl` file. The error handling above does that by skipping errors which do not apply to a header file. Another option would be to wrap the header file in a regular file for compiling, but this seems cumbersome.

evnu:
Do you want the fixes as an additional commit or squashed into preexisting commits?

evnu:
I added  9715cb8 to address the issues here, sans the header file work-around. Regarding your comment on `l:text` being redundant, I interpreted that as "just index the list itself". Is that interpretation correct?

w0rp:
@evnu Yeah, it looks good to me. :+1: There's a merge conflict to sort out now, and that's about it.

evnu:
@w0rp great, thanks for the review! I will sort out the merge conflict tomorrow and rebase.

w0rp:
I'll merge this and rename the `erlc_flags` option to `erlc_options` to match the new standard in #257.

PR #262, opened 2017-01-15
Fix SwiftLint

KenjiTakahashi:
1. Should be defined for 'swift' files, not 'swiftlint'.
2. Filepath should be supplied with `--path`, otherwise `swiftlint` cmd
treats is as a subcommand to run.

### DIFF
--- a/ale_linters/swift/swiftlint.vim
+++ b/ale_linters/swift/swiftlint.vim
@@ -1,9 +1,9 @@
 " Author: David Mohundro <david@mohundro.com>
 " Description: swiftlint for swift files
 
-call ale#linter#Define('swiftlint', {
+call ale#linter#Define('swift', {
 \   'name': 'swiftlint',
 \   'executable': 'swiftlint',
-\   'command': g:ale#util#stdin_wrapper . ' .swift swiftlint',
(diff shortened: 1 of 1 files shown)


ale_linters/swift/swiftlint.vim
```diff
@@ -1,9 +1,9 @@
 " Author: David Mohundro <david@mohundro.com>
 " Description: swiftlint for swift files
 
-call ale#linter#Define('swiftlint', {
+call ale#linter#Define('swift', {
 \   'name': 'swiftlint',
 \   'executable': 'swiftlint',
-\   'command': g:ale#util#stdin_wrapper . ' .swift swiftlint',
+\   'command': g:ale#util#stdin_wrapper . ' .swift swiftlint lint --path',
```
w0rp: I wonder, swiftlint apparently supports reading from stdin. Can we remove the use of `stdin_wrapper` here and use `--use-stdin` instead? That will likely work better.

KenjiTakahashi: Right, I must have missed it when going through the docs. Updated now, appears to work fine as well.

w0rp:
Looks good to me. :+1:

PR #265, opened 2017-01-18
Fix using variable g:ale_python_mypy_options that was not defined

aswins:


w0rp:
Cheers! You could have also forced a push to your previous branch to update the old pull request.

PR #266, opened 2017-01-19
Add option to open loclist/quicklist when there are errors

yfery:
I copied PR #137, and tries to complete it by correcting some issues and
adding vader tests.

About tests, first time with vader, can you give some feedback if this
is what you expected in PR #137.


autoload/ale/list.vim
```diff
@@ -0,0 +1,46 @@
...
+        " errors, close the window.' but that doesn't seem to work that way
+        " lclose seems to be needed
+        if g:ale_set_quickfix
+            cwindow
+            " Only for vader testing
+            let g:ale_quickfix_opened = 1
+        elseif g:ale_set_loclist
+            lwindow
+            " Only for vader testing
+            let g:ale_loclist_opened = 1
```
w0rp: You shouldn't set variables to test, you should instead test if the action you performed actually happened. Testing if a variable is set like this doesn't really test anything useful.

I'm not sure myself, but I'd look into some way of calling the standard vim functions and so on to see if the window was opened.

autoload/ale/list.vim
```diff
@@ -0,0 +1,46 @@
...
+        endif
+
+        " Only close if the list is totally empty (relying on Vim's state, not our
+        " own). This keeps us from closing the window when other plugins have
+        " populated it.
+    elseif g:ale_set_quickfix && len(getqflist()) == 0
+        cclose
+    elseif len(getloclist(0)) == 0
+        lclose
+    endif
```
w0rp: We should probably create an option for keeping the window open, and default the option so it will automatically close the window. That will have to be tested too.

autoload/ale/engine.vim
```diff
@@ -144,8 +144,9 @@ function! s:HandleExit(job) abort
     let g:ale_buffer_info[l:buffer].loclist = g:ale_buffer_info[l:buffer].new_loclist
     let g:ale_buffer_info[l:buffer].new_loclist = []
 
-    if g:ale_set_loclist
-        call setloclist(0, g:ale_buffer_info[l:buffer].loclist)
+    if g:ale_set_quickfix || g:ale_set_loclist
+"         call setloclist(0, g:ale_buffer_info[l:buffer].loclist)
+call ale#list#SetLists(g:ale_buffer_info[l:buffer].loclist)
```
w0rp: The indentation is wrong here.

autoload/ale/engine.vim
```diff
@@ -144,8 +144,9 @@ function! s:HandleExit(job) abort
     let g:ale_buffer_info[l:buffer].loclist = g:ale_buffer_info[l:buffer].new_loclist
     let g:ale_buffer_info[l:buffer].new_loclist = []
 
-    if g:ale_set_loclist
-        call setloclist(0, g:ale_buffer_info[l:buffer].loclist)
+    if g:ale_set_quickfix || g:ale_set_loclist
+"         call setloclist(0, g:ale_buffer_info[l:buffer].loclist)
```
w0rp: Don't comment out old code, replace it. Changes can be easily reverted with `git`.

yfery: I will fix it

autoload/ale/engine.vim
yfery: I will fix it too

autoload/ale/list.vim
yfery: I will add the option: ale_keep_open_list

autoload/ale/list.vim
yfery: You're right. 

Before using those variables, I search a lot to find some way to test if loclist is opened or closed but I wasn't able to.
I'll keep looking a better way.

autoload/ale/list.vim
w0rp: Could you name it `g:ale_keep_list_window_open`? That should read better.

w0rp:
The code looks good. :+1: I'll document the new options and use some sentences for the `Execute` blocks to better explain them.

yfery:
Thanks! \o/

PR #268, opened 2017-01-20
Fix parsing of currently shown signs with Spanish locale

EinfachToll:
Usually, the sign column should disappear when a linter is run and it has nothing to complain about. This stopped working for me with commit https://github.com/w0rp/ale/commit/8df2444ec487903cbb7018a3550e71ab7f006288 . Not sure why, but when using the Spanish locale, there is only one space instead of two which separates the columns. This patch fixes the parsing.


EinfachToll:
Whoops, I didn't notice there is a test.

w0rp:
Cheers! :beers:

PR #270, opened 2017-01-22
Fix the go build linter

### LINKED ISSUE #234: Go build linter doesn't work
During #180 it was discussed that go build can't work properly unless you copy across all the go files in the package.

This was seemingly then forgotten and the PR merged without it being addressed.

As it is,  the go build linter doesn't work for packages with more than 1 file.
It also doesn't work for test files,  since the go build command ignores them.

DiscoViking:
Fixes #234 

In order to get around the fact that "go build" wants all your files to be in one directory,  and we don't want to copy all the files to the temp directory,  I've worked around it by building up a command using the underlying "go compile" to achieve the same result.

Open Issues:

- Won't work if your GOPATH has multiple directories.
- Won't work if your package contains .c/.h files (ie. not pure go).
- I've only tested it on OSX.  I don't have another system to test on :(
- Dependent modules must be already built.  It will not compile dependencies,  and expects them to already be present in the $GOPATH/pkg directory.

Even with those problems,  I believe this will work in MOST cases.

Things I DID test and definitely do work:

- Linting a single file.
- Linting a _test.go test file.
- Linting multiple files in different splits (and from different packages).

### DIFF
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -29,11 +29,27 @@ function! ale_linters#go#gobuild#GetCommand(buffer) abort
     let l:goarch = substitute(l:goarch, '\n\+$', '', '')
 
     " Finally build the import path.
-    " TODO: Handle GOPATHs that contain multiple directories.
-    let l:gopath = $GOPATH
-    let l:import_path = l:gopath . '/pkg/' . l:goos . '_' . l:goarch
+    "
+    " From the output of 'go help gopath':
+    " On Unix, the value is a colon-separated string.
+    " On Windows, the value is a semicolon-separated string.
+    " On Plan 9, the value is a list.
+    let l:unix_oses = ['android', 'darwin', 'dragonfly', 'freebsd', 'linux',
+                      \'nacl', 'netbsd', 'openbsd', 'solaris']
+    if index(l:unix_oses, l:goos) >= 0
+      let l:gopaths = split($GOPATH, ':')
+    elseif l:goos == 'windows'
+      let l:gopaths = split($GOPATH, ';')
+    elseif l:goos = 'plan9'
+      " No idea if vim handles list-type environment variables properly.
+      let l:gopaths = $GOPATH
+    else
+      echoerr 'Unknown value for GOOS: ' . l:goos
+    endif
 
-    return g:ale#util#stdin_wrapper . ' .go go tool compile -I ' . l:import_path . ' -o /dev/null ' . join(l:all_files)
+    let l:import_args = map(l:gopaths, '''-I '' . v:val . ''/pkg/'' . l:goos . ''_'' . l:goarch')
+
+    return g:ale#util#stdin_wrapper . ' .go go tool compile ' . join(l:import_args) . ' -o /dev/null ' . join(l:all_files)
 endfunction
 
 call ale#linter#Define('go', {
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -1,7 +1,7 @@
 " Author: dzhou121 <dzhou121@gmail.com>, Ryan Norris <rynorris@gmail.com>
 " Description: go build for Go files
 
-function! s:FindGobuildScript() abort
+function! ale_linters#go#gobuild#GetCommand(buffer) abort
     " Get absolute path to the directory containing the current file.
     " This directory by definition contains all the files for this go package.
     let l:this_package = expand('%:p:h')
@@ -36,13 +36,10 @@ function! s:FindGobuildScript() abort
     return g:ale#util#stdin_wrapper . ' .go go tool compile -I ' . l:import_path . ' -o /dev/null ' . join(l:all_files)
 endfunction
 
-let g:ale#util#gobuild_script =
-\   get(g:, 'ale_go_gobuild_script', s:FindGobuildScript())
-
 call ale#linter#Define('go', {
 \   'name': 'go build',
 \   'output_stream': 'stdout',
 \   'executable': 'go',
-\   'command': g:ale#util#gobuild_script,
+\   'command_callback': 'ale_linters#go#gobuild#GetCommand',
 \   'callback': 'ale#handlers#HandleUnixFormatAsError',
 \})
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -1,4 +1,4 @@
-" Author: dzhou121 <dzhou121@gmail.com>
+" Author: dzhou121 <dzhou121@gmail.com>, Ryan Norris <rynorris@gmail.com>
 " Description: go build for Go files
 
 function! s:FindGobuildScript() abort
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -14,12 +14,12 @@ function! ale_linters#go#gobuild#GetCommand(buffer) abort
     " We'll then pass this list to go compile and stdin_wrapper will add on
     " the temporary version of the current file.
     let l:this_file = expand('%:p:t')
-    call filter(l:all_files, 'fnamemodify(v:val, ":t") != l:this_file')
+    call filter(l:all_files, 'fnamemodify(v:val, '':t'') != l:this_file')
 
     " We're also going to need the system information in order to find the
     " correct import directory.  We'll pull these from go itself since then
     " we're guaranteed they'll match what it's looking for.
-    " The version string is of the form "go version go1.6.2 darwin/amd64".
+    " The version string is of the form 'go version go1.6.2 darwin/amd64'.
     let l:go_version_string = system('go version')
     let l:version_part = substitute(l:go_version_string, '.* ', '', '')
     let l:goos = substitute(l:version_part, '/.*', '', '')
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -7,6 +7,7 @@ function! s:FindGobuildScript() abort
     let l:this_package = expand('%:p:h')
 
     " Get a listing of all go files in the directory.
+    " TODO: Handle packages that contain c files.
     let l:all_files = globpath(l:this_package, '*.go', 1, 1)
 
     " Filter out the current file since we don't want to include it twice.
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -2,9 +2,12 @@
 " Description: go build for Go files
 
 function! ale_linters#go#gobuild#GetCommand(buffer) abort
+    " Get name of buffer the command is being run in.
+    let l:this_bufname = bufname(a:buffer)
+    "
     " Get absolute path to the directory containing the current file.
     " This directory by definition contains all the files for this go package.
-    let l:this_package = expand('%:p:h')
+    let l:this_package = fnamemodify(l:this_bufname, ':p:h')
 
     " Get a listing of all go files in the directory.
     " TODO: Handle packages that contain c files.
@@ -13,7 +16,7 @@ function! ale_linters#go#gobuild#GetCommand(buffer) abort
     " Filter out the current file since we don't want to include it twice.
     " We'll then pass this list to go compile and stdin_wrapper will add on
     " the temporary version of the current file.
-    let l:this_file = expand('%:p:t')
+    let l:this_file = fnamemodify(l:this_bufname, ':p:t')
     call filter(l:all_files, 'fnamemodify(v:val, '':t'') != l:this_file')
 
     " We're also going to need the system information in order to find the
@@ -34,17 +37,11 @@ function! ale_linters#go#gobuild#GetCommand(buffer) abort
     " On Unix, the value is a colon-separated string.
     " On Windows, the value is a semicolon-separated string.
     " On Plan 9, the value is a list.
-    let l:unix_oses = ['android', 'darwin', 'dragonfly', 'freebsd', 'linux',
-                      \'nacl', 'netbsd', 'openbsd', 'solaris']
-    if index(l:unix_oses, l:goos) >= 0
+    if has('unix')
       let l:gopaths = split($GOPATH, ':')
-    elseif l:goos ==# 'windows'
-      let l:gopaths = split($GOPATH, ';')
-    elseif l:goos ==# 'plan9'
-      " No idea if vim handles list-type environment variables properly.
-      let l:gopaths = $GOPATH
     else
-      echoerr 'Unknown value for GOOS: ' . l:goos
+      " Assume windows,  just ignore Plan 9.
+      let l:gopaths = split($GOPATH, ';')
     endif
 
     let l:import_args = map(l:gopaths, '''-I '' . v:val . ''/pkg/'' . l:goos . ''_'' . l:goarch')
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -2,7 +2,37 @@
 " Description: go build for Go files
 
 function! s:FindGobuildScript() abort
-    return g:ale#util#stdin_wrapper . ' .go go build -o /dev/null'
+    " Get absolute path to the directory containing the current file.
+    " This directory by definition contains all the files for this go package.
+    let l:this_package = expand('%:p:h')
+
+    " Get a listing of all go files in the directory.
+    let l:all_files = globpath(l:this_package, '*.go', 1, 1)
+
+    " Filter out the current file since we don't want to include it twice.
+    " We'll then pass this list to go compile and stdin_wrapper will add on
+    " the temporary version of the current file.
+    let l:this_file = expand('%:p:t')
+    call filter(l:all_files, 'fnamemodify(v:val, ":t") != l:this_file')
+
+    " We're also going to need the system information in order to find the
+    " correct import directory.  We'll pull these from go itself since then
+    " we're guaranteed they'll match what it's looking for.
+    " The version string is of the form "go version go1.6.2 darwin/amd64".
+    let l:go_version_string = system('go version')
+    let l:version_part = substitute(l:go_version_string, '.* ', '', '')
+    let l:goos = substitute(l:version_part, '/.*', '', '')
+    let l:goarch = substitute(l:version_part, '.*/', '', '')
+
+    " goarch comes with a troublesome trailing newline,  so strip that off.
+    let l:goarch = substitute(l:goarch, '\n\+$', '', '')
+
+    " Finally build the import path.
+    " TODO: Handle GOPATHs that contain multiple directories.
+    let l:gopath = $GOPATH
+    let l:import_path = l:gopath . '/pkg/' . l:goos . '_' . l:goarch
+
+    return g:ale#util#stdin_wrapper . ' .go go tool compile -I ' . l:import_path . ' -o /dev/null ' . join(l:all_files)
 endfunction
 
 let g:ale#util#gobuild_script =
@@ -10,7 +40,7 @@ let g:ale#util#gobuild_script =
 
 call ale#linter#Define('go', {
 \   'name': 'go build',
-\   'output_stream': 'stderr',
+\   'output_stream': 'stdout',
 \   'executable': 'go',
 \   'command': g:ale#util#gobuild_script,
 \   'callback': 'ale#handlers#HandleUnixFormatAsError',
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -38,9 +38,9 @@ function! ale_linters#go#gobuild#GetCommand(buffer) abort
                       \'nacl', 'netbsd', 'openbsd', 'solaris']
     if index(l:unix_oses, l:goos) >= 0
       let l:gopaths = split($GOPATH, ':')
-    elseif l:goos == 'windows'
+    elseif l:goos ==# 'windows'
       let l:gopaths = split($GOPATH, ';')
-    elseif l:goos == 'plan9'
+    elseif l:goos ==# 'plan9'
       " No idea if vim handles list-type environment variables properly.
       let l:gopaths = $GOPATH
     else
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -49,7 +49,7 @@ function! ale_linters#go#gobuild#GetCommand(buffer) abort
 
     let l:import_args = map(l:gopaths, '''-I '' . v:val . ''/pkg/'' . l:goos . ''_'' . l:goarch')
 
-    return g:ale#util#stdin_wrapper . ' .go go tool compile ' . join(l:import_args) . ' -o /dev/null ' . join(l:all_files)
+    return g:ale#util#stdin_wrapper . ' .go go tool compile ' . join(l:import_args) . ' -o '. g:ale#util#nul_file . join(l:all_files)
 endfunction
 
 call ale#linter#Define('go', {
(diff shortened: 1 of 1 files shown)


DiscoViking:
Travis build is failing because of this Vint bug: https://github.com/Kuniwak/vint/issues/121

Not sure what to do about that.  AFAIK I just need the double-quotes inside the single-quoted expression. :/

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,45 @@
...
-\   get(g:, 'ale_go_gobuild_script', s:FindGobuildScript())
+    " goarch comes with a troublesome trailing newline,  so strip that off.
+    let l:goarch = substitute(l:goarch, '\n\+$', '', '')
+
+    " Finally build the import path.
+    " TODO: Handle GOPATHs that contain multiple directories.
+    let l:gopath = $GOPATH
+    let l:import_path = l:gopath . '/pkg/' . l:goos . '_' . l:goarch
+
+    return g:ale#util#stdin_wrapper . ' .go go tool compile -I ' . l:import_path . ' -o /dev/null ' . join(l:all_files)
```
renstrom: Might want to use `g:ale#util#nul_file` instead of `/dev/null`.

https://github.com/w0rp/ale/blob/master/autoload/ale/util.vim#L21-L26

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,45 @@
...
 
-let g:ale#util#gobuild_script =
-\   get(g:, 'ale_go_gobuild_script', s:FindGobuildScript())
+    " goarch comes with a troublesome trailing newline,  so strip that off.
+    let l:goarch = substitute(l:goarch, '\n\+$', '', '')
+
+    " Finally build the import path.
+    " TODO: Handle GOPATHs that contain multiple directories.
+    let l:gopath = $GOPATH
+    let l:import_path = l:gopath . '/pkg/' . l:goos . '_' . l:goarch
```
renstrom: I suck a Vimscript. But would something like this work?

```vim
let l:import_path = ''
for gopath in split($GOPATH, ':')
  l:import_path += ' -I ' . gopath . '/pkg/' . l:goos . '_' . l:goarch
endfor

**Comment by osleg** [comment]
Worked perfectly for me!

File: ale_linters/go/gobuild.vim (hunk 2 above)
Comment (DiscoViking) [inline ale_linters/go/gobuild.vim]: Yep.  I've gone ahead and implemented this (but using map/join).
The slightly annoying thing is handling how GOPATH is handled differently on different OSes.

File: ale_linters/go/gobuild.vim (hunk 1 above)
Comment (DiscoViking) [inline ale_linters/go/gobuild.vim]: Nice catch.  👍  Fixed.

File: ale_linters/go/gobuild.vim [hunk 3]
```diff
@@ -1,17 +1,61 @@
...
+" Author: dzhou121 <dzhou121@gmail.com>, Ryan Norris <rynorris@gmail.com>
 " Description: go build for Go files
 
-function! s:FindGobuildScript() abort
-    return g:ale#util#stdin_wrapper . ' .go go build -o /dev/null'
-endfunction
+function! ale_linters#go#gobuild#GetCommand(buffer) abort
+    " Get absolute path to the directory containing the current file.
+    " This directory by definition contains all the files for this go package.
+    let l:this_package = expand('%:p:h')
```
w0rp: This expression gives you the path to the file you are looking at, but this file could be different from the buffer the command is being run for. You should get the filename for the buffer instead.

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,61 @@
...
+
+    " Get a listing of all go files in the directory.
+    " TODO: Handle packages that contain c files.
+    let l:all_files = globpath(l:this_package, '*.go', 1, 1)
+
+    " Filter out the current file since we don't want to include it twice.
+    " We'll then pass this list to go compile and stdin_wrapper will add on
+    " the temporary version of the current file.
+    let l:this_file = expand('%:p:t')
```
w0rp: Same as the above.

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,61 @@
...
+    call filter(l:all_files, 'fnamemodify(v:val, '':t'') != l:this_file')
+
+    " We're also going to need the system information in order to find the
+    " correct import directory.  We'll pull these from go itself since then
+    " we're guaranteed they'll match what it's looking for.
+    " The version string is of the form 'go version go1.6.2 darwin/amd64'.
+    let l:go_version_string = system('go version')
```
w0rp: Can this command take some time to complete? If this command can hang for some time, then you will have to implement this using the unfinished command chaining feature described here: https://github.com/w0rp/ale/issues/124

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,61 @@
...
+
+    " Finally build the import path.
+    "
+    " From the output of 'go help gopath':
+    " On Unix, the value is a colon-separated string.
+    " On Windows, the value is a semicolon-separated string.
+    " On Plan 9, the value is a list.
+    let l:unix_oses = ['android', 'darwin', 'dragonfly', 'freebsd', 'linux',
+                      \'nacl', 'netbsd', 'openbsd', 'solaris']
+    if index(l:unix_oses, l:goos) >= 0
```
w0rp: Wouldn't it be enough to use `has('unix')` here?

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,61 @@
...
+      let l:gopaths = split($GOPATH, ':')
+    elseif l:goos ==# 'windows'
+      let l:gopaths = split($GOPATH, ';')
+    elseif l:goos ==# 'plan9'
```
w0rp: Does Plan 9 even matter? Who uses Plan 9 except maybe Rob Pike?

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,61 @@
...
+      " No idea if vim handles list-type environment variables properly.
+      let l:gopaths = $GOPATH
+    else
+      echoerr 'Unknown value for GOOS: ' . l:goos
+    endif
+
+    let l:import_args = map(l:gopaths, '''-I '' . v:val . ''/pkg/'' . l:goos . ''_'' . l:goarch')
+
+    return g:ale#util#stdin_wrapper . ' .go go tool compile ' . join(l:import_args) . ' -o '. g:ale#util#nul_file . join(l:all_files)
```
w0rp: You should check out the use of `nul_file` here on Windows. I had problems before with `-o nul` where it didn't work on Windows, but curiously `-o /dev/null` did, even though `/dev/null` isn't a thing on Windows. I can't remember which program that was for.

osleg:
Found an issue with this PR
Given 3 files when an error found in 1 file it marks the lines in all 3 files, eg:
main.go
```
package main

type A struct {  <<< marked as error
  a map[string]func([]byte, chan bool)
}
```

handlers.go 
```
package main

checks := make(map[string]funct([]byte))  <<< Marked as err and an actuall error indeed
```

sub.go
```
package main

func B(b []byte, c chan bool) {  <<< marked as error
  // some func
}
```

ale_linters/go/gobuild.vim
osleg: I think this is the cause for the error I posted below?

ale_linters/go/gobuild.vim
DiscoViking: Wait,  does Vim even support Plan 9?  I don't see it anywhere on their downloads page.
You're right, can probably just ignore it.

ale_linters/go/gobuild.vim
DiscoViking: All it does is print out some strings which are hard-coded into the binaries.  It's always instant.

ale_linters/go/gobuild.vim
DiscoViking: Yep!  Much cleaner.

DiscoViking:
@osleg can you please check if the latest commit has fixed the problem you were having?

ale_linters/go/gobuild.vim
DiscoViking: I don't have a Windows machine to test on sadly. :(

osleg:
@DiscoViking it actually became worse
There is no errors in my files, the project compiles perfectly but ale now reports errors Undefined: <someType> and showing them on totally unrelated lines of code
using commit 8468286

DiscoViking:
@osleg that's super weird.  which OS are you running on?  It's still working fine for me on OSX.

osleg:
I'm on linux
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Jan 11 2017 03:42:35)
Included patches: 1-170

also here a screenshot of  how it looks like

DiscoViking:
Is this a public project I can clone and take a look at?  I can't repro this on anything I have already.

I'm going to have to keep looking at this another day.  It's like 2AM here haha 😄

osleg:
I can make it public right now. it's 1:30AM here hehe :)

osleg:
https://github.com/WallDev/PhishermansFiendServer

ale_linters/go/gobuild.vim
w0rp: Okay. I ask because the `dub` command I use to get import paths for a project for the D linter can some times take a lot of time to complete. It was a similar scenario.

ale_linters/go/gobuild.vim
w0rp: Maybe leave it until a Windows user complains then.

w0rp: Actually! This is in fact the very same linter I had that problem with, so change it back to `-o /dev/null`. Believe it or not, that works on Windows, but `-o nul` does not.

osleg:
@DiscoViking I found where the issue! I just had 2 files out of 3 open in the vim so the actually there is 2  issues 
1.  messed up files, it shows errors in wrong file
2. the "Undefined: " error returned

ale_linters/go/gobuild.vim
DiscoViking: Will do!

DiscoViking:
@osleg thanks for bisecting and finding the breaking commit!

That's a weird one though.  That commit should have literally no effect on Unix systems.

I checked out your repo and confirmed it didn't work before,  but does after.  Can you please confirm?

osleg:
if there is a fix for it i can try tomorrow

DiscoViking:
@osleg hey,  did you get a chance to confirm the issue you were having is fixed?

osleg:
Hey sorry got a lot on my head, had no time for that.
I checked it now, first of all there is a slight but feelable lag in vim
but it might be something on my side, i will check more throughly later.
The problem with errors in wrong buffers still persist though, it shown
in all the buffer at once

w0rp:
@osleg If you could figure out what is causing the lag, let us know. The `system` could slow things down.

osleg:
@w0rp @DiscoViking 
Hey, I've tested with clean config just to make sure.
The bug is still here, errors are shown in all the buffers even for files that are not open currently in vim

And here are the longest running ale functions that I found in profile, though I'm not sure the lag is still there as I'm testing on server that is half world away from me right now ^_^
```
FUNCTION  ale_linters#go#gobuild#GetCommand()
    1 Called 21 times
    2 Total time:   0.923261
    3  Self time:   0.019579
    4
    5 count  total (s)   self (s)
    6                                 " Get name of buffer the command is being run in.
    7    21              0.000193     let l:this_bufname = bufname(a:buffer)
    8                                 "
    9                                 " Get absolute path to the directory containing the current file.
   10                                 " This directory by definition contains all the files for this go package.
   11    21              0.000739     let l:this_package = fnamemodify(l:this_bufname, ':p:h')
   12 ••••••••••••••••••••••••••••
   13                                 " Get a listing of all go files in the directory.
   14                                 " TODO: Handle packages that contain c files.
   15    21              0.004190     let l:all_files = globpath(l:this_package, '*.go', 1, 1)
   16 ••••••••••••••••••••••••••••
   17                                 " Filter out the current file since we don't want to include it twice.
   18                                 " We'll then pass this list to go compile and stdin_wrapper will add on
   19                                 " the temporary version of the current file.
   20    21              0.000491     let l:this_file = fnamemodify(l:this_bufname, ':p:t')
   21    21              0.000711     call filter(l:all_files, 'fnamemodify(v:val, '':t'') != l:this_file')
   22 ••••••••••••••••••••••••••••
   23                                 " We're also going to need the system information in order to find the
   24                                 " correct import directory.  We'll pull these from go itself since then
   25                                 " we're guaranteed they'll match what it's looking for.
   26                                 " The version string is of the form 'go version go1.6.2 darwin/amd64'.
   27    21   0.908169   0.004487     let l:go_version_string = system('go version')
   28    21              0.002401     let l:version_part = substitute(l:go_version_string, '.* ', '', '')
   29    21              0.000653     let l:goos = substitute(l:version_part, '/.*', '', '')
   30    21              0.001114     let l:goarch = substitute(l:version_part, '.*/', '', '')
   31 ••••••••••••••••••••••••••••
   32                                 " goarch comes with a troublesome trailing newline,  so strip that off.
   33    21              0.000535     let l:goarch = substitute(l:goarch, '\n\+$', '', '')
   34 ••••••••••••••••••••••••••••
   35                                 " Finally build the import path.
   36                                 "
   37                                 " From the output of 'go help gopath':
   38                                 " On Unix, the value is a colon-separated string.
   39                                 " On Windows, the value is a semicolon-separated string.
   40                                 " On Plan 9, the value is a list.
   41    21              0.000310     if has('unix')
   42    21              0.000581       let l:gopaths = split($GOPATH, ':')
   43    21              0.000097     else
   44                                   " Assume windows,  just ignore Plan 9.
   45                                   let l:gopaths = split($GOPATH, ';')
   46                                 endif
   47 ••••••••••••••••••••••••••••
   48    21              0.000536     let l:import_args = map(l:gopaths, '''-I '' . v:val . ''/pkg/'' . l:goos . ''_'' . l:goarch')
   49 ••••••••••••••••••••••••••••
   50    21              0.000799     return g:ale#util#stdin_wrapper . ' .go go tool compile ' . join(l:import_args) . ' -o /dev/null ' . join(l:all_files)
   51
```
and
```
FUNCTION  ale#Lint()
    1 Called 21 times
    2 Total time:   1.422309
    3  Self time:   0.025590
    4
    5 count  total (s)   self (s)
    6                                 " Do nothing for blacklisted files.
    7    21              0.000796     if index(g:ale_filetype_blacklist, &filetype) >= 0
    8                                     return
    9                                 endif
   10 ••••••••••••••••••••••••••••
   11    21              0.000398     let l:buffer = bufnr('%')
   12    21   0.008215   0.000378     let l:linters = ale#linter#Get(&filetype)
   13 ••••••••••••••••••••••••••••
   14                                 " Initialise the buffer information if needed.
   15    21   0.002410   0.000969     call ale#engine#InitBufferInfo(l:buffer)
   16 ••••••••••••••••••••••••••••
   17                                 " Clear the new loclist again, so we will work with all new items.
   18    21              0.000254     let g:ale_buffer_info[l:buffer].new_loclist = []
   19 ••••••••••••••••••••••••••••
   20   105              0.001598     for l:linter in l:linters
   21                                     " Check if a given linter has a program which can be executed.
   22    84              0.000882         if has_key(l:linter, 'executable_callback')
   23                                         let l:executable = ale#util#GetFunction(l:linter.executable_callback)(l:buffer)
   24                                     else
   25    84              0.001024             let l:executable = l:linter.executable
   26    84              0.000183         endif
   27 ••••••••••••••••••••••••••••
   28    84              0.005874         if !executable(l:executable)
   29                                         " The linter's program cannot be executed, so skip it.
   30                                         continue
   31                                     endif
   32 ••••••••••••••••••••••••••••
   33    84   1.394812   0.007371         call ale#engine#Invoke(l:buffer, l:linter)
   34    84              0.001027     endfor
   35
```
and
```
FUNCTION  ale#engine#Invoke()
    1 Called 84 times
    2 Total time:   1.387441
    3  Self time:   0.010112
    4
    5 count  total (s)   self (s)
    6                                 " Stop previous jobs for the same linter.
    7    84   0.022314   0.001628     call s:StopPreviousJobs(a:buffer, a:linter)
    8    84   1.364719   0.008076     call s:InvokeChain(a:buffer, a:linter, 0, [])
    9
```

w0rp:
Thank you for the profiler results. :+1: Profiling is the only way to go for debugging performance issues. I had to read up and down the profiler results before I confirmed what I suspected.

```
 27    21   0.908169   0.004487     let l:go_version_string = system('go version')
```

The system call to check the Go version runs too slowly. You'll have to speed this up by running the system call asynchronously, and to do that, you'll need the command chaining feature I had been working on. The purpose of that feature is to allow you to run linters by linking several asynchronous commands together, so you run `go version` in the background first, read the output from that command, and then run `go build` in the background.

I was working on this last night, and I was trying to replace the special D Bash script with it. If I can get D code to build with the feature, then I'll consider it stable enough to merge to `master`, and then it can be used here. The linter definition would like something like this:

```vim
function! ale_linters#go#gobuild#GoVersionCheck(buffer) abort
    " Return the command to run go version here
endfunction

function! ale_linters#go#gobuild#GoBuild(buffer, go_version_output) abort
    " Read the go version output here, and return the command for go build
endfunction
...

\   'command_chain': [
\       {'callback': 'ale_linters#go#gobuild#GoVersionCheck'},
\       {'callback': 'ale_linters#go#gobuild#GoBuild'},
\   ],
```

ale_linters/go/gobuild.vim
renstrom: Why is the version necessary at all? Also, couldn't you just define it once instead of on every call to `ale_linters#go#gobuild#GetCommand()`?

joshuarubin:
I've tried addressing some issues I found with this PR in #297. I have not profiled it yet, but I'm sure it could benefit from chaining as there are 3 `system` commands before the linter is run.

I tried to do the chaining as you showed above but the master branch throws an error when preprocessing the linter because it doesn't have a check for `command_chain`. The command-chain branch is now pretty old. Is there some other branch I should be using?

w0rp:
@joshuarubin The implementation of chaining doesn't exist in the master branch yet. I was working on it last night, and I'm not ready to merge it yet. I might see if I can finish it tonight.

osleg:
@w0rp why would anyone want to run `--version` (or system calls that presumable return same result at each run) before every linter call?
Wouldn't it be better and faster to run it once and to save the version string in a variable?

w0rp:
https://github.com/w0rp/ale/issues/124

I just implemented the command chaining, finally, which I had been working on a bit-by-bit for the past few months. The DMD linter for D now shows an example of its use: https://github.com/w0rp/ale/blob/master/ale_linters/d/dmd.vim#L92

A similar thing can be done now for this linter.

I suppose the Go version check could be run once on startup, but I'm not sure how to make running it once on startup work without some difficulty. You would need to run the command in a separate job, which would require Vim and NeoVim compatibility, and then somehow delay running the Go linter until the output of the `go version` command has been received, etc. Running it every time will work, and it won't cause input lag if the command is run in a job, although it will delay showing errors and warnings slightly.

w0rp:
@DiscoViking @joshuarubin I'll let you two work out how to best finish off these adjustments to the `go build` linter. Between your changes and some extra tweaks with the support for chaining commands, which I just added, I feel like getting this linter to work reasonably well might not be far off now.

joshuarubin:
@w0rp honestly, the `go tool compile` approach will never work as well as most other linters. There are serious restrictions about what can be done (e.g. `cgo` can't be used).

The best way, IMO, would be to do the same thing that `neomake` does, by providing the linters a way to set the `cwd` for the executable to be run in. `neovim`'s job functionality can do this natively, but `vim` cannot.

Here is how `neomake` wrapped that capability to work with both: https://github.com/neomake/neomake/blob/master/autoload/neomake.vim#L145-L243

And here is how simple the `go build` linter could be with that capability: https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/go.vim#L11-L27

Basically, just running `go test -c -o /dev/null` from the proper directory.

As far as I can tell, there isn't a reliable way, with `ale`, to run a command immediately after the job is queued and before results come in. That capability could also make it possible for linters to `cd` on their own as needed without breaking other things.

w0rp:
@joshuarubin Changing directory for any linter in ALE is easy. Just write the command `'cd /some/path && other_command'`. All commands are executed in a subshell, so `cd` works.

I don't know about running a job immediately after something is queued, or what that entails. Jobs start running after a delay while typing, or shortly after a buffer is opened, etc.

It's important to note that ALE differs from other Vim linter plugins in one very significant way. It checks the text _in your Vim buffer_, not the file on disk. It's very easy to get this working correctly for tools like `eslint` with JavaScript, and more advantageous there, and harder for statically compiled languages like Go. If you can get it to work, it means you can check for errors in the file as you type, rather than just when you save the file.

joshuarubin:
@w0rp thanks, I do keep forgetting that...

The concept would still be the same, but we'd have to copy all the files for a particular go package in addition the the buffered data to a temp dir (with`$GOPATH` structure) (e.g. `${tmpdir}/src/github.com/w0rp/ale`)

We'd then have to update the `$GOPATH` that the linter's executable sees somehow. One possibility is just to run something like this `GOPATH="$tmpdir:${GOPATH}" go test -c -o /dev/null github.com/w0rp/ale`.

I'll see what I can come up with later tonight.

w0rp:
I'll see if I can fix up what's in this PR to use the command chaining feature in the meantime, so some small progress can be made. That should fix the slowness in running this as it is.

joshuarubin:
There are some significant problems with this PR still.

For example, it can't discern if a `.go` file is actually a `cgo` file that can't be compiled for `go tool compile` which causes that tool to fail.

Also, it depends on the `$GOPATH` env var being explicitly set (which go 1.8, due out in a few days, will not require).

It ignores `$GOROOT` entirely.

Using a command like `go env GOPATH GOROOT GOOS GOARCH` should make it possible to always get the correct values (and should only have to be run once per vim session). My PR splits this into 2 commands, but this would be even more efficient.

Mine uses `go list`, which in turn causes it to depend on vim's python host to parse the json. This enables the linter to determine which files to copy and if there are any `cgo` or other unusual files.

This should be unnecessary, since we want to run `go test -c` instead of `go tool compile`. We'll need to copy the complete contents of the directory into a tmpdir (with `$GOPATH` structure), but if there are large files in that dir, it could be slow, in which case we could still use `go list` to build a list of files to copy, including `cgo` files. Not sure which is worse. Not sure if we want to introduce the python json-parsing dependency.

In essence, there are enough changes required to both this PR and my PR that I would suggest you hold off on fixing it up since so much would likely have to be changed.

w0rp:
Okay then. I'm not too familiar with Go and its tools, but it sounds like the problem to solve here is kind of similar to the problem I had to solve for D. There are two steps for that linter.

1. https://github.com/w0rp/ale/blob/master/ale_linters/d/dmd.vim#L35 Run `dub describe --import-paths` in the project root, containing the configuration file, which lists the locations for every source file, including those in the project itself.
2. https://github.com/w0rp/ale/blob/master/ale_linters/d/dmd.vim#L49 Run `dmd` with the import paths taken from the output of that command, so it knows where to look for all of the files.

This works pretty well for checking D code, including semantic analysis.

It sounds to me like `go list` might serve a similar function for Go code to what `dub describe --import-paths` does for D code. If that's the case, then the solution might be similar.

w0rp:
If you need to sometimes run commands like `go version`, etc. too, you can cache the results of the command after running them once, etc., and skip running those commands the second time the linter is run. The `flake8` linter does this now: https://github.com/w0rp/ale/blob/master/ale_linters/python/flake8.vim#L23

Returning an empty string from a callback for a single command in a chain will immediately skip running that command and run the next command in the chain instead, with an empty `List` of output. (Or input, depending on your perspective.)

w0rp:
Oh, I'll also note, you don't have to rely on Python to parse JSON. You can do that in either Vim8 or NeoVim with the `json_decode` function. There is also `json_encode`, of course.

joshuarubin:
I've updated my PR to use chained commands and `go test -c` which should work with all permutations of Go sources including cgo.

w0rp:
Cool. I'll close this pull request now. We can work on the other one.

PR #273, opened 2017-01-23
Add ALEInfo command to get list of available/enabled linters

### LINKED ISSUE #228: looking for some command can show Available checkers and Currently enabled checkers
Hello, 

is there some command can show Available checkers and Currently enabled checkers?
just like `:SyntasticInfo` of Syntastic.
thx~

DiscoViking:
Fixes #228 by adding a direct analog of SyntasticInfo which displays list of available and enabled linters.

Example output:

```
:ALEInfo
 Current Filetype: go
Available Linters: ['go build', 'gofmt', 'golint', 'go vet']
  Enabled Linters: ['go build']
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ test: test-setup
 		echo "Running tests for $$vim"; \
 		echo '========================================'; \
 		echo; \
-		docker run -it $(DOCKER_FLAGS) $$vim '+Vader! test/*' || EXIT=$$?; \
+		docker run -a stderr $(DOCKER_FLAGS) $$vim '+Vader! test/*' || EXIT=$$?; \
 	done; \
 	echo; \
 	echo '========================================'; \
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ test: test-setup
 		echo "Running tests for $$vim"; \
 		echo '========================================'; \
 		echo; \
-		docker run -a stderr $(DOCKER_FLAGS) $$vim '+Vader! test/*' || EXIT=$$?; \
+		docker run -it $(DOCKER_FLAGS) $$vim '+Vader! test/*' || EXIT=$$?; \
 	done; \
 	echo; \
 	echo '========================================'; \
--- a/autoload/ale/linter.vim
+++ b/autoload/ale/linter.vim
@@ -103,7 +103,7 @@ function! ale#linter#Define(filetype, linter) abort
     call add(s:linters[a:filetype], l:new_linter)
 endfunction
 
-function! s:LoadLinters(filetype) abort
+function! ale#linter#GetAll(filetype) abort
     if a:filetype ==# ''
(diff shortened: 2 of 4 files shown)


DiscoViking:
Some combination of exactly how we're running the tests causes them vim to hang forever.

This hangs:
`docker run -a stderr $DOCKER_FLAGS vim-v8.0.0000 '+Vader! test/*.vader'`

This works fine (but with messy output):
`docker run -it $DOCKER_FLAGS vim-v8.0.0000 '+Vader! test/*.vader'`

As does:
`docker run -t -a stdout $DOCKER_FLAGS vim-v8.0.0000 '+Vader! test/*.vader'`
So I think it's to do with having the pseudo-tty.

Worth noting that:
`docker run -a stderr $DOCKER_FLAGS vim-v8.0.0000 '+Vader! test/test_ale_info.vader'`
Also works fine.  So it's nothing specific to do with my new tests in particular,  but something to do with running them all at once.

Did we hit some threshold of output where vim is trying to pipe output into a full output buffer or something?

DiscoViking:
I'm going to suggest that vim doesn't like being run not through a tty,  and hence my fix here is actually sensible.  
It has the unfortunate downside of spewing some garbage output,  but I'm not really sure what else to do.

w0rp:
I don't like the output of the tests here. There's a lot of garbage in there.

w0rp:
The problem with running the tests isn't because of the site of the input, but because of how you are redirecting the output of the command test checking it. There is a much easier way to handle this which will work without modifying the `Makefile`. Do this instead:

```vim
Before:
  let g:testlinter1 = {'name': 'testlinter1', 'executable': 'testlinter1', 'command': 'testlinter1', 'callback': 'testCB1', 'output_stream': 'stdout'}
  let g:testlinter2 = {'name': 'testlinter2', 'executable': 'testlinter2', 'command': 'testlinter2', 'callback': 'testCB2', 'output_stream': 'stdout'}

  call ale#linter#Reset()
  let g:ale_linters = {}
  let g:ale_linter_aliases = {}

After:
  unlet! g:output

Given nolintersft (Empty buffer):

Execute(ALEInfo with no linters should return the right output):
  redir => g:output
    ALEInfo
  redir END

  AssertEqual g:output, "\n Current Filetype: nolintersft\nAvailable Linters: []\n  Enabled Linters: []"

...
```

You can rewrite the rests of the tests that way. You can capture the output of the command in variable and check the variable with AssertEqual instead of trying to capture the text in a clipboard and pasting it. Then you don't need a pseudo TTY for the test suite.

DiscoViking:
@w0rp awesome.  Didn't know you can use `redir` like that.  I've made the change. 👍

w0rp:
Looks good to me. :+1:

w0rp:
It could use a documentation entry now. I might write one later.

PR #275, opened 2017-01-23
Add clang-tidy linter for cpp.

vdeurzen:
I've added a clang-tidy linter for cpp code.

### DIFF
--- a/ale_linters/cpp/clangtidy.vim
+++ b/ale_linters/cpp/clangtidy.vim
@@ -1,0 +1,15 @@
+" Author: vdeurzen <tim@kompiler.org>
+" Description: clang-tidy linter for cpp files
+
+" Set this option to change the clang-tidy options for warnings for C.
+if !exists('g:ale_cpp_clangtidy_options')
+    let g:ale_cpp_clangtidy_options = '-std=c++14 -Wall'
+endif
+
+call ale#linter#Define('cpp', {
+\   'name': 'clangtidy',
+\   'output_stream': 'stdout',
+\   'executable': 'clang-tidy',
+\   'command': 'clang-tidy %s -- ' . g:ale_cpp_clangtidy_options,
+\   'callback': 'ale#handlers#HandleGCCFormat',
+\})
--- a/ale_linters/cpp/clangtidy.vim
+++ b/ale_linters/cpp/clangtidy.vim
@@ -10,6 +10,6 @@ call ale#linter#Define('cpp', {
 \   'name': 'clangtidy',
 \   'output_stream': 'stdout',
 \   'executable': 'clang-tidy',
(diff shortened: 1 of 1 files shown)


ale_linters/cpp/clangtidy.vim
```diff
@@ -0,0 +1,15 @@
...
+" Set this option to change the clang-tidy options for warnings for C.
+if !exists('g:ale_cpp_clangtidy_options')
+    let g:ale_cpp_clangtidy_options = '-std=c++14 -Wall'
+endif
+
+call ale#linter#Define('cpp', {
+\   'name': 'clangtidy',
+\   'output_stream': 'stdout',
+\   'executable': 'clang-tidy',
+\   'command': g:ale#util#stdin_wrapper . ' -- ' . g:ale_cpp_clangtidy_options,
```
w0rp: You need to pass the file extension as the first argument to `stdin_wrapper`. You will also need to add on `clang-tidy`, the actual command you are wrapping itself is missing here. Can you put `--` before other options? `--` usually marks the end of all options in GNU land.

vdeurzen: Interestingly it was actually working with the given code. I'm now looking at the `stdin-wrapper` script but from the way that seems to work it will put the filename at the end. According to the clang-tidy documentation the filename has to come _before_ the `--`. Can I achieve that behavior with the given set-up?

w0rp: It will put the filename at the end, but the first argument to the program has to be a file extension. so you could write `' .cpp -- '`.

vdeurzen: @w0rp Any ideas for achieving the right argument order? Are you open to a more complicated `stdin-wrapper` script? I can probably achieve it by changing the wrapper, but that would quite likely affect other linters using it. Alternative would be to introduce a new wrapper...

w0rp: I'm not sure. I don't want to modify the wrapper script too much to support this program. Can `clang-tidy` be called with `-` to specify stdin input, etc.? If so, then you don't need to use the wrapper script.

w0rp: If you can figure out some way to allow a special argument for both the Bash and Batch scripts to specify that the filename argument should come at the front of the list of arguments instead of at the end, without breaking everything, then go for it.

w0rp: I have now completely replaced the wrapper script with support for a format string which lets you place an automatic temporary file anywhere in the string, including multiple times. I'll edit this in now, and merge this linter.

vdeurzen: @w0rp Cool, I'll try it out tomorrow.

Issue #278, opened 2017-01-25
Error C++ in file C


simepy:
Hello,

I'm a new in this tool but I have a error with the linter.
My code:
![image](https://cloud.githubusercontent.com/assets/12037541/22286603/0433d7c6-e2f0-11e6-91c3-9261c6ddefe0.png)
Error:
![image](https://cloud.githubusercontent.com/assets/12037541/22286653/2de7bae2-e2f0-11e6-9f33-31d7952ec119.png)
Manual:
![image](https://cloud.githubusercontent.com/assets/12037541/22286703/584fea84-e2f0-11e6-90fa-6599416056d4.png)
My Flags options:
![image](https://cloud.githubusercontent.com/assets/12037541/22286999/721dd5e2-e2f1-11e6-90dc-3978376c37a2.png)

NVIM 0.1.7

I don't know is my mistake or is a real bug ?

Thanks in advance

w0rp:
@simepy Check this again. It looks like the GCC C linter was set to use C++ as the syntax instead of C.

simepy:
@w0rp Ok, this error is fix 👍 Thanks, 

But I have a other error in the same line:
"Implicit declaration of function 'getpagesize' is invalid in C99 [-Wimplicit-function-declaration]"
In line with sbrk.

Compile manual with (-Wall include -Wimplicit-function-declaration):
gcc myFile.c -W -Wall -Werror
I have Not error but ALE detect this error...

w0rp:
You can change your GCC options with `g:ale_c_gcc_options` See `:help g:ale_c_gcc_options`. The default setting is to use `-std=c11`, which might not be what you want.

simepy:
Yes, This is the problem, because I change this option with g:ale_c_gcc_options and nothing change, ALE take default setting and use -std=c11 and -Wall.
![image](https://cloud.githubusercontent.com/assets/12037541/22586146/6e2b75c8-e9fb-11e6-9462-fe3920714293.png)

simepy:
Is my fault, I use clang and not GCC.
Thanks for your time and ALE 👍

PR #284, opened 2017-02-01
Tell gcc to use C, not C++, when linting C source files

EinfachToll:
A regression bug introduced (most likely by accident) here: #232

Maybe related to #278


w0rp:
Looks like a simple mistake. Thank you for fixing this. :+1:

PR #291, opened 2017-02-02
Fix mypy linter line numbers

ddfisher:
The mypy linter is currently assigning the column of the error as the line number (at least for me running under neovim 0.1.7).  I'm not entirely sure why this is the case, but unconditionally matching the column number fixes the problem.  The column number should always be there anyway, as mypy is always called with the `--show-column-numbers` flag.

### DIFF
--- a/ale_linters/python/mypy.vim
+++ b/ale_linters/python/mypy.vim
@@ -19,12 +19,12 @@ endfunction
 function! g:ale_linters#python#mypy#Handle(buffer, lines) abort
     " Look for lines like the following:
     "
-    " file.py:4: error: No library stub file for module 'django.db'
+    " file.py:4:0: error: No library stub file for module 'django.db'
     "
     " Lines like these should be ignored below:
     "
-    " file.py:4: note: (Stub files are from https://github.com/python/typeshed)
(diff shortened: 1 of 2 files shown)


w0rp:
Hmm, I seem to recall that my version of `mypy` could return output with only the line numbers, and no column numbers, which is why the `?` symbols were there to make the column numbers optional. I also seem to recall that I hit the same issue when I was trying it on my work machine. I'll see if I had anything for this already.

w0rp:
Yes, if you get errors about missing modules, the errors do not include column numbers, at least on my machine. The bug here is in the regular expression, which is too greedy in the filename pattern, so the line number is included in the filename match when a column is present. I have pushed a fix for this now which will fix both this issue, and also make it so it will still handle output without column numbers, with tests for both.

w0rp:
BTW, on an entirely unrelated note, I have toyed with the idea of improving the `mypy` linter, which is very dumb at the moment, so it can detect the location of configuration files in parent directories and try to automatically figure out which paths to import based on that.

From my tinkering, it seems like the only option which will really work is to require a `mypy.ini` file in an ancestor directory from the file you are editing, and then require that this INI file explicitly lists the paths to every module you care about. Then ALE could find that file easily, include it in the command line, and then read that to set the import paths for all of the modules.

I'll open an issue for it, and probably just do it eventually.

PR #281, opened 2017-01-27
Enable Rubocop to exclude files based on config

derekprior:
When using `--stdin`, Rubocop requires that you also pass the associated
file name. ALE was previously passing `_` as the filename. By passing
the actual relative path to the file and enabling the
`--force-exclusion` option, we can get Rubocop to respect excluded files
in the configuration.

Closes #197

PR #293, opened 2017-02-02
Fix pattern for RuboCop

### LINKED ISSUE #287: Rubocop doesn't seem to be running
I know the gem is installed, because I can run it from the command line and it works just fine. Ale is working on js files just fine, but I never see rubocop run. I've been testing in a file that I know should kick off some errors and warnings.

I'm not sure what additional information I can give, other than rubocop versions 0.47.1, 0.40.0 and 0.28.0 are installed. I don't have any specific settings set for ale.

yous:
This fixes a regression of #281. Fixes #287. The filename changed from `_` to `%s`, so the pattern should not have `_`.

### DIFF
--- a/ale_linters/ruby/rubocop.vim
+++ b/ale_linters/ruby/rubocop.vim
@@ -4,9 +4,9 @@
 function! ale_linters#ruby#rubocop#Handle(buffer, lines) abort
     " Matches patterns line the following:
     "
-    " <path>/_:47:14: 83:29: C: Prefer single-quoted strings when you don't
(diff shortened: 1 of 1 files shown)


jtmkrueger:
I can confirm this fixes my issue at [#287](https://github.com/w0rp/ale/issues/287).

derekprior:
I believe this is superseded by #299 which also uses a command callback to make sure we inject the correct filename into the command.

yous:
Right. I didn't notice that this won't work when filename changes. Closing in favor of #299.

PR #294, opened 2017-02-02
Add vint support for Neovim commands

tpict:
This PR stops vint returning errors like `E492: Not an editor command: tnoremap <silent> <C-s> <C-\><C-n>` when running ale under Neovim.


w0rp:
I'll take this now, so it will work better for NeoVim users. Cheers! :beers:

Issue #124, opened 2016-10-19
Implement command chaining to replace the dmd-wrapper script and more


w0rp:
An option to be used in place of `command` or `command_callback` for when commands depend on the results of other commands should be added called `command_chain`. The option should be a List of Strings of function names or Funcref values, where the first function in the chain should accept the buffer number the command is being run for, and subsequent commands should accept both the buffer number and the lines of output from previous commands. Each function should return a command to be run.

Implementing this option will allow us to replace the current dmd-wrapper script with the following:

``` vim
function! GetDUBCommand(buffer)
    return 'dub describe'
endfunction

function! GetDMDCommand(buffer, dub_output_lines)
    " Build import paths arguments here
    return 'dmd ...'
endfunction

ale#linters#Define('d', {
\ ...
\   'command_chain' ['GetDUBCommand', 'GetDMDCommand'],
\})
```

This chaining of commands could be useful for other linters in future. The implementation of this chaining of commands is probably the hairy part, and will require some picking apart of how linters are run.

w0rp:
I have a branch open for this now, `command-chaining`, which needs some work on it. I'll stop now, as RSI prevents me from working any more at the moment. I bet I can get this done within the next week.

w0rp:
I have merged the majority of the work needed for this to `master` now. It just needs a little bit more work and some tests for it.

w0rp:
I finally got this to work! This took some effort, but unless I accidentally screwed something up, I'm pretty happy with the results. The DMD linter now uses this new feature, and should hopefully run on Windows now.

I will update the documentation later with some advice on how to use the new API for chaining commands after I'm happy enough that it works.

PR #297, opened 2017-02-03
try fixing go build

joshuarubin:
try addressing some issues I found with #270 

1. only use plain go and go test files, `go tool compile` can't compile cgo or other non-native go files
1. use a `/dev/null` that works in windows (no idea if the whole linter will work in windows)
1. ensure `$GOROOT` is considered along with each `$GOPATH`
1. ensure all dependencies are installed and up-to-date (without `go test -i` files in `GOPATH/pkg` that were compiled by different versions of go or if they didn't exist at all would cause the linter to fail)

also note that this linter could probably benefit from chained commands in #124 but they are not ready yet on the master branch


w0rp:
You have to use exactly the filename `/dev/null` for `go build` on Windows. It's weird, but that's apparently just how it works, based on reports from others.

Thank you for the pull request. The issues are being ironed out slowly in the other pull request, but it's always good to have more information.

joshuarubin:
ok, cached some system calls
fixed /dev/null
fixed `go test -i` to install for the correct package

joshuarubin:
updated to use chained commands and `go test -c`

DiscoViking:
I am for this.
Now that command-chaining is a thing,  copying all the files to a temp directory is no longer too expensive,  so there's no need to hack around with `go tool compile` anymore.

I'm still seeing some issues with errors appearing in the wrong buffers on this branch though.  Similar to what we were seeing in my PR.
Perhaps there's something slightly off about the error format which is causing ale to not match them to the right buffers?

DiscoViking:
Oh,  is it just because the go compiler returns errors for the entire package,  not just the one file?
Looks like ALE doesn't do any checking that the error is actually reported for the file we're currently in.

Two possible solutions:

1. The general way: Change `HandleUnixFormat` to discard errors which aren't for this file.
2. The local way: Pipe the `go test -c` output through grep for the current buffer filename or something.

I think (1) is better as long as it won't break anything.

joshuarubin:
Yeah, it will return errors for any file in the package. It would be pretty easy to filter out errors for other files in our own callback handler (which then delegates to `ale#handlers#HandleUnixFormatAsError`. However, I like your idea that maybe `HandleUnixFormat` should discard errors for files other than the buffer file -- that would be my preference.

What are your thoughts @w0rp?

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,126 @@
...
+
+  " get all files for the package
+  let l:files = s:pkgfiles(l:pkginfo)
+
+  " copy the files to a temp directory with $GOPATH structure
+  let l:tempdir = tempname()
+  let l:temppkgdir = l:tempdir . '/src/' . s:pkgimportpath(l:pkginfo.Dir)
+  call mkdir(l:temppkgdir, "p", 0700)
+
+  return 'cp ' . join(l:files, ' ') . ' ' . shellescape(l:temppkgdir) . ' && echo ' . shellescape(l:tempdir)
```
w0rp: I'm concerned about the cost of copying all of the files to a temporary directory.

Is it possible to specify import paths to the files instead? If you can tell the Go compiler where the source files are, you shouldn't need to copy them.

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,126 @@
-" Author: dzhou121 <dzhou121@gmail.com>
+" Author: Joshua Rubin <joshua@rubixconsulting.com>
 " Description: go build for Go files
 
-function! s:FindGobuildScript() abort
-    return g:ale#util#stdin_wrapper . ' .go go build -o /dev/null'
+" inspired by work from dzhou121 <dzhou121@gmail.com>
+
+function! s:goenv()
```
w0rp: Could you name your functions in PascalCase, per the Google style guide? https://google.github.io/styleguide/vimscriptguide.xml?showone=Naming#Naming

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,126 @@
...
+  if exists('s:go_env')
+    return s:go_env
+  endif
+
+  let l:env = systemlist('go env GOPATH GOROOT')
```
w0rp: You could put this in an initial command, too. You can write `'go env GOPATH GOROOT'` as the return value for a command callback, and you can return `''` from the callback to skip that command on the second call after that. Then the command will only be run once, but it won't block input in Vim on the first run, which is a little bit nicer. Take a look at the `flake8` linter for an example.

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,126 @@
...
+  \ 'GOROOT': l:env[1],
+  \}
+
+  return s:go_env
+endfunction
+
+let s:splitchar = ':'
+if !has('unix')
+  let s:splitchar = ';'
+endif
```
w0rp: You can write this:

```vim
let s:pathsep = has('unix') ? ':' : ':'
```

`pathsep` matches `os.pathsep`, the Python variable.

osleg:
Imho while that's a good idea performance wise we still can benefit alot from checks of other files, for example changing signature of interface will provide us with usable llist where the old signature was used so we can catch them instantly

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,126 @@
...
+
+" get a list of all source directories from $GOPATH and $GOROOT
+function! s:srcdirs()
+  let l:env = s:goenv()
+  let l:paths = split(l:env.GOPATH, s:splitchar)
+  call add(l:paths, l:env.GOROOT)
+  return l:paths
```
w0rp: Could you add a single blank line before any `return` statement, `if` statement, or loop which doesn't begin a new level of indentation?

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,126 @@
...
+  let l:bufname = resolve(bufname(a:buffer))
+  let l:pkgdir = fnamemodify(l:bufname, ':p:h')
+  let l:importpath = s:pkgimportpath(l:pkgdir)
+  let l:temppkgdir = l:tempdir . '/src/' . l:importpath
+
+  " the basename of the go file
+  let l:fname = fnamemodify(l:bufname, ':t')
+
+  " write the buffer to the tempdir
+  call writefile(getbufline(a:buffer, 1, '$'), l:temppkgdir . '/' . l:fname)
```
w0rp: The `stdin_wrapper` script effectively does this, but in the background. I'm not sure if it will work in this circumstance.

Ideally, you will be able to list the paths to all of the source files except the one being edited, and then run `go test` on the source file with the `stdin_wrapper` script. I don't know if Go has strict rules about files being in exactly the right directory or not.

ale_linters/go/gobuild.vim
```diff
@@ -1,17 +1,126 @@
...
 
 call ale#linter#Define('go', {
 \   'name': 'go build',
-\   'output_stream': 'stderr',
 \   'executable': 'go',
-\   'command': g:ale#util#gobuild_script,
+\   'command_chain': [
+\     {'callback': 'ale_linters#go#gobuild#GoList',     'output_stream': 'stdout'},
+\     {'callback': 'ale_linters#go#gobuild#CopyFiles',  'output_stream': 'stdout'},
+\     {'callback': 'ale_linters#go#gobuild#GetCommand', 'output_stream': 'stderr'},
```
w0rp: I wouldn't bother aligning the second key here. If you change any of the function names, you'll have to re-align all of the lines again, and that leads to messier diffs.

ale_linters/go/gobuild.vim
joshuarubin: I'm using `go list` to get a list of only the source files (so if there are other large files in the package directory, they will not be copied).

However, to answer your question, no. Go works with package paths and it expects all the source files for that package to be in that one path. `go tool compile` does accept file paths, but it has many drawbacks and is not a standard way to check go files.

w0rp:
Check out the TravisCI output. It's complaining that some strings could be using single quotes instead, and also complaining with my custom check which demands that every function end with the `abort` keyword. Using the `abort` keyword reduces the amount of noise when something goes wrong with Vim functions.

ale_linters/go/gobuild.vim
joshuarubin: `stdin_wrapper` can't work because we can't tell it to use the specific tempdir that was created and Go can't check multiple paths for files in a single package.

DiscoViking:
Eep,  just started a new package and started getting this error:
```
Error detected while processing function <SNR>113_HandleExitVim[1]..<SNR>113_HandleExit[23]..<SNR>113_InvokeChain[23]..ale_linters#go#gobuild#CopyFiles[5]..<SNR>103_pkgfiles:
line    3:
E715: Dictionary required
E715: Dictionary required
E715: Dictionary required
E715: Dictionary required
E715: Dictionary required
E715: Dictionary required
E715: Dictionary required
E715: Dictionary required
E715: Dictionary required
E715: Dictionary required
E715: Dictionary required
-- More --
```

Guessing that's because there's no files to copy.  We should guard against that.

autoload/ale/sign.vim
```diff
@@ -67,10 +67,14 @@ endfunction
 " Given a loclist, combine the loclist into a list of signs such that only
 " one sign appears per line. Error lines will take precedence.
 " The loclist will have been previously sorted.
-function! ale#sign#CombineSigns(loclist) abort
+function! ale#sign#CombineSigns(buffer, loclist) abort
     let l:signlist = []
 
     for l:obj in a:loclist
+        if has_key(l:obj, 'bufnr') && l:obj.bufnr >= 0 && l:obj.bufnr != a:buffer
```
joshuarubin: @w0rp I needed to prevent signs from other files from appearing in the primary buffer. what do you think of this?

It may be able to be simplified to just `if l:obj.bufnr != a:buffer`, but I wasn't sure if `bufnr` is always set.

w0rp: I don't think errors from other files should be set in the first place. If you want to see problems in those files, go look at those files. Checking a file should only return problems for that one file.

w0rp: If we were going to allow linters to set warnings or errors for other files, it shouldn't be done without seriously thinking about what the impact of that will be. More likely than not, it will break everything in strange ways.

osleg:
As I pointed in my previous message they might not be showed in the other file (which still might be a great feature) but populated in location/quickfix list

autoload/ale/sign.vim
joshuarubin: That's not the way quickfix/location list work. Go will produce errors for any file in the package. Using loclist is a great way to quickly jump to the line in another file that is causing the package the first file is in not to compile.

The errors show up properly in loclist, this is great. Discarding these errors in the lint handler is taking away useful information from the user.

The problem is that without the change I put in this file, signs from errors/warnings in other buffers get put in the linted file's buffer and shows that there is an error that doesn't exist.

w0rp:
What happens if you have both files open, and two jobs are fighting to set loclist items for the same file? Those are the kinds of problems we'd have to look at. For now, only use the warnings and errors for the file you're editing.

autoload/ale/sign.vim
joshuarubin: I tried modifying things to set the signs in another buffer, but those changes were too deep and I felt beyond the scope of this PR. I am happy to only update the signs in the current buffer, I just don't want signs from another buffer showing up in my current one.

loclist, as I said, is already able to handle multiple files just fine.

w0rp: I'd take it away for now. If we're going to set warnings in some different buffer than the one the linter is being run for, we should make sure it works well. You should see the loclist items, you should see signs, you shouldn't get duplicates if two jobs end up setting the same errors, etc.

joshuarubin: Taking this out causes signs that belong to other buffers to appear in the current one with the `go build` linter.

w0rp:
Is there any other way to check for errors besides copying all of the files to a temporary directory? What if you have a great number of source files to work with? Is it possible to specify import paths for the files and pass that to the Go compiler?

autoload/ale/sign.vim
joshuarubin: Also, I don't see how this line could could hurt things. It is a good compromise IMO until/if true multibuffer support can be added.

joshuarubin:
> Is there any other way to check for errors besides copying all of the files to a temporary directory? 

No.

> What if you have a great number of source files to work with? Is it possible to specify import paths for the files and pass that to the Go compiler?

No.

w0rp:
Huh, well I guess copying the files to a temporary directory will have to do.

autoload/ale/sign.vim
w0rp: I think the safest option is to not output warnings or errors file files other than the file you are checking for now. Supporting errors for buffers other than the one you are working on requires some serious consideration. Will it break the echo line as the cursor moves? What effect will it have on the warning/error count? Will it break the commands for jumping between warnings/errors?

Remove the warning and errors for other buffers for now, and we can develop that at another time. I think we should get something working for checking for some simple Go errors first, which doesn't suffer from module import problems, and then think about showing more warnings/errors later.

joshuarubin:
Idiomatic Go generally keeps packages pretty small. While I'm not sure if it's the best sample size, I just calculated the average package size of Go 1.7.4 source.

```zsh
$ for pkg in $(go list ./...); do total=$(du -ks ${pkg} | awk '{print $1}'); echo "$total\t$pkg"; done | awk '{sum+=$1}END{printf "Sum=%d\nCount=%d\nAve=%.2f\n",sum,NR,sum/NR}'
Sum=61100
Count=226
Ave=270.35
```

Those numbers are in Kb.

```zsh
$ for pkg in $(go list ./...); do total=$(wc -l ${pkg}/*.go | grep 'total$'); echo "$total\t$pkg"; done | awk '{sum+=$1}END{printf "Sum=%d\nCount=%d\nAve=%.2f\n",sum,NR,sum/NR}'
Sum=929815
Count=226
Ave=4114.23
```

Those are the line numbers in Go `1.7.4` source.

The Go `1.7.4` source averages `11.05` `.go` files per package and `7.88` if you exclude test files.

I just ran line counts over my company's source tree and the average line count is `1674.45`. We average `6` `.go` files per package, `4.12` if you exclude test files.

It shouldn't be too much copying in most cases.

w0rp:
Yeah, that sounds okay. Plus, it's happening in the background, so it's not too costly.

autoload/ale/sign.vim
joshuarubin: Removing these warnings/errors is harmful for this linter. Your package might not compile, yet there may be no indication why.

What about adding a new flag like `g:ale_experimental_multibuffer` that is strictly opt-in? If it is true, then multibuffer would work as it does presently in this diff. If not, this linter will strip errors/warnings for other buffers.

This would give you a good, nonintrusive way to start testing the feature while requiring minimal changes here.

To answer your questions:

> Will it break echo line as cursor moves?

No. Tested.

> What effect will it have on the warning/error count?

Counts are the total number of errors/warnings that the linter returned. This seems correct in my view as my statusline will show that an error or warning exists.

> Will it break the commands for jumping between warnings/errors?

I use [unimpaired](https://github.com/tpope/vim-unimpaired) for jumping around using `]l` and `[l` for example. Those work correctly. Is there another way I can test?

w0rp: The warnings and errors for other files should go in a separate pull request. We can get just showing the errors or warnings in the current file to work now, and work on the rest later.

joshuarubin: OK

joshuarubin:
ok, removed multibuffer changes

w0rp:
Cool! :+1: I'll let Travis check that. I'll merge this pull request, and other Go users can start testing this.

Feel free to open another pull request after that with the multiple buffer changes. If we can set up some decent Vader tests and figure out the details, it could be useful.

joshuarubin:
Thanks. Will do.

PR #299, opened 2017-02-03
Fix Rubocop filename handling

derekprior:
In my previous change, I updated the Rubocop linter to pass the filename
to Rubocop. This change was tested on a file I expected Rubocop to
ignore and the experience in vim was as I expected. However, I soon
found that ALE wasn't finding errors in files that should not be
ignored. After investigation, I found a few issues that this commit
fixes:

1. We were not properly passing the current filename. We now use
   `expand` to get the filename.
2. The regular expression used in the callback was expecting the static
   value of `_` for the filename in output. We now use a looser regular
   expression that begins matching on the first `:`.
3. The linter was defined statically. By using the current filename when
   defining the command the linter would always use the filename of the
   first Ruby file the user opened. We now use a `command_callback` to
   inject the proper filename.

I tested these changes on a configuration with included and excluded
files and found it to work as I expected. Apologies for the earlier
incorrect change.

### DIFF
--- a/ale_linters/ruby/rubocop.vim
+++ b/ale_linters/ruby/rubocop.vim
@@ -6,7 +6,7 @@ function! ale_linters#ruby#rubocop#Handle(buffer, lines) abort
     "
     " <path>/_:47:14: 83:29: C: Prefer single-quoted strings when you don't
     " need string interpolation or special symbols.
-    let l:pattern = '\v_:(\d+):(\d+): (.): (.+)'
+    let l:pattern = '\v:(\d+):(\d+): (.): (.+)'
     let l:output = []
 
     for l:line in a:lines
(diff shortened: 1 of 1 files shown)


w0rp:
Do you know what was wrong with the `%s` filename formatting? That is supposed to just work.

Your changes use `expand('%')` which could yield incorrect results if the callback for running the command executes in a different buffer than the buffer it was launched from. I don't think that's likely to happen, but I generally recommend that the buffer number provided to the functions is used to get the filename, or any other information about a buffer.

w0rp:
Defining the command as a string with '%s' shouldn't have made it use the same filename every time. Did you check which command was actually being run somehow? The string isn't modified with a filename, it's used as a template. It's probably better to write the command in a callback anyway. I will eventually change all of the built-in linters so they are written this way.

derekprior:
I think that part of the change was unnecessary. I will change it back to
%s and retest.

w0rp:
I'll merge this now and change it so it uses `bufname` instead of `expand`.

Issue #122, opened 2016-10-18
Ability to toggle linting on/off


jwilliams108:
As a feature request, it would be useful to have the ability to toggle linting on or off. Sometimes I'm working on a codebase that is full of lint errors, but it is infeasible or impossible for me to fix them. The preference there would be to temporarily disable linting while I'm working on those files.

w0rp:
I'll think about this.

hauleth:
@w0rp what do you think about not creating `autocmd`s by default but allow user to setup that? This would solve this issue as you could just use `augroup Name | au! | augroup END` to disable linters. Other option would be check inside lint function for buffer variable.

w0rp:
I'd rather have the `autocmd` commands by default, and users can use the global options to turn them off. It's better if simply installing the plugin does something useful.

amadeus:
I can throw my hat in the ring of 'I would really like the ability to execute a command to turn Ale on and off'.

It's especially useful in my workflow where I generally have a single vim instance in a project directory with many buffers open that I am working on, and sometimes I need to view various files in dependency directories, and the sign column goes crazy, because those deps don't operate with the same linting requirements.

I.E. An ideal workflow would:

```
:AleEnable
:AleDisable
```

or even

```
:ALEToggle
```

To go the Syntastic route.

w0rp:
It'll happen eventually. Either someone else will do it, or I'll do it when I finally get the time.

w0rp:
This has been implemented by now, and I just documented the command, `ALEToggle`.

talmobi:
Just what I was looking for, great! And thank you!!

SurenderLohia:
// in my .vimrc
`map <leader>at :ALEToggle<CR>`

PR #303, opened 2017-02-04
add ALEToggle command

dflupu:
Adds a command to toggle ale on or off, along with an ale_enabled flag that may be set before ale is loaded. Closes issue #122


dflupu:
I've added a call to `ale#cleanup#Buffer` when toggling off. As far as I can tell, that should stop active jobs along with cleaning up `g:ale_buffer_info`.

w0rp:
@dflupu Yep, that'll do it. :+1:

Issue #256, opened 2017-01-12
Rust linter crashes.


ivegotasthma:
I just pulled the latest changes that include the `carge` linter. When I open a file I receive these errors.

```
NVIM 0.1.7
Build type: Release
Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -Wconversion -O2 -DNDEBUG -DDISABLE_LOG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -DHAVE_CONFIG_H -D_GNU_SOURCE -I/build/neovim/src/build/config -I/build/neovim/src/neovim-0.1.7/src -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/build/neovim/src/build/src/nvim/auto -I/build/neovim/src/build/include
Compiled by builduser

Optional features included (+) or not (-): +acl   +iconv    +jemalloc +tui      
For differences from Vim, see :help vim-differences

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

-----

Error detected while processing function <SNR>45_HandleExitNeoVim[1]..<SNR>45_HandleExit[27]..ale_linters#rust#rustc#handle_rustc_errors:
line   10:
[...]

PR #305, opened 2017-02-05
Fix rust linting with cargo when multiple targets are present

### LINKED ISSUE #302: Cargo doesn't run when there is a bin and a lib
`cargo rustc` can only run when there is only one target so when there is multiple targets, there is no linting with cargo
Here is the cargo error: 
```
cargo rustc -- --error-format=json -Z no-trans
error: extra arguments to `rustc` can only be passed to one target, consider filtering
the package by passing e.g. `--lib` or `--bin NAME` to specify a single target
```

I think you should use `cargo build --message-format json`

notkild:
Fix #302

### DIFF
--- a/ale_linters/rust/cargo.vim
+++ b/ale_linters/rust/cargo.vim
@@ -15,7 +15,7 @@ endfunction
 call ale#linter#Define('rust', {
 \   'name': 'cargo',
 \   'executable_callback': 'ale_linters#rust#cargo#GetCargoExecutable',
-\   'command': 'cargo rustc -- --error-format=json -Z no-trans',
+\   'command': 'cargo build --message-format=json -q',
 \   'callback': 'ale_linters#rust#rustc#HandleRustcErrors',
-\   'output_stream': 'stderr',
+\   'output_stream': 'stdout',
 \})
--- a/ale_linters/rust/rustc.vim
+++ b/ale_linters/rust/rustc.vim
@@ -17,12 +17,20 @@ function! ale_linters#rust#rustc#HandleRustcErrors(buffer_number, errorlines) ab
         endif
 
         let l:error = json_decode(l:errorline)
+        if !has_key(l:error, 'code') && !has_key(l:error, 'message')
+            continue
+        endif
+
+        if has_key(l:error, 'message')
+            let l:error = l:error.message
+        endif
 
         if !empty(l:error.code) && index(g:ale_rust_ignore_error_codes, l:error.code.code) > -1
             continue
         endif
 
         for l:span in l:error.spans
+            let l:span.file_name = fnamemodify(l:span.file_name, ':t')
             if l:span.is_primary &&
                 \   (l:span.file_name ==# l:file_name || l:span.file_name ==# '<anon>')
                 call add(l:output, {


ale_linters/rust/cargo.vim
```diff
@@ -3,19 +3,89 @@
...
 function! ale_linters#rust#cargo#GetCargoExecutable(bufnr) abort
-    if ale#util#FindNearestFile(a:bufnr, 'Cargo.toml') !=# ''
-        return 'cargo'
-    else
-        " if there is no Cargo.toml file, we don't use cargo even if it exists,
-        " so we return '', because executable('') apparently always fails
-        return ''
-    endif
+	if ale#util#FindNearestFile(a:bufnr, 'Cargo.toml') !=# ''
+		return 'cargo'
```
w0rp: Could you change your tab indents to four spaces, so it matches the rest of the codebase?

w0rp: It looks like these changes are just from re-indenting everything.

ale_linters/rust/cargo.vim
```diff
@@ -3,19 +3,89 @@
...
+	else
+		" if there is no Cargo.toml file, we don't use cargo even if it exists,
+		" so we return '', because executable('') apparently always fails
+		return ''
+	endif
+endfunction
+
+function! ale_linters#rust#cargo#HandleCargoErrors(buffer_number, errorlines) abort
```
w0rp: The indentation here makes this hard to read. How does this differ from the rustc handler?

notkild: Yup sorry I will do that

notkild: And error message from cargo build are different, JSON error is contained in a 'message' object so I have to copy function from rustc to fix differences

notkild:
Is this good now?

ale_linters/rust/cargo.vim
notkild: Here is output from `cargo rustc --lib -- --error-format=json -Z no-trans`:
```
{"message":"expected one of `!` or `::`, found `hyper`","code":null,"level":"error","spans":[{"file_name":"src/crawl/mod.rs","byte_start":1529,"byte_end":1534,"line_start":7,"line_end":7,"column_start":4,"column_end":9,"is_primary":true,"text":[{"text":"se hyper::Url;","highlight_start":4,"highlight_end":9}],"label":null,"suggested_replacement":null,"expansion":null}],"children":[],"rendered":null}
```

And here is the output from `cargo build --message-format json`:
```
{"message":{"children":[],"code":null,"level":"error","message":"expected one of `!` or `::`, found `hyper`","rendered":null,"spans":[{"byte_end":1534,"byte_start":1529,"column_end":9,"column_start":4,"expansion":null,"file_name":"src/crawl/mod.rs","is_primary":true,"label":null,"line_end":7,"line_start":7,"suggested_replacement":null,"text":[{"highlight_end":9,"highlight_start":4,"text":"se hyper::Url;"}]}]},"package_id":"unibot 0.1.0 (path+file:///home/notkild/projects/rust/unibot)","reason":"compiler-message","target":{"kind":["lib"],"name":"libunibot","src_path":"/home/notkild/projects/rust/unibot/src/lib.rs"}}
```

w0rp:
Isn't this error handling function almost exactly the same as the function for `rustc`? Can't we fix the existing handler so it can work with both? It looks like you could set `l:error` to `l:error.message` if such a key exists, and then the rest is the same.

notkild:
I think there is a way but I am pretty bad at vimscript but I will try to find another way to do that

notkild:
Ok so I think I have found a solution. What do you think about that?

w0rp:
Looks good to me. :+1: I'll merge this, and let Rust users give it a go.

notkild:
Yup, if there any issue, send me a mail :smile:

vimeitor:
This commit breaks linting for me.

`
Error detected while processing function <SNR>98_HandleExitVim[1]..<SNR>98_HandleExit[27]..ale_linters#rust#rustc#HandleRustcErrors:
`

`
E121: Undefined variable: code
E116: Invalid arguments for function empty(l:error.code) && index(g:ale_rust_ignore_error_codes, l:error.code.code) > -1
E15: Invalid expression: !empty(l:error.code) && index(g:ale_rust_ignore_error_codes, l:error.code.code) > -1
`

notkild:
Are you using rustc to lint code?

vimeitor:
I haven't touched the default config.

w0rp:
See issue #256. I just pushed a commit which might fix this now.

vimeitor:
Almost :)

`
Error detected while processing function <SNR>100_HandleExitVim[1]..<SNR>100_HandleExit[27]..ale_linters#rust#rustc#HandleRustcErrors:
E715: Dictionary required
`

w0rp:
Try again one more time...

vimeitor:
Seems to work wonderfully now!

Thank you for the work both of you =)

w0rp:
Cool!

It was just down to a bug in a condition above not checking if the output looks like Cargo output or `rustc` output correctly. I juggled things around a bit until it started working again.

w0rp:
@notkild BTW, if you have some example output from `cargo` I could use for testing, that would be useful. We could easily write some Vader tests to cover the function for handling Rust errors.

notkild:
Yup if you want, I will make some different error to see different output

notkild:
Here are some examples of error and warning.

[error1.txt](https://github.com/w0rp/ale/files/758140/error1.txt)
[error2.txt](https://github.com/w0rp/ale/files/758143/error2.txt)
[error3.txt](https://github.com/w0rp/ale/files/758141/error3.txt)
[warning1.txt](https://github.com/w0rp/ale/files/758142/warning1.txt)

w0rp:
Thanks for the output! I missed that. I created some Vader tests now with some output I took from my machine. I might add those into test cases later.

notkild:
Good idea, I can give you some output when cargo compile `feature`, it give us JSON output that is different than error or warning

PR #311, opened 2017-02-07
Support linters that return warnings/errors for multiple files

joshuarubin:
This support must be explicitly enabled with:

```vim
let g:ale_experimental_multibuffer = 1
```

* Currently it works correctly for loclist/quickfix
* It does not attempt to update Signs on buffers other than the one that caused the linting
* Counts are derived from the number of errors the linter returned in total, not just for the file that caused the linting

### DIFF
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -201,6 +201,16 @@ function! s:FixLocList(buffer, loclist) abort
     let l:last_line_number = ale#util#GetLineCount(a:buffer)
 
     for l:item in a:loclist
+        if g:ale_experimental_multibuffer
+            let l:bufnr = a:buffer
+
+            if has_key(l:item, 'bufnr') && l:item.bufnr >= 0
+              let l:bufnr = l:item.bufnr
+            endif
+
+            let l:last_line_number = ale#util#GetLineCount(l:bufnr)
+        endif
+
         if l:item.lnum == 0
             " When errors appear at line 0, put them at line 1 instead.
             let l:item.lnum = 1
--- a/autoload/ale/sign.vim
+++ b/autoload/ale/sign.vim
@@ -67,10 +67,14 @@ endfunction
 " Given a loclist, combine the loclist into a list of signs such that only
 " one sign appears per line. Error lines will take precedence.
 " The loclist will have been previously sorted.
-function! ale#sign#CombineSigns(loclist) abort
+function! ale#sign#CombineSigns(buffer, loclist) abort
     let l:signlist = []
 
     for l:obj in a:loclist
+        if g:ale_experimental_multibuffer && has_key(l:obj, 'bufnr') && l:obj.bufnr >= 0 && l:obj.bufnr != a:buffer
+          continue
+        endif
+
         let l:should_append = 1
 
         if l:obj.lnum < 1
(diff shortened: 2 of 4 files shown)


w0rp:
You might want to roll this up into a single commit again before continuing.

joshuarubin:
Done

w0rp:
You should use `git rebase` against `master` instead of `git merge`, or you'll end up with a confusing commit history.

autoload/ale/sign.vim
```diff
@@ -67,10 +67,14 @@ endfunction
 " Given a loclist, combine the loclist into a list of signs such that only
 " one sign appears per line. Error lines will take precedence.
 " The loclist will have been previously sorted.
-function! ale#sign#CombineSigns(loclist) abort
+function! ale#sign#CombineSigns(buffer, loclist) abort
     let l:signlist = []
 
     for l:obj in a:loclist
+        if g:ale_experimental_multibuffer && has_key(l:obj, 'bufnr') && l:obj.bufnr >= 0 && l:obj.bufnr != a:buffer
```
w0rp: You need to use `!` here for the flag. You want the inverse. I think you can safely skip the `has_key` and `>= 0` checks. The API for commands is that they must set a valid buffer number.

w0rp: You can take it as a given that the `bufnr` is set. The API for commands is that they must set `bufnr`.

autoload/ale/engine.vim
```diff
@@ -201,6 +201,16 @@ function! s:FixLocList(buffer, loclist) abort
...
 
     for l:item in a:loclist
+        if g:ale_experimental_multibuffer
+            let l:bufnr = a:buffer
+
+            if has_key(l:item, 'bufnr') && l:item.bufnr >= 0
+              let l:bufnr = l:item.bufnr
+            endif
+
+            let l:last_line_number = ale#util#GetLineCount(l:bufnr)
```
w0rp: Can you cache the line count in a map? In the `master` version of the code, the line count check is only done once. It could be expensive to call this function for every item.

w0rp: You should include a check here which skips over any items from other buffers `if !g:ale_experimental_multibuffer`. Then you don't have to worry about checking that in any of the linters.

autoload/ale/engine.vim
```diff
@@ -201,6 +201,16 @@ function! s:FixLocList(buffer, loclist) abort
     let l:last_line_number = ale#util#GetLineCount(a:buffer)
 
     for l:item in a:loclist
+        if g:ale_experimental_multibuffer
+            let l:bufnr = a:buffer
+
+            if has_key(l:item, 'bufnr') && l:item.bufnr >= 0
+              let l:bufnr = l:item.bufnr
+            endif
```
w0rp: We can assume that `bufnr` is set appropriately, as that's the API for the command callbacks.

w0rp: We could consider changing the API so `bufnr` can be excluded, and then write this:

```vim
if !has_key(l:item, 'bufnr')
    let l:item.bufnr = a:buffer
endif
```

w0rp:
Something occurred to me today for this. `setloclist(0, list)` sets the loclist for the current window. If the buffer exists in another window, then the loclist items probably won't be set in that buffer. The loclist items might have to be split up and set in the different windows.

Additionally, the list items might have to be set up in the `g:ale_buffer_info` objects for each buffer, and there can be some complications there to handle.

w0rp:
This is out of date, so I'll close this one. I'm not opposed to the idea if it's workable.

Issue #149, opened 2016-10-25
Support additional lint levels beyond warning and error


euclio:
Some linters emit low-severity messages or additional information that is not considered a warning or an error. It would be useful to be able to represent these within ALE. For example, Neomake has 4 levels (error, warning, message, info).

w0rp:
We'll have to look into this. I started initially only supporting the two flags that loclist supports, `'E'` and 'W'`. We can add arbitrary keys into the loclist items, so maybe we could store more information on the objects for use in the statusline, etc.

blueyed:
Does the loclist really support only "E" and "W"?

Related: https://github.com/neomake/neomake/issues/584

w0rp:
I just had a look at `:help setqflist()`, which is where `:help setloclist()` sends you. It says only this.

```
type    single-character error type, 'E', 'W', etc.
```

I guess by "etc." you can set other error types as long as you use a single character.

blueyed:
Yep.

euclio:
I'd like to try implementing this. Should I just add support for info and message levels?

w0rp:
Yeah, go for it. Some work will need to be done for each linter, where the different error levels are available, and some work will be needed for signs, the statusline, etc.

w0rp:
Type `'I'`, and style errors and warnings through `sub_type` '`style'` are now supported for signs, but _nothing else yet_. Everything else still needs to be adjusted.

w0rp:
I just pushed a commit so the status lines will continue to work if the other problem types are set. Now  `ale#statusline#Count()` returns a `Dictionary` containing the keys `total`, `error`, `warning`, `info`, `style_error`, and `style_warning`, mapping to the counts for each. Two extra keys `0` and `1` are set so the Dictionary will continue to function with the airline extension here: https://github.com/vim-airline/vim-airline/blob/06351e86df61b79bd00971d6c2f33c286a048b22/autoload/airline/extensions/ale.vim#L11 The sum of `error` and `style_error` are set for `0`, and the sum of `warning`, `style_warning`, and `info` are set for `1`.

So the status functions should now

w0rp:
I pushed a commit for setting different syntax highlights for "info" items, style warnings, and style errors. Info and style warnings map to the highlights for warnings for now, and style errors map to the error highlights.

w0rp:
[...]

PR #312, opened 2017-02-07
add gometalinter

joshuarubin:
It would be nice to find a way to put this combine several lint commands within the `gobuild` linter so that files are only copied once. This linter has a different handler pattern though.

### DIFF
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -25,7 +25,6 @@ function! ale_linters#go#gometalinter#Handler(buffer, lines) abort
   let l:output = []
 
   for l:line in a:lines
-    echomsg l:line
     let l:match = matchlist(l:line, s:handler_pattern)
 
     if len(l:match) == 0
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -1,0 +1,71 @@
+let s:SplitChar = has('unix') ? ':' : ':'
+
+let g:ale_linters#go#gometalinter#args = get(g:, 'ale_linters#go#gometalinter#args', [
+\   '--fast',
+\   '--tests',
+\ ])
+
+function! s:Args() abort
+  return join(map(copy(g:ale_linters#go#gometalinter#args), 'shellescape(v:val)'), ' ')
+endfunction
+
+function! ale_linters#go#gometalinter#GetCommand(buffer, copy_output) abort
+  let l:tempdir = a:copy_output[0]
+  let l:importpath = ale_linters#go#gobuild#PackageImportPath(a:buffer)
+  let l:gopaths = [ l:tempdir ]
+  call extend(l:gopaths, split(g:ale_linters#go#gobuild#go_env.GOPATH, s:SplitChar))
+
+  return 'GOPATH=' . shellescape(join(l:gopaths, s:SplitChar)) . ' gometalinter ' . s:Args() . ' ' . shellescape(l:tempdir . '/src/' . l:importpath)
+endfunction
+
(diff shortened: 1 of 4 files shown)


autoload/ale/util.vim
```diff
@@ -90,6 +90,17 @@ function! ale#util#LocItemCompare(left, right) abort
...
+    " put warnings after errors (for the same line) since the text that shows
+    " when the cursor is moved will show only the first entry
+
+    if a:left['type'] < a:right['type']
+      return -1
+    endif
+
+    if a:left['type'] > a:right['type']
+      return 1
+    endif
```
w0rp: This change will sort the items in the wrong order, which is important for the movement commands, etc. Instead of getting items in column order `1, 2, 3, 4, 5`, you can see items in order `E3, E4, W1, W2, W5`, etc.

You should move these lines down below the column sorting, so it's the third key items are sorted on. Then `E` will take priority over `W` if there are both warnings and errors for the same column, if that's the problem. The way the items are sorted, the cursor is supposed to show information for a problem near the cursor. It doesn't matter what the severity is.

ale_linters/go/gometalinter.vim
```diff
@@ -0,0 +1,70 @@
...
+
+call ale#linter#Define('go', {
+\   'name': 'gometalinter',
+\   'executable': 'gometalinter',
+\   'command_chain': [
+\     {'callback': 'ale_linters#go#gobuild#GoEnv', 'output_stream': 'stdout'},
+\     {'callback': 'ale_linters#go#gobuild#GoList', 'output_stream': 'stdout'},
+\     {'callback': 'ale_linters#go#gobuild#CopyFiles', 'output_stream': 'stdout'},
+\     {'callback': 'ale_linters#go#gobuild#WriteBuffers', 'output_stream': 'stdout'},
+\     {'callback': 'ale_linters#go#gometalinter#GetCommand', 'output_stream': 'stdout'},
```
w0rp: All of these `output_stream` options are also the default value. I'll document this eventually, if I haven't already.

w0rp: It might be an idea to move functions shared between two linter definitions into the `autoload/ale/handlers` directory. At the moment, ALE loads the defnitions for a given filetype like so:

```
execute 'silent! runtime! ale_linters/' . a:filetype . '/*.vim'
```
If anyone ever changes the way linters are loaded so only the list of enabled linters are loaded, which has tried before, then this code will break when only this linter is enabled for `go`.

ale_linters/go/gometalinter.vim
```diff
@@ -0,0 +1,70 @@
...
+    let l:buffer = ale_linters#go#gobuild#FindBuffer(l:match[1])
+
+    if l:buffer == -1
+      continue
+    endif
+
+    if !get(g:, 'ale_experimental_multibuffer', 0) && l:buffer != a:buffer
+      " strip lines from other buffers
+      continue
+    endif
```
w0rp: You could perhaps start implementing this flag in the function which fixes problems with the loclist items. If the flag isn't on, you could automatically remove items for other buffers in that function.

autoload/ale/util.vim
w0rp: If you're going to sort on the error type, you should take issue #149 into account. Eventually, I would like to support error types in some way beyond "warning" and "error". 'E' seems to be the first single character which could appear in lexicographic order, but 'I' comes before 'W' in lexicographic order, and 'W' is more severe than 'I', so that wouldn't be right.

w0rp:
I'll close this one, as it's out of date. A simpler implementation might now be possible, as #333 has been implemented.

PR #306, opened 2017-02-05
Add standard linter

ahmedelgabri:
First try to add [`standard`](http://standardjs.com/) for #301

PR #313, opened 2017-02-08
add javascript standard linter

joshuarubin:
very similar to eslint


w0rp:
I only just realised, someone already implemented this, and there was a pull request waiting to go, which I just merged now. Check out PR #306. If there are any amendments you would like to make, you can make them in a new pull request.

Issue #292, opened 2017-02-02
Golang seems ignores vendor


DAddYE:
First of all, this plugin seems amazing!

I have just one small issue with **go** 

One of the linters ignores the `/vendor` directory in the root path so always complain that it can't find any import because they live in `/vendor` rather than `$GOPATH`

w0rp:
There is a pull request open for fixing `go build` issues which may address the problem you are having: https://github.com/w0rp/ale/pull/270 Some more work is needed on it before it will work.

w0rp:
@DAddYE A pull request has just been merged which thoroughly includes and copies a whole host of files. Could you check if this has fixed your issues?

w0rp:
I believe the work from @DiscoViking and @joshuarubin has fixed this now.

DAddYE:
Thanks guys!

joshuarubin:
Unfortunately, `vendor/` is a specific case that does not yet fully work. I'm trying to figure it out right now, but things keep getting more and more complicated with this linter...

Lets say we have this dir structure in your `$GOPATH`:

```
foo/
├── bar/
│   ├── bar.go
├── baz/
│   ├── baz.go
├── foo.go
├── vendor/
├   ├── qux/
└   └   └ qux.go
```

Lets say `foo/bar/bar.go` has this:

```go
type Bar struct {
    Qux qux.Qux
}
```

And `foo/baz/baz.go` has this:

```go
type Baz struct {
    Qux qux.Qux
}
```

And `foo/foo.go` has this:

```go
    var f bar.Bar
    var b baz.Baz

    f.Qux = b.Qux
```

If we lint `foo/bar/bar.go` then the copied directory structure looks like this:

```
foo/
├── bar/
│   ├── bar.go
└── foo.go
```

This will fail to build with the current linter.

The reason for this is that the copied structure is in the primary `$GOPATH`. It didn't get the vendor directory, so it may not build at all, but lets say that `qux` was in the secondary `$GOPATH`.

What the compiler actually sees is roughly this:

```go
// foo/bar/bar.go
type Bar struct {
    Qux $GOPATH.qux.Qux
}

// foo/baz/baz.go
type Baz struct {
    Qux $GOPATH.foo.vendor.qux.Qux
}
```

And the failure is in `foo/foo.go` when executing `f.Qux = b.Qux` which fails because they are not the same type.

[...]

PR #317, opened 2017-02-08
go build fixes

joshuarubin:
This is not yet ready to merge, but I wanted to share it with others.

There was a significant bug in the last `go build` linter where a package that imports another package, that is a direct parent of itself, would state that the parent package didn't exist and couldn't be imported.

An example of this would be package `github.com/joshuarubin/foo/bar` that would depend on `github.com/joshuarubin/foo`.

Additionally, several other optimizations have been implemented. Most significantly, files are no longer copied, but rather symlinked. `go list` is also no longer required. This should speed things up a fair bit, but also breaks windows compatibility.

I'd be happy to work with windows testers to get things working.

I'd also like to get a better idea of how this works with non-`sh` style shells like `csh`, `ksh`, `fish` etc.

### DIFF
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -44,12 +44,15 @@ function! ale_linters#go#gobuild#ParentImportPaths(buffer) abort
   let l:importpath = ale_linters#go#gobuild#PackageImportPath(a:buffer)
   let l:output = []
 
-  while stridx(l:importpath, '/') >= 0
+  while 1
     call add(l:output, l:importpath)
+
+    if stridx(l:importpath, '/') == -1
+      break
+    endif
+
     let l:importpath = fnamemodify(l:importpath, ':h')
   endwhile
-
-  call add(l:output, l:importpath)
 
   return l:output
 endfunction
@@ -63,7 +66,13 @@ function! ale_linters#go#gobuild#PkgFiles(buffer) abort
 
   while index(s:SrcDirs(), l:pkgdir) == -1
     call extend(l:files, glob(l:pkgdir . '/*', 1, 1))
-    let l:pkgdir = fnamemodify(l:pkgdir, ':h')
+    let l:tmppkgdir = fnamemodify(l:pkgdir, ':h')
+
+    if l:pkgdir ==# l:tmppkgdir
+      break
+    endif
+
+    let l:pkgdir = l:tmppkgdir
   endwhile
 
   call map(l:files, 'fnamemodify(resolve(v:val), '':p'')')
@@ -128,6 +137,13 @@ function! s:SymlinkFilesCmd(srcdir, destdir, files) abort
 endfunction
 
 function! ale_linters#go#gobuild#CopyFiles(buffer, goenv_output) abort
+  if !empty(a:goenv_output)
+    let g:ale_linters#go#gobuild#go_env = {
+    \ 'GOPATH': a:goenv_output[0],
+    \ 'GOROOT': a:goenv_output[1],
+    \}
+  endif
+
   let l:tempdir = tempname()
   let l:temppkgdir = l:tempdir . '/src/' . ale_linters#go#gobuild#PackageImportPath(a:buffer)
   call mkdir(l:temppkgdir, 'p', 0700)
@@ -144,8 +160,6 @@ function! ale_linters#go#gobuild#CopyFiles(buffer, goenv_output) abort
       call remove(l:files.files, l:idx)
     endif
   endfor
-
-  " TODO(jrubin) test when a package depends on a child package
 
   " symlink the files to a temp directory with $GOPATH structure
   return s:SymlinkFilesCmd(l:files.srcdir, l:tempdir, l:files.files) . ' ; echo ' . shellescape(l:tempdir)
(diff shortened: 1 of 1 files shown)


joshuarubin:
Just copying my issue from #292:

Unfortunately, `vendor/` is a specific case that does not yet fully work. I'm trying to figure it out right now, but things keep getting more and more complicated with this linter...

Lets say we have this dir structure in your `$GOPATH`:

```
foo/
├── bar/
│   ├── bar.go
├── baz/
│   ├── baz.go
├── foo.go
├── vendor/
├   ├── qux/
└   └   └ qux.go
```

Lets say `foo/bar/bar.go` has this:

```go
type Bar struct {
    Qux qux.Qux
}
```

And `foo/baz/baz.go` has this:

```go
type Baz struct {
    Qux qux.Qux
}
```

And `foo/foo.go` has this:

```go
    var f bar.Bar
    var b baz.Baz

    f.Qux = b.Qux
```

If we lint `foo/bar/bar.go` then the copied directory structure looks like this:

```
foo/
├── bar/
│   ├── bar.go
└── foo.go
```

This will fail to build with the current linter.

The reason for this is that the copied structure is in the primary `$GOPATH`. It didn't get the vendor directory, so it may not build at all, but lets say that `qux` was in the secondary `$GOPATH`.

What the compiler actually sees is roughly this:

```go
// foo/bar/bar.go
type Bar struct {
    Qux $GOPATH.qux.Qux
}

// foo/baz/baz.go
type Baz struct {
    Qux $GOPATH.foo.vendor.qux.Qux
}
```

And the failure is in `foo/foo.go` when executing `f.Qux = b.Qux` which fails because they are not the same type.

I've been racking my brain about how to fix this issue. There is no issue unless there is a vendor directory. However, how to handle copying all the necessary files is not trivial.

I can symlink the vendor directory easily.

The problem is how to identify all the packages that depend on packages in that vendor directory. Once they are identified, they all have to be copied to the lint directory too.

I'm going to take a break. Let me know if you guys come up with something.

ale_linters/go/gobuild.vim
```diff
@@ -116,42 +104,116 @@ function! ale_linters#go#gobuild#GetCommand(buffer, copy_output) abort
...
-  return 'GOPATH=' . shellescape(join(l:gopaths, s:SplitChar)) . ' go test -c -o /dev/null ' . shellescape(l:importpath)
+function! s:SymlinkFilesCmd(srcdir, destdir, files) abort
+  " symlink the files to a temp directory with $GOPATH structure
+  let l:cmds = []
+
+  for l:file in a:files
+    " TODO(jrubin) this will not work on windows
+    " TODO(jrubin) no idea if this works on windows or if this will work on
+    " shells like csh, ksh, fish, etc.
+    call add(l:cmds, 'ln -s ' . shellescape(a:srcdir . '/src/' . l:file) . ' ' . shellescape(a:destdir . '/src/' . l:file))
```
w0rp: `cp` wouldn't have worked on Windows anyway, I just realised. I wonder if a copy should be made on Windows, and a symlink should be used on Unix systems.

joshuarubin: That was my thinking

DiscoViking:
Wondering if it's time to bail out and go for the simpler approach of just:

1. Write the current buffer into `.ale.<pid>.<filename>.go` (or something relatively unique) in the actual package directory.
2. Pass `go test -c` the list of files in the package,  with the current file replaced by the temporary one we just saved.  Should be no problem since the new file is in the correct directory.
3. Delete the temporary file.

This should work in all possible current and future cases I think including cgo and vendoring.
The only downside is that we're putting the new file not in a temp directory,  so if ale dies and fails to clean up afterwards it will leave it lying around in the actual repo.
This might seem dirty (and it is),  but IMO it's a small price to pay for having the thing just actually work.

@w0rp @joshuarubin thoughts?

joshuarubin:
None of the go tools work with individual files, just packages (`go test` included).
I do wonder if we can make this all a lot easier with [build tags](https://golang.org/pkg/go/build/#hdr-Build_Constraints) though. If we could write the buffer to a temporary file in the package and somehow disable the on-disk file with a build tag, that would do it.

I don't know how to do this without modifying the original file though.

renstrom:
Just thinking out loud here, but it seems like we're going through extreme lengths to be able to get on-the-fly checking. Maybe we should take this upstream to see if there's any improvements that can be done to the actual build/lint tools to make our lives easier? Like running `go build` in the actual project directory but with some flag to replace an arbitrary file with `stdin`? That would basically solve all our problems if I understand them correctly.

w0rp:
Going through extreme lengths to get on-the-fly checking means going through extreme lengths to get something useful.

renstrom:
>Going through extreme lengths to get on-the-fly checking means going through extreme lengths to get something useful.

I'm not saying it's not useful. I'm just saying that maybe can get help from upstream tools to add support for some feature(s) to make this easier. For example, what if we could replace everything we do with something like this:

    $ go build --replace-file-with-stdin=foo.go

Now it would build normally, except instead of using `foo.go` on disk, it will use what's in `stdin` as if it was `foo.go`. If that makes sense?

w0rp:
Yeah, that would be useful. Some tools provide some support like this. `eslint` provides an option for taking input from stdin, and also specifying what the filename being tested is. That's generally the best scenario.

w0rp:
This is being superseded by a new implementation, which will be a lot simpler. See #418 and #333.

Issue #315, opened 2017-02-08
Support enabling/disabling buffer input for commands


w0rp:
ALE sends the current text buffer as input into the stdin channel of jobs that are run, if the command is defined with `command` or `command_callback`, or if the command is the last item in the chain for `command_chain`. This is what is needed for the majority of linters.

Some linters may want to send the Vim buffer to earlier commands in a chain, or disable sending Vim buffer data completely, say if the command actually reads from the file on disk. To this end, commands not in a chain, and commands in a chain, should support a `'read_buffer': 0 || 1` option.

`read_buffer` should be on by default for single commands, or the last command in a chain, and off in all other cases. It should be possible to specify the `read_buffer` for each Dictionary in a `command_chain`. This will enable, for example, running one program to read code from a file you are editing, parsing those errors, and then running another command based on that output to generate the file List of errors.

PR #320, opened 2017-02-09
[For Discussion] Very basic alternative implementation of gobuild linter

DiscoViking:
Note: This is obviously not a complete implementation,  nor am I even proposing this is the correct solution. Just displaying a very simple demonstration of the alternative implementation I mentioned in #317 for discussion.

Positives:

- Should work for any project,  present or future,  including cgo and vendor.

Negatives:

- Can pollute the source directory if ALE fails to clean up after itself.


joshuarubin:
```
$ go test -h
test [build/test flags] [packages] [build/test flags & test binary flags]
...
```

`go test` expects to be passed packages, not files.

DiscoViking:
The docstring is incomplete.  It handles files too,  as does `go build`.

`go build` actually does mention it in it's docs:

```
Build compiles the packages named by the import paths,
along with their dependencies, but it does not install the results.

If the arguments to build are a list of .go files, build treats
them as a list of source files specifying a single package.
```

joshuarubin:
Well in that case... I'll have to play around with how it works, but this should be _MUCH_ simpler. Thanks!!!

w0rp:
I'll see if I can implement issue #315 to help with this. This is one type of linter configuration I think is just a good way to solve this particular kind of problem. For linters like this, I'd like like to offer the option to not send any data via stdin, when you really don't need it.

w0rp:
I have implemented #315 now.

Now you can add `'read_buffer': 0,` to the options for the linter, or for any `command_chain` item, and this will disable sending the Vim buffer as input to the command, which uses a temporary file instead.

ale_linters/go/gobuild.vim
```diff
@@ -1,214 +1,51 @@
...
-    \   'col': l:match[3] + 0,
-    \   'text': l:match[4],
-    \   'type': 'E',
-    \   'nr': -1,
-    \})
-  endfor
+    call writefile(getbufline(a:buffer, 1, '$'), l:temp_file)
 
-  return l:output
+    return 'go test -c ' . ' -o /dev/null ' . join(l:all_files) . ' ' . l:temp_file . '; rm ' . l:temp_file
```
w0rp: It's worth nothing that this won't work on Windows, because `rm` is not a valid command on Windows.

w0rp: I believe you need to escape the filename here with `fnameescape`. It's best to do this right where the filename is added in to a command string.

w0rp:
One disadvantage of creating temporary files is that they can be left behind, say if Vim is killed before the `rm` command runs. Some thought needs to be put into cleaning up temporary files as best as we can.

One idea I had was to write an `ale#engine` function which lets you register a temporary file for deletion. The engine would then automatically remove temporary files after jobs have finished, on `VimLeavePre`, etc. This would ease the burden of writing cleanup code.

ale_linters/go/gobuild.vim
DiscoViking: Yep,  totally aware of these restrictions.  As mentioned this PR is a very hastily put together demonstration of the concept.
Seems like it might have legs,  so I'll go ahead and flesh it out a bit. 😄

DiscoViking:
Yeah that would totally work.
I was imagining it might become a step in the command_chain after the linting step.

But having it handled by the engine might be nicer since then even if the command_chain fails,  it can still be cleaned up later.

w0rp:
@DiscoViking I have implemented support for managing temporary files with ALE now. Now any temporary files can be scheduled for deletion after jobs finish or buffers are removed. See `:help ale#engine#ManageFile()` for more information.

DiscoViking:
@w0rp I've added the use of `'read_buffer': 0` and `ManageFile`.  Also added in some of the innovations @joshuarubin made in his PR like calling `go test -i` first to install dependent packages (I'm quite unsure about this since I feel like a linter shouldn't have any permanent effects on your environment),  and filtering out lines not meant for the current buffer.

I also tested with a very simple cgo project which just included one .h file and used a variable defined in it.  It "just works" now since we don't have to worry about copying the header anywhere. 😄 

Any thoughts from either of you?  Especially on the `go test -i` thing.

w0rp:
I think I'm okay with this. I can't comment too much, as I'm not that familiar with the Go tools. If @joshuarubin is happy, then I am happy.

DiscoViking:
Don't merge this.  I've noticed the temp files getting left around frequently whilst I'm working.  Needs debugging.

w0rp:
@DiscoViking That can be an issue. That's why ALE typically doesn't create temporary files outside of directories like `/tmp`. They can end up appearing inside of source control, and that can be a pain.

w0rp:
This will be superseded by #418.

PR #327, opened 2017-02-10
Pass the buffer's filename to Credo

jparise:
By default, Credo attributes input from STDIN as though it came from a
file named `stdin`. This change passes the buffer's filename, too, so
that Credo can use that information when applying its configuration.

This is a nice improvement because files like `mix.exs` are normally
excluded from Credo-based linting. Previously, ALE would show lint
warnings for those files as they were edited. Now, they are correctly
honor the Credo configuration and don't produce lint output.

### DIFF
--- a/ale_linters/elixir/credo.vim
+++ b/ale_linters/elixir/credo.vim
@@ -3,8 +3,8 @@
 function! ale_linters#elixir#credo#Handle(buffer, lines) abort
   " Matches patterns line the following:
   "
-  " stdin:19: F: Pipe chain should start with a raw value.
-  let l:pattern = '\v^stdin:(\d+):?(\d+)?: (.): (.+)$'
+  " lib/filename.ex:19:7: F: Pipe chain should start with a raw value.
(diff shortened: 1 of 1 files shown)


ale_linters/elixir/credo.vim
```diff
@@ -3,8 +3,8 @@
 function! ale_linters#elixir#credo#Handle(buffer, lines) abort
   " Matches patterns line the following:
   "
-  " stdin:19: F: Pipe chain should start with a raw value.
-  let l:pattern = '\v^stdin:(\d+):?(\d+)?: (.): (.+)$'
+  " lib/filename.ex:19:7: F: Pipe chain should start with a raw value.
+  let l:pattern = '\v^.+:(\d+):?(\d+)?: (.): (.+)$'
```
jparise: This could also be `'\v:(\d+):?(\d+)?: (.): (.+)$'` (just anchor at the tail). Let me know which is preferred.

w0rp: I'll accept this, but you might want to try a few Windows filenames including the drive letter. These have caused some problems before.

PR #328, opened 2017-02-11
Give user opportunity to modify the filetype backlist, which can prevent ALE linters running.

tenfyzhong:
Use g:ale_filetype_blacklist if user has defined it.


w0rp:
This is already supported via the `g:ale_linters` option. There should be only one way to do this. See `:help g:ale_linters`.

Issue #227, opened 2016-12-16
Clang++ support for modern c++(>= C++11)


ehzawad:
As clang++ pretty much similar to g++, I think it is not hard to integrate with this plugin..Apple swift, SONY PlayStation etc use clang++..

http://clang.llvm.org/cxx_status.html

Issue #134, opened 2016-10-21
Clang linter support for both c and c++


ehzawad:
Clang is more interesting and reliable compiler these days...

http://llvm.org/releases/3.9.0/tools/clang/docs/ReleaseNotes.html

And Clang is pretty much similar with GCC, So implementation will not bit different. But Clang++ for c++ works with different header files $CPLUS_INCLUDE_PATH that's why -stdlib=libc++ -lc++abi flags strictly necessary..

```
bash
# cpp
clang++ -std=c++14 -stdlib=libc++ -lc++abi -ldl -latomic -lpthread"
#c
clang -std=c11 -lm -lpthread -latomic -Wall -Werror -fcolor-diagnostics -pedantic-errors"
```

w0rp:
Yep, sounds good to me. If someone wants to open a pull request for it, go for it.

PR #329, opened 2017-02-11
Add clang to cpp linters.

tomotanakamura:
I've added clang for CPP language.


w0rp:
Looks okay to me. We should probably replace the `command` string with a `command_callback`, which will let users change the options for the linter dynamically. It's probably not a big deal for this linter.

w0rp:
Cheers! :beers:

tomotanakamura:
Cheers! 🍺 
#134 and #227 were resolved.Would you please close them?

w0rp:
Done.

PR #336, opened 2017-02-13
Highlighting should not filter linter messages without column info

ninjawithaneedle:


w0rp:
Cheers. That's a simple mistake on my part.

w0rp:
If you like, you can add a Vader test for that. If not, I'll add one later. You can call he function with a List and ensure that the List isn't modified.

ninjawithaneedle:
Sure. I am not really familiar with vader. Will try and get one added.

PR #342, opened 2017-02-14
Update line marker pattern for new gfortran

Exteris:
On my system the gfortran linter did not work, I suspect this to be due to a change in the output format.
I've generalized the regex slightly so that it matches my output also.


Exteris:
I copied the output from 3 different gcc's into this new test, I'm quite confident it'll work well now.
Getting support for very old gcc was a bit tricky.

We might need to do it again for GCC 7, due soon.

w0rp:
Cool, that'll do fine. The most important thing is to add some output to Vader test cases as modifications are made to handlers. Then when someone else comes along and eventually changes something else, it won't break anything which came from before. Cheers! :beers:

PR #345, opened 2017-02-15
Add yamllint option to pass in more options

AlexMasterov:
Need more options.

I'm not sure that the global variables is needed here. (if they are not used explicitly)
```viml
if exists('g:ale_yaml_yamllint_executable')
  return g:ale_yaml_yamllint_executable
endif
```


w0rp:
Yeah, I see nothing wrong with this.

Cheers! :beers:

PR #347, opened 2017-02-15
WIP: Support columns on php handler

adriaanzon:
This isn't 100% reliable because php doesn't give columns but only the invalid string, but it should work most of the time. Syntastic does it this way too.

### DIFF
--- a/ale_linters/php/php.vim
+++ b/ale_linters/php/php.vim
@@ -4,8 +4,9 @@
 function! ale_linters#php#php#Handle(buffer, lines) abort
     " Matches patterns like the following:
     "
-    " Parse error: parse error in - on line 7
-    let l:pattern = 'Parse error:\s\+\(.\+\) on line \(\d\+\)'
+    " PHP Parse error:  syntax error, unexpected ';', expecting ']' in - on line 15
+    let l:pattern = ':\s\+\(.\+unexpected ''\(.\{-}\)''.*\) in - on line \(\d\+\)'
+
     let l:output = []
 
     for l:line in a:lines
@@ -18,9 +19,9 @@ function! ale_linters#php#php#Handle(buffer, lines) abort
         " vcol is needed to indicate that the column is a character.
         call add(l:output, {
         \   'bufnr': a:buffer,
-        \   'lnum': l:match[2] + 0,
+        \   'lnum': l:match[3] + 0,
         \   'vcol': 0,
-        \   'col': 1,
+        \   'col': match(getline(l:match[3]), l:match[2]) + 1,
         \   'text': l:match[1],
         \   'type': 'E',
         \   'nr': -1,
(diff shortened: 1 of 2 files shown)


test/test_php_handler.vader
```diff
@@ -0,0 +1,59 @@
...
+  unset($foo $bar)
+
+Execute(The php handler should find columns correctly):
+  call ale#Lint()
+  call ale#engine#WaitForJobs(2000)
+
+  AssertEqual
+  \ [
+  \   {'group': 'ALEError', 'id': 4, 'pos1': [2, 12, 1], 'priority': 10},
+  \   {'group': 'ALEError', 'id': 5, 'pos1': [2, 12, 1], 'priority': 10}
```
adriaanzon: Is it correct that there are 2 matches here? I manually ran `getmatches()` and copied the result over to the test.

adriaanzon: I also see Travis fails at this test. I should probably only check the values of `pos1` , right?

adriaanzon:
Ok, this doesn't parse error messages like

```
PHP Parse error: Invalid numeric literal in - on line 5
```

I'll look into it once I get home.

test/test_php_handler.vader
w0rp: Yeah, you should map these values here. The tests can fail when you run the entire test sweet as the IDs can change. I have another test which checks the output of `getmatches()` you can check out.

ale_linters/php/php.vim
```diff
@@ -18,9 +19,9 @@ function! ale_linters#php#php#Handle(buffer, lines) abort
         " vcol is needed to indicate that the column is a character.
         call add(l:output, {
         \   'bufnr': a:buffer,
-        \   'lnum': l:match[2] + 0,
+        \   'lnum': l:match[3] + 0,
         \   'vcol': 0,
-        \   'col': 1,
+        \   'col': match(getline(l:match[3]), l:match[2]) + 1,
```
w0rp: Hmm, I wonder, could some characters from the string being used here be interpreted as special regular expression characters instead? It might be best to prefix the string with a `nomagic` flag. Maybe `'\V'`. See `:help /magic`

w0rp:
I'm okay with the general idea of this. It's been done before, and it works. It's good that you added test cases for it. That always helps.

ale_linters/php/php.vim
adriaanzon: I think i should have used `stridx()` here, which doesn't match against regular expressions.

adriaanzon:
Alright, I have now pushed something that works, but I'd still like to clean up the pattern.

The tests still fail though but this time it says:

```
(2/2) [EXECUTE] (X) [{'group': 'ALEWarning', 'pos1': [1, 10, 1]}, {'group': 'ALEError', 'pos1': [2, 10, 1]}]
 should be equal to [{'group': 'ALEError', 'pos1': [2, 12, 1]}, {'group': 'ALEError', 'pos1': [2, 12, 1]}]
```

It worked locally. Do you know what this could be?

w0rp:
I'll take this, but I'll remove the test which runs the actual PHP linter. Those tests tend to rely on too many external programs to work reliably. The tests for the handler are really the tests that are needed if you change the handler.

Cheers! :beers:

PR #348, opened 2017-02-16
Add hdevtools linter for haskell

robadmedo:
This adds support for the hdevtools haskell linter
https://github.com/hdevtools/hdevtools

The output for hdevtools is near identical to the ghc output so this
also extracts the ghc handler into the handle file and adds tests

### DIFF
--- a/ale_linters/haskell/ghc.vim
+++ b/ale_linters/haskell/ghc.vim
@@ -1,10 +1,6 @@
 " Author: w0rp <devw0rp@gmail.com>
 " Description: ghc for Haskell files
 
-if exists('g:loaded_ale_linters_haskell_ghc')
-    finish
-endif
-
 call ale#linter#Define('haskell', {
 \   'name': 'ghc',
 \   'output_stream': 'stderr',
--- a/ale_linters/haskell/ghc.vim
+++ b/ale_linters/haskell/ghc.vim
@@ -5,63 +5,12 @@ if exists('g:loaded_ale_linters_haskell_ghc')
     finish
 endif
 
-let g:loaded_ale_linters_haskell_ghc = 1
-
-function! ale_linters#haskell#ghc#Handle(buffer, lines) abort
-    " Look for lines like the following.
-    "
-    " /dev/stdin:28:26: Not in scope: `>>>>>'
-    let l:pattern = '^[^:]\+:\(\d\+\):\(\d\+\): \(.\+\)$'
-    let l:output = []
-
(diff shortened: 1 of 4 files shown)


ale_linters/haskell/hdevtools.vim
```diff
@@ -0,0 +1,9 @@
+" Author: w0rp <devw0rp@gmail.com>
```
w0rp: You can credit yourself here.

ale_linters/haskell/ghc.vim
```diff
@@ -5,69 +5,18 @@ if exists('g:loaded_ale_linters_haskell_ghc')
     finish
 endif
```
w0rp: I think these lines are a relic from long ago you can remove now.

autoload/ale/handlers.vim
```diff
@@ -220,3 +220,49 @@ function! ale#handlers#HandleStyleLintFormat(buffer, lines) abort
 
     return l:output
 endfunction
+
+function! ale#handlers#HandleGhcFormat(buffer, lines) abort
+    " Look for lines like the following.
+    "
+    " /dev/stdin:28:26: Not in scope: `>>>>>'
+     "Appoint/Lib.hs:8:1: warning:
+    let l:pattern = '^[^:]\+:\(\d\+\):\(\d\+\):\s\{-}\(warning\|error\)\(.\+\)$'
```
w0rp: Can we be confident about the changes to the pattern here? I don't use GHC myself. As long as we're sure it will work with the output of GHC across different versions and so on, then I'm happy.

There are a few test cases below, and that's good. :+1:

ale_linters/haskell/ghc.vim
robadmedo: Aha. That's probably why `ale#utils#Reset()` in an after block of the vader tests didn't work

autoload/ale/handlers.vim
robadmedo: So it turns out that the ghc 8.0 (the version I use locally) did in fact slightly tweak the output and so I've changed the handler and accompanying tests to ensure that this works with ghc 7.

I dropped the comment about matching `" /dev/stdin:28:26: Not in scope: `>>>>>'` because that's always been there [since this file was created](https://github.com/w0rp/ale/commit/57ef2c98339af5fd40e99e44c420815f9ed0805f)  but never would have matched the regex

w0rp:
Cheers! :beers:

PR #350, opened 2017-02-16
Fix clang++ flag typo

qantik:
Clang++ compiler flag indicating the language should be `c++` instead of `c`.


w0rp:
Good catch. Cheers! :beers:

PR #353, opened 2017-02-18
Add sml support via smlnj

pauloalem:
Hey there

This pull request adds support for SML/NJ.
The compiler outputs some multiline messages and column error info which I do not attempt to parse. But it does work for the simpler error messages which are most of the errors I get because I only do simple SML stuff.
Is this enough?

Thanks!

### DIFF
--- a/ale_linters/sml/smlnj.vim
+++ b/ale_linters/sml/smlnj.vim
@@ -1,0 +1,42 @@
+" Author: Paulo Alem    <paulo.alem@gmail.com>
+" Description: Rudimentary SML checking with smlnj compiler
+
+if exists('g:loaded_ale_sml_smlnj_checker')
+    finish
+endif
+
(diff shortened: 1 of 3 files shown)


ale_linters/sml/smlnj.vim
```diff
@@ -0,0 +1,42 @@
+" Author: Paulo Alem    <paulo.alem@gmail.com>
+" Description: Rudimentary SML checking with smlnj compiler
+
+if exists('g:loaded_ale_sml_smlnj_checker')
+    finish
+endif
+
+let g:loaded_ale_sml_smlnj_checker = 1
```
w0rp: I'll drop this part. This is a relic in some files, and isn't needed any more.

w0rp:
This looks okay to me. If someone else can offer any improvements, I'm sure they'll come along and do it.

Cheers! :beers:

PR #357, opened 2017-02-21
escape dot in maker regex

bounceme:


w0rp:
Looks good to me. Could you add some Vader tests for this handler, so it won't be broken in future?

w0rp:
Nice. Cheers! :beers:

PR #360, opened 2017-02-22
Fix Credo's line-matching pattern

jparise:
In d3e7d3d5, the line matching pattern was changed to handle filenames
other than `stdin`. Unfortunately, this broke the pattern's ability to
reliably extract both line and column numbers because the latter is an
optional match and the filename portion was very greedy. This resulted
in line numbers being discarded (treated as part of the filename) and
column numbers being interpreted as line numbers.

This change simplifies the pattern to only anchor on the line's suffix,
ignoring the filename portion entirely.

Alternatively, we could use vim's `\f` ("file name characters") class,
but that could still run into problems when `:`'s naturally appear in
the filename.

### DIFF
--- a/ale_linters/elixir/credo.vim
+++ b/ale_linters/elixir/credo.vim
@@ -4,7 +4,7 @@ function! ale_linters#elixir#credo#Handle(buffer, lines) abort
   " Matches patterns line the following:
   "
   " lib/filename.ex:19:7: F: Pipe chain should start with a raw value.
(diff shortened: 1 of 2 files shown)


jparise:
One other strategy would be to explicitly match on `.ex` / `.exs` file extensions. That would probably work 99% of the time; it depends on how users name their files.

w0rp:
An HTTP error broke the build. I'll merge this and fix it if the build breaks again. Cheers! :beers:

PR #364, opened 2017-02-25
Add the pylint symbol name for the msg_id of the error/warning.

jsivak:
Would like to see the Pylint symbol in the error/warning message. This makes it easier to add "ignore" comments in the code. Since Pylint is recommending that the symbol be used instead of the msg_id, this change shows the symbol as well.


w0rp:
I had to look up what "symbol" meant in terms of `pylint`, but I get it now.

Looks fine to me. Cheers! :beers:

PR #367, opened 2017-02-28
proselint: Add more supported filetypes

polyzen:
This should cover all supported filetypes (haven't found a better list):
https://github.com/amperser/proselint/commit/273d59cdab9fb9eb081ffa60d9e86d384c17a74e#diff4794aebae1680ac140fcfb9aeb737267

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,15 +98,15 @@ name. That seems to be the fairest way to arrange this table.
 | Swift | [swiftlint](https://swift.org/) |
 | Tex | [proselint](http://proselint.com/) |
 | Texinfo | [proselint](http://proselint.com/)|
-| Text^^ | [proselint](http://proselint.com/) |
+| Text^ | [proselint](http://proselint.com/) |
 | TypeScript | [tslint](https://github.com/palantir/tslint), typecheck |
 | Verilog | [iverilog](https://github.com/steveicarus/iverilog), [verilator](http://www.veripool.org/projects/verilator/wiki/Intro) |
 | Vim | [vint](https://github.com/Kuniwak/vint) |
-| Vim help | [proselint](http://proselint.com/)|
+| Vim help^ | [proselint](http://proselint.com/)|
 | XHTML | [proselint](http://proselint.com/)|
 | YAML | [yamllint](https://yamllint.readthedocs.io/) |
 
-* *^^ No text linters are enabled by default.*
+* *^ No linters for text or Vim help filetypes are enabled by default.*
 
 <a name="usage"></a>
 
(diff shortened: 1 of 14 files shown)


polyzen:
May want to have Vim help file linting disabled by default.

README.md
```diff
@@ -93,13 +97,16 @@ name. That seems to be the fairest way to arrange this table.
...
+| Text^ | [proselint](http://proselint.com/) |
 | TypeScript | [tslint](https://github.com/palantir/tslint), typecheck |
 | Verilog | [iverilog](https://github.com/steveicarus/iverilog), [verilator](http://www.veripool.org/projects/verilator/wiki/Intro) |
 | Vim | [vint](https://github.com/Kuniwak/vint) |
+| Vim help^ | [proselint](http://proselint.com/)|
+| XHTML | [proselint](http://proselint.com/)|
 | YAML | [yamllint](https://yamllint.readthedocs.io/) |
 
-* *^^ No text linters are enabled by default.*
+* *^ No linters enabled by default.*
```
w0rp: You can use a `^` marker instead, but you should change the sentence back to what it was here. This message is supposed to tell you that for any plaintext or near-plaintext filetypes, no linters will be run by default.

We can probably enable the Vim help linters by default. The only reason the plaintext linters are off by default is because they can be very annoying when you are editing files with unknown filetypes, etc.

autoload/ale/linter.vim
```diff
@@ -22,6 +22,7 @@ let s:default_ale_linter_aliases = {
 " Only cargo is enabled for Rust by default.
 let s:default_ale_linters = {
 \   'csh': ['shell'],
+\   'help': [],
```
w0rp: It _might_ be okay to remove this line and just enable linting for Vim help files by default. I'm not entirely sure about this either way.

polyzen: People might find it excessive that :help windows are suddenly getting linted. Sure, a small fraction have proselint installed, but it is possible that more linters will be added for this filetype.

README.md
polyzen: If Vim help files are to be disabled by default, do you prefer "No linters for text or Vim help filetypes are enabled by default"? This is what I had initially, but what was committed seems cleaner.

autoload/ale/linter.vim
w0rp: Okay, sounds good to me.

README.md
w0rp: Yeah, that sounds good. Keep them both off by default, and update the message to indicate which filetypes are disabled by default.

polyzen: Done.

w0rp:
Cheers! :beers:

PR #369, opened 2017-02-28
Use latest C++ standard

equal-l2:
The linter for C++ should use the standard for C++.


w0rp:
Yep. Someone used some bad copy and paste again. Cheers! :beers:

Travis CI seems to have randomly failed here. I'll just merge it and deal with whatever breaks on `master`. It'll probably be fine.

PR #370, opened 2017-02-28
Fix PHP column matching for unexpected single quotes

adriaanzon:
Unexpected single quotes resulted in an empty match, because PHP
surrounds the errors with quotes, and we check for the next quote to be
the ending delimiter.

For example: an unexpected string 'foo' would be presented as
`unexpected ''foo''`, and then the match would be `''`. The inner part
of that match is an empty string.

This adds a check for the keyword "expecting". Any quote after
"expecting" won't be matched, so we can use greedy matching instead of
non-greedy.

I also cleaned up the pattern a bit and added tests for the column matches.


w0rp:
Looks good to me. :+1: As long as we keep adding one more examples, any further changes shouldn't break existing matches. Cheers! :beers:

PR #374, opened 2017-03-02
WIP: Add support for error details

### LINKED ISSUE #225: Feature to show additional error detail (similar to elm-vim :ElmErrorDetail)
Currently, the only error text I see when I have a linting error is a one-line summary in the status bar.  Some errors have additional information that would be very helpful to see too.  For example, here is a full Elm error:

```
The 2nd argument to function `equal` is causing a mismatch.

33|                     Expect.equal (Engine.Manifest.getInventory manifest) [ ( "item2", attrs "item2" ) ]
                                                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Function `equal` is expecting the 2nd argument to be:

    List { description : String, name : String }

But it is:

    List ( String, Attributes )

Hint: I always figure out the type of arguments from left to right. If an
argument is acceptable when I check it, I assume it is "correct" in subsequent
checks. So the problem may actually be in how previous arguments interact with
the 2nd.
```

In Ale, all that I get is the top line:
`The 2nd argument to function `equal` is causing a mismatch.`

It would be super helpful to be able to run some command that could pop-up the rest of the error.

The elm-vim library does exactly this with the :ElmErrorDetail command. I'm not sure what the "pop-up" window is called in vim - it fills the bottom of the screen and I can "scroll" it, and press enter to make it go away.

 I imagine this feature would be useful in other languages too that have extended error reporting.

Thank you.

(Originally posted at https://github.com/ElmCast/elm-vim/issues/105#issuecomment-267297052)

jwoudenberg:
Fixes #225.

Hey there! I made a start adding error detail support to Ale. I'm pretty new to writing Vimscript, so any feedback is appreciated! The idea is that there's an additional detail field linters can set. A special command and keybinding can be set to display that detail field for the line the cursor is on, if it exists. This makes the addition completely backwards compatible with the way Ale currently works.

Does this seem the correct way forward?

I've added a Vader test for this feature, but I'm having some issue running the tests. It seems that once I run `make` once, the test files are somehow cached. Additional runs do not see to reflect changes made in test files. Is this a known problem and is there a work around?

Thanks!

### DIFF
--- a/ale_linters/elm/make.vim
+++ b/ale_linters/elm/make.vim
@@ -25,6 +25,7 @@ function! ale_linters#elm#make#Handle(buffer, lines) abort
                     \    'col': l:error.region.start.column,
                     \    'type': (l:error.type ==? 'error') ? 'E' : 'W',
                     \    'text': l:error.overview,
+                    \    'detail': l:error.overview . "\n\n" . l:error.details
                     \})
                 endif
             endfor
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -38,34 +38,32 @@ function! ale#cursor#TruncatedEcho(message) abort
     endtry
 endfunction
 
+function! s:FindItemAtCursor() abort
+    let l:info = get(g:ale_buffer_info, bufnr('%'), {'loclist': []})
+    let l:pos = getcurpos()
+    let l:index = ale#util#BinarySearch(l:info.loclist, l:pos[1], l:pos[2])
+    let l:loc = l:index >= 0 ? l:info.loclist[l:index] : {}
+
+    return [l:info, l:loc]
+endfunction
+
 function! ale#cursor#EchoCursorWarning(...) abort
     " Only echo the warnings in normal mode, otherwise we will get problems.
     if mode() !=# 'n'
         return
     endif
 
-    let l:buffer = bufnr('%')
+    let [l:info, l:loc] = s:FindItemAtCursor()
 
-    if !has_key(g:ale_buffer_info, l:buffer)
-        return
-    endif
-
-    let l:pos = getcurpos()
-    let l:loclist = g:ale_buffer_info[l:buffer].loclist
-    let l:index = ale#util#BinarySearch(l:loclist, l:pos[1], l:pos[2])
-
-    if l:index >= 0
-        let l:loc = l:loclist[l:index]
+    if !empty(l:loc)
         let l:msg = s:GetMessage(l:loc.linter_name, l:loc.type, l:loc.text)
         call ale#cursor#TruncatedEcho(l:msg)
-        let g:ale_buffer_info[l:buffer].echoed = 1
-    else
+        let l:info.echoed = 1
+    elseif get(l:info, 'echoed')
         " We'll only clear the echoed message when moving off errors once,
         " so we don't continually clear the echo line.
-        if get(g:ale_buffer_info[l:buffer], 'echoed')
-            echo
-            let g:ale_buffer_info[l:buffer].echoed = 0
-        endif
+        echo
+        let l:info.echoed = 1
     endif
 endfunction
 
@@ -94,27 +92,19 @@ function! ale#cursor#EchoCursorWarningWithDelay() abort
     endif
 endfunction
 
-
 function! ale#cursor#ShowCursorDetail(...) abort
-    " Only show the details in normal mode, otherwise we will get problems.
+    " Only echo the warnings in normal mode, otherwise we will get problems.
     if mode() !=# 'n'
         return
     endif
 
-    let l:buffer = bufnr('%')
+    let [l:info, l:loc] = s:FindItemAtCursor()
 
-    if !has_key(g:ale_buffer_info, l:buffer)
-        return
-    endif
-
-    let l:pos = getcurpos()
-    let l:loclist = g:ale_buffer_info[l:buffer].loclist
-    let l:index = ale#util#BinarySearch(l:loclist, l:pos[1], l:pos[2])
-
-    if l:index >= 0
-        let l:loc = l:loclist[l:index]
+    if !empty(l:loc)
         if has_key(l:loc, 'detail')
             echo l:loc.detail
+        else
+            echo  l:loc.text
         endif
     endif
 endfunction
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -93,3 +93,28 @@ function! ale#cursor#EchoCursorWarningWithDelay() abort
         let s:cursor_timer = timer_start(10, function('ale#cursor#EchoCursorWarning'))
     endif
 endfunction
+
+
+function! ale#cursor#ShowCursorDetail(...) abort
+    " Only show the details in normal mode, otherwise we will get problems.
+    if mode() !=# 'n'
+        return
+    endif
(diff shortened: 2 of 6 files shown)


autoload/ale/cursor.vim
```diff
@@ -93,3 +93,28 @@ function! ale#cursor#EchoCursorWarningWithDelay() abort
...
+
+    let l:buffer = bufnr('%')
+
+    if !has_key(g:ale_buffer_info, l:buffer)
+        return
+    endif
+
+    let l:pos = getcurpos()
+    let l:loclist = g:ale_buffer_info[l:buffer].loclist
+    let l:index = ale#util#BinarySearch(l:loclist, l:pos[1], l:pos[2])
```
w0rp: A lot of this code is the same as the above. I think it can be moved into another function called by both functions. I suggest the following, which I haven't tested extensively:

```diff
diff --git a/autoload/ale/cursor.vim b/autoload/ale/cursor.vim
index 9aaa8e9..2931952 100644
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -38,34 +38,32 @@ function! ale#cursor#TruncatedEcho(message) abort
     endtry
 endfunction
 
+function! s:FindItemAtCursor() abort
+    let l:info = get(g:ale_buffer_info, bufnr('%'), {'loclist': []})
+    let l:pos = getcurpos()
+    let l:index = ale#util#BinarySearch(l:info.loclist, l:pos[1], l:pos[2])
+    let l:loc = l:index >= 0 ? l:info.loclist[l:index] : {}
+
+    return [l:info, l:loc]
+endfunction
+
 function! ale#cursor#EchoCursorWarning(...) abort
     " Only echo the warnings in normal mode, otherwise we will get problems.
     if mode() !=# 'n'
         return
     endif
 
-    let l:buffer = bufnr('%')
+    let [l:info, l:loc] = s:FindItemAtCursor()
 
-    if !has_key(g:ale_buffer_info, l:buffer)
-        return
-    endif
-
-    let l:pos = getcurpos()
-    let l:loclist = g:ale_buffer_info[l:buffer].loclist
-    let l:index = ale#util#BinarySearch(l:loclist, l:pos[1], l:pos[2])
-
-    if l:index >= 0
-        let l:loc = l:loclist[l:index]
+    if !empty(l:loc)
         let l:msg = s:GetMessage(l:loc.linter_name, l:loc.type, l:loc.text)
         call ale#cursor#TruncatedEcho(l:msg)
-        let g:ale_buffer_info[l:buffer].echoed = 1
-    else
+        let l:info.echoed = 1
+    elseif get(l:info, 'echoed')
         " We'll only clear the echoed message when moving off errors once,
         " so we don't continually clear the echo line.
-        if get(g:ale_buffer_info[l:buffer], 'echoed')
-            echo
-            let g:ale_buffer_info[l:buffer].echoed = 0
-        endif
+        echo
+        let l:info.echoed = 0
     endif
 endfunction
 ```

Then you could use `let [l:info, l:loc] = s:FindItemAtCursor()` in your new function here.

File: autoload/ale/cursor.vim [hunk 2]
```diff
@@ -93,3 +93,28 @@ function! ale#cursor#EchoCursorWarningWithDelay() abort
...
+
+    if l:index >= 0
+        let l:loc = l:loclist[l:index]
+        if has_key(l:loc, 'detail')
+            echo l:loc.detail
+        endif
```
w0rp: I wonder if anything should be echoed if no extra details are available.

w0rp: I think maybe if `detail` isn't defined, then echo the regular message. That should also let us support echoing a full message if the message is truncated.

ale_linters/elm/make.vim
```diff
@@ -25,6 +25,7 @@ function! ale_linters#elm#make#Handle(buffer, lines) abort
                     \    'col': l:error.region.start.column,
                     \    'type': (l:error.type ==? 'error') ? 'E' : 'W',
                     \    'text': l:error.overview,
+                    \    'detail': l:error.overview . "\n\n" . l:error.details
```
w0rp: We should maybe consider just putting the full error details in the text value. I'm not sure.

jwoudenberg: Yeah, I can see the appeal in that, it would definitely be simpler. The thing I have against that in the case of Elm, is that `elm-make` already splits the message into an overview and a detail. Any default truncation we'd implement for linters generically is never going to do as good a job across all linters.

Even for other linters that do not split their output in this way, it might be useful to make a linter-specific split in the wrapper file.

jwoudenberg:
Hey @w0rp, thanks so much for your detailed feedback! I think I addressed all your concerns. Let me know if you agree and whether you have any others.

Apart from still having some trouble running the tests, I have a more fundamental problem with the detail tests I was hoping you might have an idea for: To make newlines show up naturally in the detailed test output, I need to use `echo` rather than `echomsg`. This prevents me from using the redirect trick used by other tests for checking if the right thing was logged. Do you know another trick for this and could you perhaps point me in the direction of a solution?

doc/ale.txt
```diff
@@ -1142,6 +1142,7 @@ ALEPrevious                                                       *ALEPrevious*
 ALEPreviousWrap                                               *ALEPreviousWrap*
 ALENext                                                               *ALENext*
 ALENextWrap                                                       *ALENextWrap*
+ALENextWrap                                                       *ALENextWrap*
```
polyzen: ALEDetail*

jwoudenberg: Thanks @polyzen, I removed the duplicate. The actual documentation for ALEDetail is below.

autoload/ale/cursor.vim
```diff
@@ -38,34 +38,32 @@ function! ale#cursor#TruncatedEcho(message) abort
...
+        let l:info.echoed = 1
+    elseif get(l:info, 'echoed')
         " We'll only clear the echoed message when moving off errors once,
         " so we don't continually clear the echo line.
-        if get(g:ale_buffer_info[l:buffer], 'echoed')
-            echo
-            let g:ale_buffer_info[l:buffer].echoed = 0
-        endif
+        echo
+        let l:info.echoed = 1
```
w0rp: There's a bug here, but I'll fix it.

jwoudenberg:
Thanks!

w0rp:
I tweaked the test fail to make it hopefully easier to maintain in future, and addressed one issue I know will come up. The message will now be echoed with `shortmess`'s truncation temporary turned _off_, which is the inverse of the usual behaviour. If a message is very long, you should see a prompt to press Enter.

jwoudenberg:
@w0rp Thanks for the changes and fixes. I think I discovered a problem though: the detail logging messes up newlines in the detailed error message. Instead of showing up as simple newlines, I see `^@` symbols in the output. I think this is a limitation of using `echomsg` (`echo` does print newlines as newlines).

w0rp:
@jwoudenberg I didn't spot that. I have pushed a commit now which should fix that. Now it uses `:echo` again, with shortmess turned off, and the tests include newline characters too.

haasDev:
How do I see the details?  Is there a command that was added?

w0rp:
Yeah, see `:help ALEDetail`. There's a plug mapping for it too, should you want to create keybinding for it.

PR #375, opened 2017-03-02
Update sign regex for LANG = it_IT.UTF-8

manuelvio:
I've noticed that signs weren't unplaced and, learning that this was an issue depending on locale and :sign place (see #220), I've fixed the regular expression used to match those messages in my locale (it_IT.UTF-8).


w0rp:
Yep, makes sense. This is a common issue, and it get slightly better with time. I'll modify the Vader tests to include this example too. Thanks for the PR! Cheers! :beers:

PR #377, opened 2017-03-02
Add hamllint linter for Haml

patricklewis:


patricklewis:
Hi, I tried to follow the conventions for existing linters as best I could. haml-lint doesn't support reading from stdin so I set this up to use the temporary file support provided by ALE.

Please let me know if any additional changes are needed before this is ready to be accepted.

w0rp:
Yeah, this looks just fine to me. The `%t` temporary file magic is there for the purpose you used it for. :+1: Cheers! :beers:

PR #378, opened 2017-03-02
Adds semistandard for javascript

mrak:
`semistandard` is identical to `standard`, but enforces semicolon usage. This is more or less a direct copy of the existing `standard` integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ name. That seems to be the fairest way to arrange this table.
 | Haskell | [ghc](https://www.haskell.org/ghc/), [hlint](https://hackage.haskell.org/package/hlint) |
 | HTML | [HTMLHint](http://htmlhint.com/), [proselint](http://proselint.com/), [tidy](http://www.html-tidy.org/) |
 | Java | [javac](http://www.oracle.com/technetwork/java/javase/downloads/index.html) |
-| JavaScript | [eslint](http://eslint.org/), [jscs](http://jscs.info/), [jshint](http://jshint.com/), [flow](https://flowtype.org/), [standard](http://standardjs.com/)
+| JavaScript | [eslint](http://eslint.org/), [jscs](http://jscs.info/), [jshint](http://jshint.com/), [flow](https://flowtype.org/), [standard](http://standardjs.com/), [semistandard](https://github.com/Flet/semistandard)
 | JSON | [jsonlint](http://zaa.ch/jsonlint/) |
 | LaTeX | [chktex](http://www.nongnu.org/chktex/), [lacheck](https://www.ctan.org/pkg/lacheck) |
 | Lua | [luacheck](https://github.com/mpeterv/luacheck) |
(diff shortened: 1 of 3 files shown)


w0rp:
I think instead of separately supporting this, we should follow the same advice used for Syntastic on the project's GitHub page. https://github.com/Flet/semistandard#vim Configure the executable file for `standard` instead, and assume the rest is compatiable.

tnguyen14:
@w0rp is there a way I could configure the executable based on projects? Certain projects use standard, while others use semi, yet others just use eslint.

w0rp:
See `:help g:ale_pattern_options` for setting different settings for different filesystem paths.

tnguyen14:
That will allow you to manually set the patterns. What I think would make more sense is a more dynamic way to determine the linters. For eg., use eslint if there's a `.eslintrc` file, jshint if there's a `.jshintrc` file in the project root etc.

Here's an example of how I'm doing that for syntastic https://github.com/tnguyen14/dotfiles/blob/master/home/.vimrc#L310-L327

w0rp:
I'll open an issue for adding autocmd events for before linting or fixing is done. That will let people do the above, and more.

w0rp:
See #623

tnguyen14:
Thanks. Should this logic be done on entering a buffer, instead of for every lint event? Wouldn't that be expensive?

w0rp:
If you want that, then use a `BufEnter` event.

johnelliott:
This was the most related issue to my problem of using airbnb eslint, semistandard, and standard across project directories. 

Presented for example is the linter configuration from my .virmc to set the right executables and linters. It's not very elegant, but works for me :)

```vim
" Ale, async linter
let g:ale_linters = { 'javascript': ['eslint', 'standard'] }
"let g:ale_javascript_standard_executable = 'semistandard'

" Have es-lint installed globally even when `npm bin` returns local path

" Find local .eslint if it exists
let local_eslint = finddir('node_modules', '.;') . '/.bin/eslint'
if matchstr(local_eslint, "^\/\\w") == ''
  let local_eslint = getcwd() . "/" . local_eslint
endif
" Set eslint executable if we found it
if executable(local_eslint)
  let g:ale_javascript_eslint_executable = local_eslint
else
  let g:ale_javascript_eslint_executable = '$(npm bin)/eslint'
endif
" Find local semistandard  if it exists
let local_standard = finddir('node_modules', '.;') . '/.bin/semistandard'
" Find local semistandard  if it exists
if local_standard == ''
  let local_standard = finddir('node_modules', '.;') . '/.bin/standard'
endif
" Set standard/semistandard executable if we found it
if matchstr(local_standard, "^\/\\w") == ''
  let local_standard = getcwd() . "/" . local_standard
endif
if executable(local_standard)
  let g:ale_javascript_standard_executable = local_standard
else
  let g:ale_javascript_standard_executable = '$(npm bin)/standard'
endif
" http://stackoverflow.com/questions/28573553 Updated for w0rp/ale
autocmd FileType javascript let b:ale_linters = findfile('.eslintrc', '.;') != '' ? { 'javascript': ['eslint'] } : { 'javascript': ['standard'] }
```

PR #383, opened 2017-03-03
Add support for nix linting

alibabzo:
This adds support for linting nix files for the [Nix package manager](https://nixos.org/nix/).

I'm not a regex wizard, so there might well be a better way of pattern matching the output of `nix-instantiate`.

### DIFF
--- a/ale_linters/nix/nix.vim
+++ b/ale_linters/nix/nix.vim
@@ -1,0 +1,34 @@
+" Author: Alistair Bill <@alibabzo>
+" Description: nix-instantiate linter for nix files
+
+function! ale_linters#nix#nix#Handle(buffer, lines) abort
+
+    let l:pattern = '^\(.\+\): \(.\+\), at .*:\(\d\+\):\(\d\+\)$'
+    let l:output = []
+
+    for l:line in a:lines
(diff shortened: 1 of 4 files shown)


ale_linters/nix/nix.vim
```diff
@@ -0,0 +1,34 @@
...
+        \   'text': l:match[1] . ': ' . l:match[2],
+        \   'type': l:match[1] =~# '^error' ? 'E' : 'W',
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+call ale#linter#Define('nix', {
+\   'name': 'nix',
```
w0rp: We might want to change the name for the linter to `nix-instantiate`, along with the filename. I'll leave it how it is for now.

test/test_nix_handler.vader
```diff
@@ -0,0 +1,26 @@
+Execute(The nix handler should parse nix-instantiate error messages correctly):
```
w0rp: You'll need to make sure there's an `After` block here which resets the global linters list, or it could mess up the rest of the test suite run. I might just add it in.

w0rp:
Cheers! :beers:

PR #388, opened 2017-03-09
add slim-lint

doits:
this pr adds support for [slim-lint](https://github.com/sds/slim-lint) - it works nearly like haml-lint, so this is basically a copy of it

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ name. That seems to be the fairest way to arrange this table.
 | SASS | [sass-lint](https://www.npmjs.com/package/sass-lint), [stylelint](https://github.com/stylelint/stylelint) |
 | SCSS | [sass-lint](https://www.npmjs.com/package/sass-lint), [scss-lint](https://github.com/brigade/scss-lint), [stylelint](https://github.com/stylelint/stylelint) |
 | Scala | [scalac](http://scala-lang.org) |
+| Slim | [slim-lint](https://github.com/sds/slim-lint)
 | SML | [smlnj](http://www.smlnj.org/) |
 | Swift | [swiftlint](https://swift.org/) |
 | Tex | [proselint](http://proselint.com/) |
--- a/ale_linters/slim/slimlint.vim
+++ b/ale_linters/slim/slimlint.vim
@@ -1,0 +1,33 @@
+" Author: Markus Doits - https://github.com/doits
+" Description: slim-lint for Slim files, based on hamllint.vim
+
+function! ale_linters#slim#slimlint#Handle(buffer, lines) abort
+    " Matches patterns like the following:
+    " <path>:5 [W] LineLength: Line is too long. [150/120]
+    let l:pattern = '\v^.*:(\d+) \[([EW])\] (.+)$'
+    let l:output = []
+
+    for l:line in a:lines
+        let l:match = matchlist(l:line, l:pattern)
+
+        if len(l:match) == 0
+            continue
+        endif
+
+        call add(l:output, {
+        \   'bufnr': a:buffer,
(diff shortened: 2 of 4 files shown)


w0rp:
An entry will be needed inside the Vim documentation too, in `doc/ale.txt`. Could you add some Vader tests like the existing tests in the `handlers` directory?

doits:
I added the doc entry and Vader test.

Writing the test was very hard though, because on running `make` it always showed some very weird errors after changing my new `.vader` file and finally I found out it somehow worked with an old file (there were parsing errors I removed already). I had to manually install `Vader` and run it myself locally.

w0rp:
Cool, looks good to me. Cheers! :beers:

If you can paste the errors you saw here, I might be able to help and get it running on your machine.

doits:
@w0rp it was as simply that when I did a `make`, it showed some errors:

```
[...]
  Starting Vader: /testplugin/test/handler/test_slim_handler.vader
    (1/1) [EXECUTE] The slim handler should parse lines correctly
    (1/1) [EXECUTE] (X) Vim(call):E115: Missing quote: 'This li)
      > /tmp/vAobDKI/37, line 19
[...|
```

But `test_slim_handler.vader` was changed already and did not have a line with that missing quote anymore (it was there before). It looked like docker was working with old/stale content of that file and it took me a while to realize that ...

w0rp:
Thanks! I'll look into that.

doits:
It is even in that state right now. When I run `make` in current master branch, it gives me the following output:

```
shopt -s globstar; \
        docker images -q w0rp/ale | grep ^107e4efc4267 > /dev/null || \
                docker pull w0rp/ale
bash: line 0: shopt: globstar: invalid shell option name

Running tests for vim-v8.0.0000

Vim: Warning: Output is not to a terminal
Vim: Warning: Input is not from a terminal
Syntax error: After

Running tests for vim-v8.0.0027

Vim: Warning: Output is not to a terminal
Vim: Warning: Input is not from a terminal
Syntax error: After

Running Vint to lint our code
Vint warnings/errors follow:


Running custom checks


make: *** [test] Error 1
```

I think I introduced that syntax error myself while developing the vader test, but obviously it is now gone already and should not be present in clean master branch.

w0rp:
>bash: line 0: shopt: globstar: invalid shell option name

I think there's a problem with your environment. Well, that's not totally accurate. I think you can't run the tests in your environment. I suspect you're running an old version of Bash, so that's one issue. Are you using a Mac? I think Mac OSX still uses a Bash version before 4.0, so the `globstar` option won't work on Mac, which means that the test suite won't run on Mac.

I might see if there's an easy way for me to use something like recursive glob matches in Bash 3 and so on.

w0rp:
I just pushed a commit now which just adds a series of non-recursive glob matches. That should run on your machine. Let me know if you spot any more problems.

doits:
The glob error is gone, but still no luck:

```
docker images -q w0rp/ale | grep ^107e4efc4267 > /dev/null || \
                docker pull w0rp/ale

Running tests for vim-v8.0.0000

Vim: Warning: Output is not to a terminal
Vim: Warning: Input is not from a terminal
Syntax error:










Running tests for vim-v8.0.0027

Vim: Warning: Output is not to a terminal
Vim: Warning: Input is not from a terminal
Syntax error:










Running Vint to lint our code
Vint warnings/errors follow:


Running custom checks


make: *** [test] Error 1
```

I have docker running with docker-machine, and after I stopped my docker-machine and started it again (`docker-machine stop; docker-machine start`), it worked. So at least I solved it for now, not sure why (there were no running images or so, at least not per `docker ps -a`).

PR #389, opened 2017-03-10
Write erlc output files to the temporary directory

jparise:
In particular, if we're working with a leex (.xrl) or yecc (.yrl) source
file, erlc would otherwise generate the corresponding .erl file in the
current directory (often the project root), which is generally not what
we want.

Unconditionally writing erlc output to a temporary directory also
matches [Flycheck's behavior](https://github.com/flycheck/flycheck/blob/master/flycheck.el#L7014).

### DIFF
--- a/ale_linters/erlang/erlc.vim
+++ b/ale_linters/erlang/erlc.vim
@@ -3,7 +3,8 @@
 let g:ale_erlang_erlc_options = get(g:, 'ale_erlang_erlc_options', '')
 
 function! ale_linters#erlang#erlc#GetCommand(buffer) abort
-    return 'erlc ' . g:ale_erlang_erlc_options . ' %t'
+    let l:temp_dir = has('win32') ? $TMP : $TMPDIR
+    return 'erlc -o ' . l:temp_dir . ' ' . g:ale_erlang_erlc_options . ' %t'
 endfunction
 
 function! ale_linters#erlang#erlc#Handle(buffer, lines) abort
--- a/ale_linters/erlang/erlc.vim
+++ b/ale_linters/erlang/erlc.vim
@@ -3,8 +3,9 @@
 let g:ale_erlang_erlc_options = get(g:, 'ale_erlang_erlc_options', '')
 
 function! ale_linters#erlang#erlc#GetCommand(buffer) abort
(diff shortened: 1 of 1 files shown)


ale_linters/erlang/erlc.vim
```diff
@@ -3,7 +3,8 @@
 let g:ale_erlang_erlc_options = get(g:, 'ale_erlang_erlc_options', '')
 
 function! ale_linters#erlang#erlc#GetCommand(buffer) abort
-    return 'erlc ' . g:ale_erlang_erlc_options . ' %t'
+    let l:temp_dir = has('win32') ? $TMP : $TMPDIR
```
jparise: This is also how `elm/make.vim` determines the temporary directory. It might be useful to move this logic into a common `ale#utils` function.

w0rp: I recommend using the Vim `tempname()` function instead. Files created with `tempname()` will be automatically deleted when Vim is closed.

There is also a function used in `ale/engine.vim` which automatically deletes files as best as it can when a lint cycle ends, as well as another one for generating a temporary filename using `tempname()` securely. That second one might need to be moved to `ale/util.vim`.

w0rp: It also lets you create and manage _directories_, in a similar fashion.

ale_linters/erlang/erlc.vim
```diff
@@ -3,7 +3,8 @@
...
+    return 'erlc -o ' . l:temp_dir . ' ' . g:ale_erlang_erlc_options . ' %t'
```
w0rp: You should probably `fnameescape` the directory name here, right as you concatenate it into the string. Otherwise, bad things could happen on Windows.

ale_linters/erlang/erlc.vim
jparise: Are we walking about `CreateTemporaryFileForJob`? That could be useful here, although it's not an exported function (yet). I _do_ see the `ManageFile` and `ManageDirectory` functions too.

jparise: `TemporaryFilename` also looks useful but would also need to be exported.

w0rp:
Yeah, looks good to me. Cheers! :beers:

PR #393, opened 2017-03-13
Fix pattern of output for coffeelint

naoina:
Sometimes output of coffeelint contains comma `,` like `max_line_length` error below.

```
Line exceeds maximum allowed length Length is 122, max is 120.
```

This PR fixes it.


w0rp:

PR #399, opened 2017-03-18
Add support for nim check

baabelfish:


w0rp:

PR #404, opened 2017-03-20
Fix problems with nim check

baabelfish:
- Multi file errors are not shown in the same buffer
- Fixes parsing of error type that contain ':'

### DIFF
--- a/ale_linters/nim/nimcheck.vim
+++ b/ale_linters/nim/nimcheck.vim
@@ -3,7 +3,7 @@
 
 
 function! ale_linters#nim#nimcheck#Handle(buffer, lines) abort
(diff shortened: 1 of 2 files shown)


ale_linters/nim/nimcheck.vim
```diff
@@ -3,6 +3,7 @@
 
 
 function! ale_linters#nim#nimcheck#Handle(buffer, lines) abort
+    let l:buffer_filename = fnameescape(fnamemodify(bufname(a:buffer), ':p:t'))
```
w0rp: I think you can do away with `fnameescape` here and below. You only need to escape filenames if they are being used in a command line string.

w0rp:
Cheers! :beers:

nkala:
:beers:

baabelfish:
:beers:

PR #407, opened 2017-03-22
Add the dogma checker for elixir.

archSeer:
It's almost a direct copy of the Credo checker.

https://github.com/lpil/dogma


w0rp:
I changed the formatting and added a test to cover the handler.

PR #408, opened 2017-03-22
Make cargo linter lighter

idanarye:
- Use `cargo check` instead of `cargo build`.
- Use `--frozen` to avoid locking the project.

### DIFF
--- a/ale_linters/rust/cargo.vim
+++ b/ale_linters/rust/cargo.vim
@@ -1,6 +1,8 @@
 " Author: Daniel Schemala <istjanichtzufassen@gmail.com>
 " Description: rustc invoked by cargo for rust files
 
+let g:ale_rust_cargo_use_check = get(g:, 'ale_rust_cargo_use_check', 0)
+
 function! ale_linters#rust#cargo#GetCargoExecutable(bufnr) abort
     if ale#util#FindNearestFile(a:bufnr, 'Cargo.toml') !=# ''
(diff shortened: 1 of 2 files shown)


idanarye:
Using || will require some heavier scripting(`cargo check` exits 101 if there is no such command, if there is no `Cargo.toml`, or it found errors - we'll need some more parsing), which will make Windows support a pain.

Checking the version requires an extra command, which will have to run with `system()` because we need it to be synchronous. Sucks if you use vim on windows...

So - a configurable option sounds it is. I'll add `--frozen` regardless of that option though - it's still good for `cargo build`, and it's there from version 1.11.0.

ale_linters/rust/cargo.vim
```diff
@@ -11,10 +13,18 @@ function! ale_linters#rust#cargo#GetCargoExecutable(bufnr) abort
     endif
 endfunction
 
+function! ale_linters#rust#cargo#GetCommand(buffer) abort
+    let l:command = g:ale_rust_cargo_use_check
+    \   ? 'check'
```
idanarye: Just wanted to note I hate this indentation style. Since it's Vimscript, Vim's default autoindent behavior should be considered the defacto official style...

w0rp: I follow Donald Knuth's style for binary operators across many lines, and place the two ternary symbols on the left also.

PR #409, opened 2017-03-23
Document support for hdevtools

lucaskolstad:
Support for the hdevtools Haskell linter was added in PR #348 and I've been successfully using it since - this patch documents the support in the README.md.


lucaskolstad:
Not sure why the build fails from a README change. Here is the section of output from travis indicating the only failure I believe, though I'm not familiar enough with the project to understand how it could have been caused by the change or how to fix it.

```
 Starting Vader: /testplugin/test/test_ale_toggle.vader
    (1/1) [EXECUTE] ALEToggle should reset everything and then run again
    (1/1) [EXECUTE] (X) [] should be equal to [{'lnum': 2, 'bufnr': 3, 'vcol': 0, 'pattern': '', 'valid': 1, 'nr': -1, 'type': 'E', 'col': 3, 'text': 'foo bar'}]
  Success/Total: 0/1
```

lucaskolstad:
I pushed a new commit that was the exact same change to rerun travis and the tests passed the second time. Not sure whether this is a testing bug or travis or what.

w0rp:
I'll keep an eye on that. There might be a bug in the test suite.

w0rp:
Cheers! :beers:

Issue #272, opened 2017-01-23
Vim exits visual mode when loclist is open


megapctr:
When there are some errors and loclist is open, I can't use visual mode for longer than a fraction of a second. It seems as if ale was doing something constantly.

My ale config:
```
let g:ale_open_list = 1
```

Steps to reproduce:

- Use the above setting
- Open any file with some code
- Introduce an error to that code (loclist window should show up)
- Enter visual mode
- Wait a fraction of a second

Expected result:
- Vim stays in visual mode

Actual result:
- Vim switches to normal mode

Additional notes:
- Replace and insert modes seem to work fine. Visual mode is the only problematic mode
- The issue occurs only when there are errors in loclist window

kaidiren:
me too! @w0rp 😣

ninjawithaneedle:
Looks like this is cause by these lines:
https://github.com/w0rp/ale/blob/master/plugin/ale.vim#L64
https://github.com/w0rp/ale/blob/d7ed49f84964aebdaa180b553e83943cd85c5d20/autoload/ale/list.vim#L38

Ale is set to trigger on `BufEnter` (as default).
When `ale_open_list` is set, after linting is done, `lopen` is called to open location list. To bring focus back to the original window, `wincmd p` is called. This is a `BufEnter` event and triggers ale again. Causing an infinite loop. 

Incidentally, `wincmd p` causes vim to get out of visual mode.

ninjawithaneedle:
@megapctr  @kaidiren  Setting this in vimrc should temporarily solve the problem.

    let g:ale_lint_on_save = 1
    let g:ale_lint_on_enter = 0

ninjawithaneedle:
@w0rp What do you think of changing
https://github.com/w0rp/ale/blob/master/plugin/ale.vim#L64
to just `BufRead`. May be rename `ale_lint_on_enter` to `ale_lint_on_read` then?

w0rp:
It would need some testing to make sure that error checking happens when you look at buffer after changes have been made to it via some other tool, say `git checkout`, etc. If `BufEnter` is needed, then some more work will need to be done to handle this edge case.

PR #411, opened 2017-03-23
#272 Check if quickfix window is open before reopening

lucaskolstad:
This change fixes issue #272 by adding a check for whether the quickfix window is already open, and avoids unnecessarily reopening it if it is open.

This prevents an unnecessary BufEnter event being fired over and over when the quickfix window is reopened each time causing the bug under certain settings combinations.

When `g:ale_lint_on_enter = 1` and `g:ale_open_list = 1` we have a situation where a detected error opens the quickfix window, causing a BufEnter event, reopening the quickfix window, causing another BufEnter event...... into an infinite loop that uses a lot of cpu time every 300ms to run checks no matter what other settings you have.

I used the `ale#list#IsQuickfixOpen()` function in the same `list.vim` file to check if the quickfix window is open. Before this I believe it was only used in tests? Both the location list and quickfix window have `buftype == 'quickfix'` so it should detect both hopefully without any changes.

### DIFF
--- a/autoload/ale/list.vim
+++ b/autoload/ale/list.vim
@@ -36,10 +36,12 @@ function! ale#list#SetLists(buffer, loclist) abort
     if len(a:loclist) > 0 || g:ale_keep_list_window_open
         let l:winnr = winnr()
 
-        if g:ale_set_quickfix
-            copen
-        elseif g:ale_set_loclist
-            lopen
+        if !ale#list#IsQuickfixOpen()
+          if g:ale_set_quickfix
+              copen
+          elseif g:ale_set_loclist
+              lopen
+          endif
         endif
 
         " If focus changed, restore it (jump to the last window).
--- a/test/test_list_opening.vader
+++ b/test/test_list_opening.vader
@@ -49,6 +49,10 @@ Execute(The quickfix window should open for just the loclist):
   call ale#list#SetLists(bufnr('%'), g:loclist)
   Assert ale#list#IsQuickfixOpen()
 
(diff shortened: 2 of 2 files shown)


lucaskolstad:
This test failure is non-deterministic. Locally on my machine I have the test case that last failed on travis failing on 5/10 runs of `make test` even on the upstream master not just this branch and it only ever fails on just one specific case for `vim-v8.0.0027` and never any failures for `vim-v8.0.000` in the tests.

I've added the test case change you requested and squashed to one commit - but I was getting this non-deterministic failure before squashing, on both the master branch that doesn't even have this commit and this PR branch, and it's a different test case than the one I altered anyways.

Also the previous PR of mine changing the README triggered a non-deterministic test failure on an entirely different test case as well so I think it's just completely unrelated potentially? Though I can't reproduce that previous failure at all now.

Below is the single test case that I have repeated failure on and that fails in the travis tests:

```
  Starting Vader: /testplugin/test/test_command_chain.vader
    (1/1) [EXECUTE] Check the results of running the chain
    (1/1) [EXECUTE] (X) The final callback was not called
  Success/Total: 0/1
```

I think that if I amended this commit with no edit a few times and force pushed to trigger more travis builds I could get it to pass checks after a few attempts. But I'll leave it as is so you can see what happened on the travis logs as an example case showing it's not just my local system - and you can retrigger it yourself from travis if you want to.

w0rp:
Looks good. I'll fix the test issue later. Cheers! :beers:

PR #414, opened 2017-03-25
Add support for ASM files using GCC

lucaskolstad:


lucaskolstad:
Ya I saw it was very close to identical for the handling in `ale#handlers#HandleGCCFormat()`, and I actually tried using that with no custom handler at first but it wasn't recognizing any of the error lines.

The trouble I found was that the assembler error output format is different than the `C` format. It doesn't have column numbers, only row numbers. An error line looks like `{standard input}:42: Error: ...`, and the GCC handler, I *think*, could have handled something like `{standard input}:42:4: Error: ...`. The regex groupings get messed up when the column number is missing.

I'm happy to change things however you like, but I'm not sure how to without changing the regex in `ale#handlers#HandleGCCFormat()` to permit both forms somehow. And tbh I just don't know how. I'm less confident about making a change there without affecting others since I was hesitant about how to even implement it. I tried to avoid it entirely with a custom handler.

w0rp:
@lucaskolstad That makes sense. It's better to have duplicated code than to break existing code, in this case.

I might see about changing the GCC handler later so it supports output with only the line numbers too. The key to changing that is just adding more lines of input for testing it in the Vader tests. Once that's done, I'll see about applying the same function here.

PR #423, opened 2017-03-28
Allow commands to be chained using a `|`

adriaanzon:
I tried to add the following to my vimrc

```vim
command! ALEEnable if !g:ale_enabled | ALEToggle | endif
command! ALEDisable if g:ale_enabled | ALEToggle | endif
```

But it didn't work because commands weren't registered with the `-bar` flag


w0rp:
It might be an idea to open another pull request for adding the two commands you're trying to implement there, too. They could be useful for others.

w0rp:
Cheers! :beers:

PR #424, opened 2017-03-28
Add :ALEEnable and :ALEDisable commands

adriaanzon:
Here are the commands from #423


w0rp:
Nice. You're in luck too, as the major documentation restructuring I just pushed now didn't cause any conflicts here.

Cheers! :beers:

PR #425, opened 2017-03-29
Add options to facilitate linting only in normal mode

### LINKED ISSUE #420: Add feature g:disable_when_inserting
It would be nice to have a built in flag for disabling linting while user is inserting test (in insert mode). I don't really like watching red flags/text while typing.

taylorskalyo:
Close #420 

__ale\_lint\_on\_text\_changed:__
Allow setting to 'insert' or 'normal' to lint when text is changed only in insert or normal mode respectively.

__ale\_lint\_on\_insert\_leave:__
This flag can be set to 1 to enable linting when leaving insert mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,7 +368,7 @@ options off.
 
 ```vim
 " Write this in your vimrc file
-let g:ale_lint_on_text_changed = 0
+let g:ale_lint_on_text_changed = 'never'
 " You can disable this option too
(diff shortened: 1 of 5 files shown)


taylorskalyo:
Thanks for the feedback! This is my first real foray into vim scripting, so I appreciate the guidance.

w0rp:
Cheers! :beers:

PR #431, opened 2017-03-30
Check for existence of g:ale_emit_conflict_warnings before checking value

### LINKED ISSUE #426: Undefined variable: g:ale_emit_conflict_warnings
when I open a file, I get this error
what should I do ?~
```
Error detected while processing /Users/levy/.vim/bundle/ale/after/plugin/ale.vim:
line    7:
E121: Undefined variable: g:ale_emit_conflict_warnings
E15: Invalid expression: !g:ale_emit_conflict_warnings
```

janclarin:
Fixes #426. Added some documentation, too.

**Problem**
`g:ale_emit_conflict_warnings` does not get set to a value ([default is 1]( https://github.com/w0rp/ale/blob/master/plugin/ale.vim#L48)) in some circumstances, so when [its value is checked](https://github.com/w0rp/ale/blob/master/after/plugin/ale.vim#L7) in after/plugin/ale.vim, it is undefined. 

**This Solution**
Check for existence before checking the value.

**Another solution**
Another solution aside from checking `if exists('g:ale_emit_conflict_warnings')` would be to ensure `g:ale_emit_conflict_warnings` is set before any `finish` calls in plugin/ale.vim. However, that would leave `g:ale_emit_conflict_warnings` as a random variable that is set but never used since ale would [technically not work](https://github.com/w0rp/ale/blob/master/plugin/ale.vim#L31).

### DIFF
--- a/after/plugin/ale.vim
+++ b/after/plugin/ale.vim
@@ -4,7 +4,9 @@ endif
 
 let g:loaded_ale_after = 1
 
-if !g:ale_emit_conflict_warnings
+" Check if the flag is available and set to 0 to disable checking for and
+" emitting conflicting plugin warnings.
+if exists('g:ale_emit_conflict_warnings') && !g:ale_emit_conflict_warnings
     finish
 endif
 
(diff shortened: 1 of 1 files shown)


w0rp:
I think this is an acceptable solution. I'll merge this so anyone with the problem will stop seeing any bad errors. I'll cover this with tests later.

janclarin:
No problem, glad I could help! 👍

levylll:
@janclarin well done!

PR #434, opened 2017-03-30
Remove 'col' from linters where it is hardcoded to 1

adriaanzon:
When 'col' is 1, the first column will get highlighted for no reason. It
should be 0 (which is the default).

In the scalac linter there was also a check about the outcome of
`stridx`. It would set `l:col` to 0 if it was -1, and then it uses
`'col': l:col + 1` to convert the outcome of `stridx` to the actual
column number. This will make 'col' equals 1 when there is no match. We
can remove the check because `-1 + 1 = 0`.

I also encountered some old comments which I deleted in the second commit.


w0rp:
Yep, looks good to me. Those `1` values are just a relic.

Cheers! :beers:

Issue #427, opened 2017-03-29
Feature request: ALE should prefer buffer-local variables over global variables (helpful for dynamic linter configuration)


smkent:
### Background

I'm trying out ALE as a replacement for [Syntastic](https://github.com/vim-syntastic/syntastic).

I work with both Python 2 and 3, and I need ALE to select the correct version of `flake8` to use based on the file I'm working in. As suggested by the ALE documentation, I can set the executable and arguments used to invoke `flake8` as follows:

```vim
let g:ale_python_flake8_executable = 'python3'   " or 'python' for Python 2
let g:ale_python_flake8_args = '-m flake8'
```

I'd rather do this automatically, e.g. by parsing the shebang line (shebang parsing borrowed from Syntastic):

```vim
function! BufferParseShebang()
    let l:line = getline(1)
    if line =~# '^#!'
        let l:line = substitute(line, '\v^#!\s*(\S+/env(\s+-\S+)*\s+)?', '', '')
        let l:exe = matchstr(line, '\m^\S*\ze')
        let l:args = split(matchstr(line, '\m^\S*\zs.*'))
        return { 'exe': l:exe, 'args': l:args }
    endif
    return { 'exe': '', 'args': [] }
endfunction

function! ALEPythonVersionDetect()  " {{{
    let l:shebang_info = BufferParseShebang()
    if l:shebang_info['exe'] =~# '\m\<python[2-3]'
        let g:ale_python_flake8_executable = l:shebang_info['exe']
        let g:ale_python_flake8_args = '-m flake8'
    endif
endfunction  " }}}
augroup ale_python_detect
    autocmd!
    autocmd filetype python call ALEPythonVersionDetect()
augroup END
```

### Problem

The above code correctly sets the version of Python to use with `flake8` when I open a new file, but the linter settings are stored as global variables in vim. If I open a file written in one version of Python and another file written in a different version of Python, ALE won't select the correct linter to use for each file (since the settings are global).

To illustrate this:
* Put the above chunk of code in your `.vimrc`
* Place the following in a file called `py2`:
```python
#!/usr/bin/env python2
print 'test in python 2'
```
* Place the following in a file called `py3`:
```python
#!/usr/bin/env python3
print 'test in python 3'
```
* Open `py2`: `vim py2`. Note that ALE correctly shows no `flake8` errors.
* Open a new split with `py3`. Note that ALE correctly shows an error about the invalid `print` statement syntax (since Python 3 requires parentheses with `print`).
[...]

PR #437, opened 2017-03-30
Adds options to foodcritic linter

coredump:
Adds a way to pass command line options to the foodcritic command and
documentation about it.

### DIFF
--- a/ale_linters/chef/foodcritic.vim
+++ b/ale_linters/chef/foodcritic.vim
@@ -1,4 +1,5 @@
 " Author: Edward Larkey <edwlarkey@mac.com>
+" Author: Jose Junior <jose.junior@gmail.com>
 " Description: This file adds the foodcritic linter for Chef files.
 
 function! ale_linters#chef#foodcritic#Handle(buffer, lines) abort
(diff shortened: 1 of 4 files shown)


ale_linters/chef/foodcritic.vim
```diff
@@ -30,10 +31,15 @@ function! ale_linters#chef#foodcritic#Handle(buffer, lines) abort
...
+" Support options!
+let g:ale_chef_foodcritic_options = get(g:, 'ale_chef_foodcritic_options', '')
+
 call ale#linter#Define('chef', {
 \   'name': 'foodcritic',
 \   'executable': 'foodcritic',
-\   'command': 'foodcritic %t',
+\   'command': 'foodcritic '
+\       . g:ale_chef_foodcritic_options
+\       . ' %t',
```
w0rp: While you're at it, could you replace `'command'` here with `'command_callback'` and a function which returns the string? I'm planning ahead for #427. I'll change all of the linters in the repository so they work like that eventually.

w0rp: It might be an idea to add a test case for this, too, once the callback is in. There are a few existing tests like that. Maybe it's a bit excessive, but it can't hurt.

coredump:
Did some changes based on your review. Unsure if the test is on the right place and it's actually a pretty simple test so :man_shrugging:

w0rp:
Cheers! :beers:

PR #448, opened 2017-04-03
Better Unix pattern matching

rafaelrinaldi:
While working on adding a new linting engine to ale I noticed things were not working properly for some reason even though the program was spitting out the errors in standard unix format. It turns out that the spec doesn't seem to say anything about whitespaces therefore not all programs implement this properly.

To put it simply, we're only covering the "green" use case right now:

```diff
- file:line:column: body
+ file:line:column:body
```

This PR makes the whitespace optional so it doesn't break error message parsing.


w0rp:
I have fixed this now, with some tests.

rafaelrinaldi:
@w0rp Great work! Will try to test it as soon as possible.

PR #451, opened 2017-04-06
Thread support for nimcheck

mathsaey:
Added a '--threads:on' switch to the nim check command so that it doesn't produce errors when using one of the various modules that require threads.


w0rp:
Cheers! :beers:

PR #453, opened 2017-04-06
Add Ruby MRI linter

BrandonRoehl:
Add a linter for ruby that just uses ruby's standard with the general ruby command
```bash
ruby -c -w -T1 buffer.rb
```
This way you don't need to install or configure a gem like rubocop

### DIFF
--- a/ale_linters/ruby/ruby.vim
+++ b/ale_linters/ruby/ruby.vim
@@ -1,0 +1,46 @@
+" Author: Brandon Roehl - https://github.com/BrandonRoehl
+" Description: Ruby MRI for Ruby files
+
+function! ale_linters#ruby#ruby#Handle(buffer, lines) abort
+	" Matches patterns line the following:
+	"
+	" test.rb:3: warning: parentheses after method name is interpreted as an argument list, not a decomposed argument
+	" test.rb:8: syntax error, unexpected keyword_end, expecting end-of-input
+	let l:pattern = '\v^' . bufname(a:buffer) . ':(\d+): (warning: )?(.+)$'
+	let l:column = '\v^(\s+)\^$'
+
+	let l:output = []
+
+	for l:line in a:lines
+		let l:match = matchlist(l:line, l:pattern)
+
+		if len(l:match) == 0
+			let l:match = matchlist(l:line, l:column)
+			if len(l:match) != 0
+				let l:output[len(l:output) - 1]['col'] = len(l:match[1])
+			endif
+		else
+			call add(l:output, {
+						\   'bufnr': a:buffer,
+						\   'lnum': l:match[1] + 0,
+						\   'col': 0,
+						\   'text': l:match[2] . l:match[3],
+						\   'type': empty(l:match[2]) ? 'E' : 'W',
+						\})
+		endif
+	endfor
+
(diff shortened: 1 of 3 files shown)


ale_linters/ruby/ruby.vim
```diff
@@ -0,0 +1,46 @@
+" Author: Brandon Roehl - https://github.com/BrandonRoehl
+" Description: Ruby MRI for Ruby files
+
+function! ale_linters#ruby#ruby#Handle(buffer, lines) abort
+	" Matches patterns line the following:
```
w0rp: You should use 4 spaces for indentation in this codebase.

ale_linters/ruby/ruby.vim
```diff
@@ -0,0 +1,46 @@
...
+		let l:match = matchlist(l:line, l:pattern)
+
+		if len(l:match) == 0
+			let l:match = matchlist(l:line, l:column)
+			if len(l:match) != 0
+				let l:output[len(l:output) - 1]['col'] = len(l:match[1])
+			endif
+		else
+			call add(l:output, {
+						\   'bufnr': a:buffer,
```
w0rp: Have a look at some other files and try to match the same indentation style in this file.

ale_linters/ruby/ruby.vim
```diff
@@ -0,0 +1,46 @@
+" Author: Brandon Roehl - https://github.com/BrandonRoehl
+" Description: Ruby MRI for Ruby files
+
+function! ale_linters#ruby#ruby#Handle(buffer, lines) abort
+	" Matches patterns line the following:
+	"
+	" test.rb:3: warning: parentheses after method name is interpreted as an argument list, not a decomposed argument
+	" test.rb:8: syntax error, unexpected keyword_end, expecting end-of-input
+	let l:pattern = '\v^' . bufname(a:buffer) . ':(\d+): (warning: )?(.+)$'
```
w0rp: Does the filename need to be escsped here? Some characters could be interpreted as special regex characters.

w0rp: If it's possible to escape the characters, this could work well, potentially.

ale_linters/ruby/ruby.vim
```diff
@@ -0,0 +1,46 @@
...
+						\   'type': empty(l:match[2]) ? 'E' : 'W',
+						\})
+		endif
+	endfor
+
+	return l:output
+endfunction
+
+function! ale_linters#ruby#ruby#GetCommand(buffer) abort
+	return 'ruby -w -c -T1 ' . bufname(a:buffer)
```
w0rp: This checks the file on disk. Can you check input from stdin instead, or use `%t` to check a temporary file, or is there some reason why that's not possible? (Say, import errors.)

If you can only check files, which is not ideal, use the `lint_file` option below for `ale#linter#Define`.

BrandonRoehl:
I changed the code style, made it use the tmp file %t, and added a Vader test for the handler. Waiting for travis to run the tests...

BrandonRoehl:
All checks passed for Vader tests

w0rp:
Cool, thanks! Cheers! :beers:

PR #456, opened 2017-04-08
Added a missing space to the nimcheck command

mathsaey:
I forgot to add a leading space when adding `--threads:on` to the nimcheck command.
My apologies for making this stupid mistake :).

### DIFF
--- a/ale_linters/nim/nimcheck.vim
+++ b/ale_linters/nim/nimcheck.vim
@@ -52,7 +52,7 @@ endfunction
 
 
 function! ale_linters#nim#nimcheck#GetCommand(buffer)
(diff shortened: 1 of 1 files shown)


mathsaey:
I looked into writing some basic tests for nimcheck (based on ` ale/test/c_tests/test_gcc.vader
`), but I'm not sure if this is currently possible with Vader.

It seems that Vader stores the contents of the `Given` buffer as `/some/temp/path/[Vader-workbench]`. This causes `nimcheck` to automatically return `'Error: invalid module name: ''[Vader-workbench]` and abort checking when it attempts to check this file.

I'm not very familiar with Vader (or vimscript for that matter), so I might be missing something here; I'd be happy to look into it further if you have any idea on how this can be fixed.

w0rp:
That's okay. I'll merge this now so it will fix the bug, and give it a go later.

w0rp:
Cheers! :beers:

Issue #257, opened 2017-01-13
Standardise options for linters


w0rp:
Some of the available linters have configurable arguments for the commands that are run, and the executables, but not all of them. The linters should be updated to allow the executables and arguments to be configurable as much as possible, and all use the same naming convention. Existing options which do not follow the naming convention should be deprecated.

## What should be done

Options should be created for each linter with the following conventions:

* `g:ale_<filetype>_<linter_name>_executable` - The executable to run for a given linter.
* `g:ale_<filetype>_<linter_name>_options` - Configurable options for a linter.
* `g:ale_<filetype>_<linter_name>_use_global` - For linters which can use local executables, an option which can be set to `1` to force ALE to always use the global executable, instead of whatever it finds.

Existing variables like `g:ale_html_tidy_args` should be supported for some time until they are ultimately removed, via fallback options. The newer `_options` variables should override whatever is set via previous options.

The variables should be explicitly defined at the top of linter files so the current values can be seen by using `:echo` with autocompletion and so on, so you don't have to read the code or documentation to figure out what the value is.

The current official linters should all be modified to use functions for executables and callbacks instead of strings. This will allow the linters to dynamically update when the options are changed.

Unit tests should be written for each callback for executables and commands.

## Linters to Update

- [ ] rubocop
- [ ] pylint
- [ ] Actually write the rest of them here

Rtzq0:
@w0rp thanks so much for doing this, I got tied up closing other people's issues instead of my own :-)

w0rp:
I have fixed this now. I now have a definition for "until some time." It is "Until version 2.0.0." The project should follow semver semantics.

PR #458, opened 2017-04-09
Added shellcheck executable and options variable

Szero:
Hello, I wanted to use the -x shellcheck argument and I noticed that adding custom options to shellcheck wasn't implemented yet, so I did it, conforming to #257


w0rp:

PR #460, opened 2017-04-10
Typo correcting 3.0.7 -> 0.3.7 for `vint --no-color`

TheLonelyGhost:
Since there is no vint v3.0.7, I assumed (and confirmed) that the change was to target v0.3.7 of vint instead.

Comparing [0.3.7](https://github.com/Kuniwak/vint/blob/v0.3.7/vint/linting/cli.py#L75) with [0.3.6](https://github.com/Kuniwak/vint/blob/v0.3.6/vint/linting/cli.py#L75), it looks like the intended target version was 0.3.7 after all.

Did not add myself as a contributor since this only corrects a minor typo.


w0rp:
Yep, that's just a simple mistake. Cheers! :beers:

PR #462, opened 2017-04-11
Add puppet-lint config options and documentation

bitcrush:
This adds variables for specifying the puppet-lint executable and additional command line arguments. It also removes the hard-coded '--no-autoloader_layout-check' argument.


w0rp:
I'll take this, but I'll set the default value for the options to what is is currently, so it won't break the linter for anyone.

Issue #418, opened 2017-03-26
Make the `go build` linter check files on disk


w0rp:
The `go build` linter doesn't work very well with Vim buffers, and is pretty complicated. It should be much easier to maintain and generally work better with the `lint_file` option instead. The `go build` linter should be changed to building only against the file on disk instead.

w0rp:
I don't understand how `go build` is supposed to work. If anyone knows Go, feel free to give this a go.

breed808:
Goodness, I hadn't meant for that commit to directly reference this.
Well, cat's out of the bag, you may as well review the changes in the most recent commit.

I've stripped most of the previous code and replaced it with the `HandleUnixFormatAsError`. I kept the `GoEnv` function, as the `GOPATH` is needed. Not so sure on the `GOROOT`, so I left that in.
Let me know if I've removed any necessary code, or if anything added needs cleaning (hint: I'm not overly familiar with vimscript).

Once your happy I can send this as a pull request.

w0rp:
@breed808 Yeah, open a pull request. I'd try checking errors in other files first. Say if you have A imports B, and there are errors in B. That's worth testing, as some users have had problems like that before.

breed808:
Looks like the linter is picking up errors in B, but placing the error in A at the same line number.

test.go
```
package main

import "testale"

func main() {
     testale.test()
}
```

testale.go
```
package testale

func test() {
    fmt.Println("Hello World!")
}
```

![testale](https://cloud.githubusercontent.com/assets/3103817/24943997/6c5f50bc-1f9c-11e7-98de-74199595152a.png)

`go test` output is `src/testale/testing.go:4: undefined: fmt in fmt.Println`

Would it be best to copy the `HandleUnixFormat` handler into the `go build` linter and have it set the error column to 1 if the filename doesn't match the current file?

w0rp:
I thought that might happen. You have two options. You can either change the function so it filters out errors from other files, maybe copy it into the linter Vim script so you don't break anything else, or you can find some way of reporting issues on the import lines.

I'd recommend filtering the errors out for now. The C linters will now report errors in a very simple way on include lines, but C uses textual includes instead of modules, so a different story.

The third option is to allow linting one file to set errors for another file. That will take some work. Some work has been done for that already. I wouldn't worry about that now.

I would like to have a function at some point which, given two strings, will tell you if two paths represent the same file on disk.

w0rp:
[...]

PR #465, opened 2017-04-12
Make `go build` linter check files on disk

breed808:
Went with the easier option mentioned in #418: `gobuild` linter will omit errors from imported packages.

Copied the `HandleUnixFormat` handler and set type to 'E'. Copied a small amount of code from the Rust handlers in order to get a filename that can be checked by the newly created handler.

Tested this on a couple of go scripts. File `A` that imports `B` will display errors from `A`, but not from code imported from `B`.

### DIFF
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -1,4 +1,4 @@
-" Author: Joshua Rubin <joshua@rubixconsulting.com>
+" Author: Joshua Rubin <joshua@rubixconsulting.com>, Ben Reedy <https://github.com/breed808>
 " Description: go build for Go files
 
 " inspired by work from dzhou121 <dzhou121@gmail.com>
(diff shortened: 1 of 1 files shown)


w0rp:
Cheers! :beers:

dzhou121:
This PR just breaks the go build linter, when there is multiple files in a package. 

If you go build on files, you have to pass in all the files belong to the same package; or you build against a package. This PR just reverts all the efforts of previous work.

@w0rp The go build linter is complicated because this is the only way. The complication is not caused by on-the-fly check. You need all the files belong to the same package even if you check files on disk.

breed808:
@dzhou121 apologies for that, I hadn't tested multiple files under the same package name.

Perhaps the best option would be to revert the PR and modify the previous `gobuild` linter to check files on disk?

w0rp:
Yes. Prefix the command with `cd`, and that should fix it. See my comments on the gometalinter issue.

PR #466, opened 2017-04-12
Some fixes in markdown documents

adriaanzon:
Changes in GitHub's markdown parser caused the lists to render incorrectly. See https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown

I also fixed some other issues I noticed like heading sizes and broken links.


w0rp:
Cheers! :beers:

PR #467, opened 2017-04-12
PHP: Fix double errors and support fatal errors

adriaanzon:
Because the `-d display_errors=1` argument is used on the PHP linter, errors show up 2 times like this:

```
PHP Parse error:  syntax error, unexpected end of file in - on line 9

Parse error: syntax error, unexpected end of file in - on line 9
Errors parsing -
```

Both of the lines were matched. This fixes it by looking for `PHP Parse error` specifically. 

I wonder if we can remove `-d display_errors=1`  or only look at the first line, but I don't know if either of those options have any side effects.

I also added support for "Fatal errors" by adding `Fatal` in addition to `Parse` to the pattern.


w0rp:
Cheers! :beers:

PR #470, opened 2017-04-13
Fix gobuild linter with multiple source files

breed808:
Resolve issue mentioned in #465 where `gobuild` linter would fail with multiple source files belonging to the same package.


w0rp:
Yep, that'll do it. Cheers! :beers:

Issue #419, opened 2017-03-26
Add support for gometalinter, against files on disk


w0rp:
Using the feature implemented for #333, add support for running `gometalinter`, where available.

It might be time to decide the default list of linters to run for a language dynamically. If you have gotmetalinter installed, you might not want `go build` to run.

saschagrunert:
Yeah support for gometalinter would be great 👍🏽

breed808:
I've attempted to write a linter for `gometalinter`, which hasn't been too difficult.
I've set it to be disabled by default, but the major issue I've found is that [gometalinter doesn't support linting single files](https://github.com/alecthomas/gometalinter/issues/119).

I'm familiar enough with vim-ale that I can pass a temporary file with `%t` or the full path to the opened file with `%s`, but `gometalinter` will need the directory containing the file. I'm not certain how to do this, or what best practice would be.

Thoughts?

saschagrunert:
What about running gometalinter in the current vims pwd and then extracting the "lint issues" to see if it is relevant for the current open buffers?

w0rp:
@breed808 I recommend adding the directory containing the file in the command manually, and then parsing the output to extract the errors for the current file. (Filter out the errors from other files.)

If you can accomplish this by changing the directory, then you can write this:

```vim
" Put whatever other arguments on the end here.
return ale#util#BufferCdString(a:buffer) . 'gometalinter ...'
```

Otherwise, you could write this, if `gometalinter` accepts a directory name:

```vim
return  'gometalinter ... ' . fnameescape(fnamemodify(bufname(a:buffer), ':p:h'))
```

Filtering for filenames should be roughly the same as it is with the `go build` linter.

Remember to set `'lint_file': 1`.

w0rp:
This has now been implemented, thanks to @breed808 and a little bit of fiddling around from me. Some more Vader tests are still needed, but everyone can try this out now. See `:help ale-go-options` first for information on turning this linter on. It is disabled by default.

PR #471, opened 2017-04-13
Add gometalinter linter for go files

breed808:
Referring to #419 for this.
Added the `gometalinter` to ALE, but left it disabled by default (see `g:ale_go_gometalinter_enabled`) as it conflicts with a number of existing ALE linters for Go (`golint`, `govet`, `gosimple`, `staticcheck`, etc).

Vader tests for this haven't been written yet, but it is on my to-do list (along with the tests for the `gobuild` linter).

### DIFF
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -1,0 +1,61 @@
+" Author: Ben Reedy <https://github.com/breed808>
+" Description: Adds support for the gometalinter suite for Go files
+
+" Disable gometalinter by default
+if !exists('g:ale_go_gometalinter_enabled')
+    let g:ale_go_gometalinter_enabled = 0
+endif
+
+if !exists('g:ale_go_gometalinter_options')
+    let g:ale_go_gometalinter_options = ''
+endif
+
+function! ale_linters#go#gometalinter#GetCommand(buffer) abort
+    if g:ale_go_gometalinter_enabled == 1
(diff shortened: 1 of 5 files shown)


ale_linters/go/gometalinter.vim
```diff
@@ -0,0 +1,61 @@
...
+if !exists('g:ale_go_gometalinter_enabled')
+    let g:ale_go_gometalinter_enabled = 0
+endif
+
+if !exists('g:ale_go_gometalinter_options')
+    let g:ale_go_gometalinter_options = ''
+endif
+
+function! ale_linters#go#gometalinter#GetCommand(buffer) abort
+    if g:ale_go_gometalinter_enabled == 1
```
w0rp: Instead of doing this, you should edit the default list of enabled linters for `Go`. Look for `s:default_ale_linters` in `autoload/ale/linter.vim`. That will update the output of `:ALEInfo` appropriately. (It will show up in the list of available linters, but not in the enabled list.)

Write a paragraph about this at the top of the Go integration file. Look at the Rust file for an example of some documentation like that.

w0rp:
What's most important is having a test where you feed the handler function some lines of output, with the current filename and another filename, and then check that the correct errors are returned.

w0rp:
I had some time, so I made the changes for using `g:ale_linters` for enabling the linter instead, plus the documentation. The tests can come later.

PR #474, opened 2017-04-14
Add linter for Crystal lang

jordanandree:
Adds linter for [Crystal lang](https://crystal-lang.org) using the packaged `build` command:

```sh
$ crystal build -f json --no-codegen -o /dev/null
```
the `-o` argument will implement `g:ale#util#nul_file`

also adds Vader tests


jordanandree:
@w0rp ah totally missed that configuration option! Put up another PR with that switched out since it was producing buggy behavior.

PR #475, opened 2017-04-14
add goflycheck for go files

dzhou121:
Add a new linter goflycheck for Golang. This linter is able to read stdin of the file buffer and take care of Golang package complexity from ale. 

https://github.com/dzhou121/goflycheck

### DIFF
--- a/ale_linters/go/goflycheck.vim
+++ b/ale_linters/go/goflycheck.vim
@@ -1,0 +1,9 @@
+" Author: dzhou121 <dzhou121@gmail.com>
+" Description: goflycheck for Go files
+
+call ale#linter#Define('go', {
+\   'name': 'goflycheck',
+\   'executable': 'goflycheck',
+\   'command': 'goflycheck %s',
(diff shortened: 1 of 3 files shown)


ale_linters/go/goflycheck.vim
```diff
@@ -0,0 +1,9 @@
+" Author: dzhou121 <dzhou121@gmail.com>
+" Description: goflycheck for Go files
+
+call ale#linter#Define('go', {
+\   'name': 'goflycheck',
+\   'executable': 'goflycheck',
+\   'command': 'goflycheck %s',
```
w0rp: If you're checking the file on disk, you should add the `'lint_file': 1` flag below. Otherwise the linter will run while you're typing.

w0rp:
Oh, this is the first argument which gives you the filename, and then it also reads stdin, and it's a program you wrote? I did not expect that.

I don't think this is worth supporting right now, officially. I'd recommend adding the linter configuration to your own `.vim` directory, and maybe getting some others to field test it first.

PR #476, opened 2017-04-14
Improve flake8 message severity

### LINKED ISSUE #430: Feature request: Distinguish style warnings/errors from normal warnings/errors
I'm trying out ALE as a replacement for [Syntastic](https://github.com/vim-syntastic/syntastic).

Syntastic distinguishes between regular errors/warnings and style-related errors/warnings (such as whitespace issues).

It would be nice if ALE distinguished between these as well. Currently, style errors such as too many blank lines in `flake8` show up as errors while more important issues like undefined variables show up as warnings.

ALE screenshot:

screenshot

Syntastic screenshot:

screenshot2

Chronial:
(party) fixes #430 

The `g:ale_python_flake8_error_codes` is deliberately undocumented – It seems unlikely that someone might want to modify it. but it's overrideable just in case.

### DIFF
--- a/ale_linters/python/flake8.vim
+++ b/ale_linters/python/flake8.vim
@@ -7,6 +7,34 @@ let g:ale_python_flake8_executable =
 let g:ale_python_flake8_args =
 \   get(g:, 'ale_python_flake8_args', '')
 
+" E999: Synax Error
+"
+" F601: dictionary key name repeated with different values
+" F602: dictionary key variable name repeated with different values
+" F621: too many expressions in an assignment with star-unpacking
+" F622: two or more starred expressions in an assignment (a, *b, *c = d)
+" F631: assertion test is a tuple, which are always True
+"
+" F701: a break statement outside of a while or for loop
+" F702: a continue statement outside of a while or for loop
+" F703: a continue statement in a finally block in a loop
+" F704: a yield or yield from statement outside of a function
+" F705: a return statement with arguments inside a generator
+" F706: a return statement outside of a function/method
+" F707: an except: block as not the last exception handler
+"
+" F811: redefinition of unused name from line N
+" F821: undefined name name
+" F822: undefined name name in __all__
+" F823: local variable name ... referenced before assignment
+let g:ale_python_flake8_error_codes =
+\   get(g:, 'python_flake8_error_codes', [
+\     'E999',
+\     'F601', 'F602', 'F621', 'F622', 'F631',
+\     'F701', 'F702', 'F703', 'F704', 'F705', 'F706', 'F707',
+\     'F811', 'F821', 'F823', 'F831',
+\   ])
+
 " A map from Python executable paths to semver strings parsed for those
 " executables, so we don't have to look up the version number constantly.
 let s:version_cache = {}
@@ -72,5 +100,5 @@ call ale#linter#Define('python', {
 \       {'callback': 'ale_linters#python#flake8#VersionCheck'},
 \       {'callback': 'ale_linters#python#flake8#GetCommand'},
 \   ],
-\   'callback': 'ale#handlers#HandlePEP8Format',
+\   'callback': 'ale#handlers#HandleFlake8Format',
 \})
--- a/autoload/ale/handlers.vim
+++ b/autoload/ale/handlers.vim
@@ -76,8 +76,6 @@ endfunction
 function! ale#handlers#HandlePEP8Format(buffer, lines) abort
     " Matches patterns line the following:
     "
-    " Matches patterns line the following:
-    "
     " stdin:6:6: E111 indentation is not a multiple of four
     " test.yml:35: [EANSIBLE0002] Trailing whitespace
     let l:pattern = '^' . s:path_pattern . ':\(\d\+\):\?\(\d\+\)\?: \[\?\(\([[:alpha:]]\)[[:alnum:]]\+\)\]\? \(.*\)$'
@@ -109,6 +107,45 @@ function! ale#handlers#HandlePEP8Format(buffer, lines) abort
         \   'col': l:match[2] + 0,
         \   'text': l:code . ': ' . l:match[5],
         \   'type': l:match[4] ==# 'E' ? 'E' : 'W',
+        \   'nr': -1,
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+
+function! ale#handlers#HandleFlake8Format(buffer, lines) abort
+    " Matches patterns line the following:
+    "
+    " stdin:6:6: E111 indentation is not a multiple of four
+    "
+    " Uses g:ale_python_flake8_error_codes to decide the severity of the
+    " message
+    let l:pattern = '^' . s:path_pattern . ':\(\d\+\):\?\(\d\+\)\?: \(\([[:alpha:]]\)[[:alnum:]]\+\) \(.*\)$'
+    let l:output = []
+
+    for l:line in a:lines
+        let l:match = matchlist(l:line, l:pattern)
+
+        if len(l:match) == 0
+            continue
+        endif
+
+        let l:code = l:match[3]
+        if (l:code ==# 'W291' || l:code ==# 'W293')
+                    \ && !g:ale_warn_about_trailing_whitespace
+            " Skip warnings for trailing whitespace if the option is off.
+            continue
+        endif
+
+        call add(l:output, {
+        \   'bufnr': a:buffer,
+        \   'lnum': l:match[1] + 0,
+        \   'vcol': 0,
+        \   'col': l:match[2] + 0,
+        \   'text': l:code . ': ' . l:match[5],
+        \   'type': index(g:ale_python_flake8_error_codes , l:code) == -1 ? 'W' : 'E',
         \   'nr': -1,
         \})
     endfor


w0rp:
Overriding the option could be useful. If you add some extra `flake8` plugins, you might want to configure some codes as warnings or errors.

ale_linters/python/flake8.vim
```diff
@@ -7,6 +7,34 @@ let g:ale_python_flake8_executable =
...
+" F811: redefinition of unused name from line N
+" F821: undefined name name
+" F822: undefined name name in __all__
+" F823: local variable name ... referenced before assignment
+let g:ale_python_flake8_error_codes =
+\   get(g:, 'python_flake8_error_codes', [
+\     'E999',
+\     'F601', 'F602', 'F621', 'F622', 'F631',
+\     'F701', 'F702', 'F703', 'F704', 'F705', 'F706', 'F707',
+\     'F811', 'F821', 'F823', 'F831',
```
w0rp: Could you write one error code per line? That will lead to better diffs in the future, when someone else edits this.

autoload/ale/handlers.vim
```diff
@@ -116,6 +114,45 @@ function! ale#handlers#HandlePEP8Format(buffer, lines) abort
...
+            continue
+        endif
+
+        call add(l:output, {
+        \   'bufnr': a:buffer,
+        \   'lnum': l:match[1] + 0,
+        \   'vcol': 0,
+        \   'col': l:match[2] + 0,
+        \   'text': l:code . ': ' . l:match[5],
+        \   'type': index(g:ale_python_flake8_error_codes , l:code) == -1 ? 'W' : 'E',
```
w0rp: I have some suggestions here, after thinking about this a little more.

I reckon we should create a `Dictionary` instead of a `List`, and the values in the dictionaries should be the error types. That way, you could map some things to errors, other things to warnings, and then later some things to style warnings. (`'S'`) That will be useful for configuring extra `flake8` plugins.

I also think we should look at the start of the error code to distinguish between warnings or errors, so some things will be handled correctly by default. Your changes here could turn some errors into warnings. So errors beginning with `'W`' should be treated as warnings by default, and maybe treat everything else like an error by default.

We should probably provide no codes by default. Reading all of the `F` errors above, I would call them _errors_, not _warnings_. It's probably a matter of opinion, so it should be configurable. I recommend merging the default `Dictionary` with the user-configured `Dictionary`.

I'd write these changes into the `HandlePEP8Format` function above, change the global variable name to `g:ale_python_error_codes`, and document it in the Python documentation. Add some Vader tests for it. (Set some codes, call the function for handling lines of output, check the objects returned.)

ale_linters/python/flake8.vim
Chronial: I disagree that that would improve the diffs, because that means that you have to change two code lines (the comment and the list entry) that are now even further apart. But this is your codebase :)

autoload/ale/handlers.vim
Chronial: The distinction between errors and warnings I did here is not very opinion based: Everything that will keep the code from running correctly is an error, everything else is a warning. I would argue that that is the most sensible distinction, because that means you can't disagree with errors and need to fix them, but you might disagree with warnings and ignore them.

> Reading all of the F errors above, I would call them errors, not warnings.

You might have misunderstood my code – the list contains everything that is an error. Everything else is a warning. So all these F errors are actually classified as errors.

> I also think we should look at the start of the error code to distinguish between warnings or errors, so some things will be handled correctly by default.

That doesn't work. In Flake8 the letter at the start mostly informs you where the error is from. All the `E` messages are generated by `pycodestyle` (former `pep8`), so are usually style problems, not errors.

> We should probably provide no codes by default.

I strongly disagree with that, as that just provides a useless default configuration. Message examples:
* E112 	expected an indented block (this a syntax error)
* E121 	continuation line under-indented for hanging indent
* E201 	whitespace after ‘(‘
* E999 Generic Syntax Error
* F823 	local variable name ... referenced before assignment (this is always a runtime error)
* F701 	a break statement outside of a while or for loop (this is a syntax error)
* F401 	module imported but unused

So are E and F messages errors or warnings?

> I'd write these changes into the HandlePEP8Format function above, change the global variable name to g:ale_python_error_codes, and document it in the Python documentation. 

You seem to be assuming that all python tools share these error codes? That is not correct. `flake8` inherits `pycodestyle`'s (former `pep8`) list and extends it, but pylint completely does its own thing.

> I reckon we should create a Dictionary instead of a List, and the values in the dictionaries should be the error types. That way, you could map some things to errors, other things to warnings, and then later some things to style warnings. ('S') That will be useful for configuring extra flake8 plugins.

I don't quite see the point of this – what is the difference between a warning and a style warning?

ale_linters/python/flake8.vim
w0rp: It means if you remove only one code, you will see one line being changed, and if you add one code, you will see one line being added. It's easier to track in diffs.

autoload/ale/handlers.vim
w0rp: Everything that begins with `E` from `flake8` and that begins with `W` from `flake8` should be output as errors and warnings, by default, respectively. The rest is a matter of configuration.

w0rp: The `'S'` warnings are a feature to come later. That's so different signs can be used, different highlight syntax can be used, whatever. Everything can be just `'W'` and `'E'` for now.

w0rp: The flake8 errors and warnings are a superset for pep8/pycodestyle. So it makes sense to write these changes into the PEP8 handler, and use the same settings for both. It shouldn't cause any issues.

w0rp: Remove all of the default error code mappings, and treat codes beginning with `F` as errors by default. You want something like this.

```vim
let l:type = get(
\    g:ale_python_error_codes , 
\    l:code, 
\    l:code =~# '\v^(E|F)' ? 'E' : 'W'
\),
```

That will cover current and future error codes.

ale_linters/python/flake8.vim
Chronial: But the comment and the list entry have to be on different lines, or is there a vimscript feature I don't know about?

w0rp: I'm asking for this:

```vim
[
\   'E999',
\   'F601',
\   'F602',
\   'F621', 
\   'F622', 
\   'F631',
```

One line for each code. However, referring to the comments below, I think you should remove all of the default codes here.

autoload/ale/handlers.vim
Chronial: How does your solution improve #430? Wouldn't it result in the same output that smkent was complaining about?

It's also a solution that I'm personally not interested in at all. To get what I would consider sensible behavior (errors for errors, warnings for style issues) in this situation, I would have to list [all of pycodestyle's error codes](https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes) in my vimrc.

> The flake8 errors and warnings are a superset for pep8/pycodestyle. So it makes sense to write these changes into the PEP8 handler, and use the same settings for both. It shouldn't cause any issues.

What about [ansible](https://github.com/w0rp/ale/blob/ecbb27680563a50d4dd981f968d659ef20bfbe30/ale_linters/ansible/ansible-lint.vim#L8) and [pylint](https://github.com/w0rp/ale/blob/e97dada261c4a69f43c5a6c34d349ad6246fe34c/ale_linters/python/pylint.vim)?

w0rp: Because what is a "warning" or "error" depends on the user's perspective, the only default we can set is to make all errors, as defined by pycodestyle errors, and all warnings, as defined by pycodestyle, warnings. So anything beginning with `E` is an error, and anything beginning with `W` is a warning. We should stick with what pycodestyle says by default.

The defaults can then be changed to your liking.

Chronial: So I am understanding correctly that you are not interested in merging a PR that provides a quick fix for the situation in #430 in the default configuration and just want to classify everything but the 10 warning types of pycodestyle as an error?

w0rp: Correct.

Chronial: That's a shame :(. Since your solution only complicates the situation for me compared to my current fix (I would have to list a lot more codes in my config and update that whenever a minor check is added), I am not interested in providing a PR according to your design.

w0rp: Okay then.

Chronial:
@w0rp any chance you are willing to revisit this?

The current situation is this:
* `F401 module imported but unused` is a non-style error.
* `E112 expected an indented block` is a style error. This is actually a syntax error.

w0rp:
Sure, I'll make those updates.

Chronial:
Thx :). For the same reason all of these should be warnings: (that is every error starting with `F4`)
>F401 | module imported but unused
>F402 | import module from line N shadowed by loop variable
>F403 | ‘from module import *’ used; unable to detect undefined names
>F404 | future import(s) name after other statements
>F405 | name may be undefined, or defined from star imports: module
>F406 | ‘from module import *’ only allowed at module level
>F407 | an undefined __future__ feature name was imported

Same goes for these (these are *not* all errors starting with `F8`):
> F812 | list comprehension redefines name from line N
> F831 | duplicate argument name in function definition
> F841 | local variable name is assigned to but never used

PR #477, opened 2017-04-15
Flow: Fix case where 1st msg points to different file

ravicious:
Take a look at [an example gist](https://gist.github.com/ravicious/ceaf1fd5c60791f527c8edc4a34e45d8). We assume that we have ale v1.2.0 installed and we open `index.js`. `error.json` represents the JSON output of Flow.

The current implementation takes [the first location information](https://github.com/w0rp/ale/blob/c9a5d9845b9bff9174c4bb3b67a9f7bfda190ee6/ale_linters/javascript/flow.vim#L50-L53) that it spots in `errors[0].message` elements. If you look at [`error.json`](https://gist.github.com/ravicious/ceaf1fd5c60791f527c8edc4a34e45d8#file-error-json), this is not correct in this case, as `errors[0].message[0].loc` points to a different file (`foo.js`) than the one opened in the current buffer (`index.js`).

This patch ensures that we only take the location information into consideration if the location source is equal to the buffer path.

Sorry for not writing any tests. 😐 I'm not familiar with Vader, but even if I was, I'd first need to come up with a more straightforward example that doesn't need React to be installed.

### DIFF
--- a/test/handler/test_flow_handler.vader
+++ b/test/handler/test_flow_handler.vader
@@ -113,14 +113,12 @@ Execute(The flow handler should process errors correctly.):
   let g:expected = [
   \ {
   \   'lnum': 417,
-  \   'bufnr': 347,
   \   'type': 'E',
   \   'col': 10,
   \   'text': 'number: This type is incompatible with the expected return type of array type',
   \ },
   \ {
   \   'lnum': 419,
-  \   'bufnr': 347,
   \   'type': 'W',
   \   'col': 3,
   \   'text': 'unreachable code:',
--- a/test/handler/test_flow_handler.vader
+++ b/test/handler/test_flow_handler.vader
@@ -1,4 +1,15 @@
 Before:
+  runtime ale_linters/javascript/flow.vim
+
+After:
+  unlet! g:flow_output
+  unlet! g:expected
+  unlet! g:actual
+  call ale#linter#Reset()
+
+Execute(The flow handler should process errors correctly.):
+  e! /home/w0rp/Downloads/graphql-js/src/language/parser.js
+
   let g:flow_output = {
   \ "flowVersion": "0.39.0",
   \ "errors": [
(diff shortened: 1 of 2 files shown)


w0rp:
You can remove the `bufnr': a:buffer,` line in the file you are editing here now. That will be set automatically by ALE later on, and that will make the tests easier to write.

ravicious:
Cool, thanks! I'll look into it and the failing test.

ravicious:
So far I've been trying to just fix the failing test and I'm stuck at setting the filename for the buffer.

I tried adding

```vimscript
file /home/w0rp/Downloads/graphql-js/src/language/parser.js
```

as the first line of the execute block, but it doesn't help. I checked to what string `expand('#' . a:buffer . ':p')` resolves to in `ale_linters#javascript#flow#Handle` and it seems that it resolves to an empty string (I checked that by setting a global variable in that function and then logging it in the vader file after executing the function).

I see that `347` is passed as the buffer number to `ale_linters#javascript#flow#Handle`, so I'd need to somehow set the file for buffer 347 to `/home/w0rp/Downloads/graphql-js/src/language/parser.js`, but I don't know if it's even possible.

Setting `loc.source` property of the error message to an empty string instead of `/home/w0rp/Downloads/graphql-js/src/language/parser.js` fixes the test, but I'm not sure if that's the correct approach.

w0rp:
@ravicious Have you tried passing `bufnr('')` as he first argument of the `Handle` function? If `:file` doesn't work in Vader, try using `:e!` instead.

ravicious:
Passing `bufnr('')` and using `:e!` fixes (along with removing `bufnr`) fixes the failing test! 👍

I'm going to write a new test for my patch.

test/handler/test_flow_handler.vader
```diff
@@ -1,4 +1,15 @@
...
+After:
+  unlet! g:flow_output
+  unlet! g:expected
+  unlet! g:actual
+  call ale#linter#Reset()
+
+Execute(The flow handler should process errors correctly.):
+  e! /home/w0rp/Downloads/graphql-js/src/language/parser.js
+
   let g:flow_output = {
```
ravicious: Inspired by the ESLint handler tests, I moved the output from the `After` block into the `Execute` block.

test/handler/test_flow_handler.vader
```diff
@@ -100,31 +111,124 @@ Before:
...
-  \   'bufnr': 347,
   \   'type': 'W',
   \   'col': 3,
   \   'text': 'unreachable code:',
   \ },
   \]
 
   AssertEqual g:expected, g:actual
+
+Execute(The flow handler should fetch the correct location for the currently opened file, even when it's not in the first message.):
```
ravicious: I'm not sure about this test description, it's kinda lengthy, but I wasn't able to come up with anything more succinct. Should I break it into two lines somehow?

w0rp: Nah, that's fine. It's better to have the Vader lines be long and descriptive, than being hard to understand.

w0rp:
Cheers! :beers:

PR #479, opened 2017-04-16
Fix htmlhint variable names

adriaanzon:
Thought: It might be nice to also have a function that sets `ale_...` variables, to further avoid these errors but also just because it's nicer when you don't have to specify the variable name twice, like you do when using `let ... = get(...)`.


w0rp:
That's my bad. That looks good to me.

I considered adding a function for setting variables, but it seems like overkill. I tend to avoid hiding logic in small functions until I'm positive it's an improvement. I'll keep considering it.

PR #482, opened 2017-04-17
Update flow.vim

fgarcia:
The current exact equality operator breaks all my errors. One side of the operator uses a full file path, and the other side uses a relative path according the current workdir. 

WARNING: I have not tested this with Vader. I just needed a quick fix to solve my current problem (all my flow errors lack a line number). This fixes my problem and I thought an untested pull request is more helpful than a plain issue.


w0rp:
I have fixed this now with a new version of `ale#path#IsBufferPath` and a Vader test.

PR #486, opened 2017-04-18
Add rpm spec file linter (rpmlint)

jasontibbitts:
This is a rough attempt at adding a linter for RPM spec files using rpmlint.  It appears to work well enough, but note the following:

* I basically don't know what I'm doing.  I have many years of programming experience but outside of setting variables in my .vimrc, vim isn't something I've done before.
* Some rpmlint errors do not have line numbers.  I just made them show up at line 1 but perhaps there's a better way to handle that.
* The vim modeline reads 'spec' but everyone's going to call them "RPM spec files".  I'm not sure where that should actually go in the documentation.
* Editing vim help files is... weird.  I tried to get things to line up the right way but I'm not sure if it's done the right way.
* I added an entry for the test suite but I didn't try actually running it.  I guess the CI setup will yell at me if it doesn't run properly.

So, anyway, it's a first try.  Please let me know what I can do to make it better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ name. That seems to be the fairest way to arrange this table.
 | Puppet | [puppet](https://puppet.com), [puppet-lint](https://puppet-lint.com) |
 | Python | [flake8](http://flake8.pycqa.org/en/latest/), [mypy](http://mypy-lang.org/), [pylint](https://www.pylint.org/) |
 | reStructuredText | [proselint](http://proselint.com/)|
+| RPM spec | [rpmlint](https://github.com/rpm-software-management/rpmlint) |
 | Ruby | [rubocop](https://github.com/bbatsov/rubocop), [ruby](https://www.ruby-lang.org) |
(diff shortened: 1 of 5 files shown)


jasontibbitts:
Well crap, I check back and now the checks failed.  The log doesn't make it obvious but I would guess I screwed up the test I added.  I guess I'll try to figure out how to run it myself.

jasontibbitts:
I played around with it a bit but no luck.  I am not seeing where the syntax error in my test is, and while I can get the test suite to start, it just hangs after the first incidence of:

```
docker run -a stderr --rm -v /home/tibbs/work/ale:/testplugin -v /home/tibbs/work/ale/test:/home w0rp/ale vim-v8.0.0000 +Vader! test/*.vader test/*/*.va   der test/*/*/*.vader test/*/*/*/*.vader
Vim: Warning: Output is not to a terminal
Vim: Warning: Input is not from a terminal
```

If you have any hints, that would be great.

w0rp:
Cheers! :beers:

jasontibbitts:
Thanks!

PR #487, opened 2017-04-18
Marks persist when switching from lintable to not lintable buffer

vavaka:
Hi

Please find below steps to reproduce persistent marks problem:
1. Open lintable file which contains marks
2. From this file switch to not lintable file
You should notice that some letters in not lintable files are highlighted. This happens because Ale does not cleared marks info on buffer switch.

### DIFF
--- a/autoload/ale/filetypes.vim
+++ b/autoload/ale/filetypes.vim
@@ -58,3 +58,8 @@ function! ale#filetypes#GuessExtension(filetype) abort
     " Return an empty string if we don't find anything.
     return ''
 endfunction
+
+function! ale#filetypes#IsLintable(filetype) abort
+    let l:linters = ale#linter#Get(a:filetype)
+    return len(l:linters) != 0
+endfunction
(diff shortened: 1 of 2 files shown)


vavaka:
> The ale_enabled check is needed, so that shouldn't be removed.
`ale_enabled` has not been removed. I just added one more condition for clearing matches state in order it happened when current buffer filetype has no associated linter.

w0rp:
Ah sorry, my bad.

I think this is not quite complete, as the same problem might come up when you open another file which _is_ lintable. I'll think about this.

w0rp:
Yes, the problem occurs with files you can lint as well, as I suspected. A little bit of refactoring and an extra call in the existing cleanup function will fix it. I'll push a commit soon.

w0rp:
Here is the commit: https://github.com/w0rp/ale/commit/d8dcc56af58d1545cf3a7a9f6160d8ba6ad39923

That should cover this case, and changing to any other file in general.

PR #490, opened 2017-04-19
Add the Reek checker for Ruby.

elebow:
Add support for Reek, a Ruby code smell detector.

### DIFF
--- a/ale_linters/ruby/reek.vim
+++ b/ale_linters/ruby/reek.vim
@@ -13,7 +13,6 @@ function! ale_linters#ruby#reek#Handle(buffer, lines) abort
     for l:error in l:errors
         for l:location in l:error.lines
             call add(l:output, {
-            \    'bufnr': a:buffer,
             \    'lnum': l:location,
             \    'type': 'W',
             \    'text': s:BuildText(l:error),
--- a/ale_linters/ruby/reek.vim
+++ b/ale_linters/ruby/reek.vim
@@ -1,0 +1,48 @@
+" Author: Eddie Lebow https://github.com/elebow
+" Description: Reek, a code smell detector for Ruby files
+
+function! ale_linters#ruby#reek#Handle(buffer, lines) abort
+    if len(a:lines) == 0
+        return []
+    endif
+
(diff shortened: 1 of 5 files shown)


ale_linters/ruby/reek.vim
```diff
@@ -0,0 +1,48 @@
...
+        endfor
+    endfor
+
+    return l:output
+endfunction
+
+function! s:BuildText(error) abort
+    let l:text = a:error.smell_type . ':'
+
+    if get(g:, 'ale_ruby_reek_show_context', 0)
```
w0rp: You should set some default values for these at the top of the file, like in the other files, and then use `ale#Var` here. `ale#Var` is being used so all settings can be overridden in each buffer.

ale_linters/ruby/reek.vim
```diff
@@ -0,0 +1,48 @@
...
+    endif
+
+    let l:errors = json_decode(a:lines[0])
+
+    let l:output = []
+
+    for l:error in l:errors
+        for l:location in l:error.lines
+            call add(l:output, {
+            \    'bufnr': a:buffer,
```
w0rp: You don't have to set `bufnr` if you don't want to. That key will be set automatically by ALE later on.

test/handler/test_reek_handler.vader
```diff
@@ -0,0 +1,67 @@
+Execute(The reek handler should parse JSON correctly):
+    runtime ale_linters/ruby/reek.vim
+
+    let g:ale_ruby_reek_show_context = 1
+    let g:ale_ruby_reek_show_wiki_link = 0
```
w0rp: You should split each of these checks into separate `Execute` blocks. That way, the tests can fail later for 1 out of 4 cases, and we can spot which one is broken.

elebow:
That should address the feedback. Do you want me to squash the fixup commits?

w0rp:
Nah, I'd rather have a bunch of commits in a pull request. Then I check check out "changes since your last review" in GitHub. I tend to squash merge pull requests.

w0rp:
Cheers! 🍻

PR #496, opened 2017-04-22
Haskell:  new ghc-mod linter

wizzup:
Since `hdevtools` seem to have [problem](https://github.com/hdevtools/hdevtools/issues/49) with `stack`, I would love to propose `ghc-mod` as alternative linter.

I made this by modify the `hdevtools` linter. ` ale/ale_linters/haskell/hdevtools.vim
`
Never develop vim plugin before, I have no idea how to full-test.

I have use it for a while and it seem to work. I have [eagletmt/ghcmod-vim
](https://github.com/eagletmt/ghcmod-vim) and lint message seem to be the same.

Please take a look

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ name. That seems to be the fairest way to arrange this table.
 | Go | [gofmt -e](https://golang.org/cmd/gofmt/), [go vet](https://golang.org/cmd/vet/), [golint](https://godoc.org/github.com/golang/lint), [gometalinter](https://github.com/alecthomas/gometalinter), [go build](https://golang.org/cmd/go/), [gosimple](https://github.com/dominikh/go-tools/tree/master/cmd/gosimple), [staticcheck](https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck) |
 | Haml | [haml-lint](https://github.com/brigade/haml-lint)
 | Handlebars | [ember-template-lint](https://github.com/rwjblue/ember-template-lint) |
-| Haskell | [ghc](https://www.haskell.org/ghc/), [hlint](https://hackage.haskell.org/package/hlint), [hdevtools](https://hackage.haskell.org/package/hdevtools) |
+| Haskell | [ghc](https://www.haskell.org/ghc/), [ghc-mod](https://github.com/DanielG/ghc-mod), [hlint](https://hackage.haskell.org/package/hlint), [hdevtools](https://hackage.haskell.org/package/hdevtools) |
 | HTML | [HTMLHint](http://htmlhint.com/), [proselint](http://proselint.com/), [tidy](http://www.html-tidy.org/) |
 | Java | [javac](http://www.oracle.com/technetwork/java/javase/downloads/index.html) |
 | JavaScript | [eslint](http://eslint.org/), [jscs](http://jscs.info/), [jshint](http://jshint.com/), [flow](https://flowtype.org/), [standard](http://standardjs.com/), [xo](https://github.com/sindresorhus/xo)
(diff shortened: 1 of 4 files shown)


wizzup:
> Could you provide some example lines of output ..

Sure, this is test code.

```
$ cat check1.hs 
-- Error: missing import
import Missing

$ cat check2.hs 
my_variable :: Int
my_variable = 0

-- Warning: defined but not use
myFunc :: Int -> Int
myFunc x = succ x

-- Warning: Top level binding missing signature
main = print my_variable
```

And this is ghc-mod output

```
$ ghc-mod check check1.hs 
check1.hs:2:1:Failed to load interface for ‘Missing’Use -v to see a list of the files searched for.

$ ghc-mod check check2.hs 
check2.hs:6:1:Warning: Defined but not used: ‘myFunc’
check2.hs:9:1:Warning: Top-level binding with no type signature: main :: IO ()
```

wizzup:
Sorry, I am trying to rebase with upstream and end up with ugly commit graph.
If you accept this PR, hope you can squash them to single commit.

w0rp:
Maybe you can add the `lint mode` part in another pull request. I don't know what `lint mode` is, as I'm not a Haskell user. I'll add some tests for the output you have pasted above and merge this pull request later. Thanks for the info!

wizzup:
I just found that ghc-mod document indicate that it use `hlint` internally so I think check-mode is enough.

For the mode, from the `ghc-mod --help`, it seem like lint-mode use `hlint` and check-mode use `ghc` internally.

This might make `ghc-mod` unnecessary form ale point-of-view since there are `ghc` and `hlint` linter already.

ghc-mod seem to have lots of functionally but I don't know if ale will benefit from any of it

```
Available commands:
  lang                     List all known GHC language extensions
  flag                     List GHC -f<foo> flags
  debug                    Print debugging information. Please include the
                           output in any bug reports you submit
  debug-component          Debugging information related to cabal
                           component resolution
  boot                     Internal command used by the emacs frontend
  root                     Try to find the project directory.
  legacy-interactive       ghc-modi compatibility mode
  list                     List all visible modules
  modules                  List all visible modules
  dumpsym                  
  find                     List all modules that define SYMBOL
  doc                      Try finding the html documentation directory for the
                           given MODULE
  lint                     Check files using `hlint'
  browse                   List symbols in a module
  check                    Load the given files using GHC and report
                           errors/warnings, but don't produce output files
  expand                   Like `check' but also pass `-ddump-splices' to GHC
  info                     Look up an identifier in the context of FILE (like
                           ghci's `:info')
  type                     Get the type of the expression under (LINE,COL)
  split                    Split a function case by examining a type's
                           constructors
  sig                      Generate initial code given a signature
  auto                     Try to automatically fill the contents of a hole
  refine                   Refine the typed hole at (LINE,COL) given EXPR
```
What do you think? Should you keep ghc-mod?

w0rp:
Hmm, I'm not sure myself. Maybe some other Haskell users have an opinion.

w0rp:
I'll hold off on writing the tests for this until we can decide if we want this or not.

shym:
I would be happy to get `ghc-mod` support since it provides (in my experience, maybe I missed something) better warning / error reports than `hdevtools` (and than `ghc -fno-code -v0` ; maybe ghc requires other arguments to be more complete).
The `lint` subcommand is indeed redundant with hlint, only the formatting of the messages is different.

wizzup:
@shym You were right, I found that `ghc-mod` linter give warning `Defined but not used: 'var'` but `ghc` linter is not.

PR #497, opened 2017-04-23
Add erb linter

wikimatze:


ale_linters/eruby/erubylint.vim
```diff
@@ -0,0 +1,27 @@
+" Author: Matthias Guenther - https://wikimatze.de
+" Description: erb-lint for eruby/erb files
+
+function! ale_linters#eruby#erubylint#Handle(buffer, lines) abort
```
w0rp: I think the existing function in `ale_linters/ruby/ruby.vim` will handle these errors better. It has some slightly more complex code for handling column numbers. I recommend moving that existing handler to a new file `autoload/ale/handlers/ruby.vim`, renaming it appropriately for autoload semantics, and calling it from here. Update the tests appropriately.

That way, we'll be able to parse more errors here, and we'll get better test coverage.

ale_linters/eruby/erubylint.vim
```diff
@@ -0,0 +1,27 @@
...
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+call ale#linter#Define('eruby', {
+\   'name': 'erubylint',
+\   'executable': 'erb',
+\   'command': 'erb -P -x %t | ruby -c 2>&1',
```
w0rp: If you want to, you can use the `output_stream` option to read from stderr instead, so you don't need to redirect stderr to stdout.

wikimatze: Awesome idea, accomplished via https://github.com/w0rp/ale/pull/497/commits/a199c9b781fc846d8ef69a24391711f58e796b2a

ale_linters/eruby/erubylint.vim
wikimatze: Got, if I'm with https://github.com/w0rp/ale/pull/497/commits/a94e6c7f282e8b6951263f409f73e87001232175 on the right way, I'll try to add tests.

autoload/ale/handlers/ruby.vim
```diff
@@ -0,0 +1,25 @@
...
+        \   'type': '',
+        \   'text': l:match[2] . ':' . l:match[3]
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+function! ale#handlers#ruby#HandleErbFormationAsError(buffer, lines) abort
+    return s:HandleErbError(a:buffer, a:lines)
```
w0rp: You can just cut and paste your code above into this function.

w0rp:
Also, you might want to enable some more Ruby flags for your linter here, like `-w`, etc. Have a look at the existing Ruby linter.

wikimatze:
@w0rp thanks for the review, everything is working on https://github.com/w0rp/ale/pull/497/commits/b2522184110612869a1234134127a2e2efdee76c.

w0rp:
Yep, looks good to me. Cheers! :beers:

Issue #385, opened 2017-03-06
Brakeman Integration


jtmkrueger:
I run [brakeman](https://github.com/presidentbeef/brakeman) regularly and it would be amazing if it could work with ale. I'm not sure if static analysis is within the scope of ale though, so this may be a non-starter.

w0rp:
If someone wants to add this, go for it. This might require me implementing #333 first.

jtmkrueger:
I'll read through a few of the other linters and see if I can put something together. Good to know that ale is going to be allowed to do static analysis! Would the best idea be to work off that `only-run-on-save`  branch?

w0rp:
@jtmkrueger I'd ignore that branch. I might start again on it.

jtmkrueger:
Cool @w0rp I'll just work off master then.

elebow:
The tricky part about brakeman is that not only does the file need to be stored on disk, but the checker needs to be passed the path of (or invoked from) the root of a Rails application.

@w0rp how do you feel about an approach like the following for detecting the root of the Rails app?

```vim
function! ale_linters#ruby#brakeman#GetCommand(buffer) abort
  " Find the nearest dir contining "app", "db", and "config", and assume it is
  " the root of a Rails app.
  let l:app_parent    = fnamemodify(ale#path#FindNearestDirectory(a:buffer, "app"), ":p")
  let l:db_parent     = fnamemodify(ale#path#FindNearestDirectory(a:buffer, "db"), ":p")
  let l:config_parent = fnamemodify(ale#path#FindNearestDirectory(a:buffer, "config"), ":p")

  if (l:app_parent != l:db_parent) || (l:db_parent != l:config_parent)
    return ''
  end

  let l:rails_root = l:app_parent

  " This is incorrect, since a valid Rails app could exist at the filesystem root.
  " But I think it's a safe assumption for the vast majority.
  if l:rails_root == '/'
    return ''
  end

  return 'brakeman -f json -p ' . l:rails_root
endfunction
```

w0rp:
This general kind of approach can be okay, but you will need to replace the `systemlist` call with with the `command_chain` feature. See `:help ale-command-chain`. (I just added the tag for that right _now_).

w0rp:
Actually, you shouldn't make any calls to `find`. It won't work on Windows, and it will probably mess up on Mac OSX, because the BSD tools are crazy. Have a look at functions like `ale#path#FindNearestFile` and friends. They might help you.

[...]

PR #509, opened 2017-04-27
[RFC] Add Brakeman for Ruby on Rails (references #385)

elebow:
This is a first attempt at adding support for Brakeman, a static analyzer for Ruby on Rails apps.

Supporting Brakeman raises two main challenges that I ask for some input on:

1. The tool must be run against a directory containing a Rails application—it cannot check a single file. I believe this has been addressed satisfactorily by traversing directories upward until we find the presense of `app`, `db`, and `config`, which indicates the root of a [conventional] Rails application. This is testable in a straightforward manner by including some fixtures in the test directory.

2. The output may refer to multiple files in the application. This means we need to ignore any items that do not refer to the file currently in the buffer. Easy enough to write, but so far very difficult to test. I'm not sure how to account for the buffer name in the test runner, and it doesn't appear that Vader has any provision for stubbing functions. Any advice on how to implement this test?

The linter support itself appears to be working correctly in this branch.

### DIFF
--- a/ale_linters/ruby/brakeman.vim
+++ b/ale_linters/ruby/brakeman.vim
@@ -46,29 +46,21 @@ endfunction
 function! s:FindRailsRoot(buffer) abort
     " Find the nearest dir contining "app", "db", and "config", and assume it is
     " the root of a Rails app.
+    for l:name in ['app', 'config', 'db']
+        let l:dir = fnamemodify(
+        \   ale#path#FindNearestDirectory(a:buffer, l:name),
+        \   ':h:h'
+        \)
 
-    let l:path = fnamemodify(bufname(a:buffer), ':p:h')
-
-    while l:path !=? '/'
-        let l:absolute_directories = glob(l:path . '/*/', 0, 1)
-        let l:directories = map(l:absolute_directories,
-        \   {key, value -> substitute(value, '.*\/\(.\{-}/\)', '\1', '') })
-
-        if index(l:directories, 'app/') != -1 &&
-        \  index(l:directories, 'config/') != -1 &&
-        \  index(l:directories, 'db/') != -1
-            break
+        if l:dir !=# '.'
+        \&& isdirectory(l:dir . '/app')
+        \&& isdirectory(l:dir . '/config')
+        \&& isdirectory(l:dir . '/db')
+            return l:dir
         endif
+    endfor
 
-        let l:path = fnamemodify(l:path, ':h')
-    endwhile
-
-    " This is technically incorrect, since a valid Rails app could exist at the filesystem root.
-    if l:path ==? '/'
-        return ''
-    else
-        return l:path
-    endif
+    return ''
 endfunction
 
 call ale#linter#Define('ruby', {
--- a/ale_linters/ruby/brakeman.vim
+++ b/ale_linters/ruby/brakeman.vim
@@ -42,12 +42,16 @@ function! s:FindRailsRoot(buffer) abort
     " Find the nearest dir contining "app", "db", and "config", and assume it is
     " the root of a Rails app.
 
-    let l:path = fnamemodify(bufname(a:buffer), ':p')
+    let l:path = fnamemodify(bufname(a:buffer), ':p:h')
 
     while l:path !=? '/'
-        if strlen(finddir('app', l:path)) > 0 &&
-        \  strlen(finddir('db', l:path)) > 0 &&
-        \  strlen(finddir('config', l:path)) > 0
+        let l:absolute_directories = glob(l:path . '/*/', 0, 1)
+        let l:directories = map(l:absolute_directories,
+        \   {key, value -> substitute(value, '.*\/\(.\{-}/\)', '\1', '') })
+
+        if index(l:directories, 'app/') != -1 &&
+        \  index(l:directories, 'config/') != -1 &&
+        \  index(l:directories, 'db/') != -1
             break
         endif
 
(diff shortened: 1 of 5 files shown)


jtmkrueger:
@elebow While I'm not sure how you'll overcome some of those testing issues, I think it's easy to get your new linter added to the [readme](https://github.com/w0rp/ale/blob/master/README.md) in this pull request so when it's accepted it'll be immediately discoverable. Should be a simple addition.

Great work on this!

ale_linters/ruby/brakeman.vim
```diff
@@ -0,0 +1,83 @@
...
+
+    return 'brakeman -f json -q -p ' . l:rails_root
+endfunction
+
+function! s:FindRailsRoot(buffer) abort
+    " Find the nearest dir contining "app", "db", and "config", and assume it is
+    " the root of a Rails app.
+
+    " Return the cached location, if any.
+    let l:path_cached = getbufvar(a:buffer, 'ruby_brakeman_rails_root_cached', '')
```
w0rp: Don't bother caching it. Finding the directory won't be that expensive, and we'll probably run into caching issues.

ale_linters/ruby/brakeman.vim
```diff
@@ -0,0 +1,83 @@
...
+    if l:path_cached !=? '' && l:path_cached !=? '/'
+      return l:path_cached
+    endif
+
+    let l:path = fnamemodify(bufname(a:buffer), ':p')
+
+    while l:path !=? '/'
+        if strlen(finddir('app', l:path)) > 0 &&
+        \  strlen(finddir('db', l:path)) > 0 &&
+        \  strlen(finddir('config', l:path)) > 0
```
w0rp: We should find a way to search upwards in paths for multiple matches in one go.

ale_linters/ruby/brakeman.vim
```diff
@@ -0,0 +1,83 @@
+" Author: Eddie Lebow https://github.com/elebow
+" Description: Brakeman, a static analyzer for Rails security
+
+function! ale_linters#ruby#brakeman#Handle(buffer, lines) abort
+    let l:result = json_decode(join(a:lines, ''))
+
+    let l:output = []
+
+    for l:warning in l:result.warnings
+        if !s:MatchFileWithoutRailsRoot(l:warning.file, a:buffer)
```
w0rp: I need to write that function for matching filenames against a buffer. This is a common pattern.

ale_linters/ruby/brakeman.vim
adriaanzon: I don't know about multiple matches, but you can search upwards by appending `;` to the path in the second argument of `finddir`. But that doesn't ensure the matches are in the same directory when calling it multiple times.

ale_linters/ruby/brakeman.vim
w0rp: This function exists now. Try using `ale#path#IsBufferPath(a:buffer, l:warning.file)`. You can pass a relative path or an absolute path, and `1` will be returned if the filenames represent the same file. (It should work like that anyway.)

elebow:
I've addressed the first two feedback items.

As for searching upwards for the three key directories when finding the rails root, I've reduced the filesystem accesses to once per directory level. But this comes at the cost of some significant complexity, and personally I would prefer to revert 5e62403.

ale_linters/ruby/brakeman.vim
```diff
@@ -0,0 +1,75 @@
...
+        let l:path = fnamemodify(l:path, ':h')
+    endwhile
+
+    " This is technically incorrect, since a valid Rails app could exist at the filesystem root.
+    if l:path ==? '/'
+        return ''
+    else
+        return l:path
+    endif
+endfunction
```
w0rp: This works. I might try this in future:

```vim
function! s:FindRailsRoot(buffer) abort
    for l:name in ['app', 'config', 'db']
        let l:dir = fnamemodify(
        \   ale#path#FindNearestDirectory(a:buffer, l:name),
        \   ':h:h'
        \)

        if l:dir !=# '.'
        \&& isdirectory(l:dir . '/app')
        \&& isdirectory(l:dir . '/config')
        \&& isdirectory(l:dir . '/db')
            return l:dir
        endif
    endfor

    return ''
endfunction
```

w0rp:
@elebow Check out the test result. There's something wrong with the JSON parsing.

elebow:
Yes, I wasn't able to finish writing the handler test.

Since the handler discards any lines that do not match the buffer filename, and also assumes that all the filenames in the linter output are relative to the root of the Rails application, I'm not sure how to implement the test.

w0rp:
Use `silent file! filename` to set the filename for the buffer. Use a `cd!` command with the `/testplugin/test` directory or whatever the subdirectory is. Look for similar lines in other tests.

elebow:
That should do it for the tests. Thanks for the help, everyone!

But now vint is complaining about the lambda in FindRailsRoot, even though it is valid. Apparently, the parser that vint uses does not yet support this new syntax: https://github.com/Kuniwak/vint/issues/209.

I'll try to replace the function with the alternative suggested in https://github.com/w0rp/ale/pull/509#discussion_r114495476.

test/handler/test_brakeman_handler.vader
```diff
@@ -7,20 +7,23 @@ After:
     call ale#linter#Reset()
 
 Execute(The brakeman handler should parse JSON correctly):
+  cd! /testplugin/test/ruby_fixtures/valid_rails_app/app/models
```
w0rp: I might add a second `cd` line so the test runs when you run the file directly with `:Vader`. This is good. :+1:

w0rp:
Cheers! :beers: 

Thanks for all of your work on this, @elebow.

PR #512, opened 2017-04-27
add xo reference for javascript linters on readme

dansdantas:


w0rp:
```  
Starting Vader: /testplugin/test/test_ale_toggle.vader
    (1/1) [EXECUTE] ALEToggle should reset everything and then run again
    (1/1) [EXECUTE] (X) [] should be equal to [{'lnum': 2, 'bufnr': 3, 'vcol': 0, 'pattern': '', 'valid': 1, 'nr': -1, 'type': 'E', 'col': 3, 'text': 'foo bar'}]
  Success/Total: 0/1
```

That was the build error, for future reference. Seems like that might periodically fail. I'll keep an eye on that.

w0rp:
Cheers! :beers:

dansdantas:
I was kind worried but thanks for accepting. =D

PR #514, opened 2017-04-27
Add tags for buffer-local variants of each linter option

adriaanzon:
I blindly went ahead and added buffer-local variables for each linter option to the documentation. I also added an example to use the buffer-local variables with shellcheck, but I noticed that the shellcheck linter doesn't use the `ale#Var` function. This may also be the case in other linters, so consider this pull request a WIP while I'm going to check that.

I also made some stylistic changes.

### DIFF
--- a/doc/ale-handlebars.txt
+++ b/doc/ale-handlebars.txt
@@ -6,9 +6,8 @@ ALE Handlebars Integration                             *ale-handlebars-options*
 ember-template-lint                          *ale-handlebars-embertemplatelint*
 
 g:ale_handlebars_embertemplatelint_executable
-                              \ *g:ale_handlebars_embertemplatelint_executable*
-
-  Type: |String|
+                                *g:ale_handlebars_embertemplatelint_executable*
+  Type: |String|                  *b:ale_handlebars_embertemplatelint_executable*
   Default: `'ember-template-lint'`
 
   ALE will look for ember-template-lint executable in ancestor node_modules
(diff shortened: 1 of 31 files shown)


adriaanzon:
Shouldn't `ale_linters_sh_shellcheck_exclusions` be called `ale_sh_shellcheck_exclusions`?

doc/ale-handlebars.txt
```diff
@@ -6,9 +6,8 @@ ALE Handlebars Integration                             *ale-handlebars-options*
 ember-template-lint                          *ale-handlebars-embertemplatelint*
 
 g:ale_handlebars_embertemplatelint_executable
-                              \ *g:ale_handlebars_embertemplatelint_executable*
-
-  Type: |String|
+                                *g:ale_handlebars_embertemplatelint_executable*
+  Type: |String|                  *b:ale_handlebars_embertemplatelint_executable*
```
adriaanzon: This one was a bit weird to format but I've done it like in other places where there is a tag for `g:` and `b:` variables: One line between the variable name and the beginning of Type / Default.

Like this one: https://github.com/adriaanzon/ale/blob/7230cbe9e43d64a02212ae4803abff874e39b8d6/doc/ale.txt#L366

w0rp:
Yes, `ale_linters_sh_shellcheck_exclusions` should be called `ale_sh_shellcheck_exclusions`. We'll have to add some tests for that and set up a fallback. Then we can remove the code for the old option completely in 2.0.

w0rp:
Cheers! :beers:

PR #522, opened 2017-04-30
Glue for Vale

chew-z:
Following [discussion here](https://github.com/w0rp/ale/issues/518#issuecomment-298181571) I have added simple glue for [vale linter](https://github.com/ValeLint/vale) for markdown and text.

Vale is cool linter for natural language and it works well with ale. The only trick is to use option `--output=line` to adjust to `ale#handlers#unix#HandleAsWarning` function.


w0rp:
Have a look at the Travis output. It's complaining about your double-quoted strings. I recommend installing `vint`, which ALE integrates with.

chew-z:
I have `vint` installed but mine wasn't complaining about double quotes. Anyway I have changed to single quotes and it works.

PR #524, opened 2017-05-01
shell: Don't default to fish

adriaanzon:
When `$SHELL` was set to /usr/bin/fish, the shell linter defaulted to that. Since fish is not a POSIX compatible shell, `sh` scripts and the shell linter won't work with fish.

### DIFF
--- a/ale_linters/sh/shell.vim
+++ b/ale_linters/sh/shell.vim
@@ -11,7 +11,7 @@ endif
 if !exists('g:ale_sh_shell_default_shell')
     let g:ale_sh_shell_default_shell = fnamemodify($SHELL, ':t')
 
-    if g:ale_sh_shell_default_shell ==# ''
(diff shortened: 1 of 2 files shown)


ale_linters/sh/shell.vim
```diff
@@ -11,7 +11,7 @@ endif
 if !exists('g:ale_sh_shell_default_shell')
     let g:ale_sh_shell_default_shell = fnamemodify($SHELL, ':t')
 
-    if g:ale_sh_shell_default_shell ==# ''
+    if g:ale_sh_shell_default_shell ==# '' || g:ale_sh_shell_default_shell ==# 'fish'
```
w0rp: Cool. Change the documentation for the setting to explain this.

w0rp:
Cheers! :beers:

Issue #37, opened 2016-10-03
Attempting to add support for PHP


scwood:
Hello, I am trying to add support for php with `php -l`. I've added a directory called `php` to `ale_linters` and added a file called `php.vim`.

Here's my code so far:

```viml
if exists('g:loaded_ale_linters_php_php')
    finish
endif

let g:loaded_ale_linters_php_php = 1

function! ale_linters#php#php#Handle(buffer, lines)
    " Matches patterns like the following:
    "
    " Parse error: parse error in - on line 7
    let pattern = 'Parse error: \(.\+\) on line \(\d\+\)'
    let output = []

    for line in a:lines
        let l:match = matchlist(line, pattern)

        if len(l:match) == 0
            continue
        endif

        " vcol is Needed to indicate that the column is a character.
        call add(output, {
        \   'bufnr': a:buffer,
        \   'lnum': l:match[1] + 0,
        \   'vcol': 0,
        \   'col': 0,
        \   'text': l:match[0],
        \   'type': 'E',
        \   'nr': -1,
        \})
    endfor

    return output
endfunction

call ALEAddLinter('php', {
\   'name': 'php',
\   'executable': 'php',
\   'output_stream': 'stderr',
\   'command': 'php -l -',
\   'callback': 'ale_linters#php#php#Handle',
\})
```

When I run `cat myfile.php | php -l` I get sytnax errors. But in the buffer nothing shows up.

w0rp:
Hmm, the problem here isn't obvious to me yet. I might have a go at adding this to the repo myself and credit you as the author. It looks okay, so it's probably some minor bug.

scwood:
Okay, let me know if there's something that needs to be changed. I would like to keep contributing. Great plugin.

w0rp:
I have pushed your code now with some minor changes. `php` needed `--` instead of `-` to read from stdin, which is a bit odd, but it works. The second error was that the regex matches start at index `1`, and not `0`. That's a common mistake, which I didn't notice until I started playing around with it.

scwood:
Thanks for explaining those changes. I have the latest code, but I'm still
not seeing any linting feedback in my buffer.

Here's the code I'm using:

```php
<?php
echo "Hello!" // this needs a semi-colon
```

w0rp:
[...]

PR #533, opened 2017-05-06
Add PHP 7 compatibility

pbogut:
PHP 7+ displays error slightly different. In PHP5 errors where displayed both on stderr and stdout (which also depends on php.ini, I think).

For example one could get output like this:

```
PHP Parse error:  syntax error, unexpected ';', expecting ']' in - on line 15
Parse error:  syntax error, unexpected ';', expecting ']' in - on line 15
```

Because of that PHP ale linter was ignoring second version. However, since PHP7
errors are always displayed as `Parse error: ...` without the `PHP ` prefix.

This pull request is changing pattern to match both formats and additionaly it's deduping list, so older PHP versions won't report same error twice.

### DIFF
--- a/ale_linters/php/php.vim
+++ b/ale_linters/php/php.vim
@@ -5,7 +5,7 @@ function! ale_linters#php#php#Handle(buffer, lines) abort
     " Matches patterns like the following:
     "
     " PHP Parse error:  syntax error, unexpected ';', expecting ']' in - on line 15
-    let l:pattern = '\vPHP %(Fatal|Parse) error:\s+(.+unexpected ''(.+)%(expecting.+)@<!''.*|.+) in - on line (\d+)'
+    let l:pattern = '\v^%(Fatal|Parse) error:\s+(.+unexpected ''(.+)%(expecting.+)@<!''.*|.+) in - on line (\d+)'
 
     let l:output = []
 
@@ -23,7 +23,7 @@ endfunction
 call ale#linter#Define('php', {
 \   'name': 'php',
 \   'executable': 'php',
-\   'output_stream': 'both',
-\   'command': 'php -l -d display_errors=1 --',
+\   'output_stream': 'stdout',
+\   'command': 'php -l -d error_reporting=E_ALL -d display_errors=1 --',
 \   'callback': 'ale_linters#php#php#Handle',
 \})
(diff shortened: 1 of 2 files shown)


adriaanzon:
Hmm, this was working fine for me on PHP 7.

Maybe your php.ini settings are different. What is your value of `error_reporting`? We may just need to alter that setting when the linter gets run, like we do with `display_errors` [here](https://github.com/w0rp/ale/blob/master/ale_linters/php/php.vim#L27).

w0rp:
I'm okay with this, but I'll let @adriaanzon check it.

pbogut:
I had this problem at work, but now when I'm checking it on my local machine it is displaying both lines with and without `PHP` prefix.
Today I was testing it with docker image `php:7-cli`, so I didn't even notice this it is working on my PC. Still, there is the same problem with docker version as I had at work.

Docker settings:
`error_reporting = E_ALL & ~E_DEPRECATED & ~E_STRICT & ~E_NOTICE`
If I run it with `-d error_reporting=-1` I can only see `Parse error:`. If I try `-d error_reporting=0` then no error is displayed.
I guess its not `error_reporting` but maybe it is coused by different php.ini option...

adriaanzon:
I've tinkered around and it looks like `PHP Parse error: ...` is printed to stderr, and `Parse error: ...` to stdout.

I also looked at linters for other editors, and they all seem to just use stdout, and look for `^(Parse|Fatal)...`.

I also had a look at the history of the file, especially #37 and #235. With `--display_errors=1` it will output to stdout, so I think it's enough to remove the line that changes `output_stream` to 'both', and change beginning of the pattern to `\v^%(Fatal|Parse) `

adriaanzon:
Just throwing this in here. This shows where the output goes with various combinations of options. Red is stderr, green is stdout. 

image

By the looks of it, we may also need to add `-d error_reporting=E_ALL`, in case people have a falsy `error_reporting`.

pbogut:
@adriaanzon should we close this PR then and create the new one?

adriaanzon:
You could also revert the commit here and then we'll squash it when merging.

pbogut:
I will do that later today.

pbogut:
@adriaanzon I've updated it as you suggested. I guess it needs to be reviewed once again.

adriaanzon:
Thanks!

PR #546, opened 2017-05-10
Add cpplint linter

dawikur:
Added note to `README.md`, `doc/ale.txt` and `doc/ale-cpp.txt`.

This will handle the simplest case running `cpplint` with default options.
Marking everything as warning; this gives good starting point :)


w0rp:
I've been away for a few days.

Would you like to rebase this one and add a test case for the new error handler?

dawikur:
Done! :smile:

PR #551, opened 2017-05-12
Add checkstyle linter

### LINKED ISSUE #550: [FEATURE] Support for Java checkstyle
http://checkstyle.sourceforge.net

A tool for enforcing a configurable list of styles on your Java project.

It comes with a default preset that matches the Google style guide.

The stdout looks easily parseable, I'd be happy to take a crack at implementing this if someone could point me in the right direction.

meunierd:
I put this together based off of the rubocop linter. Checkstyle doesn't actually accept stdin, I'm not sure if that's a problem?

Closes #550


meunierd:
Whoops, I just noticed some errors provide column numbers, I'm going to update this.

meunierd:
@w0rp I see that the stdin actually comes from a tmp file, is it possible to actually just pass that path instead?

w0rp:
If you put `%t` in your command string, the temporary file will be passed an argument there automatically, instead of using stdin. You can update your code to do that.

meunierd:
Updated, works perfectly. Thanks!

PR #552, opened 2017-05-12
Support using relative paths in commands, and add an option for flow to use them

dcosson:
I wanted to open this to get feedback on if this approach is something that could be accepted, and if so I'll write tests for it to get this merged.

This exposes a `%r` interpolation variable for commands, which will fill in the relative path instead of absolute.

My use case is that I use docker in development, and for more intensive linting tools like flow (which needs to have a large part of my js build environment, have all node_modules installed, etc.) it's nice to install all of that once in my docker environment, rather than setting it up both on my host machine and in docker.

The paths in my docker env are different i.e. not nested in `/Users/username`, but with relative paths it works fine (I have a tiny wrapper script that I set as `g:ale_javascript_flow_executable` which proxies the command from Ale into my running dev docker container, though I think there's also a `flow ide`  command that I haven't looked much into that could potentially be used in the future and would likely still need relative paths coming from the editor).

### DIFF
--- a/ale_linters/javascript/flow.vim
+++ b/ale_linters/javascript/flow.vim
@@ -7,6 +7,9 @@ let g:ale_javascript_flow_executable =
 let g:ale_javascript_flow_use_global =
 \   get(g:, 'ale_javascript_flow_use_global', 0)
 
+let g:ale_javascript_flow_use_relative_paths =
+\   get(g:, 'ale_javascript_flow_use_relative_paths', 0)
+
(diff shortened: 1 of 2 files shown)


w0rp:
It depends on what the paths are relative to. Some linters change directory to a project directory before running the command. In those cases, a path which is relative to the buffer won't work.

I'd recommend writing a function which accepts a base  directory and another path, and returning a relative path computed from that, like the existing functions which return absolute paths. The `%r` formatting won't work everywhere, so I'd take that away.

PR #559, opened 2017-05-16
Add xmllint linter

q12321q:
Add xmlllint as a new linter for xml.
I'll try to add some vader tests but I need to check how.

### DIFF
--- a/ale_linters/xml/xmllint.vim
+++ b/ale_linters/xml/xmllint.vim
@@ -3,7 +3,7 @@
 
 " CLI options
 let g:ale_xml_xmllint_executable = get(g:, 'ale_xml_xmllint_executable', 'xmllint')
-let g:ale_xml_xmllint_options = get(g:, 'ale_xml_xmllint_options', '--noout')
+let g:ale_xml_xmllint_options = get(g:, 'ale_xml_xmllint_options', '')
 
 function! ale_linters#xml#xmllint#GetExecutable(buffer) abort
     return ale#Var(a:buffer, 'xml_xmllint_executable')
@@ -12,7 +12,7 @@ endfunction
 function! ale_linters#xml#xmllint#GetCommand(buffer) abort
     return ale#Escape(ale_linters#xml#xmllint#GetExecutable(a:buffer))
     \   . ' ' . ale#Var(a:buffer, 'xml_xmllint_options')
-    \   . ' -'
+    \   . ' --noout -'
 endfunction
 
 function! ale_linters#xml#xmllint#Handle(buffer, lines) abort
--- a/ale_linters/xml/xmllint.vim
+++ b/ale_linters/xml/xmllint.vim
@@ -5,11 +5,14 @@
 let g:ale_xml_xmllint_executable = get(g:, 'ale_xml_xmllint_executable', 'xmllint')
 let g:ale_xml_xmllint_options = get(g:, 'ale_xml_xmllint_options', '--noout')
 
+function! ale_linters#xml#xmllint#GetExecutable(buffer) abort
+    return ale#Var(a:buffer, 'xml_xmllint_executable')
+endfunction
+
 function! ale_linters#xml#xmllint#GetCommand(buffer) abort
-    return printf('%s %s -',
-    \   g:ale_xml_xmllint_executable,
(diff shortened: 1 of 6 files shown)


ale_linters/xml/xmllint.vim
```diff
@@ -0,0 +1,70 @@
...
+    "       ^
+    let l:pattern_column_token = '\v^\s+\^$'
+
+    let l:output = []
+
+    for l:line in a:lines
+
+        " Parse error/warning lines
+        let l:match_message = matchlist(l:line, l:pattern_message)
+        if len(l:match_message)
```
w0rp: `if !empty(l:match_message)` is preferred here.

ale_linters/xml/xmllint.vim
```diff
@@ -0,0 +1,70 @@
...
+          let l:line = l:match_message[2] + 0
+          let l:type = 'E'
+          if match(l:match_message[3], 'warning') != -1
```
w0rp: You could use `=~#` instead.

w0rp: A ternary might be better. 

```vim
let l:type = l:match_message[3] =~# 'warning' ? 'W' : 'E'
```

ale_linters/xml/xmllint.vim
```diff
@@ -0,0 +1,70 @@
...
+            let l:type = 'W'
+          endif
+          let l:text = l:type . ' ' . l:match_message[3] . ' : ' . l:match_message[4]
+
+          call add(l:output, {
+          \   'bufnr': a:buffer,
```
w0rp: `bufnr` is set automatically, so you can remove this line.

ale_linters/xml/xmllint.vim
```diff
@@ -0,0 +1,70 @@
...
+          \   'text': l:text,
+          \   'type': l:type,
+          \})
+
+          continue
+        endif
+
+        " Parse column position
+        let l:match_column_token = matchlist(l:line, l:pattern_column_token)
+        if len(l:output) && len(l:match_column_token)
```
w0rp: `!empty` again.

ale_linters/xml/xmllint.vim
```diff
@@ -0,0 +1,70 @@
...
+          let l:previous = l:output[len(l:output)-1]
```
w0rp: Add spaces around binary operators like `-`.

ale_linters/xml/xmllint.vim
```diff
@@ -0,0 +1,70 @@
...
+        endif
+
+    endfor
+
+    return l:output
+endfunction
+
+call ale#linter#Define('xml', {
+\   'name': 'xmllint',
+\   'executable': g:ale_xml_xmllint_executable,
```
w0rp: Use a callback function here, and use `ale#Var` for the variable. Then it will be possible to modify the executable variable dynamically, and with different values for each buffer.

ale_linters/xml/xmllint.vim
```diff
@@ -0,0 +1,70 @@
+" Author: q12321q <q12321q@gmail.com>
+" Description: This file adds support for checking XML code with xmllint.
+
+" CLI options
+let g:ale_xml_xmllint_executable = get(g:, 'ale_xml_xmllint_executable', 'xmllint')
+let g:ale_xml_xmllint_options = get(g:, 'ale_xml_xmllint_options', '--noout')
+
+function! ale_linters#xml#xmllint#GetCommand(buffer) abort
+    return printf('%s %s -',
+    \   g:ale_xml_xmllint_executable,
```
w0rp: Call your executable callback here, and apply `ale#Escape` to your executable, in case your executable path contains spaces or other special characters.

q12321q:
The '--noout' option display all the xml in the stdout so it's almost mandatory to have this flag.
does the option 'ale_xml_xmllint_options' exhaustively list all the options or can I hard code it in the command?

w0rp:
@q12321q Most other linters put mandatory options into the command string, and default the options variables to an empty string. You can take that out of there. Make sure you do test setting other options.

q12321q:
@w0rp I did all the changes and added two vader tests. Do you see anything else to add?

PR #561, opened 2017-05-16
Allow overriding rubocop executable.

meunierd:
Rubocop configuration can get pretty hairy, to the point that you actually want one per-project rather than a global rubocop. The two standard ways for the ruby ecosystem to handle this is either through binstubs (eg. `bin/rubocop`) or bundler `bundle exec rubocop`.

### DIFF
--- a/ale_linters/ruby/rubocop.vim
+++ b/ale_linters/ruby/rubocop.vim
@@ -25,14 +25,19 @@ function! ale_linters#ruby#rubocop#Handle(buffer, lines) abort
 endfunction
 
 function! ale_linters#ruby#rubocop#GetCommand(buffer) abort
-    return ale#Var(a:buffer, 'ruby_rubocop_executable')
+    let l:unescaped = ale#Var(a:buffer, 'ruby_rubocop_executable')
+    let l:executable = ale#Escape(l:unescaped)
+    if l:unescaped =~? 'bundle$'
+        let l:executable = l:executable . ' exec rubocop'
+    endif
+    return l:executable
     \   . ' --format emacs --force-exclusion '
     \   . ale#Var(a:buffer, 'ruby_rubocop_options')
     \   . ' --stdin ' . bufname(a:buffer)
 endfunction
 
 function! ale_linters#ruby#rubocop#GetExecutable(buffer) abort
-    let l:executable = split(ale#Var(a:buffer, 'ruby_rubocop_executable'))[0]
+    let l:executable = ale#Var(a:buffer, 'ruby_rubocop_executable')
     if executable(l:executable)
         return l:executable
     endif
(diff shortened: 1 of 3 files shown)


ale_linters/ruby/rubocop.vim
```diff
@@ -36,6 +37,10 @@ if !exists('g:ale_ruby_rubocop_options')
     let g:ale_ruby_rubocop_options = ''
 endif
 
+if !exists('g:ale_ruby_rubocop_executable')
+    let g:ale_ruby_rubocop_executable = 'rubocop'
+endif
+
 call ale#linter#Define('ruby', {
 \   'name': 'rubocop',
 \   'executable': 'rubocop',
```
w0rp: You should change this to `executable_callback` and add a function, so ALE which check if your override is executable.

ale_linters/ruby/rubocop.vim
```diff
@@ -25,7 +25,8 @@ function! ale_linters#ruby#rubocop#Handle(buffer, lines) abort
 endfunction
 
 function! ale_linters#ruby#rubocop#GetCommand(buffer) abort
-    return 'rubocop --format emacs --force-exclusion '
+    return ale#Var(a:buffer, 'ale_ruby_rubocop_executable')
```
w0rp: You should apply `ale#Escape` to this value here, in case the path to the executable contains spaces, or whatever.

meunierd:
Added tests and `executable_callback` 👍

meunierd:
Updated.

ale_linters/ruby/rubocop.vim
```diff
@@ -25,20 +25,37 @@ function! ale_linters#ruby#rubocop#Handle(buffer, lines) abort
...
+        let l:executable = l:executable . ' exec rubocop'
+    endif
+    return l:executable
+    \   . ' --format emacs --force-exclusion '
     \   . ale#Var(a:buffer, 'ruby_rubocop_options')
     \   . ' --stdin ' . bufname(a:buffer)
 endfunction
 
+function! ale_linters#ruby#rubocop#GetExecutable(buffer) abort
+    let l:executable = ale#Var(a:buffer, 'ruby_rubocop_executable')
```
w0rp: You just need to `return` here, as ALE uses `if executable(...)` internally. I'll fix this later.

w0rp:
Cheers! :beers:

meunierd:
🥂 It looks like flake8 does the same thing: https://github.com/w0rp/ale/blob/master/ale_linters/python/flake8.vim#L28

w0rp:
Ah, that's slightly different. It's looking for an executable in virtualenv, and if that fails it fails back on a possible global executable.

PR #566, opened 2017-05-19
Improve performance when using gometalinter

svanharmelen:
Before this change when I opened a big project that had 6000+ warnings/errors it took ages to get the actual warnings/errors and it caused my CPU to be busy for quite some time. The call to gometalinter alone took about 24 seconds, but after that vim was struggling as well.

After this change the gometalinter call just takes 2 seconds and nothing noticeable happens with the CPU and/or vim.

### DIFF
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -6,7 +6,7 @@ if !exists('g:ale_go_gometalinter_options')
 endif
 
 function! ale_linters#go#gometalinter#GetCommand(buffer) abort
-    return 'gometalinter '
+    return 'gometalinter --include=''^' . expand('%:p') . '.*$'' '
     \   . ale#Var(a:buffer, 'go_gometalinter_options')
     \   . ' ' . ale#Escape(fnamemodify(bufname(a:buffer), ':p:h'))
 endfunction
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

ale_linters/go/gometalinter.vim
```diff
@@ -6,7 +6,7 @@ if !exists('g:ale_go_gometalinter_options')
 endif
 
 function! ale_linters#go#gometalinter#GetCommand(buffer) abort
-    return 'gometalinter '
+    return 'gometalinter --include=''^' . expand('%:p') . '.*$'' '
```
w0rp: Someone will probably run into shell escaping issues on Windows at some point with some suggestions. Windows is crazy.

svanharmelen: Yeah indeed, but I doubt this will cause issues... I used the exact same code in `vim-go` quite some time ago and until now never got a complaint.

And since your not a Go user... FYI Vim in combination with `vim-go` is probably the most used setup for Go developers:

image
https://blog.golang.org/survey2016-results

So I think we're good, but of course ping me if an related issue does come in. Thanks 👍

PR #570, opened 2017-05-21
feat(options): add ale_fix_* options

tunnckoCore:
Add 3 new options for better control of ALEFix: `ale_fix_on_insert_leave`, `ale_fix_on_save` and `ale_fix_on_enter`. Their defaults are the same as their `ale_lint_*` equivalents.

- ale_fix_on_insert_leave = 0
- ale_fix_on_enter = 1
- ale_fix_on_save = 1

Also added docs.

Sorry, but don't have tests because I'm new to Vim world, can you handle them?

### DIFF
--- a/plugin/ale.vim
+++ b/plugin/ale.vim
@@ -89,6 +89,15 @@ let g:ale_lint_on_save = get(g:, 'ale_lint_on_save', 1)
 " This flag can be set to 1 to enable linting when the filetype is changed.
 let g:ale_lint_on_filetype_changed = get(g:, 'ale_lint_on_filetype_changed', 1)
 
+" This flag can be set to 1 to enable fixing when leaving insert mode.
+let g:ale_fix_on_insert_leave = get(g:, 'ale_fix_on_insert_leave', 0)
+
+" This flag can be set to 0 to disable fixing when the buffer is entered.
+let g:ale_fix_on_enter = get(g:, 'ale_fix_on_enter', 1)
+
+" This flag can be set to 1 to enable fixing when a buffer is written.
+let g:ale_fix_on_save = get(g:, 'ale_fix_on_save', 1)
+
 " This flag may be set to 0 to disable ale. After ale is loaded, :ALEToggle
 " should be used instead.
 let g:ale_enabled = get(g:, 'ale_enabled', 1)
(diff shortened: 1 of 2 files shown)


tunnckoCore:
@w0rp ookey, i'll change them. It would be good to add your thoughts about ALEFix on the docs too :)

plugin/ale.vim
```diff
@@ -89,6 +89,15 @@ let g:ale_lint_on_save = get(g:, 'ale_lint_on_save', 1)
 " This flag can be set to 1 to enable linting when the filetype is changed.
 let g:ale_lint_on_filetype_changed = get(g:, 'ale_lint_on_filetype_changed', 1)
 
+" This flag can be set to 1 to enable fixing when leaving insert mode.
+let g:ale_fix_on_insert_leave = get(g:, 'ale_fix_on_insert_leave', 0)
+
+" This flag can be set to 0 to disable fixing when the buffer is entered.
+let g:ale_fix_on_enter = get(g:, 'ale_fix_on_enter', 0)
```
w0rp: I'd remove "fix on enter" completely. A file shouldn't be modified just by looking at it.

w0rp:
For fixing files when leaving insert mode or saving files, we should probably be a little more clever. The `autocmd` commands for checking and fixing the files will be competing with one another. If both are on at the same time, we should make it so files are checked only _after_ fixing them, which won't be straightforward to handle. It might not be possible at all. Vim doesn't let you modify a buffer which isn't open, but we can read a buffer which isn't open. So, we could fix and then check a file if the buffer is open, but we can only check a file if the buffer is closed, unless we are writing the changes to disk and then reloading them when the buffer opens again.

I suggest passing an even argument to `Fix` like for what is currently done for `Queue`, and then saving that event string in the buffer Dictionary for fixing files. Then we can load that string back again when files are fixed, and we can test that string for how files should be checked. `'lint_on_save'`, etc.

I can work on all of this later. Here are the two workflows.

For fixing files when leaving insert mode:

1. Leave insert mode.
2. Fix a file, and also check it.
3. Apply changes to buffer when buffer is open.
4. Check file again, running only the linters with `lint_file` set to 0.

For fixing files when saving them:

1. Save a file.
2. Fix a file, and also check it.
3. Save the result of fixing the file to disk.
4. If the buffer ever opens again, reload the buffer and run all linters for the file.

plugin/ale.vim
tunnckoCore: okey i'll remove that line and the docs part for it

tunnckoCore:
Mostly, don't know. As I said I'm new to Vim, so don't understand the whole things too deeply.

w0rp:
I'll give the handling of saving files and so on a go when I get some time.

w0rp:
I'll close this as I'm working through these options now. I'm not sure about fixing files on enter yet. I have added fixing files on save for now, which is the hardest to implement. I'll work on storing the buffer variables for `ale#Var` in case the variables are read after the buffer is deleted.

tunnckoCore:
Cool! No problem. It's great that progress is happening.

PR #571, opened 2017-05-21
feat(fixer): add Prettier fixer (using Prettier-ESLint CLI) + docs

tunnckoCore:
I did a PR to Prettier-ESLint ([merged as v4](https://github.com/prettier/prettier-eslint-cli/releases/tag/v4.0.0)) that made it exact mirror of the original Prettier CLI, so here i just use it for better integration.

If someone don't want it he can just use `g:ale_javascript_prettier_executable = 'prettier'` and it will use the original one.

Docs added too, adapted from ESLint fixer.

### DIFF
--- a/autoload/ale/handlers/prettier.vim
+++ b/autoload/ale/handlers/prettier.vim
@@ -1,0 +1,49 @@
+
+" Author: tunnckoCore (Charlike Mike Reagent) <mameto2011@gmail.com>
+" Description: Integration between Prettier and ESLint.
+
+" Here we use `prettier-eslint` intetionally,
+" because from v4 it is direct mirror of `prettier` - mimics
+" it's flags and etc.
+
+let g:ale_javascript_prettier_executable =
+\   get(g:, 'ale_javascript_prettier_executable', 'prettier-eslint')
+
+let g:ale_javascript_prettier_options =
+\   get(g:, 'ale_javascript_prettier_options', '')
+
+function! ale#handlers#prettier#GetExecutable(buffer) abort
+    if ale#Var(a:buffer, 'javascript_prettier_use_global')
+        return ale#Var(a:buffer, 'javascript_prettier_executable')
+    endif
+
+    " Look for the kinds of paths that create-react-app generates first.
+    let l:executable = ale#path#ResolveLocalPath(
+    \   a:buffer,
+    \   'node_modules/prettier-eslint-cli/index.js',
+    \   ''
+    \)
+
+    if !empty(l:executable)
+        return l:executable
+    endif
+
+    return ale#path#ResolveLocalPath(
+    \   a:buffer,
+    \   'node_modules/.bin/prettier-eslint',
+    \   ale#Var(a:buffer, 'javascript_prettier_executable')
+    \)
+endfunction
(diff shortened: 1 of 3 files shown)


autoload/ale/handlers/prettier.vim
```diff
@@ -0,0 +1,49 @@
...
+" it's flags and etc.
+
+let g:ale_javascript_prettier_executable =
+\   get(g:, 'ale_javascript_prettier_executable', 'prettier-eslint')
+
+let g:ale_javascript_prettier_options =
+\   get(g:, 'ale_javascript_prettier_options', '')
+
+function! ale#handlers#prettier#GetExecutable(buffer) abort
+    if ale#Var(a:buffer, 'javascript_prettier_use_global')
```
tunnckoCore: btw, locally it fails saying that "it is not set" or sort of like that.

@w0rp, it's the same when use the `eslint` fixer, so the problem isn't here.

tunnckoCore: the error

```
Error detected while processing function ale#fix#Fix[21]..<SNR>51_RunFixer[6]..ale#handlers#eslint#Fix[7]..ale#handlers#eslint#GetExecutable[1
]..ale#Var:                                                                                                                                   
line    3:                                                                                                                                    
E716: Key not present in Dictionary: ale_javascript_eslint_use_global                                                                         
E116: Invalid arguments for function getbufvar(str2nr(a:buffer), l:full_name, g:[l:full_name])                                                
E15: Invalid expression: getbufvar(str2nr(a:buffer), l:full_name, g:[l:full_name])
```

w0rp: The problem came from when the handler for `eslint` has been loaded, but not the linter, because the variable was being initialised in the linter file by mistake. You will have copy and pasted the bug into this file. I have fixed that for `eslint` now.

You'll need to initialize `g:ale_javascript_prettier_use_global` to `0`.

w0rp:
I just realised this is using `prettier-eslint`. We should use `prettier` instead, and then you can combine that with `eslnit --fix` yourself, if you want.

mvolkmann:
I like the idea of using `prettier-eslint` instead of just `prettier`.

w0rp:
Add both.

autoload/ale/handlers/prettier.vim
```diff
@@ -0,0 +1,49 @@
+
+" Author: tunnckoCore (Charlike Mike Reagent) <mameto2011@gmail.com>
+" Description: Integration between Prettier and ESLint.
+
+" Here we use `prettier-eslint` intetionally,
```
alexlafroscia: Heads up -- you've got a typo here. `intetionally` -> `intentionally`

w0rp:
I have merged this via the command line now, and I've split it up into fixers for both `prettier` and `prettier` eslint, with separate registry entries and options.

tunnckoCore:
Not make any sense. Because they are basically the same thing. One just have more better behaving which should be recommended to be used.

w0rp:
There are two, so you can use either. They have different paths in `node_modules`.

PR #578, opened 2017-05-25
Translate pylint output column to 1-based index

### LINKED ISSUE #575: Cannot ale_previous_wrap with Pylint
I'm unable to get to a previous pylint error under certain circumstances.

To reproduce this bug on the following code, make pylint your only Python linter, move the cursor to the `pass` error, then try to go to previous error twice:
~~~Python
import random
def get_random():
    pass
    return random.random()
~~~

Current behavior:  I cannot reach the `import random` line via ale_previous_wrap until I move the cursor to the beginning of the `def` line.

Expected behavior:  I should be able to reach the `import random` line via ale_previous_wrap without having to set my cursor to the beginning of the `def` line myself.

More Information:  I believe that the problem lies in the way pylint reports the column where the error is and the way ALE expects the column.  Pylint reports 0-based columns, but ALE expects 1-based columns (at least, that's what I've gathered based on what is written under `:help cursor` and the way ALE is parsing PEP8 linter information ([ale#handlers#python#HandlePEP8Format](https://github.com/w0rp/ale/blob/master/autoload/ale/handlers/python.vim#L4) seems to expect flake8 column information, which is 1-based)).

Suggested fix:  Make a new python handler for pylint that calls ale#handlers#python#HandlePEP8Format and adds 1 to the column before returning.

I wouldn't mind implementing the suggested fix, but I wanted to get some expert opinion before submitting a pull request.

Here's my ALEInfo:
~~~
 Current Filetype: python
Available Linters: ['flake8', 'mypy', 'pylint']
  Enabled Linters: ['pylint']
 Linter Variables:
let g:ale_python_pylint_executable = 'pylint'
let g:ale_python_pylint_options = ''
let g:ale_python_pylint_use_global = 0
 Global Variables:
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = '✗'
let g:ale_echo_msg_format = '[%linter%] %s'
let g:ale_echo_msg_warning_str = '!'
let g:ale_enabled = 1
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_linter_aliases = {}
let g:ale_linters = {'python': ['pylint']}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_error = '✗'
let g:ale_sign_warning = '!'
let g:ale_statusline_format = ['✗ %d', '! %d', '⬥ ok']
  Command History:
(finished - exit code 20) ['/bin/bash', '-c', '''pylint''  --output-format text --msg-template="{path}:{li
ne}:{column}: {msg_id} ({symbol}) {msg}" --reports n ''/home/okuda/test.py''']
~~~

Here's my pylint version:
~~~
No config file found, using default configuration
pylint 1.7.1, 
astroid 1.5.2
Python 3.5.2 (default, Sep 14 2016, 11:28:32) 
[GCC 6.2.1 20160901 (Red Hat 6.2.1-1)]
~~~

nOkuda:
This should fix #575; also added vader tests to ensure that translation
is working properly.

### DIFF
--- a/ale_linters/python/pylint.vim
+++ b/ale_linters/python/pylint.vim
@@ -32,10 +32,43 @@ function! ale_linters#python#pylint#GetCommand(buffer) abort
     \   . ' %s'
 endfunction
 
+function! ale_linters#python#pylint#Handle(buffer, lines) abort
+    " Matches patterns like the following:
+    "
+    " test.py:4:4: W0101 (unreachable) Unreachable code
+    let l:pattern = '\v^[^:]+:(\d+):(\d+): ([[:alnum:]]+) \((.*)\) (.*)$'
+    let l:output = []
+
+    for l:match in ale#util#GetMatches(a:lines, l:pattern)
+        "let l:failed = append(0, l:match)
+        let l:code = l:match[3]
+
+        if (l:code ==# 'C0303')
+        \ && !ale#Var(a:buffer, 'warn_about_trailing_whitespace')
+            " Skip warnings for trailing whitespace if the option is off.
+            continue
+        endif
+
+        if l:code ==# 'I0011'
+            " Skip 'Locally disabling' message
+             continue
+        endif
+
+        call add(l:output, {
+        \   'lnum': l:match[1] + 0,
+        \   'col': l:match[2] + 1,
+        \   'text': l:code . ': ' . l:match[5],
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

PR #581, opened 2017-05-26
Fix file name checking in rust handler

agatan:
In Rust handler, file name matching is not correct; the handler reports errors in `src/haserrors/mod.rs` while editing `src/noerrors/mod.rs`.

In the handler, the name of buffer and the name in the error message are striped using `fnamemodify(..., ':t')`, but they should be compared using suffix match.

### DIFF
--- a/autoload/ale/handlers/rust.vim
+++ b/autoload/ale/handlers/rust.vim
@@ -50,7 +50,7 @@ function! ale#handlers#rust#HandleRustErrorsForFile(buffer, full_filename, lines
 
             if (
             \   l:span.is_primary
-            \   && (l:span_filename ==# l:filename || l:span_filename ==# '<anon>')
+            \   && (a:full_filename =~ (l:span.file_name . '$') || l:span_filename ==# '<anon>')
             \)
                 call add(l:output, {
                 \   'lnum': l:span.line_start,
--- a/autoload/ale/handlers/rust.vim
+++ b/autoload/ale/handlers/rust.vim
@@ -46,11 +46,9 @@ function! ale#handlers#rust#HandleRustErrorsForFile(buffer, full_filename, lines
         endif
 
         for l:span in l:error.spans
-            let l:span_filename = fnamemodify(l:span.file_name, ':t')
-
             if (
             \   l:span.is_primary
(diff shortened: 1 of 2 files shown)


autoload/ale/handlers/rust.vim
```diff
@@ -50,7 +50,7 @@ function! ale#handlers#rust#HandleRustErrorsForFile(buffer, full_filename, lines
 
             if (
             \   l:span.is_primary
-            \   && (l:span_filename ==# l:filename || l:span_filename ==# '<anon>')
+            \   && (a:full_filename =~ (l:span.file_name . '$') || l:span_filename ==# '<anon>')
```
w0rp: Could you try using `ale#path#IsBufferPath(a:buffer, l:span.file_name)` here, and let me know if that works? Matching with regex won't work in every case, as some other files might be similar enough that they match.

w0rp: You may have to add `silent file ...` to the tests after changing this, so the tests run with the buffer set to a filename we should match against.

agatan: Thank you for your fast reply!
`ale#path#IsBufferPath` does not work here. In my experience, `IsBufferPath`'s work is like this:

0. cwd = `/tmp/sample`, error file = `/tmp/sample/src/haserrors/mod.rs`, current file = `/tmp/sample/src/haserrors/mod.rs`
1. Get the file name of current buffer as `l:buffer_filename` ( = `/tmp/sample/src/haserrors/mod.rs`)
2. Apply `fnamemodify` to get enclosing directory of `l:buffer_filename` ( = `/tmp/sample/src/haserrors/` )
3. Concat the directory and error file name using `ale#path#GetAbsPath` ( = `/tmp/sample/src/haserrors/src/haserrors/mod.rs` )

`cargo` reports error with relative file path from the current directory, so the reported file name is `src/haserrors/mod.rs`.
(If you are in `src/haserrors`, the above solution works correctly)

Anyway, I will try to remove unused variables and add a test for the silent case.

w0rp:
This will do for now. The regex matching of the filename isn't perfect, as certain characters in filenames will be treated as pattern characters, but this will be better that what's there at the moment.

I would like to work on writing a function which can solve this problem easily for any linter at some point. The same problem is being solved in different ways in different files.

Cheers! :beers:

PR #585, opened 2017-05-27
kotlin linter support for maven/pom.xml

gavocanov:
Added support for getting the classpath from a pom.xml using maven if present, same as the javac implementation.


w0rp:
Cheers! :beers:

gavocanov:
Emacs uses https://github.com/shyiko/ktlint, seems much faster and nicer overall, would be worth taking in consideration.

w0rp:
I'll create an issue for tracking that.

w0rp:
#593 There's the issue.

PR #588, opened 2017-05-28
Set qflist/loclist window title properly ...

cs86661:


autoload/ale/list.vim
```diff
@@ -14,11 +14,17 @@ endfunction
 function! ale#list#SetLists(buffer, loclist) abort
     if g:ale_set_quickfix
         call setqflist(a:loclist)
+        if has('patch-7.4.2200')
```
w0rp: These checks are redundant. The plugin only works with Vim 8 and up.

autoload/ale/list.vim
```diff
@@ -14,11 +14,17 @@ endfunction
...
+            call setqflist([], 'r', {'title': expand('%')})
```
w0rp: You should include the title option in the call above, instead of making two calls.

`expand('%')` won't work all the time, as the buffer errors are being set for can be different from the current buffer. You should get the buffer's filename with `expand('#' . a:buffer)`. You may want to add either `':p'` or `':t'` to get the absolute path or the basename.

cs86661:
I don't know how to make "Vader tests" ? I am a newbie on GitHub. :(

I have re-commit it. And it must call setqflist()/setloclist() twice for setting it's title ...

autoload/ale/list.vim
```diff
@@ -14,16 +14,19 @@ endfunction
 function! ale#list#SetLists(buffer, loclist) abort
     if g:ale_set_quickfix
         call setqflist(a:loclist)
+        call setqflist([], 'r', {'title': getbufinfo(a:buffer)[0].name})
     elseif g:ale_set_loclist
         " If windows support is off, bufwinid() may not exist.
         if exists('*bufwinid')
             " Set the results on the window for the buffer.
             call setloclist(bufwinid(str2nr(a:buffer)), a:loclist)
+            call setloclist(bufwinid(str2nr(a:buffer)), [], 'r', {'title': getbufinfo(a:buffer)[0].name})
```
w0rp: Won't this clear the list when the title is set? `'r'` replaces the list. I believe you can use this instead to set the list and title in one go.

```vim
call setloclist(
\   bufwinid(str2nr(a:buffer)), 
\   a:loclist, 
\   ' ', 
\   {'title': getbufinfo(a:buffer)[0].name}
\)
```

cs86661:
I've tried, and it really can't set title in the same time ...
```
If the optional {what} dictionary argument is supplied, then
		only the items listed in {what} are set. The first {list}
		argument is ignored.
```

w0rp:
Okay. I'll add some tests for this.

w0rp:
The API for NeoVim is different, so I had to account for that. I wish they could have just made the API _the same_, no matter how good or bad it is. (It's good enough.)

PR #595, opened 2017-05-30
go: Remove `staticcheck` and `go build` defaults

### LINKED ISSUE #594: go: disable staticcheck by default
One of the linters that's enabled by default for Go is `staticcheck`. Though tremendously useful `staticcheck` can use a massive amount of RAM and CPU due to how it works. This will become especially visible as your codebase grows.

For a not entirely trivial `main.go` with 172 lines (including white space and comments) `staticcheck` takes about 20s to run spiking all cores which makes my laptop largely unresponsive and as a consequence the UI starts to stutter since my system is strapped for resources. Since running `staticcheck` happens quite a bit this results on a fairly consistent high load rendering my system useless until I disabled it.

I would recommend to disable `staticcheck` by default. I've switched my config to

```vimL
let g:ale_linters = {'go': ['gometalinter', 'gofmt']}
let g:ale_go_gometalinter_options = '--fast'
```

The `--fast` flag to `gometalinter` ensure it only runs linters that are known to be pretty quick. This gets you the benefit of running quite a few linters and then running something like staticcheck as pre-flight to integration tests instead. The linters that get disabled when running with `--fast` can be found here: https://github.com/alecthomas/gometalinter/blob/9af851db896f6a62ceee990ba8a437ad807d1292/config.go#L103. It might be worth drawing some inspiration from that for the default linters for Go in ALE.

daenney:
Fixes #594.

I've opted to just remove `go build` and `staticcheck` from the default linters right now to ensure things stay snappy. A `go build` of a larger project can take quite some time. I thought about changing the default according to the configuration I showed in #594 but that seemed more risky as it's a larger departure from the current behaviour which risks more easily breaking on people.

I've updated the documentation to reflect the change and added a comment about using `gometalinter` instead with the `--fast` flag.

### DIFF
--- a/autoload/ale/linter.vim
+++ b/autoload/ale/linter.vim
@@ -23,7 +23,7 @@ let s:default_ale_linter_aliases = {
 " rpmlint is disabled by default because it can result in code execution.
 let s:default_ale_linters = {
 \   'csh': ['shell'],
-\   'go': ['go build', 'gofmt', 'golint', 'gosimple', 'go vet', 'staticcheck'],
+\   'go': ['gofmt', 'golint', 'go vet'],
 \   'help': [],
 \   'rust': ['cargo'],
 \   'spec': [],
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

daenney:
🍻

PR #600, opened 2017-05-30
Remove -X flag from perl defaults.

oalders:
"-X Disables all warnings regardless of use warnings or $^W".  See
"perldoc perlrun" or http://perldoc.perl.org/perlrun.html

With the current defaults, warnings are squashed.  For example:

$ perl -X -Mwarnings -c -e'BEGIN { 42 + undef }'
-e syntax OK

$ perl -Mwarnings -c -e'BEGIN { 42 + undef }'
Use of uninitialized value in addition (+) at -e line 1.
-e syntax OK

So, it's not clear from the current defaults whether Ale wants to remove
warnings or enable them.  As it stands, it's trying to do both and the
disabling appears to win.

This commit enables warnings by default.

I'm not familiar with Vader, but if this needs some tests, I'm happy to dig in and try to figure that out.

### DIFF
--- a/ale_linters/perl/perl.vim
+++ b/ale_linters/perl/perl.vim
@@ -5,7 +5,7 @@ let g:ale_perl_perl_executable =
 \   get(g:, 'ale_perl_perl_executable', 'perl')
 
 let g:ale_perl_perl_options =
-\   get(g:, 'ale_perl_perl_options', '-X -c -Mwarnings -Ilib')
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

oalders:
Thank you. I started using Ale last week and it has really been a huge improvement for me.  Also, fast merge!  🚀

w0rp:
The speed of merging can depend on whether or not I'm looking at my inbox at the time, etc.

PR #601, opened 2017-05-30
Include span label in rust lints

### LINKED ISSUE #597: rust/cargo only gives brief "mismatched types" error on type errors
When using the rust or cargo linters, the short error shown in the statusline is usually helpful. The exception is for type mismatches, where it will simply show ` mismatched types`, without communicating what those types are. neomake had the same issue (neomake/neomake#793), which was I believe was eventually solved by using the JSON output format (neomake/neomake#942; @JelteF or @fuine may be able to give more details).

To see an example of this, consider the following rust program:
```rust
fn main() {
    let foo: bool = 42;
}
```

Compiling it gives the following error:
```
error[E0308]: mismatched types
 --> test.rs:2:21
  |
2 |     let foo: bool = 42;
  |                     ^^ expected bool, found integral variable
  |
  = note: expected type `bool`
  = note:    found type `{integer}`
```

Unfortunately, the statusline given by ale just shows `mismatched types`. Given how common type errors are in Rust development, this hinders development speed (even navigating to the line and opening the location window does not help -- it also does not show the types).

### Expected behavior

ale statusline shows 

```
cargo: mismatched types (expected `bool` found `{integer}`)
```

jonhoo:
This turns relatively unhelpful error messages like

    mismatched types

into more expressive messages along the lines of

    mismatched types: expected bool, found integral variable

Fixes #597.

I would add a test to `test/handler/test_rust_handler.vader`, but I can't seem to find where the files it operates on are sourced from?


w0rp:
There are no files to source for `test_rust_handler.vader`. The tests in that file call the autoload functions with a `List` of lines for the JSON output, and check the `List` of objects that are returned. To cover this, you need to add some lines which contain spans with labels, and check that the labels are used in the messages.

jonhoo:
All right @w0rp, I think it's ready to merge.

w0rp:
Cheers! :beers:

PR #603, opened 2017-05-31
make prettier options enabled

craftgear:
Just moving from neomake to ale for linting javascript.
I love how It's fast and no-brainer to use!

One thing though, seems that ale#Escape make prettier's options unusable.
Here's a tiny patch for it.

Thanks for you great work.


w0rp:
Cheers! :beers:

craftgear:
🍻

PR #613, opened 2017-06-02
Add prettier to list of JS checkers

jackfranklin:
Thought that Ale didn't support Prettier, but when looking through the code I found `prettier.vim` so it looks like it does, so made this PR. If I've got something wrong then apologies for the noise!

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ name. That seems to be the fairest way to arrange this table.
 | HTML | [HTMLHint](http://htmlhint.com/), [proselint](http://proselint.com/), [tidy](http://www.html-tidy.org/) |
 | Java | [checkstyle](http://checkstyle.sourceforge.net), [javac](http://www.oracle.com/technetwork/java/javase/downloads/index.html) |
 | JavaScript | [eslint](http://eslint.org/), [jscs](http://jscs.info/), [jshint](http://jshint.com/), [flow](https://flowtype.org/), [standard](http://standardjs.com/), [xo](https://github.com/sindresorhus/xo)
-| JSON | [jsonlint](http://zaa.ch/jsonlint/) |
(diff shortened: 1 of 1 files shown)


jackfranklin:
I've just noticed Prettier is a _fixer_, not a _checker_, which makes sense. I wonder if it should be listed somewhere in the README though? It would be easy for someone to think Ale doesn't support it.

adriaanzon:
Yeah, it would make sense to me for the first section of the readme to have a *linters* and *fixers* subsection. The other fixers still need to be listed too though, because the fixing feature is still quite new.

I would also vote for a dedicated `ale_fixers` folder. Currently they live in `autoload/ale/handlers`, together with the generalized handlers. Any thoughts about that @w0rp? Have you considered doing that?

w0rp:
The `ale_fixers` directory doesn't need to exist, as the fixers are loaded differently. The way fixing works, it's basically the opposite of linting.

For the linters, ALE searches `runtimepath` to load everything it can find, and you can filter the list of linters down if you only want to run a few. ALE tries to check your file with anything it can find by default.

For fixing files, ALE doesn't load anything automatically. You pass a series of short names registered in the registry, function names, Funcref values, or lambdas, and ALE runs everything you gave it to fix the files. So, ALE doesn't need to search the `runtimepath`.

What we could do is move more of the functions for fixing things into the `fix` directory, so the first party function names would all be `ale#fix#...`.

w0rp:
I think we should add a second table to the README. List all of the tools for linting things in one table, and all of the tools for fixing things in the other table.

w0rp:
Hmm, actually, just put them all in the same table.

adriaanzon:
Makes sense. I only glanced over how fixers are registered, and that was what came to my mind first.  

> What we could do is move more of the functions for fixing things into the fix directory, so the first party function names would all be ale#fix#....

Yep, that would be fine too 👍

w0rp:
I have added these to the README and the main online documentation file now.

PR #615, opened 2017-06-02
Suggest :ALELint instead of ale#Lint()

adriaanzon:


w0rp:
I missed that. Cheers! :beers:

PR #620, opened 2017-06-04
nim: optionally check original file instead of a buffer copy

stefantalpalaru:
Most Nim projects have a project/file-specific "nim.cfg" which is ignored when running "nim check" on a copy in an unrelated temp dir.

This pull request implements an option to switch between original file and temporary copy checking, defaulting to the former which is guaranteed to always work properly.

### DIFF
--- a/ale_linters/nim/nimcheck.vim
+++ b/ale_linters/nim/nimcheck.vim
@@ -54,7 +54,7 @@ function! ale_linters#nim#nimcheck#GetCommand(buffer) abort
         \   . ' --threads:on --verbosity:0 --colors:off --listFullPaths %t'
     else
         return 'nim check'
-        \   . ' --threads:on --verbosity:0 --colors:off --listFullPaths %s'
+        \   . ' --verbosity:0 --colors:off --listFullPaths %s'
     endif
 endfunction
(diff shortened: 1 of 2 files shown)


w0rp:
Does nim support an argument for manually specifying the location of a configuration file? If it does, this can be fixed by setting that configuration file explicitly. See the JSHint linter for an example.

stefantalpalaru:
> Does nim support an argument for manually specifying the location of a configuration file?

No, it doesn't: https://nim-lang.org/docs/nimc.html#compiler-usage-configuration-files

The only arguments is has are for disabling specific steps in the search, not for adding extra paths and filenames.

It also doesn't allow specifying a project directory on the command line. All we can do is copy the whole project in the temp dir (but it would be hard to guess the project dir for files in nested subdirs, duplicating some compiler logic in the process) or just check the original file in its place.

w0rp:
In that case, set `lint_file` to `1` for Nim so it only checks for errors against the file on disk. If we have to work with the file, then do that.

w0rp:
I have done that now. Now we will only check Nim files on disk, which is the only thing that works well.

PR #621, opened 2017-06-05
Support dscanner for D linting

Calrama:


Calrama:
@w0rp There were no vader tests findable with `egrep -ri dmd` for the dmd linter I could adopt.

w0rp:
Have a look at the tests in `test/handler`

PR #632, opened 2017-06-07
Search python fixers in virtual environment

jlesquembre:
For the python linters, ALE will search first in a virtualenv first. This commit extends the python fixers to do the same, first try to find the fixer command in a virtualenv.
If the command is not found (nor in the virtualenv nor globally), don't do anything. Now if the fixer command doesn't exist, `ALEFix` returns an empty file.

### DIFF
--- a/autoload/ale/fixers/yapf.vim
+++ b/autoload/ale/fixers/yapf.vim
@@ -2,12 +2,17 @@
 " Description: Fixing Python files with yapf.
 
 function! ale#fixers#yapf#Fix(buffer) abort
+    let l:executable = ale#handlers#python#GetExecutable(a:buffer, 'yapf')
+    if empty(l:executable)
+        return 0
+    endif
+
     let l:config = ale#path#FindNearestFile(a:buffer, '.style.yapf')
     let l:config_options = !empty(l:config)
     \   ? ' --style ' . ale#Escape(l:config)
     \   : ''
 
     return {
-    \   'command': 'yapf --no-local-style' . l:config_options,
+    \   'command': ale#Escape(l:executable) . ' --no-local-style' . l:config_options,
     \}
 endfunction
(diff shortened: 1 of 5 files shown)


autoload/ale/handlers/python.vim
```diff
@@ -45,3 +45,24 @@ function! ale#handlers#python#HandlePEP8Format(buffer, lines) abort
...
+        if executable(l:ve_executable)
+            return l:ve_executable
+        endif
+    endif
+
+    if executable(a:cmd_name)
+        return a:cmd_name
+    endif
+
+    return v:null
```
w0rp: You can return an empty string here. I recommend only using `v:null` for JSON.

autoload/ale/fixers/yapf.vim
```diff
@@ -1,13 +1,18 @@
 " Author: w0rp <devw0rp@gmail.com>
 " Description: Fixing Python files with yapf.
 
-function! ale#fixers#yapf#Fix(buffer) abort
+function! ale#fixers#yapf#Fix(buffer, lines) abort
```
w0rp: You can remove the `lines` argument here and return `0` instead. You can return `0` to skip any fixer.

w0rp: The same applies to the other files.

autoload/ale/fixers/yapf.vim
```diff
@@ -1,13 +1,18 @@
...
+    endif
+
     let l:config = ale#path#FindNearestFile(a:buffer, '.style.yapf')
     let l:config_options = !empty(l:config)
     \   ? ' --style ' . ale#Escape(l:config)
     \   : ''
 
     return {
-    \   'command': 'yapf --no-local-style' . l:config_options,
+    \   'command': l:executable . ' --no-local-style' . l:config_options,
```
w0rp: Apply `ale#Escape` to `l:executable` right here, and in other files. The path to virtualenv might contain special characters.

jlesquembre:
I did the changes, but for the tests I need some more time, never used vader before.

w0rp:
Cheers! :beers:

PR #636, opened 2017-06-08
ale_fix_on_save, accept list of extensions instead of a boolean

jlesquembre:
I would like to fix on save only some files, e.g. I want to fix automatically on save javascript files, but I want to manually execute the fixers for python files.

Currently `g:ale_fix_on_save` accepts 0 or 1. This PR proposes to accept a list of extensions, e.g.: `['js', 'py']` to fix on save only some files. If you want to maintain the old behavior and fix on save all the files, you could  set `g:ale_fix_on_save` to `['*']`. For backwards compatibility, `g:ale_fix_on_save` still accepts 0 and 1, which is the same that setting it to `[]` or `['*']`

### DIFF
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -178,3 +178,17 @@ function! ale#Escape(str) abort
 
     return shellescape (a:str)
 endfunction
+
+" Parses the ale_fix_on_save value, before it was a number. If 0 returns `[]`,
+" if 1 returns `['*']`
+function! ale#ParseFixOnSave(value) abort
+    if type(a:value) == type(0)
+        if a:value == 0
+            return []
+        elseif a:value == 1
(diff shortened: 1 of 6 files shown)


w0rp:
This isn't the right way to handle this. We should support `b:ale_fix_on_save` instead, and then filetypes can be configured with `g:ale_pattern_options`. See `:help g:ale_pattern_options`.

w0rp:
We could consider also supporting `g:ale_filetype_options`, mapping filetypes to options instead of filename patterns.

jlesquembre:
Thanks for pointing me to `g:ale_pattern_options`, didn't know that option. I agree, it would be better to add support for `b:ale_fix_on_save`

PR #637, opened 2017-06-08
Improve elm linter

jwoudenberg:
Currently some elm compiler errors are not returned in json format. An example of this is the case of importing a module that does not exist. The current behavior of the elm-make ale linter is to simply not show an error in this case requiring one to manually run elm-make to figure out what's broken.

This PR adds a fallback. In case of a non-parseable error it takes the error-message as is and attaches it on the first line.

### DIFF
--- a/ale_linters/elm/make.vim
+++ b/ale_linters/elm/make.vim
@@ -21,7 +21,6 @@ function! ale_linters#elm#make#Handle(buffer, lines) abort
 
                 if l:file_is_buffer
                     call add(l:output, {
-                    \    'bufnr': a:buffer,
                     \    'lnum': l:error.region.start.line,
                     \    'col': l:error.region.start.column,
                     \    'type': (l:error.type ==? 'error') ? 'E' : 'W',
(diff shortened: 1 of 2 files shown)


ale_linters/elm/make.vim
```diff
@@ -29,9 +30,22 @@ function! ale_linters#elm#make#Handle(buffer, lines) abort
...
                 endif
             endfor
+        elseif l:line !=# 'Successfully generated /dev/null'
+            call add(l:unparsed_lines, l:line)
         endif
     endfor
 
+    if len(l:unparsed_lines) > 0
+        call add(l:output, {
+        \    'bufnr': a:buffer,
```
w0rp: This key doesn't need to be added.

ale_linters/elm/make.vim
```diff
@@ -29,9 +30,22 @@ function! ale_linters#elm#make#Handle(buffer, lines) abort
...
+        \    'lnum': 0,
+        \    'col': 0,
```
w0rp: You don't need to set this key if there's no column.

ale_linters/elm/make.vim
```diff
@@ -29,9 +30,22 @@ function! ale_linters#elm#make#Handle(buffer, lines) abort
...
             endfor
+        elseif l:line !=# 'Successfully generated /dev/null'
+            call add(l:unparsed_lines, l:line)
         endif
     endfor
 
+    if len(l:unparsed_lines) > 0
+        call add(l:output, {
+        \    'bufnr': a:buffer,
+        \    'lnum': 0,
```
w0rp: You should set the line number to `1` if the error is at the first line.

jwoudenberg:
Thanks for the review @w0rp. I've got a lot going on this week, will return to this next!

jwoudenberg:
Thank you for the review!

w0rp:
Cheers! :beers:

PR #640, opened 2017-06-09
Perlcritic column number and rule names

shumphrey:
This Adds the column number for perlcritic linting errors.
It also adds a config option to display the perlcritic rule name for each error.


w0rp:
Looks fine to me.

w0rp:
Cheers! :beers:

PR #648, opened 2017-06-14
Add fusion-lint, documentation, and tests

RyanSquared:
This adds `fusion-lint` to the list of linters, including documentation over options and a test for the command string callback. I'm not able to verify if the tests pass for the callback due to a weird quirk with Vader and UltiSnips; however, extensive manual testing showed they should work.

### DIFF
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -33,6 +33,8 @@ CONTENTS                                                         *ale-contents*
       erlc................................|ale-erlang-erlc|
     fortran...............................|ale-fortran-options|
       gcc.................................|ale-fortran-gcc|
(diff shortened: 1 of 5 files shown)


doc/ale.txt
```diff
@@ -33,6 +33,8 @@ CONTENTS                                                         *ale-contents*
       erlc................................|ale-erlang-erlc|
     fortran...............................|ale-fortran-options|
       gcc.................................|ale-fortran-gcc|
+		fusionscript..........................ale-fuse-options
```
w0rp: The formatting isn't right here. Use `||` for linking to the options, and align the entries like the other ones.

PR #651, opened 2017-06-14
Fix for Crystal support

### LINKED ISSUE #650: Crystal linter fails on opening new file
```
 Current Filetype: crystal
Available Linters: ['crystal']
  Enabled Linters: ['crystal']
 Linter Variables:

 Global Variables:

let g:ale_echo_cursor = 1
let g:ale_enabled = 1
let g:ale_lint_delay = 600
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_linter_aliases = {}
let g:ale_linters = {}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_error = '>>'
let g:ale_sign_warning = '--'
  Command History:
```

When opening a new buffer, e.g., `vim something/foo.cr`, the crystal linter runs on the file (as I would expect). The problem is that it fails because the target file, `./something/foo.cr`, does not 
exist yet until I save that buffer.

I looked into changing the crystal linter to accept the current buffer contents instead of reading from a file, but there's a prerequisite issue still being discussed at the moment (crystal-lang/crystal#4561).

My suggested solution is for the linting not to run if the `lint_file` attribute is set for the linter and the actual file does not exist.

TheLonelyGhost:
Fixes #650.

If the linter requires the target file (e.g., `lint_file` attribute is set), skip the linting for that file if it does not yet exist. This can happen if you open vim at a location for a new file and the buffer for that file has not been written out to disk yet.

Also includes a fix for Crystal compiler output defaulting to colorized output now, which should be removed for display in the vim output.

### DIFF
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -1,4 +1,4 @@
-" Author: w0rp <devw0rp@gmail.com>
+" Author: w0rp <devw0rp@gmail.com>, David Alexander <opensource@thelonelyghost.com>
 " Description: Primary code path for the plugin
 "   Manages execution of linters when requested by autocommands
 
@@ -86,6 +86,11 @@ function! ale#Lint(...) abort
     if has_key(s:should_lint_file_for_buffer, l:buffer)
         unlet s:should_lint_file_for_buffer[l:buffer]
         let l:should_lint_file = 1
+    endif
+
+    " Don't lint files if the file does not exist
+    if l:should_lint_file && filereadable(expand('#' . l:buffer . ':p'))
+      let l:should_lint_file = 0
     endif
 
     " Initialise the buffer information if needed.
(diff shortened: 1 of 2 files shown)


TheLonelyGhost:
I did notice a slight hit to performance, but that was the easiest way I found to check if the file exists. Definitely open to suggestions.

TheLonelyGhost:
Alright. Now I'm confused. WIth `empty(glob(expand('#' . l:buffer . ':p')))` it worked without any test failures, but `filereadable(expand('#' . l:buffer . ':p')))` results in some issues.

Wat? o_O

autoload/ale.vim
```diff
@@ -85,7 +85,8 @@ function! ale#Lint(...) abort
     " Check if we previously requested checking the file.
     if has_key(s:should_lint_file_for_buffer, l:buffer)
         unlet s:should_lint_file_for_buffer[l:buffer]
-        let l:should_lint_file = 1
+        " Lint files if they exist.
+        let l:should_lint_file = filereadable(expand('#' . l:buffer . ':p'))
```
blueyed: Couldn't ALE also use temporary files in this case, like with modified files?!

w0rp: No. That's a check for linters which cannot use temporary files.

PR #660, opened 2017-06-16
handle column number in javac linter

oaue:
This retrieve the column number in javac output.


w0rp:
Cheers! :beers:

w0rp:
I might see about looking at the symbol names for determining the end positions for errors later. Then ALE will be able to highlight the whole word.

PR #663, opened 2017-06-18
TSLint: distinguish warnings from errors

mrpeabody:
Currently TSLint linter displays everything as errors. After looking at the source code, I've modified the regex to detect tslint message type so that ALE can show warnings correctly.

### DIFF
--- a/ale_linters/typescript/tslint.vim
+++ b/ale_linters/typescript/tslint.vim
@@ -14,19 +14,21 @@ endfunction
 function! ale_linters#typescript#tslint#Handle(buffer, lines) abort
     " Matches patterns like the following:
     "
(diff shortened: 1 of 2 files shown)


mrpeabody:
Added the test :)

w0rp:
Cheers! :beers:

mrpeabody:
By the way, wanted to tell you, this plugin is awesome, I'm positive in a year or so syntastic will be history :D

w0rp:
Syntastic was a great inspiration for this plugin, and LCD's comments on what to do for improving on what was there really helped a lot. With the help of others, eventually ALE should get support for all of the same linters and more.

PR #666, opened 2017-06-19
Support project's local phpcs installation

Firehed:
This change allows use of the PHPCS linter when installed as a project's dev dependency, instead of only a global install. It firsts checks for `vendor/bin/phpcs` as the executable path before falling back to the previous behavior (`phpcs` anywhere on the $PATH). This is similar to how many Node-based linters work, but with the paths from PHP's Composer.

### DIFF
--- a/ale_linters/php/phpcs.vim
+++ b/ale_linters/php/phpcs.vim
@@ -14,12 +14,14 @@ function! ale_linters#php#phpcs#GetExecutable(buffer) abort
 endfunction
 
 function! ale_linters#php#phpcs#GetCommand(buffer) abort
+    let l:executable = ale_linters#php#phpcs#GetExecutable(a:buffer)
+
     let l:standard = ale#Var(a:buffer, 'php_phpcs_standard')
     let l:standard_option = !empty(l:standard)
     \   ? '--standard=' . l:standard
     \   : ''
 
-    return ale_linters#php#phpcs#GetExecutable(a:buffer)
+    return ale#Escape(l:executable)
     \   . ' -s --report=emacs --stdin-path=%s ' . l:standard_option
 endfunction
 
--- a/ale_linters/php/phpcs.vim
+++ b/ale_linters/php/phpcs.vim
@@ -3,13 +3,22 @@
 
 let g:ale_php_phpcs_standard = get(g:, 'ale_php_phpcs_standard', '')
 
+function! ale_linters#php#phpcs#GetExecutable(buffer) abort
+    return ale#path#ResolveLocalPath(
+    \   a:buffer,
+    \   'vendor/bin/phpcs',
+    \   'phpcs'
+    \)
+endfunction
+
 function! ale_linters#php#phpcs#GetCommand(buffer) abort
     let l:standard = ale#Var(a:buffer, 'php_phpcs_standard')
     let l:standard_option = !empty(l:standard)
     \   ? '--standard=' . l:standard
     \   : ''
 
-    return 'phpcs -s --report=emacs --stdin-path=%s ' . l:standard_option
+    return ale_linters#php#phpcs#GetExecutable(a:buffer)
+    \   . ' -s --report=emacs --stdin-path=%s ' . l:standard_option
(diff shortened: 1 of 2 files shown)


ale_linters/php/phpcs.vim
```diff
@@ -1,15 +1,24 @@
...
+endfunction
+
 function! ale_linters#php#phpcs#GetCommand(buffer) abort
     let l:standard = ale#Var(a:buffer, 'php_phpcs_standard')
     let l:standard_option = !empty(l:standard)
     \   ? '--standard=' . l:standard
     \   : ''
 
-    return 'phpcs -s --report=emacs --stdin-path=%s ' . l:standard_option
+    return ale_linters#php#phpcs#GetExecutable(a:buffer)
```
w0rp: You should `ale#Escape` the executable here, as the path might contain spaces, etc.

ale_linters/php/phpcs.vim
```diff
@@ -1,15 +1,24 @@
...
 " Description: phpcs for PHP files
 
 let g:ale_php_phpcs_standard = get(g:, 'ale_php_phpcs_standard', '')
 
+function! ale_linters#php#phpcs#GetExecutable(buffer) abort
+    return ale#path#ResolveLocalPath(
+    \   a:buffer,
+    \   'vendor/bin/phpcs',
+    \   'phpcs'
+    \)
```
w0rp: Could you make the global executable configurable and add a `use_global` option while you're at it, similar to eslint and others?

Firehed:
Thanks for the feedback! It's my first time writing any vimscript (outside of vimrc); I'll try to address your concerns and update the pull shortly.

ale_linters/php/phpcs.vim
```diff
@@ -1,15 +1,28 @@
...
+" Author: jwilliams108 <https://github.com/jwilliams108>, Eric Stern <https://github.com/firehed>
 " Description: phpcs for PHP files
 
 let g:ale_php_phpcs_standard = get(g:, 'ale_php_phpcs_standard', '')
 
+call ale#Set('php_phpcs_executable', 'phpcs')
+call ale#Set('php_phpcs_use_global', 0)
+
+function! ale_linters#php#phpcs#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'php_phpcs', [
```
w0rp: You fell into the "pit of success" here. I might rename this because it turns out it works for more than just node programs.

partounian:
I'm using root/sage (8.5.1 latest production release) which is a WP theme. This reason this is important is because I also have template overrides. The template overrides are done with WP Coding Standards that way diffs can be checked against the originals when needed, and the rest of the theme follows the Roots guidelines (PSR-2 with slight modifications). I have PHPCS scripts for both of them. 

What is the best way to handle linting for this situation? 

(btw just switched to ale from neomake and loving the performance boost so far)

Firehed:
PHPCS should pick up a `phpcs.xml` in the project root automatically (or, at least, it does for me). However there's a `g:ale_php_phpcs_standard` variable that you can set in your vimrc that will let you explicitly pass a `--standard=<that value>` flag to the command.

partounian:
Yes but what if the project has two different set of styles? For example the main dir all subdirectories follow a variant of PSR-2 except for files in `/templates` and `/woocommerce`.

Firehed:
Based on [this page](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-default-configuration-file), it looks like PHPCS may pick up on directory-relative ruleset files. Failing that, there appears to be at least some support for specifying included/excluded files to lint in the phpcs.xml file, although it's not clear to me from their docs if you could set up two configurations "next to each other" in that way.

You might be able to rig something up with an `ftdetect` script; something like `au BufNewFile,BufRead */templates/*.php set g:ale_php_phpcs_standard=path/to/rules.xml`. Definitely not an ideal solution, but I think PHPCS really isn't designed to allow for multiple standards within the same project.

PR #675, opened 2017-06-22
Add profile, other options to the perlcritic linter

RsrchBoy:
This should bring the `perlcritic` linter in line with the `perl`
linter, allowing for custom (and per-buffer) perlcritic options and
executable.

We also supply a "standard" perltidy profile setting.


ale_linters/perl/perlcritic.vim
```diff
@@ -1,17 +1,32 @@
...
 " Description: This file adds support for checking perl with perl critic
 
-if !exists('g:ale_perl_perlcritic_showrules')
-    let g:ale_perl_perlcritic_showrules = 0
-endif
+let g:ale_perl_perlcritic_executable =
+\   get(g:, 'ale_perl_perlcritic_executable', 'perlcritic')
+
+let g:ale_perl_perlcritic_profile =
+\   get(g:, 'ale_perl_perlcritic_profile', '.../.perlcriticrc')
```
w0rp: You should use the `ale#path#FindNearestFile` function below to find `.percriticrc` instead. See the JSHint implementation for an example.

ale_linters/perl/perlcritic.vim
```diff
@@ -1,17 +1,32 @@
...
 
 function! ale_linters#perl#perlcritic#GetCommand(buffer) abort
     let l:critic_verbosity = '%l:%c %m\n'
-    if g:ale_perl_perlcritic_showrules
+    if ale#Var(a:buffer, 'perl_perlcritic_showrules')
         let l:critic_verbosity = '%l:%c %m [%p]\n'
     endif
 
-    return "perlcritic --verbose '". l:critic_verbosity . "' --nocolor"
+    return ale_linters#perl#perlcritic#GetExecutable(a:buffer)
```
w0rp: You should apply `ale#Escape` to the executable here if it's configurable, as the executable path could contain spaces, etc.

test/test_perlcritic_linter.vader
```diff
@@ -0,0 +1,42 @@
+" NOTE: We use the 'b:' forms below both to make testing easier (no need to
+" mess with globals) and to ensure that we're properly using ale#Var()
+
+Before:
+  runtime ale_linters/perl/perlcritic.vim
+  silent noautocmd new testfile.pl
```
w0rp: I recommend using `ale#test#SetFilename` instead. `new` produces weird output when you run the Vader tests inside of Vim while editing the tests.

ale_linters/perl/perlcritic.vim
RsrchBoy: Thanks for that -- I'd mistaken the `...` as something `perlcritic` does, rather than a shell construct.

ale_linters/perl/perlcritic.vim
RsrchBoy: Done.  I've also applied it to the profile file.

w0rp:
I think there might be a commit you haven't pushed yet.

RsrchBoy:
There is :)  I'm having trouble with your last comment, concerning the use of `ale#test#SetFilename()`.  I can run the tests without a problem in the current incantation, but I can't quite seem to replicate that with `...#SetFilename()`.

I think I'm just going to bypass that and use a [`Given:` block](https://github.com/junegunn/vader.vim#given), allowing Vader to handle it.

w0rp:
Cheers! :beers:

RsrchBoy:
Thanks! :)

PR #681, opened 2017-06-22
Skip invalid signs

teasp00n:
Kotlinc was reporting some warnings about the classpath it was being called with. These didn't have line numbers associated with them so a line number of -1 was used when attempting to add the sign which caused an error. I've added a handler around the code that adds the sign so invalid signs will no longer prevent valid signs from being added. The loclist is still populated with errors/ warnings that cannot be added as signs. There might be a more efficient way to do this, I'm not overly familiar with vimscript.


w0rp:
koitlinc shouldn't be reporting errors at line -1, they should be at line 1.

Issue #392, opened 2017-03-13
Fix C/C++ header issues


w0rp:
There are a number of issues with C and C++ filetypes and header files which need to be resolved. This issue will capture all of them, so discussion doesn't get fragmented.

It is worth noting that it is not probably not possible to detect the location of header files for 100% of projects. Some manual configuration on the part of users will be needed. (See the most popular C projects on GitHub for examples.)

## TODO

- [x] Support `#include` lines for quoted headers in the same directory.
- [x] Somehow detect the location of other header files, and use `-I` to include those. (See #379.)
- [x] Solve some strange issue with capitalized header names? (See: #318. Probably just how the system reports a missing header file.)
- [x] Ensure that errors in headers _are not_ reported as errors inside C files. (See #321, #211. A filename filter for errors may be needed, or a filter for `<stdin>`, etc.)
- [x] Fix some kind of issue with `#pragma once` in C++ header files. (See #356)

klement:
> Somehow detect the location of other header files, and use -I to include those. (See #379.)

I think this is not a job for ale. User should provide the necessary `-I` and `-isystem` switches to the appropriate `ale_c_*_options/ale_cpp_*_options`. Guessing would probably break horribly in more complicated projects, e.g. self-contained build systems, cross-compile scenarios etc.

As for `#include "header"` syntax - per C standard, this type of include is implementation-specific. This typically means that the header search starts in the directory containing the file with the `#include` statement. In this case - if the actual gcc/clang command is not executed in the source code directory (when ale runs the linter), I would add an implicit `-I<c-source-directory>` to the gcc/clang command line. Also, this might need an on/off behaviour (e.g. `let g:ale_c_gcc_autoinclude_source_dir = 1` or similar), because if using a preprocessor which doesn't support `#include ""` syntax and treats it as `#include <>` instead (allowed by C standard), then this might break things..

w0rp:
For the `#include "foo"` syntax, all I did was fix it so the directory the file is in will be included in the search path, which is how `gcc` and `clang` work, which are the only supported compilers for linting. If some other compilers are added, they can be implemented differently.

For `-I` paths, it would be very useful to parse a project's configuration files, or follow a common project structure, etc., and determine how to include headers automatically. If it doesn't break anything, or it can just be turned off, it will only be benefit.

w0rp:
I think for basically any functionality, you should be able to turn it off. Everyone wants different behaviour.

klement:
[...]

PR #686, opened 2017-06-23
Add clangcheck Linter to cpp

Gagbo:
Related to #392 but does not fix it.
In order to run clang static analysis that actually finds headers in not obvious places, I added the clang-check linter to the list of cpp linters. As stated by the [doc](http://clang.llvm.org/docs/ClangCheck.html), clang-check is a wrapper of libTooling to run those static analyses. That means it supports reading compilation databases JSONs and I think we should be able to use this tool when possible (especially since clang-tidy is not packaged in Fedora, so clang-check is the easiest method of running clang analysis on cpp code in my opinion).

I am not really proficient with Vader yet, so there are no tests for the time being, but it's the only missing part to be perfectly, fully compliant with merge guidelines. Hopefully it won't take too long to do, as it is heavily inspired on clang-tidy linter


Gagbo:
Commit 622584c is related to #688 actually, but I forgot to precise the weird interactions between options in clang-tidy linter

PR #688, opened 2017-06-23
Added builddir option to clang-tidy to point to json folder

Gagbo:
Related to #392 but does not fix it (it only gives clang-tidy the possibility to read properly compilation db in order to find all includes)

Clangtidy linter currently deactivates the look for json compilation databases if the option cpp_clangtidy_options is set to a non-empty string. This is problematic when the only option we want to pass to clangtidy is actually a compilation database.

To compensate for this, another option is added to clangtidy linter, named `ale_cpp_clangtidy_builddir` that specifically only takes the path to the build directory. Setting this option will nullify all other `ale_cpp_clangtidy_options` set, since finding a compilation database means we have all the compilation flags we care about. (Also this behaviour means we can have default flags in the vimrc, that will be overwritten by compilation db flags for the specific projects/folders that do support it).

Only the vader test is missing, the only fringe case I can imagine is to verify that when both sets, clang-tidy effectively takes the builddir option. I'll look into vader folder to try and implement the test.

### DIFF
--- a/ale_linters/cpp/clangtidy.vim
+++ b/ale_linters/cpp/clangtidy.vim
@@ -30,7 +30,7 @@ function! ale_linters#cpp#clangtidy#GetCommand(buffer) abort
     if empty(l:user_builddir)
         let l:builddir_names = ale#Var(a:buffer, 'cpp_clangtidy_builddirnames')
         for l:name in l:builddir_names
-            let l:candidates = finddir(l:name, expand('%:p:h') . ';', -1)
+            let l:candidates = finddir(l:name, expand('#' . a:buffer.':p:h') . ';', -1)
             for l:candidate in l:candidates
                 if filereadable(l:candidate . '/compile_commands.json')
                     let l:user_builddir = l:candidate
--- a/ale_linters/cpp/clangtidy.vim
+++ b/ale_linters/cpp/clangtidy.vim
@@ -23,14 +23,13 @@ function! ale_linters#cpp#clangtidy#GetCommand(buffer) abort
     let l:user_options = ale#Var(a:buffer, 'cpp_clangtidy_options')
     let l:user_build_dir = ale#Var(a:buffer, 'c_build_dir')
 
-    " Build directory has the priority if
-    " both builddir and builddirnames options are defined
+    " c_build_dir has the priority if defined
     if empty(l:user_build_dir)
         let l:user_build_dir = ale#c#FindCompileCommands(a:buffer)
     endif
 
-    " We check again if user_builddir stayed empty after the builddirnames
-    " check
+    " We check again if user_builddir stayed empty after the
+    " c_build_dir_names check
     " If we found the compilation database we override the value of
     " l:extra_options
     if empty(l:user_build_dir)
--- a/ale_linters/cpp/clangtidy.vim
+++ b/ale_linters/cpp/clangtidy.vim
@@ -1,4 +1,5 @@
-" Author: vdeurzen <tim@kompiler.org>, w0rp <devw0rp@gmail.com>
+" Author: vdeurzen <tim@kompiler.org>, w0rp <devw0rp@gmail.com>,
+" gagbo <gagbobada@gmail.com>
 " Description: clang-tidy linter for cpp files
 
 " Set this option to check the checks clang-tidy will apply.
@@ -8,15 +9,26 @@ let g:ale_cpp_clangtidy_checks = get(g:, 'ale_cpp_clangtidy_checks', ['*'])
 " This will disable compile_commands.json detection.
 let g:ale_cpp_clangtidy_options = get(g:, 'ale_cpp_clangtidy_options', '')
 
+" Set this option to manually point to the build directory for clang-tidy.
+" This will disable all the other clangtidy_options, since compilation
+" flags are contained in the json
+let g:ale_cpp_clangtidy_builddir = get(g:, 'ale_cpp_clangtidy_builddir', '')
+
 function! ale_linters#cpp#clangtidy#GetCommand(buffer) abort
     let l:check_list = ale#Var(a:buffer, 'cpp_clangtidy_checks')
     let l:check_option = !empty(l:check_list)
     \   ? '-checks=' . ale#Escape(join(l:check_list, ',')) . ' '
     \   : ''
     let l:user_options = ale#Var(a:buffer, 'cpp_clangtidy_options')
-    let l:extra_options = !empty(l:user_options)
-    \   ? ' -- ' . l:user_options
-    \   : ''
+    let l:user_builddir = ale#Var(a:buffer, 'cpp_clangtidy_builddir')
+    " Build directory has the priority if both options are defined
+    if empty(l:user_builddir)
+        let l:extra_options = !empty(l:user_options)
+        \   ? ' -- ' . l:user_options
+        \   : ''
+    else
+        let l:extra_options = ' -p ' . l:user_builddir
+    endif
 
     return 'clang-tidy ' . l:check_option . '%s' . l:extra_options
 endfunction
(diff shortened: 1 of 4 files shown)


w0rp:
It would be better to search for this automatically with a list of names for the build directory, which can be changed with an option. Set the option to `['bin', 'build']` by default, and search upwards through directories to find `'bin/compile_commands.json'`.

Gagbo:
If I run CMake in the build dir and have CMAKE_EXPORT_COMPILE_COMMANDS on,
then the json is directly in the build folder and not in a bin
subdirectory. Maybe just searching upwards for build folder and then just
check for an existing compilé_commands.json ? That would mean using only 1
string for the option

w0rp:
In that case, just use `ale#path#FindNearestFile(a:buffer, 'compile_commands.json')` and set that if it's not empty. Remember to apply `ale#Escape()` to the path when writing it into the command string. Then add a test for that.

Gagbo:
I have one example case to illustrate why I finally wasn't able to work out a solution with `ale#path#FindNearestFile()`

- If the source file is in `${PROJECT_ROOT}/src/ui` and then the build dir we should find is `${PROJECT_ROOT}/build`, then using FindNearestFile will fail because running only upward search means we will look for `compile_commands.json` only in `${PROJECT_ROOT}/src/ui`, `${PROJECT_ROOT}/src`, `${PROJECT_ROOT}` and irrelevant directories after. I think we don't want to change the FindNearestFile function to run combined {up/down}wards search, as vim help really shows it's not really a good solution.

Therefore I feel like the quicker solution to this problem would be to give directly the build directory as an option. But I agree that I could : 
- use a default list with common names,
- run FindNearestDir for the names in the list,
- and then look for compile_commands.json in these folders

And have that as a default behaviour if the option is not set.

Gagbo:
I finally got over my dumbness and implemented the behaviour mentioned earlier : 
- If `ale_cpp_clangtidy_builddir` is set, then the `-p ale_cpp_clangtidy_builddir` is set to the linter and we're done
- If `ale_cpp_clangtidy_builddir` is not set, then for each `name` in the list `ale_cpp_clangtidy_builddirnames` (which defaults to `[ 'build', 'bin' ]`)
    - We look upwards for any directory named `name` and then verify if there is a `name/compile_commands.json` readable file.
    - Whenever a suitable directory with a readable file is found, we set `ale_cpp_clangtidy_builddir` to this subfolder and stop the search.

The method relies on `finddir()` giving a list ordered from the nearest subdirectory to the furthest, so that the first valid candidate is really the one we want (that is why I added a `build` folder to the root of `test_c_projects`, to check the priority is followed).

ale_linters/cpp/clangtidy.vim
```diff
@@ -8,15 +9,51 @@ let g:ale_cpp_clangtidy_checks = get(g:, 'ale_cpp_clangtidy_checks', ['*'])
...
-    \   ? ' -- ' . l:user_options
-    \   : ''
+    let l:user_builddir = ale#Var(a:buffer, 'cpp_clangtidy_builddir')
+
+    " Build directory has the priority if
+    " both builddir and builddirnames options are defined
+    if empty(l:user_builddir)
+        let l:builddir_names = ale#Var(a:buffer, 'cpp_clangtidy_builddirnames')
+        for l:name in l:builddir_names
+            let l:candidates = finddir(l:name, expand('#' . a:buffer.':p:h') . ';', -1)
```
w0rp: Have a look at the implementation of `ale#python#FindVirtualenv`. You can use the `ale#path#Upwards` function with the `isdirectory` function for implementing this. I suggest implementing this as a `ale#c#FindCompileCommands` function. It could be useful for other tools.

ale_linters/cpp/clangtidy.vim
```diff
@@ -8,15 +9,51 @@ let g:ale_cpp_clangtidy_checks = get(g:, 'ale_cpp_clangtidy_checks', ['*'])
...
     let l:user_options = ale#Var(a:buffer, 'cpp_clangtidy_options')
-    let l:extra_options = !empty(l:user_options)
-    \   ? ' -- ' . l:user_options
-    \   : ''
+    let l:user_builddir = ale#Var(a:buffer, 'cpp_clangtidy_builddir')
+
+    " Build directory has the priority if
+    " both builddir and builddirnames options are defined
+    if empty(l:user_builddir)
+        let l:builddir_names = ale#Var(a:buffer, 'cpp_clangtidy_builddirnames')
```
w0rp: I recommend calling this variable `ale_c_build_dir_names`. It can be used for a few C and C++ tools, and that name will match the `ale_virtualenv_dir_names` variable name.

ale_linters/cpp/clangtidy.vim
```diff
@@ -8,15 +9,51 @@ let g:ale_cpp_clangtidy_checks = get(g:, 'ale_cpp_clangtidy_checks', ['*'])
...
 function! ale_linters#cpp#clangtidy#GetCommand(buffer) abort
     let l:check_list = ale#Var(a:buffer, 'cpp_clangtidy_checks')
     let l:check_option = !empty(l:check_list)
     \   ? '-checks=' . ale#Escape(join(l:check_list, ',')) . ' '
     \   : ''
     let l:user_options = ale#Var(a:buffer, 'cpp_clangtidy_options')
-    let l:extra_options = !empty(l:user_options)
-    \   ? ' -- ' . l:user_options
-    \   : ''
+    let l:user_builddir = ale#Var(a:buffer, 'cpp_clangtidy_builddir')
```
w0rp: I'm not sure if this one should be left as-is, or should be named `c_build_dir`, or maybe `cpp_build_dir`. I think leave it for now.

ale_linters/cpp/clangtidy.vim
```diff
@@ -8,15 +9,51 @@ let g:ale_cpp_clangtidy_checks = get(g:, 'ale_cpp_clangtidy_checks', ['*'])
 " This will disable compile_commands.json detection.
 let g:ale_cpp_clangtidy_options = get(g:, 'ale_cpp_clangtidy_options', '')
 
+" Set this option to manually point to the build directory for clang-tidy.
+" This will disable all the other clangtidy_options, since compilation
+" flags are contained in the json
+let g:ale_cpp_clangtidy_builddir = get(g:, 'ale_cpp_clangtidy_builddir', '')
+
+let g:ale_cpp_clangtidy_builddirnames =
```
w0rp: You can initialize this in `autoload/ale/c.vim` with the function below.

test/test_c_import_paths.vader
```diff
@@ -226,3 +226,13 @@ Execute(The C++ Clang handler should include root directories for projects with
...
+
+Execute(The C++ ClangTidy handler should include json folders for projects with suitable build directory in them):
+  runtime! ale_linters/cpp/clangtidy.vim
+
+  cd test_c_projects/json_project/subdir
+  silent noautocmd file file.cpp
+
+  AssertEqual
+  \   'clang-tidy -checks=''*'' %s -p ''/testplugin/test/test_c_projects/json_project/build'''
+  \ , ale_linters#cpp#clangtidy#GetCommand(bufnr(''))
```
w0rp: :ok_hand:

ale_linters/cpp/clangtidy.vim
Gagbo: I agree that this variable should be shared betwwen c-family linters, I wasn't sure about how to do it, I'll look into this

Gagbo:
I'm so bad with these tests it's incredible.

Well, I've done the changes as proposed. It looks better this way indeed.
- I chose to keep `ale_c_build*` names, because I think there is no reason to separate the behaviours between c and cpp (and most likely with any language of the C-family supported by CMake)
- The doc has not been written yet. Should I add a 'Global options' section under cpp language in the doc like it's done for Python ?
- If we agree on this behaviour, I'll also change #686 to make use of the C-family feature.

Then a discussion can be started on the right choice of `ale_c_build_dir_names` but I don't think I'm experienced enough to propose names. Maybe starting the conversation in #392 could help have a better default list. `[ 'build', 'build/bin', 'buildLinux', 'buildDebug', 'build_debug', 'build_release', ...]`

w0rp:
Yeah, you can document the option like the Python virtualenv option.

w0rp:
Cheers! :beers:

Gagbo:
Yay !
I sneaked a doc explanation in the other PR #686, basically because the default behaviour is not really as evident as it should be with the linter.

Last step is to actually add a test for autoload/ale/c.vim, but I'll look into this a little later, for the time being the behaviour to test is actually exactly the same as in FindVirtualEnv in python projects

PR #693, opened 2017-06-25
Feature/restore display of symbol

jsivak:
It appears that since commit #6b1f0c5, the Pylint symbol is no longer displaying in the msg output during checking.

I've updated the pylint handler function to display the symbol again and updated the tests to pass given the new change.

A thing to note: I chose to display the symbol *after* the lint message. If you want to more closely mirror the Pylint "original output" and put the symbol *before* the message, then let me know and I'll update the pull request.

Thanks


w0rp:
Cheers! :beers:

PR #695, opened 2017-06-26
Change regex to better handle messages with multiple groups of parentheses.

jsivak:
With the symbolic name now displaying for pylint, I stumbled across some lint messages that contain multiple groups of ()'s.
Example actual output from pylint directly:
`foo.py:34:4: W0702 (bare-except) No exception type(s) specified`

message in vim as seen before the fix:
`[pylint] W0702: specified (bare-except) No exception type(s) [Warning]`

message in vim after the patch is applied:
`[pylint] W0702: No exception type(s) specified (bare-except) [Warning]`

(*note*, I have `let g:ale_echo_msg_format = '[%linter%] %s [%severity%]' ` in my .vimrc)

This patch and test account for extra parentheses.


w0rp:
Cheers! :beers:

PR #696, opened 2017-06-27
Remove style classification from E999

ollien:
One of the things I noticed when working with Ale and flake8 was that syntax errors were reported as "style errors." I did some digging and noticed this was done in response to #430. Here, w0rp states that he made all error codes that started with "E" and "W" have the `'style'` `subtype`. However, this neglects the E999 error code, [which is documented](http://flake8.pycqa.org/en/latest/user/error-codes.html) as an error code used to report compilation failures. As such, since this is not a stylistic error, and I have added an exception for it in `ale_linters/python/flake8.vim`

### DIFF
--- a/ale_linters/python/flake8.vim
+++ b/ale_linters/python/flake8.vim
@@ -136,7 +136,7 @@ function! ale_linters#python#flake8#Handle(buffer, lines) abort
         \   'type': 'W',
         \}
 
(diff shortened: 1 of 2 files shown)


ollien:
Done! Let me know if that's what you wanted.

w0rp:
Cheers! :beers:

PR #699, opened 2017-06-27
Elm linter shows full error ranges

jwoudenberg:
Previously when using the elm linter only the first character of each error was shown. This PR ensures the complete range is passed to ALE.

Tests were updated to reflect this change.


w0rp:
Cheers! :beers:

PR #701, opened 2017-06-28
Add puppet-lint fixer

ace13:
Adds a puppet fixer using `puppet-lint --fix`

### DIFF
--- a/autoload/ale/fixers/puppetlint.vim
+++ b/autoload/ale/fixers/puppetlint.vim
@@ -1,0 +1,14 @@
+" Author: Alexander Olofsson <alexander.olofsson@liu.se>
+" Description: puppet-lint fixer
+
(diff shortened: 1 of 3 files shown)


autoload/ale/fixers/puppetlint.vim
```diff
@@ -0,0 +1,14 @@
+" Author: Alexander Olofsson <alexander.olofsson@liu.se>
+" Description: puppet-lint fixer
+
+function! ale#fixers#puppetlint#Fix(buffer) abort
+    let l:executable = ale#Var(a:buffer, 'puppet_puppetlint_executable')
```
w0rp: You'll need to default initialise the variables above, in case someone hasn't loaded the linter, but has loaded the fixer.

w0rp:
Cheers! :beers:

PR #702, opened 2017-06-28
Add prettier-standard support

sheerun:
Disclaimer: I'm an author

prettier-standard is a fixer (not linter), for standard rules

It uses prettier underneath

https://github.com/sheerun/prettier-standard


w0rp:
Sure, I'll edit out that duplicated header.

I recommend that people use ESLint instead of standard or xo, but I'll support whatever works.

sheerun:
Thank you, I've mentioned your plugin in the README: https://github.com/sheerun/prettier-standard

There's just one issue that happens when project uses just prettier-standard and no .eslintrc. I'll add another issue so you can track it.

PR #708, opened 2017-06-29
Close the quickfix/location list window when quitting the parent buffer

smkent:

### DIFF
--- a/autoload/ale/list.vim
+++ b/autoload/ale/list.vim
@@ -69,3 +69,15 @@ function! ale#list#CloseWindowIfNeeded(buffer) abort
         lclose
     endif
 endfunction
+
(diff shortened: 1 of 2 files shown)


Saul-Mirone:
This is awesome! Anyone can tell me how to use this?

w0rp:
I recommend creating the function which is added in the diff here and writing your own `QuitPre` autocmd function. I decided this is such a general thing that it's not worth including.

cwonrails:
Thanks for the advice on how to add this back. I know the original PR didn't add the tests you mentioned @w0rp, but this is a really, really useful option and one I had to hack around to get to work with Syntastic before swapping to ale. Is there some way this can get included in the project?

w0rp:
If and only if you write tests for it.

Issue #704, opened 2017-06-28
Add option for specifying path to c++ linters


Toreno96:
I think it would be really useful feature (and easy to implement, quoting w0rp) to have possibility of specifying path to c++ linters, just like it is possible with javascript (`ale_javascript_eslint_executable` option), mainly for specifying which version of g++ (or clang++) to use.

w0rp:
It's something that can be done, with tests, in the space of an hour or so if someone has the time. I'll do it eventually if I have an hour free for it.

w0rp:
This is the same as #711.

PR #709, opened 2017-06-29
Add executable selection for cpp/gcc 

Toreno96:
Partial implementation of #704 (only GCC for now).
Tell me if I should correct something, and if you want me to implement this functionality for the rest of C++ linters.

### DIFF
--- a/ale_linters/cpp/gcc.vim
+++ b/ale_linters/cpp/gcc.vim
@@ -1,4 +1,4 @@
-" Author: geam <mdelage@student.42.fr>
+" Author: geam <mdelage@student.42.fr>, Toreno96 <toreno96@gmail.com>
 " Description: gcc linter for cpp files
 
 let s:executable = get(g:, 'ale_cpp_gcc_executable', 'gcc')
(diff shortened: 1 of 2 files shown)


ale_linters/cpp/gcc.vim
```diff
@@ -1,9 +1,11 @@
-" Author: geam <mdelage@student.42.fr>
+" Author: geam <mdelage@student.42.fr>, Toreno96 <toreno96@gmail.com>
 " Description: gcc linter for cpp files
 
+let s:executable = get(g:, 'ale_cpp_gcc_executable', 'gcc')
+
 " Set this option to change the GCC options for warnings for C.
 if !exists('g:ale_cpp_gcc_options')
-    let s:version = ale#handlers#gcc#ParseGCCVersion(systemlist('gcc --version'))
+    let s:version = ale#handlers#gcc#ParseGCCVersion(systemlist(s:executable . ' --version'))
```
w0rp: I'll have to convert this to a callback in a chain. Let me handle this later.

w0rp:
I'm going to go through these now.

PR #712, opened 2017-06-30
Be more explicit about the effect of `g:..._perlcritic_profile = ''`

RsrchBoy:
Just to prevent any confusion, the documentation now explicitly states
that setting `g:ale_perl_perlcritic_profile` to an empty string merely
disables passing an explicit profile to `perlcritic` and does not cause
`--no-profile` to be set.


w0rp:
Cheers! :beers: 

That should be clearer.

PR #720, opened 2017-07-02
Allow `hadolint` linter to run via docker image

RsrchBoy:
These changes enable the `hadolint` linter to run via the author's docker
image, if present.   Three modes are supported:

 * never use docker;
 * always use docker; and
 * use docker as a fallback.

### DIFF
--- a/ale_linters/dockerfile/hadolint.vim
+++ b/ale_linters/dockerfile/hadolint.vim
@@ -1,5 +1,9 @@
 " Author: hauleth - https://github.com/hauleth
 
+" always, yes, never
+call ale#Set('dockerfile_hadolint_use_docker', 'never')
+call ale#Set('dockerfile_hadolint_docker_image', 'lukasmartinelli/hadolint')
+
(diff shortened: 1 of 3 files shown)


w0rp:
I don't know about this. What happens when we want to run a second linter in Docker? This seems like something that should have a more general solution, if it's supported at all.

w0rp:
I guess it takes input via stdin, so there are no filename mapping issues to handle, and the author supports running the tool via a Docker image. Maybe this is okay.

RsrchBoy:
I totally agree that this begs for something more general -- if for no other reason than there's about three other linters that (for various reasons) would be a pain to install outright but are easily squirreled away inside an image.   (e.g. [sqlint](/rsrchboy/docker-sqlint), [vint](/rsrchboy/docker-vint), [proselint](/rsrchboy/docker-proselint)).

However...  That's going to require some thought to do right.  This works for this linter, right now, while not really cluttering things up for however it's handled in the general.  (There's a reason I have it defaulting to "never" :))

w0rp:
Cheers! :beers:

PR #722, opened 2017-07-03
Drop "only master branch" travis restriction

RsrchBoy:
I get that one may want to restrict where travis runs (resource contentions,
reserving travis runs for master, etc), but this makes it difficult for others
to enable travis on their forks -- unless they're willing to do all their work
on master.

Would you be amenable to this change?  Or, in the alternative, creating a
[white/black list](https://docs.travis-ci.com/user/customizing-the-build/#Building-Specific-Branches)
to allow for a broader range of branches?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@
 sudo: required
 services:
     - docker
-branches:
-    only:
-        - master
 language: python
 script: |
     ./run-tests


w0rp:
I'll take this. I set that restriction when I was creating a few unstable branches and I didn't want to see the emails where Travis complained about them.

RsrchBoy:
Awesome, thanks :)

PR #725, opened 2017-07-04
Get linter args by parsing compile_commands.json (c/gcc and cpp/gcc linters)

esran:
I haven't done much more than get this working for myself and tidy it up a bit. There's no tests and I know there are some potential issues. I'm creating the PR more to share and get criticism on my first attempt at vim scripting.

It works best for databases generated from bear. This puts full paths everywhere. When the database is generated from cmake/ninja it has relative paths for the arguments. Additionally the database is written to the build directory and not the source tree. I got round this by putting a link to the database in the root of my source tree and ensuring the current working directory is the build directory (where the paths are relative to). Perhaps a better option would be a per project setting for the location of the build directory which could be used both to find the database and as a temporary working directory when linting.

Of course this isn't true json parsing to databases with slightly different layout could trip this code up. For example, it does rely on the "file" element being after the other two elements. That has been my experience with both bear and cmake/ninja however.


ale_linters/c/gcc.vim
```diff
@@ -10,14 +10,24 @@ if !exists('g:ale_c_gcc_options')
...
+
     let l:paths = ale#c#FindLocalHeaderPaths(a:buffer)
 
     " -iquote with the directory the file is in makes #include work for
     "  headers in the same directory.
     return 'gcc -S -x c -fsyntax-only '
-    \   . '-iquote ' . ale#Escape(fnamemodify(bufname(a:buffer), ':p:h')) . ' '
-    \   . ale#c#IncludeOptions(l:paths)
-    \   . ale#Var(a:buffer, 'c_gcc_options') . ' -'
+                \   . '-iquote ' . ale#Escape(fnamemodify(bufname(a:buffer), ':p:h')) . ' '
```
w0rp: The indentation went a bit crazy here, and in other places.

ale_linters/c/gcc.vim
```diff
@@ -10,14 +10,24 @@ if !exists('g:ale_c_gcc_options')
 endif
 
 function! ale_linters#c#gcc#GetCommand(buffer) abort
+    " attempt to get args from compilation database
+    let l:args = ale#c#FindCompileArgs(a:buffer)
+
+    " if we've found compile args then just use those
+    if (!empty(l:args))
```
w0rp: The outer parentheses are redundant.

ale_linters/c/gcc.vim
```diff
@@ -10,14 +10,24 @@ if !exists('g:ale_c_gcc_options')
...
+        return 'gcc -S -x c -fsyntax-only '
+                    \	. l:args
+                    \   . ' -'
```
w0rp: Have a look through files and get the indentation to match the rest of the codebase.

autoload/ale/c.vim
```diff
@@ -89,3 +89,59 @@ function! ale#c#FindCompileCommands(buffer) abort
...
+    let l:use_args = []
+
+    let l:compile_commands_path = ale#path#FindNearestFile(a:buffer, 'compile_commands.json')
+    if (!empty(l:compile_commands_path))
+        let l:bufname = fnamemodify(bufname(a:buffer), ':p')
+        let l:command = ''
+        let l:file = ''
+
+        for elem in readfile(l:compile_commands_path)
+            let l:split = split(elem, ':')
```
w0rp: I recommend using the built in JSON decoding functions instead. See `:help json_decode`.

Break this up into two functions, and write tests for each.  Write a function that accepts the result of `json_decode` and takes the options out of it, and a second function which calls the first function. Then you can write a test case which tests finding the file and parsing the options out if it, and a series of tests for the function which doesn't do any I/O.

w0rp: Add a test which checks the behaviour for files with bad syntax.

esran: Thanks for the feedback. Bit swamped at work currently but will try and find some time to get back on this.

autoload/ale/c.vim
```diff
@@ -89,3 +89,45 @@ function! ale#c#FindCompileCommands(buffer) abort
...
 endfunction
+
+" Given a buffer number, find a compilation database (compile_commands.json)
+" and parse this into an argument string
+function! ale#c#FindCompileArgs(buffer) abort
+    let l:compile_args = {}
+
+    let l:compile_commands_path = ale#path#FindNearestFile(a:buffer, 'compile_commands.json')
+    if !empty(l:compile_commands_path)
+        let l:bufname = fnamemodify(bufname(a:buffer), ':p')
```
w0rp: The buffer name can be different from the filename for the buffer. You should use `expand('#' . a:buffer . ':p')` instead.

autoload/ale/c.vim
```diff
@@ -89,3 +89,45 @@ function! ale#c#FindCompileCommands(buffer) abort
...
+
+        for elem in json_decode(readfile(l:compile_commands_path))
+            if elem["file"] == l:bufname
```
w0rp: It might be easier to apply `ale#path#IsBufferPath(a:buffer, elem.file)` here.

autoload/ale/c.vim
```diff
@@ -89,3 +89,45 @@ function! ale#c#FindCompileCommands(buffer) abort
...
+                let l:compile_args["directory"] = elem["directory"]
```
w0rp: You can use `x.foo` in place of `x["foo"]`, here and everywhere else.

autoload/ale/c.vim
```diff
@@ -89,3 +89,45 @@ function! ale#c#FindCompileCommands(buffer) abort
...
+                "   first arg (compiler)
+                "   last arg (filename)
+                "    -o objectfile
+                "    -c
+                let l_skip = 0
+                let l:args = []
+                for arg in split(elem["command"], ' ')[2:-2]
+                    if (l_skip == 1)
+                        let l_skip = 0
+                    elseif (arg == '-o')
```
w0rp: You don't need the parentheses for if statements, etc.

w0rp: Use `==#` for string equality just about everywhere. Vint will warn you about this. `==` behaves different depending on settings.

autoload/ale/c.vim
```diff
@@ -89,3 +89,45 @@ function! ale#c#FindCompileCommands(buffer) abort
...
+            if elem["file"] == l:bufname
+                let l:compile_args["directory"] = elem["directory"]
+                " skip unwanted stuff
+                "   first arg (compiler)
+                "   last arg (filename)
+                "    -o objectfile
+                "    -c
+                let l_skip = 0
+                let l:args = []
+                for arg in split(elem["command"], ' ')[2:-2]
```
w0rp: Will this break if one of the arguments contains spaces?

autoload/ale/c.vim
```diff
@@ -89,3 +89,45 @@ function! ale#c#FindCompileCommands(buffer) abort
...
+                    if (l_skip == 1)
+                        let l_skip = 0
```
w0rp: Use `l:` as a prefix for all local variables. Vint will warn you about this.

ale_linters/c/gcc.vim
```diff
@@ -10,6 +10,17 @@ if !exists('g:ale_c_gcc_options')
 endif
 
 function! ale_linters#c#gcc#GetCommand(buffer) abort
+    " attempt to get args from compilation database
+    let l:args = ale#c#FindCompileArgs(a:buffer)
+
+    " if we've found compile args then just use those
+    if len(l:args) > 0
```
w0rp: It might be better to test for `has_key(l:args, 'args') && has_key(l:args, 'directory')`

autoload/ale/c.vim
esran: Yes. I need to create a test case locally to see how bear/cmake handle spaces in arguments and then work out how to parse whatever they output.

esran:
Thanks again for the feedback. Going to try and find some time today/tomorrow to check out vint etc. but I'm on holiday next week so it might be a while till I get back to this :(

w0rp:
@esran That's cool. I might work on this myself if it somehow becomes a priority.

w0rp:
Thank you for working on this. I think it's going to be good when it's ready.

Gagbo:
The function `ale#c#FindCompileArgs` may be what's necessary to fix #782 too. If instead of checking for the exact file with [isBufferPath](https://github.com/w0rp/ale/pull/725/files#diff-e3c040b70d14f91037db76ed32789e8cR101) we use another check to look only for directory or the right path head if the file shown in `a:buffer` is a header.

The main idea is to modify the ale/c*/*.vim files to : 
- check the extension of the file linked to `a:buffer`
- follow the normal execution if it's not a header file
- use the modified version of `ale#c#FindCompileArgs` if the buffer is a header file, and look for the right JSON entry.

Problems (in my noob opinion) left to solve are : 
- What is the expected behaviour if the headers of the projects are in a different folder than inplemention files (aka `root/include` and `root/src`) ?
- What is the expected behaviour if the header files do not have the same basename as an implementation file ? (I suppose this could happen for very very virtual classes ?)
- Also, the clang linters need to be (once again) properly overridden if we find compile arguments for this. clang linters in header files produce this (file skipped because not found in the keys of compile_commands.json) : 
```
 Current Filetype: cpp
Available Linters: ['clang', 'clangcheck', 'clangtidy', 'cppcheck', 'cpplint', 'g++']
  Enabled Linters: ['clangcheck']
 Linter Variables:

let g:ale_cpp_clangcheck_executable = 'clang-check'
let g:ale_cpp_clangcheck_options = ''
 Global Variables:

let g:ale_echo_cursor = 1
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_linter_aliases = {}
let g:ale_linters = {'c': ['clangcheck', 'clang'], 'cpp': ['clangcheck'], 'python': ['pylint']}
let g:ale_open_list = 1
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_error = '>>'
let g:ale_sign_warning = '--'
  Command History:

(executable check - success) clang-check
(started) ['/bin/bash', '-c', '''clang-check'' -analyze ''/path/to/root/src/feature/NiceHeader.h'' -p ''/path/to/root/build''']
(finished - exit code 0) ['/bin/bash', '-c', '''clang-check'' -analyze ''/path/to/root/src/feature/NiceHeader.h'' -p ''/path/to/root/build''']

<<<OUTPUT STARTS>>>
Skipping /path/to/root/src/feature/NiceHeader.h. Compile command not found.
<<<OUTPUT ENDS>>>
```

In the simple case of having header and implementation files in the same folders for everything, looking only for the baseneme in the case of header files in FindCompileArgs could fix the header issue.

w0rp:
There has to be some degree of doing something sensible which works for 80% of projects, and avoiding very strange things beyond that. It's probably sensible to find compilation flags for exact filename matches, or somewhat close filename matches, like `foo.hpp` using the flags for `foo.cpp`. Maybe some flags common to a whole project could be used for things which don't match.

Beyond that it's probably a matter of, "Your project is just too weird. Create a .vim file and source it to set the flags." There are plugins for sourcing project configuration files for that.

w0rp:
One of the unfortunate things about C++ and C is that despite lots of time being spent creating ISO standards, there are no standards for project layouts, configuration files, etc. (Maybe C++20 will actually define modules this time around.)

StoneLin0708:
How about parse compile_command.json file at ale_linters/cpp/clang.vim or ale_linters/cpp/gcc.vim

 ```
function! ale_linters#cpp#clang#GetCommand(buffer) abort
     let l:paths = ale#c#FindLocalHeaderPaths(a:buffer)
 
+    let l:cmds = ale#c#IncludeOptions(l:paths) . ale#Var(a:buffer, 'cpp_clang_options')
+
+    let l:compile_commands_json = s:GetBuildDirectory(a:buffer) . '/compile_commands.json'
+    if filereadable(l:compile_commands_json)
+        for l:elem in json_decode(readfile(l:compile_commands_json))
+            if ale#path#IsBufferPath(a:buffer, l:elem.file)
+                " ignore options
+                let l:skip = 1
+                for l:arg in split(l:elem.command,' ')
+                    if l:skip == 1
+                        let l:skip = 0
+                    elseif l:arg ==# '-o'
+                        let l:skip = 1
+                    elseif l:arg ==# '-c'
+                        let l:skip = 1
+                    else
+                        let l:cmds .= l:arg . ' '
+                    endif
+                endfor
+              endif
+        endfor
+    endif
+
     " -iquote with the directory the file is in makes #include work for
     "  headers in the same directory.
     return ale#Escape(ale_linters#cpp#clang#GetExecutable(a:buffer))
     \   . ' -S -x c++ -fsyntax-only '
     \   . '-iquote ' . ale#Escape(fnamemodify(bufname(a:buffer), ':p:h')) . ' '
-    \   . ale#c#IncludeOptions(l:paths)
-    \   . ale#Var(a:buffer, 'cpp_clang_options') . ' -'
+    \   . l:cmds . ' -'
 endfunction
```

[patch_to_cpp_clang.txt](https://github.com/w0rp/ale/files/1361654/compile_command_json_for_clang.txt)

w0rp:
The intention of the function being developed here is that it can be used anywhere.

w0rp:
Closed due to inactivity.

PR #729, opened 2017-07-05
Python `virtualenv` finder behaves badly on Windows

dandersson:
### Summary
This is a quick fix to collapse the initial path component `//` into `/` to avoid freezing Vim for several seconds when running Ale in an Windows+MSYS2 environment, also fixing potential (though exotic) issues on other systems.

I end with some thoughts on a more complete solution to fix similar issues in a few other places in the code.

### Problem description
From 65fbf1c and onwards, Vim freezes for a couple of seconds every time Ale triggers Pylint when editing Python files under Windows in an [MSYS2](http://www.msys2.org/) environment. This is the standard environment that e.g. a standard Git installation on Windows uses by default.

The `virtualenv` discovery code in `autoload/ale/python.vim` currently loops through some usual suspects (configurable via `g:ale_virtualenv_dir_names`) when naming virtual environments such as `env`, `.env`, etc., within parent directories of the buffer being edited, and in particular checks if `{virtualenv dir}/bin/activate` exists.

This is very "*nix only" since virtual environments created on Windows actually calls the `bin` subdirectory `Scripts` (why this was chosen boggles the mind, but nevertheless ([`virtualenv`](https://virtualenv.pypa.io/en/stable/userguide/#activate-script), [`venv`](https://docs.python.org/3/library/venv.html#creating-virtual-environments)), which means that the code will never work when a virtual environment has been created with a Windows-native Python installation.

This would just have been a transparent degradation in function with a minimal time penalty, but the `ale#path#Upwards` function returns a list with the `/` entry when running in an MSYS2 environment on Windows. After concatenating paths, `ale#Python#FindVirtualenv` will then check readability of `//env/bin/activate`, `//.env/bin/activate` and so on. The meaning of the `//` path within MSYS2 on Windows is unclear to me, but it might mean the network root or something (analogous to `\\`) -- in any case lookups on `//` are _really_ slow:

    $ time ls //
    ls: reading directory '//': No such file or directory

    real    0m7.561s
    user    0m0.530s
    sys     0m2.324s

So, every attempt at finding a virtual environment on Windows within an MSYS2 environment (which will never succeed using Windows-native Python since scripts are in `Scripts`, not in `bin`) adds a couple of seconds of freezing time for the current Vim session. The default list of possible paths has five entries, which multiplies the issue.

For completeness sake, running it via e.g. GVim on Windows will yield a list of parent directories using Windows-style paths and fail at finding virtual environments as well due to the `Scripts`/`bin` issue, but at least it seems to fail fast.

### Quick workaround
For anyone encountering this, not running with this pull request applied: define `g:ale_virtualenv_dir_names` to be an empty list. In your `vimrc`:

    let g:ale_virtualenv_dir_names = []

This in practice inactivates the virtual environment finding mechanism altogether by looping over an empty list.

### Solution thoughts
One thought would be to not enable the virtual environment finder by default, but I am not sure whether users would want this. Personally I tend to always launch Vim from a session where the project virtual environment is already active which is then inheritied by Pylint if I understand it correctly, but I can imagine that if I within the same Vim session would open a file from another project, this would tend to give incorrect warnings.

Making the path joining smarter than just "concatenate base path and `/bin/activate`" to instead immediately collapse multiple slashes would also solve the freeze issue in a way, I guess, rather by accident. Technically, [a leading `//` is not necessarily equivalent to a leading `/` within POSIX](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap04.html#tag_04_11) ("A pathname that begins with two successive slashes may be interpreted in an implementation-defined manner") and in this case the code really wants to find a single leading `/`, so this might be a good thing anyway.

This pull request attempts to fix this inline for the specific place that matters for the Pylint issue. However, similar code exists in at least another place in `autoload/ale/python.vim` as well as in `autoload/ale/c.vim`, so either one should duplicate this inline fix in those places or perhaps it should rather be done through a more robust new `os.path.join`-like utility function in `autoload/ale/path.vim`.

To make the virtual environment search work for Windows-native Python installations as well, the finder function should also have to check for `{virtualenv dir}/Scripts/activate`. This would currently break other assumptions made in

* `autoload/ale/fixers/mypy`
* `autoload/ale/fixers/autopep8`
* `autoload/ale/fixers/isort`
* `autoload/ale/fixers/yapf`

that also assume that scripts within virtual environments reside in the `{virtualenv dir}/bin` directory. This has not been addressed at all in this merge request.

### DIFF
--- a/autoload/ale/python.vim
+++ b/autoload/ale/python.vim
@@ -27,7 +27,7 @@ endfunction
 function! ale#python#FindVirtualenv(buffer) abort
     for l:path in ale#path#Upwards(expand('#' . a:buffer . ':p:h'))
         for l:dirname in ale#Var(a:buffer, 'virtualenv_dir_names')
-            let l:venv_dir = l:path . '/' . l:dirname
+            let l:venv_dir = l:path =~# '/$' ? l:path . l:dirname : l:path . '/' . l:dirname
 
             if filereadable(l:venv_dir . '/bin/activate')
                 return l:venv_dir


w0rp:
You didn't need to write a novel in the pull request description.

w0rp:
I recommend putting a `has` check at the top of the function and making the function return an empty string for Windows for now.

autoload/ale/python.vim
```diff
@@ -27,7 +27,7 @@ endfunction
 function! ale#python#FindVirtualenv(buffer) abort
     for l:path in ale#path#Upwards(expand('#' . a:buffer . ':p:h'))
         for l:dirname in ale#Var(a:buffer, 'virtualenv_dir_names')
-            let l:venv_dir = l:path . '/' . l:dirname
```
w0rp: If you want to convert `//` to `/`, use the `simplify` function. See `:help simplify()`.

w0rp:
```vim
let l:venv_dir = simplify(l:path . '/' . l:dirname)
```

w0rp:
I fixed this by using `simplify`.

w0rp:
Now it searches for executables in the `Scripts` directories on Windows.

w0rp:
It might need some tweaking, as `has('unix')` could return 1 with hybrid Unix and Windows installations of Vim.

dandersson:
Thanks for the quick response!

`simplify` does not help in this case, since it ignores any amount of leading path separators when collapsing paths ([seemingly by deliberate design](https://github.com/vim/vim/blob/master/src/tag.c#L3537)); e.g. `simplify('//env')` returns `//env`, which makes the issue trigger again.

dandersson:
(Forgot to mention @w0rp )

Regarding hybrid installations: I can verify that `has('unix')` returns `1` when using the Vim installation from MSYS2. To further complicate things, when using the MSYS2 native Python package (e.g. through `pacman --sync python2`), the virtual environments created will be *nix style with `bin` subdirectories, so more combinations are possible.

Supporting *nix style Python installations on *nix and Windows style Python installations on Windows seems like the most reasonable approach unless one always want to check for both `bin` and `Scripts` every time, though.

w0rp:
`simplify('foo//bar')` returns `'foo/bar'`. The leading path component isn't an empty string.

dandersson:
@w0rp 
The leading path component is not empty, but it is a single slash when checking for virtual environments in the system root. The problem occurs since

* `ale#path#Upwards(expand('#' . a:buffer . ':p:h'))` returns a list that contains a `'/'` element (reasonable)
* this becomes `l:path` when looping over parent directories (reasonable)
* when concatenating `l:path . '/' . l:dirname`, where `l:dirname` is e.g. `'env'`, the resulting string is `'//env'` (this is where the problems start)
* `simplify('//env')` just returns `'//env'`, since `simplify` does not act on repeated initial slashes
* the `filereadable` function checks if `//env/bin/activate` (once again `simplify` does not take care of the initial slash) is readable, which triggers the file system search on `//`, which is what hangs the Vim process when running from MSYS2.

---

To verify, if I add

            echo simplify(l:venv_dir . '/' . s:bin_dir . '/activate')

just before the `if filereadable(simplify(l:venv_dir . '/' . s:bin_dir . '/activate'))` check in `autoload/ale/python.vim::ale#python#FindVirtualenv` using the current commit on master (a04e73ddbcbc2371424644145941f3878f894de8), whenever Ale wants to trigger Pylint from Vim in MSYS2 I can see prints of

* `//.env/bin/activate`
* `//env/bin/activate`
* `//ve-py3/bin/activate`
* `//ve/bin/activate`
* `//virtualenv/bin/activate`

with a delay of a second or so between every printout, with Vim unresponsive until the list has been run through (after which Pylint runs and everything is fine).

w0rp:
Okay. Now I have banned `simplify()` from the codebase, in favour of a wrapper function which changes a leading `//` to `/`.

dandersson:
@w0rp 
Great! I am no longer at work so I cannot immediately test this on Windows, but it seems like 1bd9b0fbe2b43dfa6716fcbf4cee5dfa4f6b9f1c should fix everything. Thanks again!

PR #731, opened 2017-07-05
Add ale_fix_on_save and ale_fixers to debug output

gregeinfrank:
While trying to debug why I can't get fixers to work - I noticed that these global variables weren't included in the output of `ALEInfo`


w0rp:
Cheers! :beers:

Issue #724, opened 2017-07-04
Rust linter does not work on windows


daa84:
I'm no sure but looks like file path compare is not working on windows because path contains backslash characters.

https://github.com/w0rp/ale/blob/master/autoload/ale/handlers/rust.vim#L51
```rust
a:full_filename =~ (l:span.file_name . '$')
```

ALEInfo output filtered, as to much out data
```

 Current Filetype: rust
Available Linters: ['cargo', 'rustc']
  Enabled Linters: ['cargo']
 Linter Variables:

let g:ale_rust_cargo_use_check = 1
 Global Variables:

let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%s'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'always'
let g:ale_linter_aliases = {}
let g:ale_linters = {}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_column_always = 0
let g:ale_sign_error = '>>'
let g:ale_sign_offset = 1000000
let g:ale_sign_warning = '--'
let g:ale_statusline_format = ['%d error(s)', '%d warning(s)', 'OK']
let g:ale_warn_about_trailing_whitespace = 1
  Command History:

(finished - exit code 0) 'cmd /c cargo check --frozen --message-format=json -q'

<<<OUTPUT STARTS>>>
...
[...]

PR #736, opened 2017-07-07
Fix windows path check on rust linter

daa84:
Fix for #724 
I'm not sure it is good solution, but for me it works.

### DIFF
--- a/autoload/ale/handlers/rust.vim
+++ b/autoload/ale/handlers/rust.vim
@@ -22,7 +22,6 @@ endfunction
 
 " A handler function which accepts a file name, to make unit testing easier.
 function! ale#handlers#rust#HandleRustErrorsForFile(buffer, full_filename, lines) abort
-    let l:filename = fnamemodify(a:full_filename, ':t')
     let l:output = []
 
     for l:errorline in a:lines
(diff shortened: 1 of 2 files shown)


autoload/ale/handlers/rust.vim
```diff
@@ -46,9 +46,11 @@ function! ale#handlers#rust#HandleRustErrorsForFile(buffer, full_filename, lines
         endif
 
         for l:span in l:error.spans
+            " substitue used for window file paths
+            let l:file_name = substitute(l:span.file_name, '\\', '\\\\', 'g')
             if (
             \   l:span.is_primary
-            \   && (a:full_filename =~ (l:span.file_name . '$') || l:span.file_name ==# '<anon>')
+            \   && (a:full_filename =~ (l:file_name . '$') || l:span.file_name ==# '<anon>')
```
w0rp: Could you try using `ale#path#IsBufferPath` for this instead? I'm trying to use the one function for handling everything. That's step 1 towards supporting setting errors from other files in quickfix. Step two is outputting `'filename': whatever` in the Dictionary objects, and then checking the filename returned automatically.

daa84:
I replace code with this function and for me it works but find problem with this when run tests on windows - tests does not pass because ale#test#SetFilename('src/playpen.rs') converts "/" to "\" in this case

w0rp:
@daa84 That's okay. The tests don't run on Windows at the moment. I'm planning on running the tests on AppVeyor eventually.

w0rp:
Cheers! :beers:

PR #737, opened 2017-07-08
Use different reporter to support older versions of jscs

AGhost-7:
We use an older version of jscs at work (version 1). It doesn't have unix output so I added support for a different reporter.


ale_linters/javascript/jscs.vim
```diff
@@ -1,9 +1,63 @@
...
+
+    return l:command
+endfunction
+
+function! ale_linters#javascript#jscs#Handle(buffer, lines) abort
+    " Matches patterns looking like the following
+    "
+    " foobar.js: line 2, col 1, Expected indentation of 1 characters
+    "
+    let l:pattern = '^.*:\s\+line \(\d\+\),\s\+col\s\+\(\d\+\),\s\+\(.*\)$'
```
w0rp: If you hate using so many backslashes for special regular expression characters, you can also start the regular expression with `\v`.  See `:help magic`

ale_linters/javascript/jscs.vim
```diff
@@ -1,9 +1,63 @@
 " Author: Chris Kyrouac - https://github.com/fijshion
 " Description: jscs for JavaScript files
 
+call ale#Set('javascript_jscs_executable', 'jscs')
+call ale#Set('javascript_jscs_use_global', 0)
```
w0rp: Add some documentation entries for these variables. They just need to link back to the explanation of how the `node_modules` loading works like `eslint`.

ale_linters/javascript/jscs.vim
```diff
@@ -1,9 +1,63 @@
...
+
+function! ale_linters#javascript#jscs#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'javascript_jscs', [
+    \   'node_modules/.bin/jscs',
+    \])
+endfunction
+
+function! ale_linters#javascript#jscs#GetCommand(buffer) abort
```
w0rp: Could you add some tests for this function and the executable function? There are a few in the `command_callback` directory for other things.

AGhost-7:
Should be ready for another review!

w0rp:
Cheers! :beers:

PR #738, opened 2017-07-08
Use rubocop's JSON output format (resolves #339)

### LINKED ISSUE #339: Emit a warning if rubocop is using an unsupported version
It is not currently known what the oldest supported version of rubocop is, but a sufficiently old version of rubocop does not work with ALE. ALE should emit some kind of warning if rubocop is out of date.

This issue follows on from #287.

elebow:
Rubocop's emacs formatter seems to have changed format in some
not-so-ancient version. The JSON formatter should provide a more stable
interface than parsing lines with a regex.

The JSON formatter was introduced in mid-2013, so it should be safe to
assume available in any reasonably-modern environment. The oldest
currently-supported version of ruby (according to ruby-lang.org) was
not supported by rubocop until 2014.

### DIFF
--- a/ale_linters/ruby/rubocop.vim
+++ b/ale_linters/ruby/rubocop.vim
@@ -23,9 +23,11 @@ function! ale_linters#ruby#rubocop#Handle(buffer, lines) abort
     let l:output = []
 
     for l:error in l:errors['files'][0]['offenses']
+        let l:start_col = l:error['location']['column'] + 0
         call add(l:output, {
         \   'lnum': l:error['location']['line'] + 0,
-        \   'col': l:error['location']['column'] + 0,
+        \   'col': l:start_col,
+        \   'end_col': l:start_col + l:error['location']['length'] - 1,
         \   'text': l:error['message'],
         \   'type': ale_linters#ruby#rubocop#GetType(l:error['severity']),
         \})
--- a/ale_linters/ruby/rubocop.vim
+++ b/ale_linters/ruby/rubocop.vim
@@ -1,5 +1,5 @@
-" Author: ynonp - https://github.com/ynonp
-" Description: rubocop for Ruby files
+" Author: ynonp - https://github.com/ynonp, Eddie Lebow https://github.com/elebow
+" Description: RuboCop, a code style analyzer for Ruby files
 
 function! ale_linters#ruby#rubocop#GetCommand(buffer) abort
     let l:executable = ale#handlers#rubocop#GetExecutable(a:buffer)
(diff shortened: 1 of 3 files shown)


ale_linters/ruby/rubocop.vim
```diff
@@ -8,34 +8,49 @@ function! ale_linters#ruby#rubocop#GetCommand(buffer) abort
...
+        \   'col': l:error['location']['column'] + 0,
+        \   'text': l:error['message'],
+        \   'type': s:GetType(l:error['severity']),
         \})
     endfor
 
     return l:output
 endfunction
 
+function! s:GetType(severity) abort
```
w0rp: I recommend making this a global function with a the `ale_linters` naming convention, and adding unit tests for it.

ale_linters/ruby/rubocop.vim
```diff
@@ -8,34 +8,49 @@ function! ale_linters#ruby#rubocop#GetCommand(buffer) abort
...
+    let l:output = []
 
+    for l:error in l:errors['files'][0]['offenses']
         call add(l:output, {
-        \   'lnum': l:match[1] + 0,
-        \   'col': l:match[2] + 0,
-        \   'text': l:text,
-        \   'type': index(['F', 'E'], l:type) != -1 ? 'E' : 'W',
+        \   'lnum': l:error['location']['line'] + 0,
+        \   'col': l:error['location']['column'] + 0,
```
w0rp: While your're at it, does the JSON format include end column or line numbers? If so, you could include those here to improve the highlighting.

w0rp:
The `command_callback` tests need to be updated. If you have Docker installed on your system, you can use `./run-tests -q` to run the tests and see only the failing tests.

ale_linters/ruby/rubocop.vim
```diff
@@ -8,34 +8,51 @@ function! ale_linters#ruby#rubocop#GetCommand(buffer) abort
...
+      return 'W'
+    elseif a:severity ==? 'convention'
+      return 'W'
+    elseif a:severity ==? 'warning'
+      return 'W'
+    elseif a:severity ==? 'error'
+      return 'E'
+    elseif a:severity ==? 'fatal'
+      return 'E'
+    else
```
elebow: What is the best way to handle values Rubocop might return in the future?

w0rp: Get rid of the message and return `'E'`. Then the worst case scenario is that you'll set some things as errors instead.

Also, you can write your code like this.

```vim
if ...
    return ...
endif

if ...
    return ...
endif

return ...
```

w0rp: You can't echo a message here, as it will annoy people if there's some update to rubocop, and then they'll just turn the linter off.

w0rp:
I'll merge this and change that so I can close the bug for supporting old rubocop versions.

PR #742, opened 2017-07-09
Add new Haskell linter

jez:
This PR adds a new Haskell linter (based on GHC). As such, it can take advantage
of the pre-existing `HandleGHCFormat` function. Also, we update the docs.

- **Add stack-build linter for Haskell**

  The stack-build linter works better than the other two linters when you're
  working with an entire Haskell project. It builds the project entirely and
  reports any errors.

  The other two Haskell GHC linters only work on single files, which can
  result in spurious errors (for example, not being able to find imports).

- **Document all available Haskell linters**

### DIFF
--- a/ale_linters/haskell/ghc.vim
+++ b/ale_linters/haskell/ghc.vim
@@ -16,3 +16,12 @@ call ale#linter#Define('haskell', {
 \   'command': 'stack ghc -- -fno-code -v0 %t',
 \   'callback': 'ale#handlers#haskell#HandleGHCFormat',
 \})
+
+call ale#linter#Define('haskell', {
+\   'name': 'stack-build',
+\   'output_stream': 'stderr',
+\   'executable': 'stack',
+\   'command': 'stack build',
+\   'lint_file': 1,
+\   'callback': 'ale#handlers#haskell#HandleGHCFormat',
+\})
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ name. That seems to be the fairest way to arrange this table.
 | Go | [gofmt -e](https://golang.org/cmd/gofmt/), [go vet](https://golang.org/cmd/vet/), [golint](https://godoc.org/github.com/golang/lint), [gometalinter](https://github.com/alecthomas/gometalinter), [go build](https://golang.org/cmd/go/), [gosimple](https://github.com/dominikh/go-tools/tree/master/cmd/gosimple), [staticcheck](https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck) |
 | Haml | [haml-lint](https://github.com/brigade/haml-lint)
 | Handlebars | [ember-template-lint](https://github.com/rwjblue/ember-template-lint) |
-| Haskell | [ghc](https://www.haskell.org/ghc/), [ghc-mod](https://github.com/DanielG/ghc-mod), [hlint](https://hackage.haskell.org/package/hlint), [hdevtools](https://hackage.haskell.org/package/hdevtools) |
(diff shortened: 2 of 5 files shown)


jez:
After looking at it a little longer, I realized that the `stack-ghc-mod` linter is actually more in line with what I was going for:

- uses `stack` to correctly resolve project dependencies
- just *checks* the file, instead of actually building the whole thing like `stack-build` does.

That being said, I think it's valuable to have both. `stack-ghc-mod` requires an extra dependency (`ghc-mod`). `stack-build` has the nice side effect of also building your project (you can imagine that people might want to build their project on save). And above all, keeping `stack-build` has minimal maintenance cost, because it re-uses `HandleGHCFormat`.

TL;DR: `stack-ghc-mod` is more of what I wanted, but `stack-build` is still fine to add the new `stack-build` linter.

ale_linters/haskell/ghc.vim
```diff
@@ -16,3 +16,12 @@ call ale#linter#Define('haskell', {
 \   'command': 'stack ghc -- -fno-code -v0 %t',
 \   'callback': 'ale#handlers#haskell#HandleGHCFormat',
 \})
+
+call ale#linter#Define('haskell', {
```
w0rp: Could you move this to a separate file? The other two should probably also be in their own files.

jez:
> Will this generate binary files? That may surprise and annoy people.

Yes, this will. My reasoning why it wasn't a huge deal was that

- Stack's "big idea" is reproducible builds. Running `stack build` is as close to idempotent as you can get.
- ALE is async. Even if the code generation takes a long time, at least it's not blocking the UI.
- `stack-build` provides a nice alternative to `stack-ghc` and `stack-ghc-mod` in that it requires no external dev tools and works with an entire project.

Unfortunately, there's no support with stack to just typecheck: commercialhaskell/stack#977.

What do you think? Is it annoying enough that we shouldn't include it? Could we make it opt-in as opposed to on by default?

ale_linters/haskell/ghc.vim
jez: Done.

jez:
Should be ready for round 2, thanks!

w0rp:
I will summon Haskell users who have commented before.

@wizzup @azure-satellite @parsonsmatt @jwoudenberg @WhiteAbeLincoln @habibalamin Does anyone mind if a Haskell linter generates binary files?

habibalamin:
As long as it's stored in an OS-provided tempfolder, cleaned up ASAP, and is space-efficient (depending on the program being linted), I wouldn't mind too much. However, I also don't know if I *should* mind.

However, I'm not sure how a project-level linter would work when embedded in a text editor. Is it necessary to have? What can it do that a file-level linter can't?

habibalamin:
Actually, forget about storing it in a tempfolder. If it just builds the project in the same location `stack build` normally does, it shouldn't be an issue. I just don't want any stray copies of the built project laying around, but if it's building where I'm building, I'm okay with it.

w0rp:
If it builds into the regular build directory, which it should do, then it'll probably be fine.

habibalamin:
I don't have any objections, but someone more knowledgeable than me may is where I'll leave it. I'm happy with this PR being merged.

w0rp:
I'll merge this, and let people try it.

w0rp:
Cheers! :beers:

parsonsmatt:
This looks great to me! Running `build` is pretty typical. The only suggestion I would make is using `--fast` to build without optimizations, which is about twice as fast on my work project.

w0rp:
@jez Do you want to try using `--fast`? Sounds like a good idea.

jez:
Good idea. Added in #754.

PR #743, opened 2017-07-09
Correctly resolve temp files on macOS

jez:
On macOS, the `$TMPDIR` is in `/var`. However, `/var -> /private/var`.
This means that fully resolved temp filenames weren't always getting
checked against the proper prefix in `IsTempFile`.

This was affecting some of the Haskell plugins, though I'm sure it could
have affected any program that resolved past the generated `$TMPDIR/foo`
and wound up at a different filename.


w0rp:
Cheers! :beers:

PR #745, opened 2017-07-09
Build classpath/sourcepath for gradle.

pardom:
I'm a Vim/Vim Script/Ale noob, so please give me your constructive criticism. 😄 Building with all the classpaths and source dirs can be quite slow. Perhaps there is a faster way to build with included sources.

Changes:
- Adds `init.gradle` plugin/script to build dependency classpath in a similar format as the mvn task
- Returns classpath from root gradle project if `gradlew` or `gradle` are found and `pom.xml` is not.
- Returns sourcepath from by searching for root gradle project files.

### DIFF
--- a/ale_linters/kotlin/kotlinc.vim
+++ b/ale_linters/kotlin/kotlinc.vim
@@ -12,15 +12,19 @@ let g:ale_kotlin_kotlinc_module_filename = get(g:, 'ale_kotlin_kotlinc_module_fi
 let s:classpath_sep = has('unix') ? ':' : ';'
 
 function! ale_linters#kotlin#kotlinc#GetImportPaths(buffer) abort
-    " exec maven only if classpath is not set
+    " exec maven/gradle only if classpath is not set
     if ale#Var(a:buffer, 'kotlin_kotlinc_classpath') !=# ''
         return ''
     else
         let l:pom_path = ale#path#FindNearestFile(a:buffer, 'pom.xml')
-
         if !empty(l:pom_path) && executable('mvn')
             return ale#path#CdString(fnamemodify(l:pom_path, ':h'))
                         \ . 'mvn dependency:build-classpath'
+        endif
+
+        let l:classpath_command = ale#gradle#BuildClasspathCommand(a:buffer)
+        if !empty(l:classpath_command)
+            return l:classpath_command
         endif
 
         return ''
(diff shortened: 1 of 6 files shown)


ale_linters/kotlin/kotlinc.vim
```diff
@@ -23,6 +24,20 @@ function! ale_linters#kotlin#kotlinc#GetImportPaths(buffer) abort
...
         endif
 
+        let l:build_path = ale#path#FindNearestFile(a:buffer, 'build.gradle')
+
+        if !empty(l:build_path)
+            let l:gradlew_path = ale#path#FindNearestFile(a:buffer, 'gradlew')
+            
+            if !empty(l:gradlew_path)
+                return ale#path#CdString(fnamemodify(l:gradlew_path, ':h'))
+                            \ . './gradlew -I ' . s:script_path . '/init.gradle -q printClasspath'
```
w0rp: Write your line continuations like so.

```vim
foo
\   . bar
```

You can look through the codebase for examples.

pardom:
Just so I understand correctly before I continue, I should add a `gradle.vim` to `autoload/ale` and extract the logic I added to it. Should I also move `init.gradle` to that directory? I seems like a different directory would be more appropriate?

pardom:
Cool, could you direct me to some example tests that would help with my tests?

w0rp:
Have a look at files like `test/command_callback/test_brakeman_command_callback.vader` There a few files like that which change the filename to point to something inside of a directory whch contains some project files, and a then check that the command generated contains stuff based on the project the file is in. You can do the same when Gradle files are detected. You just need to `touch` some empty gradle files in the right places.

pardom:
For the `ale#gradle#FindExecutable` tests, I need to check for `gradle` in the `$PATH` via `executable('gradle')`. How can I temporarily set the `$PATH` for these tests?

pardom:
Ignore that comment ^.

pardom:
@w0rp A couple tests are failing because `ale#gradle#FindExecutable` relies on `executable('gradle')`. Can you offer a suggestion for putting `gradle` on the system env path for tests?

w0rp:
You could use `let $PATH` to change PATH in the test and then add a dummy file with an execute bit set for it. That would let you add some tests where `gradle` is or isn't executable.

pardom:
Perhaps I've done it incorrectly, but I think I've already done what you're suggesting to no effect. `test/gradle-test-files/gradle` is an executable (but 0-byte) file which is put onto the path here: https://github.com/w0rp/ale/pull/745/files#diff-a0b1fbb70b6ce5b008d549b81895012fR20

w0rp:
Try adding `test/gradle-test-files` to `PATH` instead.

pardom:
Ah, duh

pardom:
👍 Tests passing 👍 

9mxxgad

w0rp:
Cool! :+1: Looks good.

w0rp:
Cheers! :beers:

PR #751, opened 2017-07-11
Add rails_best_practices handler (resolves #655)

### LINKED ISSUE #655: support rails best practices gem
It would be really great to support [rails best practices](https://github.com/flyerhzm/rails_best_practices) as another linting tool for Ruby.

elebow:
This borrows heavily from the Brakeman and Rubocop handlers.

I left two questions below.

### DIFF
--- a/ale_linters/ruby/rails_best_practices.vim
+++ b/ale_linters/ruby/rails_best_practices.vim
@@ -1,0 +1,49 @@
+" Author: Eddie Lebow https://github.com/elebow
+" Description: rails_best_practices, a code metric tool for rails projects
+
+let g:ale_ruby_rails_best_practices_options =
+\   get(g:, 'ale_ruby_rails_best_practices_options', '')
+
+function! ale_linters#ruby#rails_best_practices#Handle(buffer, lines) abort
+    if len(a:lines) == 0
+        return []
+    endif
+
+    let l:result = json_decode(join(a:lines, ''))
+
+    let l:output = []
+
+    for l:warning in l:result
+        if !ale#path#IsBufferPath(a:buffer, l:warning.filename)
+          continue
+        endif
+
+        call add(l:output, {
+        \    'lnum': l:warning.line_number + 0,
+        \    'type': 'W',
+        \    'text': l:warning.message,
+        \})
+    endfor
+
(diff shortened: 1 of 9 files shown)


ale_linters/ruby/rails_best_practices.vim
```diff
@@ -0,0 +1,55 @@
...
+    \   : ''
+
+    let l:rails_root = ale#ruby#FindRailsRoot(a:buffer)
+
+    if l:rails_root ==? ''
+        return ''
+    endif
+
+    return ale#Escape(l:executable) . l:exec_args
+    \    . ' --silent -f json --output-file /dev/stdout '
```
elebow: Writing to `/dev/stdout` won't work on Windows, of course. Is there an equivalent mechanism?

If not, it may be worth an upstream PR to add a stdout option to the linter.

doc/ale-ruby.txt
```diff
@@ -14,6 +14,25 @@ g:ale_ruby_brakeman_options                       *g:ale_ruby_brakeman_options*
 
 
 ===============================================================================
+rails_best_practices                            *ale-ruby-rails_best_practices*
+
+g:ale_ruby_rails_best_practices_executable *g:ale_ruby_rails_best_practices_executable*
```
elebow: These lines are too long, but I'm not sure how to split them while satisfying the vim help syntax.

ale_linters/ruby/rails_best_practices.vim
w0rp: `CON` might work on Windows. Another option would be to write to `%t`, which ALE will replace with a temporary file, and then add `; cat %t` for Unix and `; type %t` for Windows. That might work.

doc/ale-ruby.txt
w0rp: Write your lines likes so instead:

```vim
g:foo
         *g:foo*
```

Only with the last character aligned to the right margin. `:ALEFixSuggest` will suggest a function for aligning your help tags to the right margin for you, which you can run with `:ALEFix`.

elebow:
That should address the feedback, but I don't have any way to test this on Windows.

PR #752, opened 2017-07-11
Brakeman: Remove unused cache var from tests

elebow:
Caching of the Rails application root was used in an initial version of the Brakeman handler, but was removed before merging. This must have slipped by.

Also fixes a typo in a comment elsewhere.


w0rp:
Yep. Cheers! :beers:

PR #753, opened 2017-07-11
Add better erlang linter

urbanserj:

### DIFF
--- a/ale_linters/erlang/erlc.erl
+++ b/ale_linters/erlang/erlc.erl
@@ -1,0 +1,116 @@
+#!/usr/bin/env escript
+
+% This is a simple linter for erlang. For compatibility with rebar3
+% erlc.erl specifies paths for libraties and other compiler options.
+
+main([Orig, Buf]) ->
+    case file_type(Buf) of
+        escript ->
+            Port = open_port(
+                {spawn_executable, "/usr/bin/env"},
+                [exit_status, nouse_stdio, {args, ["escript", "-s", Buf]}]),
+            receive
+                {Port, {exit_status, Status}} ->
+                    halt(Status)
+            end;
+        module ->
+            Default = [strong_validation, report],
+            {ProjectDir, OptsDirs} = erl_opts_i(Orig),
+            Opts = Default ++ OptsDirs ++ erl_opts(ProjectDir),
+            case compile:file(Buf, Opts) of
+                {ok, _} -> halt(0);
+                error -> halt(1)
+            end
+    end.
+
+-spec file_type(string()) -> module | escript.
+file_type(File) ->
+    {ok, Fd} = file:open(File, [raw, read]),
+    try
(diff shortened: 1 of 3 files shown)


ale_linters/erlang/erlc.erl
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env escript
```
w0rp: What is this and what does it even do? There aren't any comments anywhere.

urbanserj: Sorry, I should’ve explained this earlier.

Erlang’s ecosystem doesn’t have a linker and each project has to implement it from scratch for its own needs (e.g. [1]). Erlc [2] is a compiler for Erlang, it knows nothing about build tooling that almost everyone uses (rebar3 [3][4]). You could think of it as gcc vs. cmake. It’s not possible to run rebar3 as a linter. So the script specifies paths for libraries and other compiler options.

[1] https://github.com/vim-erlang/vim-erlang-compiler/blob/master/compiler/erlang_check.erl
[2] http://erlang.org/doc/man/erlc.html
[3] http://www.rebar3.org
[4] https://github.com/erlang/rebar3

w0rp: Could you add some comments to the Erlang script to explain this?

w0rp:
I think you should add this as a new `rebar` linter, and then one linter or the other could be enabled or disabled as needed. Otherwise, changing the existing `erlc` linter will make it so some users who do not use `rebar3`, but do have a basic Erlang compiler, won't be able to check their files with ALE any more.

urbanserj:
Should we ask an opinion of the initial author @evnu?

These linters have completely the same functionality. It’s the same compiler, but the script is setting up paths to dependencies and other options defined in the config file. It works the same in the case when your project doesn’t use rebar3 as a build tool. @w0rp, do you still insist on making it as the separate linter for ALE?

evnu:
I like this implementation. My original implementation did not take `escript` into account, and "project support" using vim variables is probably a hazzle for people already using `rebar`. Note however that I'd still like the option to pass in custom parameters, as I for one do not use `rebar`. I usually use `localvimrc` to set up my project-specific configuration, which handles Erlang configuration for ale as well.

So, I'd suggest two changes to the linter presented here:

* Allow setting configuration with `g:ale_erlang_erlc_options` as before
* The module's file extension should be `.escript`

evnu:
After having updated ale, I  see that it has another Erlang linter `syntaxerl` (see https://github.com/w0rp/ale/commit/1aea6a34ff22c9e79d20a018b4d46000c208caf2). Without having checked `syntaxerl`, do the proposed changes here overlap with `syntaxerl`?

urbanserj:
Oh wow. I didn't know about this project! It works well for my needs. I think I can close my PR at the moment. Thank you very much!

PR #754, opened 2017-07-11
Prefer --fast for stack-build

jez:
Follows discussion in #742. Commit summary:

- **Vim scripts shouldn't have hyphens**

  Especially not ones that will be autoloaded. You can't have a hyphen in a
  function name, so autoloading functions based on filename will fail.

- **Add g:haskell_stack_build_options, default: --fast**

  If we're going to use the `--fast` option, we may as well go the whole 9
  yards and let the user configure the 'stack build' flags.

- **Create documentation for stack-build options**


w0rp:
Cheers! :beers:

PR #755, opened 2017-07-11
Look for ini file to spot python project root

bugant:
When looking for the project root folder it would be better to check for some well-known init file instead of __init__.py. Indeed, with python3 it is now possible to have namespace modules where intermediate dirs are not required to include the __init__.py file.


autoload/ale/python.vim
```diff
@@ -11,11 +11,26 @@ let g:ale_virtualenv_dir_names = get(g:, 'ale_virtualenv_dir_names', [
 \   'virtualenv',
 \])
 
+
+function! ale#python#FindProjectRootIni(buffer) abort
+    for l:path in ale#path#Upwards(expand('#' . a:buffer . ':p:h'))
+        if filereadable(l:path . 'MANIFEST.in') || filereadable(l:path . '/setup.cfg') || filereadable(l:path . '/pytest.ini') || filereadable(l:path . '/tox.ini')
```
w0rp: Break your or conditions over several lines with `\||`. Look at similar code.

autoload/ale/python.vim
```diff
@@ -11,11 +11,26 @@ let g:ale_virtualenv_dir_names = get(g:, 'ale_virtualenv_dir_names', [
...
+
 " Given a buffer number, find the project root directory for Python.
 " The root directory is defined as the first directory found while searching
 " upwards through paths, including the current directory, until a path
 " containing no __init__.py files is found.
 function! ale#python#FindProjectRoot(buffer) abort
+    let l:ini_root = ale#python#FindProjectRootIni(a:buffer)
+    if !empty(l:ini_root)
+      return l:ini_root
+    endif
```
w0rp: Add blank lines before and after if statements and so on that don't start a new block.

elebow:
The comment above `ale#python#FindProjectRoot()` should be updated to mention the new behavior.

PR #756, opened 2017-07-11
Make gometalinter work again

svanharmelen:
They changed their logic to use related paths instead of absoluut paths (see [here](https://github.com/alecthomas/gometalinter/commit/a04df08be5899be3c7ad69aa379858dcb660c709#diff-04424ed7c660c10495a54e8d11be89eaR253))

This fixes the linter by also using relative paths…

### DIFF
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -10,11 +10,11 @@ endfunction
 
 function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:executable = ale_linters#go#gometalinter#GetExecutable(a:buffer)
-    let l:filename = expand('#' . a:buffer . ':p')
+    let l:filename = expand('#' . a:buffer)
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
 
     return ale#Escape(l:executable)
-    \   . ' --include=''^' . l:filename . '.*$'''
+    \   . ' --include=' . ale#Escape(l:filename)
     \   . (!empty(l:options) ? ' ' . l:options : '')
     \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
 endfunction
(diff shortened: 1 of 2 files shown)


ale_linters/go/gometalinter.vim
```diff
@@ -10,11 +10,11 @@ endfunction
 
 function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:executable = ale_linters#go#gometalinter#GetExecutable(a:buffer)
-    let l:filename = expand('#' . a:buffer . ':p')
+    let l:filename = expand('#' . a:buffer)
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
 
     return ale#Escape(l:executable)
-    \   . ' --include=''^' . l:filename . '.*$'''
+    \   . ' --include=''' . l:filename . ''''
```
w0rp: Could you change this to `' --include=' . ale#Escape(l:filename)`? If the filename contains single quotes, it won't be escaped properly here.

`--include` takes a regular expression, doesn't it? We should escape the characters in the filename. `\` and `.` will appear pretty commonly, and have special meaning in regex.

svanharmelen:
Updated to use `ale#Escape(l:filename)` instead.

About the regex part... I guess it's more pure to escape everything properly, but I doubt if it's really needed as I never saw an issue about the current value passed to the `--include` flag.

Not here and not with `vim-go`, so it feels like fixing something that isn't actually broken... If you insist on adding the escaping, I need some help in doing so as I don't know how that should be done in a nice way in vimscript.

w0rp:
The escaping will probably be needed for Windows support, where paths mostly use backslashes.

w0rp:
Use something like this:

```vim
l:path_regex = substitute(l:path, '\([.\\]\)', '\\\1', 'g')
```

Just add more characters than just `.` and `\`, put it in a function, and test that function with Vader tests.

w0rp:
I'll take this and sort the regex escaping out.

svanharmelen:
I did have look at it, but didn't manage to get a working test in place before I needed to continue with my day job.... So thanks for getting this fix in!

svanharmelen:
For anyone finding this issue and/or having issues with `gometalinter`, please first run `:GoUpdateBinaries` within [n]vim or `go get -u github.com/alecthomas/gometalinter` within your terminal. Most likely that will fix your problems...

PR #761, opened 2017-07-12
Fixup #756 (gometalinter)

svanharmelen:
The real fix was not using absolute paths anymore (so not expanding with the `:p` option).

The regex was correct and should at least include the `^` character to make sure the string starts with the given path/filename and not any possible references to the path/filename in error descriptions.


w0rp:
Cheers! :beers:

svanharmelen:
Thanks!

svanharmelen:
For anyone finding this issue and/or having issues with `gometalinter`, please first run `:GoUpdateBinaries` within [n]vim or `go get -u github.com/alecthomas/gometalinter` within your terminal. Most likely that will fix your problems...

PR #766, opened 2017-07-12
Add scalastyle linter

okkays:
Hello!

I haven't touched vimscript or vader before, so sorry for any issues that come up from me re-purposing code without fully understanding it :sweat_smile: 

Adds support for the scalastyle linter (www.scalastyle.org).

Adds an option, `g:ale_scala_scalastyle_options = '-c scalastyle-config.xml'`, which allows custom options to the executable and provides a default config file name. (scalastyle requires `-c` be specified)

Has docs, and vader tests for the Handler and Command functions.  I _think_ I covered everything in the contributing page of the wiki.  At the very least, I tried.

Thanks for taking a look!


ale_linters/scala/scalastyle.vim
```diff
@@ -0,0 +1,46 @@
+" Author: Kevin Kays - https://github.com/okkays
+" Description: Support for the scalastyle checker.
+
+let g:ale_scala_scalastyle_options =
+\   get(g:, 'ale_scala_scalastyle_options', '-c scalastyle-config.xml')
```
w0rp: Is that a common configuration file name? You should try and search for it in parent directories, and make it automatic. Look at the JSHint linter or other linters for examples.

ale_linters/scala/scalastyle.vim
```diff
@@ -0,0 +1,46 @@
...
+
+function! ale_linters#scala#scalastyle#Handle(buffer, lines) abort
+    " Matches patterns line the following:
+    "
+    " warning file=/home/blurble/Doop.scala message=Missing or badly formed ScalaDoc: Extra @param foobles line=190
+
+    let l:patterns = [
+        \ '^\(.\+\) .\+ message=\(.\+\) line=\(\d\+\)$',
+        \ '^\(.\+\) .\+ message=\(.\+\) line=\(\d\+\) column=\(\d\+\)$',
+        \]
```
w0rp: Align the continued lines on the same level of indentation, but up-indent the lines inside of the List. Have a look through the codebase for examples.

ale_linters/scala/scalastyle.vim
```diff
@@ -0,0 +1,46 @@
...
+    for l:match in ale#util#GetMatches(a:lines, l:patterns)
+        let l:args = {
+        \   'lnum': l:match[3] + 0,
+        \   'type': l:match[1] =~? 'error' ? 'E' : 'W',
+        \   'text': l:match[2]
+        \ }
+
+        let l:col = l:match[4] + 0
+        if l:col > 0
+            let l:args['col'] = l:col + 1
```
w0rp: If `l:col` is `1`, then the value here will be `2`. I think these are off by one.

ale_linters/scala/scalastyle.vim
okkays: Yup, although looking into it, it seems scalastyle.org defaults to `scalastyle_config.xml` while something (possibly sbt) has caused other projects to use `scalastyle-config.xml`.  I've tried to account for both.

ale_linters/scala/scalastyle.vim
okkays: Good catch, fixed and tested.  Scalastyle outputs zero-based column numbers, so it's now checking for the presence of a column with `empty` and adding one to get to vim's one-based indices.

okkays:
Awesome, thanks!

w0rp:
Cheers! :beers:

PR #767, opened 2017-07-12
Rubocop: handle empty 'files' array in output

elebow:
The handler previously assumed there would be at least one entry in the
'files' array in the output JSON. It looks like this in the normal case:

    "files":[{"path":"app/models/image.rb","offenses":[]}]

But if RuboCop's config excludes the specified input files, causing no
files to be linted, the output is emptier:

    "files":[]

This change causes the handler to treat that case correctly, and also
exit early if the reported offense_count is zero.

### DIFF
--- a/ale_linters/ruby/rubocop.vim
+++ b/ale_linters/ruby/rubocop.vim
@@ -28,6 +28,10 @@ function! ale_linters#ruby#rubocop#Handle(buffer, lines) abort
         \}]
     endtry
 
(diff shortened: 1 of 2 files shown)


kriansa:
It looks like we're working on the same thing: #768 😄 

I also added a comment on the last issue you were also participant, maybe it's worth checking it as well: https://github.com/w0rp/ale/issues/760#issuecomment-314957246

w0rp:
Cheers! :beers:

PR #768, opened 2017-07-13
Fix Rubocop when linting a ignored file

kriansa:
When Rubocop is linting an ignored file, it will output something like this:

```json
{"metadata":{"rubocop_version":"0.48.1","ruby_engine":"ruby","ruby_version":"2.4.1","ruby_patchlevel":"111","ruby_platform":"x86_64-darwin16"},"files":[],"summary":{"offense_count":0,"target_file_count":0,"inspected_file_count":0}}
```

Currently, ale will throw an error because it expects that the `files` key has at least one element in it. I just added this protection condition so it won't try to loop through an invalid array.

This PR also reverts the behavior introduced by 5885954, which is undesirable: whenever rubocop outputs something to stderr, the linter would display an error message, even if the content of the stdout is valid.

For more info, also check my comment [here](https://github.com/w0rp/ale/issues/760#issuecomment-314957246).

### DIFF
--- a/ale_linters/ruby/rubocop.vim
+++ b/ale_linters/ruby/rubocop.vim
@@ -14,19 +14,11 @@ function! ale_linters#ruby#rubocop#GetCommand(buffer) abort
 endfunction
 
 function! ale_linters#ruby#rubocop#Handle(buffer, lines) abort
-    if empty(a:lines)
-        return []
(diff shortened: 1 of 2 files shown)


w0rp:
I'll take the other one, as this removes some code we want to keep.

PR #769, opened 2017-07-13
Do not complain if Rubocop is displaying warning messages on stderr

kriansa:
This will make Rubocop work even when there are warning messages being outputted to stderr, which does not impact the linting.

## Before:

screenshot 2017-07-13 13 43 53

It would stop the linting, even if there's a valid linter output.

## After:

screenshot 2017-07-13 13 43 39

Now it will disconsider any err outputted to stderr.

### DIFF
--- a/ale_linters/ruby/rubocop.vim
+++ b/ale_linters/ruby/rubocop.vim
@@ -59,5 +59,4 @@ call ale#linter#Define('ruby', {
 \   'executable_callback': 'ale#handlers#rubocop#GetExecutable',
 \   'command_callback': 'ale_linters#ruby#rubocop#GetCommand',
 \   'callback': 'ale_linters#ruby#rubocop#Handle',
-\   'output_stream': 'both',
 \})


w0rp:
Apply a regular expression to the message.

w0rp:
I have taken stderr output away for now, and I have covered more cases for output which doesn't contain problems we can parse.

If we want to reliably support information users about configuration errors, changes will have to be made so it's possible for two separate Lists to be passed to linters for stdout and stderr, instead of one combined List.

kriansa:
As always, thank you for your quick response @w0rp.

The way I see it is that while there's a valid output coming from the linter, there's no reason why we would hide it to the user, even if there's some warning message related to some misconfiguration.

Maybe in the future, we could have something similar to what Neovim has on `:CheckHealth`, so ale could spot potential issues related to some linter configuration or setup. What do you think?

w0rp:
I know that what I want to avoid is showing no problems for a file with no indication of what went wrong. For ESLint, it just prints a problem at line 1 and you can see the full stack trace in `:ALEDetail`. That works pretty well. It doesn't stop you editing the file like using `throw` or `echoerr` would, and you won't be unaware that your linter isn't doing what it should.

PR #772, opened 2017-07-14
Added phpstan linter for php.

ArdisDreelath:
Hi. This is my first attempt to contributing for vim plugin. 
I hope I have met all the requirements that you have imposed and will not do much work for you.
I mostly based on existing linters for phpcs and phpmd and created similar linter for [phpstan](https://github.com/phpstan/phpstan).

### DIFF
--- a/ale_linters/php/phpstan.vim
+++ b/ale_linters/php/phpstan.vim
@@ -16,7 +16,7 @@ endfunction
 function! ale_linters#php#phpstan#Handle(buffer, lines) abort
     " Matches against lines like the following:
     "
-    " filename.php:15:message 
+    " filename.php:15:message
     let l:pattern = '^.*\.php:\(\d\+\):\(.\+\)$'
     let l:output = []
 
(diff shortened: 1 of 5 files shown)


ale_linters/php/phpstan.vim
```diff
@@ -0,0 +1,39 @@
...
+    \   . ale#Var(a:buffer, 'php_phpstan_level')
+    \   . ' --errorFormat raw'
+    \   . ' %s'
+endfunction
+
+function! ale_linters#php#phpstan#Handle(buffer, lines) abort
+    " Matches against lines like the following:
+    "
+    " filename.php:15:message
+    let l:pattern = '^.*\.php:\(\d\+\):\(.\+\)$'
```
w0rp: I just realised. I used to work at a company where we named some files `.inc` if they were library files to be included in other files. You might want to use a regex where the file extension doesn't matter.

I recommend looking at the regex used for some of the C linters. You can write a regex which looks for a possible Windows drive letter at the start, and then anything except `:` after that, up to the `:` for the line number.

w0rp:
Cheers! :beers:

PR #777, opened 2017-07-18
Use new way of installing plugin with vim 8 in README

alkino:


README.md
```diff
@@ -195,24 +195,23 @@ into a bundle directory as with pathogen, and add the repository to your
...
 ```
 
-Then, modify your `~/.vimrc` file to add this plugin to your runtime path.
+Then, modify your `~/.vim/vimrc` file to add this plugin to your runtime path.
 
 ```vim
-set nocompatible
 filetype off
 
 let &runtimepath.=',~/.vim/bundle/ale'
```
w0rp: This isn't valid with the lines above.

w0rp:
Looking at `:help packloadall`, it does look like packages are loaded automatically, but after vimrc has been processed. For some reason, no packages load for me.

alkino:
I made an error, no need to modify vimrc. If you put your package in "start" folder, you don't need to "packadd" it.
Maybe check your "packpath".

mkarpoff:
There seems to be some confusion in how packages work in vim 8. So I've been using this plugin with vim8 for a few months The installation is rather simple for the most part. To make the plugin run every time it should just be placed in the 
```bash
$HOME/.vim/plug/<some name1>/start/<some name2>/
$HOME/.vim/plug/<some name1>/opt/<some name2>/
```
If the plugin is in the opt directory then you need add 
```vim
packadd <some name2>
```
 to your vimrc. If you placed it in the start director it will be loaded automatically. 

For example the configuration that I am using with ale and vim 8 I have the plugin cloned with this command
```bash
git clone https://github.com/w0rp/ale ~/.vim/plug/git-plugins/opt/ale
```
and I have the following in my vimrc
```vim
function! s:ale_settings()
    packadd ale
   
    let g:airline#extensions#ale#enabled = 1
    let g:ale_sign_column_always=1
    let g:ale_sign_error='E>'
    let g:ale_sign_warning='W>'
    let g:ale_change_sign_column_color=1
    let g:ale_echo_msg_format='[%linter%] [%severity%] %s'
  endfunction
```
This allows me to use auto commands to load ale if there is a filetype that I want to run ale on
Coincidently (shameless self plug) I wrote/am writing set of bash scripts for installing and updating vim8 plugins [here](https://github.com/mkarpoff/vim8-plug)

alkino:
What is weird is that I haven't any "plug" directory. I use packages as explained in the doc: http://vimhelp.appspot.com/repeat.txt.html#packages

w0rp:
Installing ALE as `~/.vim/pack/anything/start/ale` works. I'll update the instructions.

mkarpoff:
I think the reason why the confusion exists is that the vim8 plugin system was designed so that a plugin was build with parts that were load always and parts that were optionally loaded. This would have plugins all formate where components where placed in the start and opt directories. For example ale could be redesigned so that it was structured like so.
```bash
$ALE_BASE/start/plugin/ale.vim
$ALE_BASE/opt/<ale optional component>
```
Then ale could have components that would be added using `packadd <component name>`. In practice though every plugin I've looked at handles optional components internally thorugh boolean flags or toggle functions.

w0rp:
I have updated the README with the correct instructions for the `pack` directory now.

Issue #780, opened 2017-07-19
Add support for Standard fixer


sumnerevans:
The Standard linter for JavaScript has a `--fix` option that can be used to fix linter errors in ALE. Create a new fixer that takes advantage of this.

@w0rp, I'll submit a PR for this.

sumnerevans:
Resolved by #781.

PR #781, opened 2017-07-19
Added fixer for Standard linter

sumnerevans:
Addresses #780.

@w0rp, is there anywhere that the available fixers are listed?

### DIFF
--- a/autoload/ale/fixers/standard.vim
+++ b/autoload/ale/fixers/standard.vim
@@ -11,8 +11,16 @@ endfunction
 function! ale#fixers#standard#Fix(buffer) abort
     let l:executable = ale#fixers#standard#GetExecutable(a:buffer)
 
+    if ale#Has('win32') && l:executable =~? 'cmd\.js$'
+        " For Windows, if we detect an standard.js script, we need to execute
+        " it with node, or the file can be opened with a text editor.
+        let l:head = 'node ' . ale#Escape(l:executable)
+    else
+        let l:head = ale#Escape(l:executable)
+    endif
+
     return {
-    \   'command': ale#Escape(l:executable)
+    \   'command': l:head
     \       . ' --fix %t',
     \   'read_temporary_file': 1,
     \}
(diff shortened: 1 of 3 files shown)


autoload/ale/fixers/standard.vim
```diff
@@ -0,0 +1,19 @@
+" Author: Sumner Evans <sumner.evans98@gmail.com>
+" Description: Fixing files with Standard.
+
+function! ale#fixers#standard#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'javascript_standard', [
+    \   'node_modules/standard/bin/cmd.js',
```
w0rp: You should probably ensure that the `cmd.js` version is executed with `node` in the `Fix` function below on Windows. See the eslint function for an example.

w0rp:
@sumnerevans The tools are just listed in the table with the linters. `:ALEFixSuggest` suggests some things from the registry. We don't really need to explicitly mention anywhere that you can also fix files with certain things, it should basically be a given.

sumnerevans:
I'm in the process of adding the Vader tests (I've never done them before so it's a good learning opportunity :smile:). If all goes well, I'll push sometime today or tomorrow.

w0rp:
If you run into trouble installing Docker for running all of the tests, your problem is probably going to be that you need to add your user to a `docker` group and log out and in again. That's the one part of setting up Docker which isn't explained too well.

sumnerevans:
@w0rp, I actually already have Docker installed and I was happy to see how easy it was to run the tests. (I also use Linux which always helps things go smoother :wink:.)

w0rp:
Cool.

sumnerevans:
@w0rp, how does it look now? I added some stuff to the eslint test directory. Should I create a different directory instead? I could also rename the `eslint-test-files` folder to `js-test-files`.

w0rp:
The Fix function looks fine. I don't think you committed the tests, though.

sumnerevans:
@w0rp, whoops. I totally forgot to `git add .`. I updated and force pushed.

w0rp:
Cheers! :beers:

PR #784, opened 2017-07-20
Allow `perlcritic` linter notes to be shown as warnings

RsrchBoy:
Some of us prefer to take `perlcritic`, well, not literally.  We
introduce a new variable, `g:ale_perl_perlcritic_as_warnings`, that
allows one to change the default handler behaviour from undefined
(effectively shows as errors) to warnings.

### DIFF
--- a/ale_linters/perl/perlcritic.vim
+++ b/ale_linters/perl/perlcritic.vim
@@ -13,6 +13,8 @@ let g:ale_perl_perlcritic_options =
 let g:ale_perl_perlcritic_showrules =
 \   get(g:, 'ale_perl_perlcritic_showrules', 0)
 
+call ale#Set('perl_perlcritic_as_warnings', 0)
+
 function! ale_linters#perl#perlcritic#GetExecutable(buffer) abort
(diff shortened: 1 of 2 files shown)


elebow:
Since perlcritic is a style checker, not a syntax validator, I would argue for always classifying its output as warnings. Or at least, making warnings the default setting of this option.

RsrchBoy:
On Wed, 19 Jul 2017, Eddie Lebow wrote:

> Since perlcritic is a style checker, not a syntax validator, I would 
> argue for always classifying its output as warnings. Or at least, 
> making warnings the default setting of this option.

I would be OK with that -- I was erring on the side of caution in 
leaving the default behaviour intact.

-- 
Chris Weyl (aka rsrchboy at github, the CPAN, and irc.freenode.org)
Ex Astris, Scientia!
GPG: 9DB3 B4D1 3392 7169 0D89  A91C 1535 F82E 8083 A84A

w0rp:
This isn't necessary. There's an existing option for remapping problem types however you want. I'm on my phone, so I can't remember the variable name, but use `: help ale` and search for `map`. It might be named `ale_type_map`.

PR #788, opened 2017-07-21
added missing visual reselection after quick/location list update

adizero:

### DIFF
--- a/autoload/ale/list.vim
+++ b/autoload/ale/list.vim
@@ -39,6 +39,13 @@ function! ale#list#SetLists(buffer, loclist) abort
     if (g:ale_open_list && !empty(a:loclist)) || g:ale_keep_list_window_open
         let l:winnr = winnr()
 
+        let l:reselect = 0
+        if mode() ==? 'v' || mode() ==# "\<c-v>"
+            let l:reselect = 1
+        elseif mode() ==? 's' || mode() ==# "\<c-s>"
+            let l:reselect = 2
+        endif
+
         if g:ale_set_quickfix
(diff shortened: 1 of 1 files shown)


adizero:
Just start async background linting and play with the buffer in the meantime. Create a visual selection for example. When the background linting finishes and quickfix/location list gets updated (does not have to change) your visual selection will be lost. The patch fixes this problem by reselecting previously selected block if there was a selection in progress. The patch distinguishes different Vim modes like visual, visual-line,visual-block, select, select-line, select-block, and restores them properly.

w0rp:
I'll have a look at this. This might be reasonable. I'd use a couple of boolean variables instead of `0`,`1`, and `2`, maybe.

autoload/ale/list.vim
```diff
@@ -39,6 +39,13 @@ function! ale#list#SetLists(buffer, loclist) abort
     if (g:ale_open_list && !empty(a:loclist)) || g:ale_keep_list_window_open
         let l:winnr = winnr()
 
+        let l:reselect = 0
```
w0rp: Instead of the values `0`, `1`, and `2`, which don't really have any meaning, use a couple of Boolean variables which explain their purpose like `l:reset_visual_selection`.

w0rp:
I haven't merged this yet, as there are no Vader tests for this code. If I think of a way to easily add some tests for this, I'll add the tests and merge this code.

w0rp:
I don't know if it's possible to write a Vader test which checks visual selections, so I'll merge this now. Cheers! :beers:

Issue #782, opened 2017-07-19
[ale-cpp] Linting in headers don't work. Add option to run clang tools in source file.


phcerdan:
Clang tools, like clang-tidy, clang-check, don't show errors when run in headers, even with a compilation_database (the compilation database doesn't include a compile command for headers)
```
Skipping /path/to/header.h. Compile command not found.
```
This is not an ALE problem, but it would be great to add options for a workaround. One option would to run the tool instead of pointing to the current header, pointing to the source file (which is compiled) including the header file.

so:
```
clang-tidy -p /path/folder_compilation_database header.h
 Skipping /path/to/header.h. Compile command not found.
clang-tidy -p /path/folder_compilation_database source_with_header.cpp
... It works, show errors in header ...
```

w0rp:
We could consider creating temporary .c files for checking header files which include the header files. There are two problems with this.

1. We can't pass any additional arguments to clang tools when using compilation databases.
2. We don't know what a valid path to the header is. Is it `foo.h`? `include/foo.h`?

We could write the temporary files with a `.cpp` extension instead. That _might_ work. I don't know if either linters only run against the files on disk. If they do, then that's not an option.

It sounds like this problem should really be fixed in `clang-tidy` and `clang-check`.

phcerdan:
It is a long standing problem, even cmake, which generates the compilation_database knows nothing about headers. There is this [clangd](https://www.eclipse.org/forums/index.php/t/1082819/) (Language protocol server by clang) that might solve most  of these problems. But still on development.

One non-automated solution would be to let the user choose the file included in the compilation database which uses the header. YouCompleteMe does some sophisticated thing to guess headers, but still it doesn't work when the class in the header uses templates ( a lot of them). Pointing to a source file will solve this, because the template is instantiated there.

It is not an easy problem, and not the job of Ale to solve it, but it would be super nice to have something, with the flexibility of your plugin.

w0rp:
Could we just stop settings the `-p` options for `clang-tidy` and `clang-check` if the file extension is `.h` or `.hpp`? There are no compile commands for headers anyway.  Maybe there is some kind parsing that could be done like in #725 for getting the include paths from the `compile_commands.json` file.

w0rp:
This is all yet another reason to want modules to become part of the C++ standard.

phcerdan:
Sounds good, and the list for headers extensions should be flexible (there are .hxx, .txx, .tcc, ...).

[...]

PR #790, opened 2017-07-21
Workaround to get linting in cpp header files with clangtidy.

phcerdan:
Provide extra logic to handle header files in cpp.
Only using clangtidy, but same workaround will work in other clang checkers.

First, provide a list of headers suffixes:
By default:
let g:cpp_clangtidy_header_suffixes = ['h', 'hpp', 'hxx', 'tcc']
And then, use a variable specific for headers:
let b:cpp_clangtidy_header_sourcefile = ''

This variable should point to a source file (a .cpp) which uses the
current header.
This source file could be set by the user when editing the buffer, with
a simple:
let b:cpp_clangtidy_header_sourcefile

Or set a g:ale_pattern_options in a project vimL file.
let g:ale_pattern_options = {
      \   'my_favourite_header\.h$': {
      \       'ale_linters': {'cpp': ['clangtidy']},
      \       'ale_cpp_clangtidy_header_sourcefile': '/path/test_header.cpp',
      \   },
      \}

Related: #782

I have written vader tests, but not run them.

### DIFF
--- a/ale_linters/cpp/clangtidy.vim
+++ b/ale_linters/cpp/clangtidy.vim
@@ -1,5 +1,5 @@
 " Author: vdeurzen <tim@kompiler.org>, w0rp <devw0rp@gmail.com>,
-" gagbo <gagbobada@gmail.com>
+" gagbo <gagbobada@gmail.com>, phcerdan <pablo.hernandez.cerdan@outlook.com>
 " Description: clang-tidy linter for cpp files
 
 call ale#Set('cpp_clangtidy_executable', 'clang-tidy')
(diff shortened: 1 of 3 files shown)


ale_linters/cpp/clangtidy.vim
```diff
@@ -9,19 +9,27 @@ call ale#Set('cpp_clangtidy_checks', ['*'])
...
 endfunction
 
-function! s:GetBuildDirectory(buffer) abort
-    " Don't include build directory for header files, as compile_commands.json
-    " files don't consider headers to be translation units, and provide no
-    " commands for compiling header files.
-    if expand('#' . a:buffer) =~# '\v\.(h|hpp)$'
-        return ''
+function! ale_linters#cpp#clangtidy#IsHeader(buffer) abort
+    let l:search = '\v\.(' . join(ale#Var(a:buffer,'cpp_clangtidy_header_suffixes'), '|') . ')$'
```
w0rp: This isn't safe, as the option values could contain special regular expression characters. I think you'll be better off using an endswith check here.

```vim

for l:suffix in ale#Var(a:buffer, 'cpp_clangtidy_header_suffixes')
    let l:extension = '.' . l:suffix

    if l:filename[-len(l:extension):] ==# l:extension
        return 1
    endif
endfor

return 0
```

ale_linters/cpp/clangtidy.vim
```diff
@@ -35,17 +43,45 @@ endfunction
...
-
-    return ale#Escape(ale_linters#cpp#clangtidy#GetExecutable(a:buffer))
-    \   . (!empty(l:checks) ? ' -checks=' . ale#Escape(l:checks) : '')
-    \   . ' %s'
-    \   . (!empty(l:build_dir) ? ' -p ' . ale#Escape(l:build_dir) : '')
-    \   . (!empty(l:options) ? ' -- ' . l:options : '')
+                \   ? ale#Var(a:buffer, 'cpp_clangtidy_options')
+                \   : ''
+
+    if !ale_linters#cpp#clangtidy#IsHeader(a:buffer)
```
w0rp: I'm really not a fan of massive `if ... else` blocks like this. It's better if you can just progressively build options and return a command at the end, so the code just flows top to bottom.

ale_linters/cpp/clangtidy.vim
```diff
@@ -35,17 +43,45 @@ endfunction
...
+            \   . (!empty(l:options) ? ' -- ' . l:options : '')
+        " If header_sourcefile: Don't use the current buffer as input! The user has to
+        " provide a source file (.cpp) to run the checker on it when editing
+        " the header.
+        " The build_dir is kept if present, the provided source file must be
+        " in the database.
+        else
+            return ale#Escape(ale_linters#cpp#clangtidy#GetExecutable(a:buffer))
+            \   . (!empty(l:checks) ? ' -checks=' . ale#Escape(l:checks) : '')
+            \   . (!empty(l:header_sourcefile) ? ' ' . ale#Escape(l:header_sourcefile) : '')
```
w0rp: You could just write `'  ' . (!empty(l:header_sourcefile) ? ale#Escape(l:header_sourcefile) : '%s')` with the way the code was structured before.

phcerdan: true! thanks.

phcerdan:
Yeah, I understand.. I will use it for a while, and see. I work most of the time in headers with templates, where there is no much support, even with IDE's. Also compiling is slow, that why I really like the fast and low false positives of clang-tidy with ale.

I understand you don't want to maintaing or train users to use that. Thanks for the review.

PR #806, opened 2017-07-26
Update neovim to 0.2.0

RsrchBoy:
Neovim 0.2.0 is required for lambdas and partials.


w0rp:
We need to keep supporting the older version of NeoVim for now. I plan on requiring 0.2.0 as a minimum at some point in the future in ALE 2.0.0. That will allow me to use lambdas for many things, which will make life easier.

PR #807, opened 2017-07-26
Proselint and standard vars

RsrchBoy:
This pull request introduces two things:  the concept of a set of "standard linter variables" and functionality, and a central means to define a `proselint` linter.  Both of these come out of work I've been doing to more capably support running linters inside containers, yet are sufficiently generic so as to be generally useful.  This PR is intended to introduce these ideas, and open things up for discussion.

The standard functionality introduced is aimed at making certain common things -- for now, executable and options -- easy.  Using a function like `ale#linter#util#SetStandardVars()` ensures that every calling linter can be expected to have these variables defined, and makes it easy to introduce new variables (e.g. `...image`, `..._use_docker`, etc) without having to go back and touch each linter.

The standard vars introduced here are:

 * `g:ale_{ft}_{linter}_executable`
 * `g:ale_{ft}_{linter}_options`

These variables have been slowly making their way into a significant number of linters, completely independent of this PR.

Documentation of these variables is centralized -- no sense in copy and pasting the same docs everywhere.  Linters using this setup can refer to the general documentation from their specific help text; e.g. the proselint linter converted does just that.

Note that the neovim tests will fail until #806 is merged and the image rebuilt.

### DIFF
--- a/ale_linters/help/proselint.vim
+++ b/ale_linters/help/proselint.vim
@@ -1,9 +1,4 @@
 " Author: Daniel M. Capella https://github.com/polyzen
 " Description: proselint for Vim help files
 
-call ale#linter#Define('help', {
-\   'name': 'proselint',
-\   'executable': 'proselint',
-\   'command': 'proselint %t',
-\   'callback': 'ale#handlers#unix#HandleAsWarning',
-\})
+call ale#proselint#Define('help')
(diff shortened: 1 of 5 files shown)


w0rp:
No. There's no need to write some functions to wrap defining the options. Defining them is one of the easiest things to do. Documenting them requires more effort, and there's no way to avoid that, as the documentation is required.

There's no point in wrapping the existing code for defining linters in more generic code which won't have much use. I'd rather copy and paste the GetExecutable and GetCommand functions into different files. They typically don't remain so simple for very long. Many linters offer many different options and behave in various useful ways.

We can't use lambda syntax in the codebase yet, as NeoVim 0.1.7 is supported as a minimum, and will be supported until 0.20 becomes more common.

PR #811, opened 2017-07-28
(work in progress) Add support for write-good linter

### LINKED ISSUE #643: Add write-good
Linter for Markdown.

sumnerevans:
Resolves #643.

This is a work in progress, but I wanted some feedback on what I have so far, and I wanted to get an answer to a question (below).

**Things that I know still need to be done:**
- [ ] Vader tests
- [x] Documentation of options for write-good linter

**Question:**
Just like with `proselint`, the `write-good` linter can be applied to many file types. The problem is that, unlike `proselint`, the `write-good` linter allows the user to specify some options (more along the lines of the `eslint` linter).

Looking at the documentation, it appears that if a linter exposes options to the user, it needs to be documented in the *Integration Documentation* under the appropriate language(s). This is a problem, since `write-good` applies to multiple languages but does not have any configuration that changes across languages (this is unlike `gcc`, for example, which has different options for C/C++).

**Should I duplicate the documentation for all of the languages for which the `write-good` linter applies?** It seems messy (if anyone changes anything in the future, they'd have to edit multiple files), but at the same time, it seems to be most consistent with the current documentation structure.


w0rp:
If the options are to be configured the same for every instances of write-good at once, then create a single option for `write-good`, document the option in the global configuration settings, and link to the help tag for the global option from the `write-good` heading for each language.

w0rp:
```
write-good                                                *ale-html-write-good*

See |ale-write-good-options|
```

You can do that.

doc/ale.txt
```diff
@@ -946,6 +947,36 @@ ALEWarningSign                                                 *ALEWarningSign*
   The highlight used for warning signs. See |g:ale_set_signs|.
 
 
+-------------------------------------------------------------------------------
+3.2. Options for write-good Linter                     *ale-write-good-options*
```
w0rp: You should either write "options for write-good," or "options for the write-good linter." The latter sounds weird without the definite article. The former sounds better.

doc/ale.txt
```diff
@@ -946,6 +947,36 @@ ALEWarningSign                                                 *ALEWarningSign*
...
+
+The options for the write-good linter are global because it does not make
+sense to have them specified on a per-language basis.
```
w0rp: This explanation doesn't quite sound right. Instead of making excuses for why this is different, explain the advantages of configuring everything in one place instead.

autoload/ale/handlers/writegood.vim
```diff
@@ -0,0 +1,58 @@
...
+    endif
+
+    let l:options = ale#Var(a:buffer, 'writegood_options')
+
+    return l:head
+    \   . (!empty(l:options) ? ' ' . l:options : '')
+    \   . ' %t'
+endfunction
+
+function! s:HandleWriteGoodFormat(buffer, lines, type) abort
```
w0rp: You can move the contents of this function into the one below.

w0rp:
I just rebased this, squashed it, added tests for it, slapped your name on it, and merged it. It's now finally in there. :+1:

sumnerevans:
@w0rp , sorry. Thanks for picking this up. I started this over the summer and then school hit. :)

PR #820, opened 2017-08-02
Rubocop: Show cop name

junegunn:
This commit adds "cop name" to the formatted output of Rubocop handler. I find it useful when I want to suppress a subset of Rubocop warnings with magic comments, e.g. `# rubocop:disable Metrics/MethodLength`. When the cop name is included in the output, I don't have to search for the name that corresponds to the message.


junegunn:
Actually I wasn't sure about how I should format the message, so I followed the style of shellcheck.

```
Double quote to prevent globbing and word splitting. [SC2086]
```

I have no particular preference.

w0rp:
Cheers! :beers:

junegunn:
Thanks!

PR #821, opened 2017-08-02
Add Tcl nagelfar linter

verbitan:
I've added support for Tcl with [nagelfar](http://nagelfar.sourceforge.net).

It's my first time making a change to a Vim plugin, so any feedback welcome.


w0rp:
Cheers! :beers:

PR #836, opened 2017-08-08
Suppress warning about .eslintignore'd file

Firehed:
`eslint` currently issues a warning when a file has been excluded by `.eslintignore`. Ale sees this warning and renders it, which is somewhat pointless and annoying. This change suppresses that warning, which makes it easier to focus on actual lint errors on the file exposed by other linters, e.g. `tslint`

I don't really care for how fragile this implementation is, but I couldn't find any flags that I could pass to eslint (either directly or via its config file) to prevent them from showing up at all. Most suggestions I found on github and SO said to use `--quiet`, but that will remove *all* warnings which is not desirable.

### DIFF
--- a/autoload/ale/handlers/eslint.vim
+++ b/autoload/ale/handlers/eslint.vim
@@ -82,6 +82,10 @@ function! ale#handlers#eslint#Handle(buffer, lines) abort
         let l:type = 'Error'
         let l:text = l:match[3]
 
+        if l:text == 'File ignored because of a matching ignore pattern. Use "--no-ignore" to override.'
+            continue
+        endif
+
         " Take the error type from the output if available.
         if !empty(l:match[4])
             let l:type = split(l:match[4], '/')[0]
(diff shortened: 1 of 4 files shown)


Firehed:
Sounds good. I'll take another pass at this. Thanks for the feedback!

autoload/ale/handlers/eslint.vim
```diff
@@ -82,6 +82,10 @@ function! ale#handlers#eslint#Handle(buffer, lines) abort
         let l:type = 'Error'
         let l:text = l:match[3]
 
+        if l:text == 'File ignored because of a matching ignore pattern. Use "--no-ignore" to override.'
```
w0rp: You should install Vint for checking your code. It will complain about the use of `==`. I am also banning `==#` and `==?` from the codebase, as they are also a frequent source of bugs. I recommend using `x is# y` instead.

Firehed:
Just checking in on this - not sure if you saw my updates. I think they address all of the feedback you provided.

w0rp:
Yeah, looks fine to me.

w0rp:
Cheers! :beers:

PR #837, opened 2017-08-09
	 Default flake8 to --format=default

matthewgrossman:
EDIT: The solution to this problem has changed, scroll down further in this PR and look at the diffs

`let g:ale_python_flake8_options = '--format=pylint'` will cause Ale to not properly parse the error message. As an example, consider file asdf.py

```python
def asdf:
```

```
bash-3.2$ flake8 --format=pylint asdf.py
asdf.py:1: [E999] SyntaxError: invalid syntax
bash-3.2$ flake8 --format=default asdf.py
asdf.py:1:9: E999 SyntaxError: invalid syntax
```
The issue here is the `pylint` format will include the square brackets around the error number, e.g `[E999]` vs `E999`. The regex setup for parsing the code for flake8 won't be able to properly parse this and won't show the error message. I added an optional opening/closing square brackets in the `pattern` to allow for this regex to catch this type of format as well.

I'm still a bit iffy on the test cases

### DIFF
--- a/ale_linters/python/flake8.vim
+++ b/ale_linters/python/flake8.vim
@@ -80,6 +80,7 @@ function! ale_linters#python#flake8#GetCommand(buffer, version_output) abort
     \   : ''
 
     let l:options = ale#Var(a:buffer, 'python_flake8_options')
+    \   . ' --format=default'
 
     return ale#Escape(ale_linters#python#flake8#GetExecutable(a:buffer))
     \   . (!empty(l:options) ? ' ' . l:options : '')
--- a/test/command_callback/test_flake8_command_callback.vader
+++ b/test/command_callback/test_flake8_command_callback.vader
@@ -19,23 +19,23 @@ Execute(The flake8 callbacks should return the correct default values):
   \ '''flake8'' --version',
   \ ale_linters#python#flake8#VersionCheck(bufnr(''))
   AssertEqual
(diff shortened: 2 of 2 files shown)


w0rp:
I think we should set `--format=default` instead, explicitly. From the output above, it looks like the pylint format doesn't include the column number for errors, but the default format does.

w0rp:
Actually, the default is just the default. So just don't set those options.

matthewgrossman:
The reason this was an issue was because I happened to stumble into a repo where `format=pylint` was an option the `setup.cfg`. It took me a bit of debugging to realize why the flake8 output messages weren't showing up in the statusline/sidebar, as there was no error message indicating a failed parse in `:ALEInfo`.

I would be in favor of making `--format=default` to ensure that Ale works no matter what the repo's config is setup as.

w0rp:
Ah, I see. That make sense. In that case, I'll reopen this, and I suggest setting `--format=default` so it's always using the format ALE expects. You can modify the Vader tests to make sure it sets the format.

matthewgrossman:
@w0rp I got the tests passing, will squash and then it look good to merge?

w0rp:
Looks okay to me. I'll move the option to the string below and maybe get rid of extra spaces in the tests. Cheers! :beers:

PR #838, opened 2017-08-09
Add linter for Idris

bonds:


doc/ale-idris.txt
```diff
@@ -0,0 +1,10 @@
+===============================================================================
+ALE Idris Integration                                   *ale-idris-options*
+
+===============================================================================
+idris                                                   *ale-idris-stack-build*
```
w0rp: The help tag is wrong here. It should be `ale-idris-idris`, as in the table of contents.

doc/ale-idris.txt
```diff
@@ -0,0 +1,10 @@
+===============================================================================
+ALE Idris Integration                                   *ale-idris-options*
```
w0rp: Align the help tags to the right margin. If you use `:ALEFixSuggest`, it will suggest a function for fixing this.

ale_linters/idris/idris.vim
```diff
@@ -0,0 +1,72 @@
...
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+call ale#linter#Define('idris', {
+\   'name': 'idris',
+\   'executable': 'idris',
+\   'command': 'idris --total --warnpartial --warnreach --warnipkg --check %t',
```
w0rp: You should make the non-essential options configurable, and add Vader tests with a callback.

bonds:
Thank you for the feedback @w0rp. I believe I've addressed your concerns, please take another look.

w0rp:
Cheers! :beers:

PR #841, opened 2017-08-09
Rubocop fixer now uses g:ale_ruby_rubocop_options

### LINKED ISSUE #840: ale_ruby_rubocop_options used for linter, but ignored for fixer
I want to use the rubocop fixer, but this is currently preventing me from doing so.

I need to add this line: `let g:ale_ruby_rubocop_options = '--except Lint/Debugger'`

to prevent rubocop from complaining about `binding.pry` statements (which I used to debug the code)
I can't add this to .rubocop.yml itself, because this is only true during development (I don't want my CI to not catch these debug statements later on

Adding the flags above, the linter no longer complains about those lines. But enabling the fixer will still automatically remove them.

Taking a quick look at the source code, and with my limited knowledge, I see that this options variable doesn't seem to be used at all in the fixer

naps62:
This fixes #840, which I just reported

It turns out the fix was really simple. However, I'm unsure how to write tests for this (I'm not all that familiar with Vader). I can try to look into this again later (got no time at the moment), or if someone else wants to take a look at it I'd also appreciate it

### DIFF
--- a/autoload/ale/fixers/rubocop.vim
+++ b/autoload/ale/fixers/rubocop.vim
@@ -7,7 +7,8 @@ function! ale#fixers#rubocop#GetCommand(buffer) abort
 
     return ale#Escape(l:executable) . l:exec_args
     \   . (!empty(l:config) ? ' --config ' . ale#Escape(l:config) : '')
-    \   . ' --auto-correct %t'
+    \   . ' --auto-correct %t '
+    \   . ale#Var(a:buffer, 'ruby_rubocop_options')
 
 endfunction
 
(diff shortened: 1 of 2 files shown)


w0rp:
Try and fix the tests. Vader files are basically just Vim scripts with some commands for grouping other commands. If you have installed Docker correctly, you can run all of the tests with the `run-tests` script.

naps62:
@w0rp had some time to look into this today. Seems to be handled now

w0rp:
Cheers! :beers:

PR #842, opened 2017-08-09
Add support for SwiftFormat as a fixer

gfontenot:
SwiftFormat is a tool that can be used to format Swift files. This commit
adds support for using SwiftFormat as a fixer from ALE. It looks for
executables in the Pods directory, then the Pods directory for a React
Native project, then finally falls back to the globally installed instance
if neither of those were found.

https://github.com/nicklockwood/SwiftFormat


w0rp:
Cheers! :beers:

PR #846, opened 2017-08-12
Make ghc-mod linter check current buffer

voanhduy1512:
Right now ghc-mod linter check temp file instead of current buffer,
which cause the problem that it can't detect cabal file and raise
missing package error.

IP/s: try to look at the test but can't find ghc-mod specific test.

### DIFF
--- a/ale_linters/haskell/ghc-mod.vim
+++ b/ale_linters/haskell/ghc-mod.vim
@@ -4,13 +4,13 @@
 call ale#linter#Define('haskell', {
 \   'name': 'ghc-mod',
 \   'executable': 'ghc-mod',
-\   'command': 'ghc-mod check %t',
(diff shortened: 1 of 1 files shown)


voanhduy1512:
Looks like ghc-mod have a `map-file` option that `Redirect one file to another`or from `stdin` [source](https://github.com/DanielG/ghc-mod/blob/master/core/GhcMod/Options/Options.hs#L159). I assume with this i can implement 2nd option as your suggestion. I will update my PR soon.

w0rp:
Cool, that might do the trick. :+1: `flake8`, `mypy`, and `eslint` have similar options.

voanhduy1512:
Ready for review 🎉

w0rp:
Cheers! :beers:

PR #848, opened 2017-08-13
Fix bug with ALEPrevious for blank lines.

ilex:
ALEPrevious (as well as ALEPreviousWrap) does not work if there is an
error in empty line like 'W391: blank line at end of file' in python.


w0rp:
Cheers! :beers:

PR #854, opened 2017-08-16
Passthrough eslint config to prettier-eslint

### LINKED ISSUE #853: Prettier-eslint fixer not using eslintrc
regarding #683 

prettier-eslint was meant to infer it's options of eslint-config
however current `prettier_eslint.vim` fixer file lacks FindConfig function and passing that to prettier-eslint
which is currently blocked by prettier-eslint-cli [not having option to passthrough eslint config path](https://github.com/prettier/prettier-eslint-cli/blob/master/other/ROADMAP.md) (unlike eslint)
so that's why prettier-eslint fixer [fallbacks to its defaults](https://github.com/prettier/prettier-eslint/blob/master/README.md#technical-details)

just keeping this issue open as would try to provide pr later

morhetz:
closes #853 
- [x] depends on https://github.com/prettier/prettier-eslint-cli/pull/91
- [x] update the documentation to indicate the minimum supported version
- [x] fallback option (legacy)

### DIFF
--- a/autoload/ale/fixers/prettier_eslint.vim
+++ b/autoload/ale/fixers/prettier_eslint.vim
@@ -1,6 +1,25 @@
 " Author: tunnckoCore (Charlike Mike Reagent) <mameto2011@gmail.com>,
-"         w0rp <devw0rp@gmail.com>
+"         w0rp <devw0rp@gmail.com>, morhetz (Pavel Pertsev) <morhetz@gmail.com>
 " Description: Integration between Prettier and ESLint.
+
+function! s:FindConfig(buffer) abort
+    for l:filename in [
+    \   '.eslintrc.js',
+    \   '.eslintrc.yaml',
+    \   '.eslintrc.yml',
+    \   '.eslintrc.json',
+    \   '.eslintrc',
+    \   'package.json',
+    \]
+        let l:config = ale#path#FindNearestFile(a:buffer, l:filename)
+
+        if !empty(l:config)
+            return l:config
+        endif
+    endfor
+
+    return ''
+endfunction
 
 call ale#Set('javascript_prettier_eslint_executable', 'prettier-eslint')
 call ale#Set('javascript_prettier_eslint_use_global', 0)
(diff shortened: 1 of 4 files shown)


w0rp:
I'll close this for now, as no release of the tool contains this argument.

w0rp:
Oh, 4.2.0 does, as of less than half an hour ago.

I'll reopen this. What you should do is leave this off by default, and add an option for turning this behaviour on. Then edit the documentation to explain this. It can be turned on by default after enough people upgrade.

autoload/ale/fixers/prettier_eslint.vim
```diff
@@ -1,24 +1,46 @@
...
+    return ''
+endfunction
+
 call ale#Set('javascript_prettier_eslint_executable', 'prettier-eslint')
 call ale#Set('javascript_prettier_eslint_use_global', 0)
 call ale#Set('javascript_prettier_eslint_options', '')
 
 function! ale#fixers#prettier_eslint#GetExecutable(buffer) abort
     return ale#node#FindExecutable(a:buffer, 'javascript_prettier_eslint', [
-    \   'node_modules/prettier-eslint-cli/index.js',
```
w0rp: Are there older versions of `prettier-eslint-cli` with this path? If so, leave this in, but move it down the List.

morhetz: Nope. There were no version with this path

w0rp: Okay, cool. :+1:

morhetz:
@w0rp 
> What you should do is leave this off by default, and add an option for turning this behaviour on. Then edit the documentation to explain this. It can be turned on by default after enough people upgrade.

Hm, this fixer was never working as it supposed to due to wring exec path and lack of eslint config passing. Best case it was working just as plain prettier.js formatter havin no `esint --fix` functionality. So there are zero to none users of it.

Wouldn't it be better/easier to engage users to upgrade `prettier-eslint-cli` package?

morhetz:
Leaving this option `off` by default would still keep fixer broken

w0rp:
If you set the argument by default, then the integration will stop working for people who are using the plugin who haven't yet upgraded `prettier-eslint-cli`.

w0rp:
New options for any tool can't be used by default until some time has passed and people have the opportunity to upgrade first.

morhetz:
Integration is not working without that option

morhetz:
It's not just a new one, it's core functionality of prettier-eslint

w0rp:
Maybe an exception could be made here, as the configuration file is pretty important.

w0rp:
I think let's go with it. Update the documentation to indicate the minimum supported version, and cover the new argument with Vader tests.

w0rp:
I would rather there weren't so many tools for doing essentially the same thing. Personally, I wish that `standard`, `xo`, `prettier-standard`, and `prettier-eslint` didn't exist. `eslint` with `eslint-plugin-prettier` is probably the best option. Running `eslint --fix` followed by `prettier` is probably the second best option.

Still, some people like this, so it's worth supporting.

morhetz:
As far as I know `eslint-plugin-prettier` would require [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier) to disable all format conflicting rules on the `eslint` side
`prettier-eslint` gives opportunity to use your preffered code style with no change
more like a drop in replacement for the `eslint --fix` inferring formatting options of the `eslintrc`

morhetz:
@w0rp could you please review when you have a time?

w0rp:
I have added tests for this and merged this now. I also had to fix an issue where an empty configuration filename would sometimes be used.

w0rp:
Cheers! :beers:

PR #860, opened 2017-08-19
Add original filename to SyntaxErl command

dmitrivereshchagin:
This little change adds original filename to SyntaxErl command. It makes linting with SyntaxErl much more clever.

It's all thanks to @urbanserj's changes in SyntaxErl 😇

### DIFF
--- a/ale_linters/erlang/syntaxerl.vim
+++ b/ale_linters/erlang/syntaxerl.vim
@@ -10,7 +10,7 @@ endfunction
 
 
 function! ale_linters#erlang#syntaxerl#GetCommand(buffer) abort
-    return ale_linters#erlang#syntaxerl#GetExecutable(a:buffer) . ' %t'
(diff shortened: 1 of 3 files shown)


dmitrivereshchagin:
Done.

w0rp:
Cheers! :beers:

w0rp:
@dmitrivereshchagin Hmm, I had a look, and it doesn't look like there's a stable release which contains this argument. It's hard to tell because there's no changelog. We might have to make this optional and turn it off for now.

w0rp:
See here: https://github.com/w0rp/ale/commit/20e64fbae092e7307c4f01d82fa0e0082c3bb1cb

I made this optional and turned it off by default, so old versions will continue to work. The option can be turned on by default in future, after enough stable versions which support the option have been out for a while.

dmitrivereshchagin:
Oh. You are right. Thank you! 🍻

PR #861, opened 2017-08-19
Syntaxerl linter feature check

dmitrivereshchagin:
This replaces `erlang_syntaxerl_use_basename` option with simple feature check based on the help output. I've decided not to use version check since SyntaxErl has no `--version` flag and at the time code in master displays only the previously released version number.


w0rp:
Cheers! :beers:

PR #862, opened 2017-08-20
add phpcbf fixer

notomo:
phpcbf is PHP's code format tool.
It's in the same repository as phpcs.(ALE already supports)
https://github.com/squizlabs/PHP_CodeSniffer
This PR adds phpcbf support.


w0rp:
Cheers :beers:

PR #863, opened 2017-08-20
Added new linter for GraphQL: gqlint

happylinks:
I created a linter for GraphQL Schemas and use it with Ale.

The linter is fairly simple so I wasn't sure if tests where needed in this case. Also I wasn't sure if I should add the help-files, as I don't have configuration options, but I added them anyway for potential future graphql tools.

Let me know if I should change something :)


w0rp:
Cheers! :beers:

happylinks:
Wow, thanks a lot!

PR #865, opened 2017-08-21
Add erubis linter

jez:
This linter works largely the same as the existing `erubylint` linter,
except it works with `erubis` instead of `erb` as the driving command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ name. That seems to be the fairest way to arrange this table.
 | Dockerfile | [hadolint](https://github.com/lukasmartinelli/hadolint) |
 | Elixir | [credo](https://github.com/rrrene/credo), [dogma](https://github.com/lpil/dogma) |
 | Elm | [elm-make](https://github.com/elm-lang/elm-make) |
-| Erb | [erb](https://github.com/jeremyevans/erubi) |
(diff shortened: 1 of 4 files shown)


w0rp:
I wonder, is there a way to install executables for Ruby projects like how you install scripts in `node_modules` for Node projects? If so, it would be useful to include something for that for both flavours of eruby files, so different projects could automatically use the appropriate tool.

jez:
As I understand it, this is already the way it works with `rbenv`. `rbenv`
manipulates your shell so that the `erubis` or `erb` executable points to
whatever's local to your current project. The project-specific binaries are
located under a separate tree on your system, rather than within the
current directory as a `gems/` folder for example.

jez:
Maybe an example is best:

    ❯ which erubis
    /Users/jez/.rbenv/shims/erubis

So basically, `rbenv` links the `erubis` binary to this "shim" which is a
simple bash script that determines the actual `erubis` binary to run.

w0rp:
I see. It might be worth making ALE explicitly look for executables in `.rbenv` at some point. If you use GVim, it will be easy to run Vim without `rbenv`. Though you might set that up with a Vim plugin for loading rbenv.

w0rp:
Cheers! :beers:

PR #872, opened 2017-08-24
Add pycodestyle Python linter support

### LINKED ISSUE #867: Pycodestyle Linter Support
Pycodestyle linter support would be useful for me and my friends at work.

I've looked into using flake8 and setting it to only lint with pycodestyle, but it doesn't look like this is possible.

Work has already been done to implement this but tests/docs would need to be added etc.
https://github.com/dmerejkowsky/vim-ale/blob/master/ale_linters/python/pycodestyle.vim

micthiesen:
Looking for some feedback before I write tests. I based this code off of the existing Python linters. Luckily pycodestyle doesn't care from which directory it's run, so it wasn't too difficult. 

Closes #867

### DIFF
--- a/ale_linters/python/pycodestyle.vim
+++ b/ale_linters/python/pycodestyle.vim
@@ -26,12 +26,12 @@ function! ale_linters#python#pycodestyle#Handle(buffer, lines) abort
     " file.py:21:26: W291 trailing whitespace
     for l:match in ale#util#GetMatches(a:lines, l:pattern)
         call add(l:output, {
-    \   'filename': l:match[1],
-    \   'lnum': l:match[2] + 0,
-    \   'col': l:match[3] + 0,
-    \   'type': l:match[6],
-    \   'text': l:match[4],
-    \})
+        \   'filename': l:match[1],
+        \   'lnum': l:match[2] + 0,
+        \   'col': l:match[3] + 0,
+        \   'type': l:match[6],
+        \   'text': l:match[4],
+        \})
     endfor
 
     return l:output
--- a/ale_linters/python/pycodestyle.vim
+++ b/ale_linters/python/pycodestyle.vim
@@ -8,14 +8,6 @@ let g:ale_python_pycodestyle_options =
 let g:ale_python_pycodestyle_use_global =
 \   get(g:, 'ale_python_pycodestyle_use_global', 0)
 
-function! s:GetDir(buffer) abort
-    let l:project_root = ale#python#FindProjectRoot(a:buffer)
-
-    return !empty(l:project_root)
-    \   ? l:project_root
-    \   : expand('#' . a:buffer . ':p:h')
-endfunction
-
 function! ale_linters#python#pycodestyle#GetExecutable(buffer) abort
     return ale#python#FindExecutable(a:buffer, 'python_pycodestyle', ['pycodestyle'])
 endfunction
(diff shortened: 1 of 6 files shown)


micthiesen:
It looks like `pycodestyle` is only being run on save, which doesn't make sense as a default since it's so fast. Any idea why? Here's my settings:

```vimscript
 Current Filetype: python
Available Linters: ['flake8', 'mypy', 'pycodestyle', 'pylint']
  Enabled Linters: ['mypy', 'pycodestyle', 'pylint']
 Linter Variables:
let g:ale_python_mypy_executable = 'mypy'
let g:ale_python_mypy_options = ''
let g:ale_python_mypy_use_global = 0
let g:ale_python_pycodestyle_executable = 'pycodestyle'
let g:ale_python_pycodestyle_options = ''
let g:ale_python_pycodestyle_use_global = 0
let g:ale_python_pylint_executable = 'pylint'
let g:ale_python_pylint_options = ''
let g:ale_python_pylint_use_global = 0
 Global Variables:
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'E'
let g:ale_echo_msg_format = '[%severity%] [%linter%] %s'
let g:ale_echo_msg_warning_str = 'W'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_linter_aliases = {}
let g:ale_linters = {'json': ['jsonlint'], 'sh': ['shellcheck'], 'javascript': ['jshint', 'jscs'], 'python': ['mypy', 'pycodestyle', 'pylint']}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_error = '>>'
let g:ale_sign_warning = '--'
```

ale_linters/python/pycodestyle.vim
```diff
@@ -0,0 +1,46 @@
...
+
+function! ale_linters#python#pycodestyle#Handle(buffer, lines) abort
+    let l:pattern = '\v^(\S*):(\d*):(\d*): ((([EW])\d+) .*)$'
+    let l:output = []
+
+    " lines are formatted as follows:
+    " file.py:21:26: W291 trailing whitespace
+    for l:match in ale#util#GetMatches(a:lines, l:pattern)
+        call add(l:output, {
+        \   'filename': l:match[1],
```
w0rp: You might not need the filename here. I dunno, there might be some plugin that follows imports.

Leave it in. It can't do any harm.

ale_linters/python/pycodestyle.vim
```diff
@@ -0,0 +1,46 @@
...
+
+function! ale_linters#python#pycodestyle#GetExecutable(buffer) abort
+    return ale#python#FindExecutable(a:buffer, 'python_pycodestyle', ['pycodestyle'])
+endfunction
+
+function! ale_linters#python#pycodestyle#GetCommand(buffer) abort
+    return ale#Escape(ale_linters#python#pycodestyle#GetExecutable(a:buffer))
+    \   . ' '
+    \   . ale#Var(a:buffer, 'python_pycodestyle_options')
+    \   . ' %s'
```
micthiesen: I got `STDIN` working with `-` so I'll use that instead.

ale_linters/python/pycodestyle.vim
micthiesen: When using `STDIN` along with the `filename`, the errors weren't recognized in the editor. Removing `filename` fixed this issue.

ale_linters/python/pycodestyle.vim
```diff
@@ -0,0 +1,46 @@
+" Author: Michael Thiesen <micthiesen@gmail.com>
+" Description: pycodestyle linting for python files
+
+let g:ale_python_pycodestyle_executable =
+\   get(g:, 'ale_python_pycodestyle_executable', 'pycodestyle')
+let g:ale_python_pycodestyle_options =
+\   get(g:, 'ale_python_pycodestyle_options', '')
+let g:ale_python_pycodestyle_use_global =
+\   get(g:, 'ale_python_pycodestyle_use_global', 0)
```
micthiesen: I'm use that, much cleaner

ale_linters/python/pycodestyle.vim
w0rp: That will have been because the filename wasn't a filename any more. if a filename was needed, say for imported files, then the trick to that would be to only set the `filename` key when the matched name wasn't some special name for stdin.

micthiesen:
@w0rp I've made a few changes and added tests. Ready for a re-review.

ale_linters/python/pycodestyle.vim
micthiesen: Gotcha that makes sense

w0rp:
Cheers! :beers:

PR #873, opened 2017-08-24
Add clang-format fixer for C/C++

renstrom:
All the paths, variables and functions use "c" instead of "cpp". Not sure what's preferred, or if I should duplicate it?

https://clang.llvm.org/docs/ClangFormat.html

### DIFF
--- a/doc/ale-cpp.txt
+++ b/doc/ale-cpp.txt
@@ -190,4 +190,23 @@ g:ale_cpp_gcc_options                                   *g:ale_cpp_gcc_options*
 
 
 ===============================================================================
+clang-format                                              *ale-cpp-clangformat*
+
+g:ale_c_clangformat_executable                 *g:ale_c_clangformat_executable*
+                                               *b:ale_c_clangformat_executable*
+  Type: |String|
+  Default: `'clang-format'`
+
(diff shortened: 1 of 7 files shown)


renstrom:
Added documentation now (I knew I missed something!). Not entirely sure I did it correct though.

doc/ale-cpp.txt
```diff
@@ -190,4 +190,23 @@ g:ale_cpp_gcc_options                                   *g:ale_cpp_gcc_options*
 
 
 ===============================================================================
+clang-format                                              *ale-cpp-clangformat*
+
+g:ale_c_clangformat_executable                 *g:ale_c_clangformat_executable*
```
w0rp: I think here you can just write `See |ale-c-clangformat|` and a short message about the options being the same.

renstrom: Does this sound okay?

```
See |ale-c-clangformat| for information about the available options.
Note that the C options are also used for C++.
```

w0rp: Yeah, that sounds fine to me.

renstrom:
Done and done!

w0rp:
Cheers! :beers:

Issue #439, opened 2017-03-31
Request: CUDA linting


sunbubble:
CUDA is a C++ extension for GPU programming.
Modern versions of **Clang** can compile it, provided you also have the Nvidia CUDA-toolkit installed, which also includes a fully functioning compiler **NVCC**.
Clang warnings and error messages are, as expected, superior.

So I request support for both:
* NVCC
* Clang (http://llvm.org/docs/CompileCudaWithLLVM.html)

CUDA file extensions are *.cu and *.cuh (although for headers *.h and *.hpp are also common)

Thank you

w0rp:
If anyone has experience with this, feel free to open a pull request. (Especially one with documentation and Vader tests.)

lucaskolstad:
I made an attempt at this for both and there was an implementation issue for each:

1. `nvcc` doesn't seem to accept input through stdin through any of the usual ways. If someone has an example how to do this, if it can do it, that would be helpful as it is a non-starter otherwise. It's required for handlers in `ale`.

2. For the `clang++` option, it is complicated by management of the cuda paths, but could work. It works for me using just the quick example implementation in PR #442, but
that is because I have `cuda` symlinked to `/usr/local/cuda` on my machine from where it was installed by my package manager into `/opt/cuda`. `clang++` just detects it automatically in that case and can find the headers and such to do the linting.

    Any user who doesn't have `cuda` installed, or doesn't have their cuda paths set up properly, would have to pass options to `g:ale_cuda_clang_options` to fix that by hand. Those options would be things like `--cuda-path=/opt/cuda` or more directly with typical link and include options.
    
    However, the issue is that if a user doesn't have this set up properly we could not detect it and would not know to not run our `clang++` linter. All we check for before running it is that we are in a file with `ft == cuda` and that `clang++` is available. We would just get an invisible linter error every time in the background like this:

```
clang-3.9: error: cannot find CUDA installation. Provide its path via --cuda-path, or pass -nocudainc to build without CUDA includes.
```

Now, I am no CUDA expert so I am unsure how common this would be for people who use ale and open cuda files to not have their cuda paths set up. It's pretty ugly to know their is a case where we would waste CPU time asking `clang++` to fail to find the `cuda` installation over and over in the background.

lucaskolstad:
We could check a few likely spots for the installation like `/usr/local/cuda` and `/opt/cuda` and add `--cuda-path=/usr/local/cuda` if we find it. However that is not very robust for two reasons:

[...]

PR #874, opened 2017-08-24
Add cuda nvcc linter

blahgeek:
Add linter for CUDA files, using nvcc. #439 

I'm new to vim scripts and ALE, please let me know if I did anything wrong. Thanks

### DIFF
--- a/ale_linters/cuda/nvcc.vim
+++ b/ale_linters/cuda/nvcc.vim
@@ -9,10 +9,13 @@ function! ale_linters#cuda#nvcc#GetExecutable(buffer) abort
 endfunction
 
 function! ale_linters#cuda#nvcc#GetCommand(buffer) abort
+    let l:output_file = tempname() . ".ii"
+    call ale#engine#ManageFile(a:buffer, l:output_file)
     return ale#Escape(ale_linters#cuda#nvcc#GetExecutable(a:buffer))
     \   . ' -cuda '
     \   . ale#c#IncludeOptions(ale#c#FindLocalHeaderPaths(a:buffer))
     \   . ale#Var(a:buffer, 'cuda_nvcc_options') . ' %s'
+    \   . ' -o ' . l:output_file
 endfunction
 
 function! ale_linters#cuda#nvcc#HandleNVCCFormat(buffer, lines) abort
--- a/ale_linters/cuda/nvcc.vim
+++ b/ale_linters/cuda/nvcc.vim
@@ -9,14 +9,15 @@ function! ale_linters#cuda#nvcc#GetExecutable(buffer) abort
 endfunction
 
 function! ale_linters#cuda#nvcc#GetCommand(buffer) abort
-    let l:output_file = tempname() . '.ii'
-    call ale#engine#ManageFile(a:buffer, l:output_file)
+    " Unused: use ale#util#nul_file
+    " let l:output_file = tempname() . '.ii'
+    " call ale#engine#ManageFile(a:buffer, l:output_file)
 
     return ale#Escape(ale_linters#cuda#nvcc#GetExecutable(a:buffer))
     \   . ' -cuda '
     \   . ale#c#IncludeOptions(ale#c#FindLocalHeaderPaths(a:buffer))
     \   . ale#Var(a:buffer, 'cuda_nvcc_options') . ' %s'
-    \   . ' -o ' . l:output_file
+    \   . ' -o ' . g:ale#util#nul_file
 endfunction
 
 function! ale_linters#cuda#nvcc#HandleNVCCFormat(buffer, lines) abort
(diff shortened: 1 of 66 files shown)


ale_linters/cuda/nvcc.vim
```diff
@@ -0,0 +1,61 @@
...
+endfunction
+
+function! ale_linters#cuda#nvcc#GetCommand(buffer) abort
+    let l:output_file = tempname() . ".ii"
+    call ale#engine#ManageFile(a:buffer, l:output_file)
+    return ale#Escape(ale_linters#cuda#nvcc#GetExecutable(a:buffer))
+    \   . ' -cuda '
+    \   . ale#c#IncludeOptions(ale#c#FindLocalHeaderPaths(a:buffer))
+    \   . ale#Var(a:buffer, 'cuda_nvcc_options') . ' %s'
+    \   . ' -o ' . l:output_file
```
w0rp: I wonder, is there an option to cancel output, or output to a null file? Most tools like this have such an option.

ale_linters/cuda/nvcc.vim
```diff
@@ -0,0 +1,61 @@
...
+call ale#Set('cuda_nvcc_options', '-std=c++11')
+
+function! ale_linters#cuda#nvcc#GetExecutable(buffer) abort
+    return ale#Var(a:buffer, 'cuda_nvcc_executable')
+endfunction
+
+function! ale_linters#cuda#nvcc#GetCommand(buffer) abort
+    let l:output_file = tempname() . ".ii"
+    call ale#engine#ManageFile(a:buffer, l:output_file)
+    return ale#Escape(ale_linters#cuda#nvcc#GetExecutable(a:buffer))
```
w0rp: I recommend putting a blank line before `return`, `if`, `while`, or `for`, unless they start a new level of indentation.

ale_linters/cuda/nvcc.vim
```diff
@@ -0,0 +1,61 @@
...
+    " Look for lines like the following.
+    "
+    " test.cu(8): error: argument of type "void *" is incompatible with parameter of type "int *"
+    let l:pattern = '\v^([^:\(\)]+):?\(?(\d+)\)?:(\d+)?:?\s*\w*\s*(error|warning): (.+)$'
+    let l:output = []
+
+    for l:line in a:lines
+        let l:match = matchlist(l:line, l:pattern)
+
+        if !empty(l:match)
```
w0rp: You might like to use the `ale#util#GetMatches` function here.

w0rp:
Check out the Travis results. I recommend checking your VimL code with Vint. 

```
Running Vint to lint our code
Vint warnings/errors follow:
ale_linters/cuda/nvcc.vim:12:38: Prefer single quoted strings (see Google VimScript Style Guide (Strings))
```

You can escape single quotes in strings by doubling your single quotes. `'''foo''' is# "'foo'"`

ale_linters/cuda/nvcc.vim
blahgeek: If "-o xxx" is omitted, it would write output to current directory. It can output to the null (/dev/null), but it cannot work on windows, am i right?

w0rp: Does `-o nul` work on Windows? There's a variable in `ale#util` for giving you the path to the null file for each platform.

w0rp: I'd explore that option first, so no files can be generated at all. If that doesn't work out, then ALE has some tools for managing temporary files so they can be automatically deleted when linting is done, which were built for handling this kind of problem.

blahgeek: I've changed the code to use `ale#util#nul_file` and it works on linux, but I am unable to test on windows for now.

ale_linters/cuda/nvcc.vim
```diff
@@ -0,0 +1,59 @@
...
+    "
+    " test.cu(8): error: argument of type "void *" is incompatible with parameter of type "int *"
+    let l:pattern = '\v^([^:\(\)]+):?\(?(\d+)\)?:(\d+)?:?\s*\w*\s*(error|warning): (.+)$'
+    let l:output = []
+
+    for l:match in ale#util#GetMatches(a:lines, l:pattern)
+        " Ignore errors that is not for this file
+        if !(fnamemodify(l:match[1], ':p:.') is# bufname(bufnr('')))
+            continue
+        endif
```
w0rp: Instead of doing this, I recommend setting the `filename` key with the filename below, but please ensure that the an _absolute path_ to a filename is used. Then ALE will be able to show problems in other files in quickfix and so on. See `:help ale-loclist-format`.

blahgeek:
@w0rp done! :P

w0rp:
Cheers! :beers:

Issue #653, opened 2017-06-15
Support setting errors for other files


w0rp:
The easiest way to present some problems is to show only the problems in the current file. This is what ALE has done for a while. For quickfix and a loclist, it's also useful to present errors for other files, which can be used to open other files to jump to where the problem came from, etc.

Modifications should be made to support presenting errors for other files.

## To Do
- [x] Fix loclist items so ALE automatically sets `item.bufnr = bufnr(item.filename)` when `filename` is set for loclist items.
- [x] Modify loclist sorting to include buffer numbers in the sorting.
- [x] Make the echo message filter items based on `bufnr`.
- [x] Make the sign column highlighting filter items based on `bufnr`.
- [x] Make problem counts filter items based on `bufnr`.
- [x] Make ballonexpr filter items based on `bufnr`.
- [x] Fix loclist jumping so it filters items based on the buffer number.
- [x] Update the quickfix list so it collects all of the problems from every buffer and de-duplicates them.
- [ ] Set problems for other buffers in those buffers when they are open.
- [x] Save problems for filenames which aren't open yet, and set them when the files open.

w0rp:
An easy way to get everything to work would be to pass the items for other buffers to the quickfix list and loclist, and send only the items for the current buffer to everything else. (And store those in the `loclist` List we use for most things.)

w0rp:
Now items from all buffers ALE has checked will be set in quickfix, when it is turned on. The window won't open unless there are some problems found when checking the current buffer. Problems found while checking the current buffer which are in other buffers will cause the window to be opened, if that option is on.

w0rp:
Now if some linter returns problems with `filename` keys, they will be filtered out automatically and set in those files when they are opened.

I wonder if the last checklist item here is worth implementing, as it gets so incredibly complicated it might be untenable. ALE could look at buffers that are already open and set problems in them, but it might be possible that the list will conflict with the items which were set when checking the buffer itself. Problems are de-duplicated in quickfix, so that's not an issue, and they are set in loclist for the window, so that's not an issue.

I think I might leave that part out for now, and see what comes the implementation so far. The next step is to go through existing linters and get them to return the `filename` key instead of using the current function for filtering items for the filename.

w0rp:
[...]

PR #875, opened 2017-08-25
Show errors from other files for ghc (#653)

takano-akio:
GHC-based linters (including `hdevtools` and `ghc-mod`) may generate errors about files imported from the current file. This patch adds support for them.

### DIFF
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -177,6 +177,7 @@ function! s:HandleExit(job_id, exit_code) abort
     let l:output = l:job_info.output
     let l:buffer = l:job_info.buffer
     let l:next_chain_index = l:job_info.next_chain_index
+    let l:wd = l:job_info.wd
 
     if g:ale_history_enabled
         call ale#history#SetExitCode(l:buffer, a:job_id, a:exit_code)
@@ -208,7 +209,7 @@ function! s:HandleExit(job_id, exit_code) abort
         call ale#history#RememberOutput(l:buffer, a:job_id, l:output[:])
     endif
 
-    let l:loclist = ale#util#GetFunction(l:linter.callback)(l:buffer, l:output)
+    let l:loclist = ale#util#GetFunction(l:linter.callback)(l:buffer, l:output, l:wd)
 
     call s:HandleLoclist(l:linter.name, l:buffer, l:loclist)
 endfunction
@@ -552,6 +553,7 @@ function! s:RunJob(options) abort
         \   'buffer': l:buffer,
         \   'output': [],
         \   'next_chain_index': l:next_chain_index,
+        \   'wd': getcwd(),
         \}
     endif
 
--- a/autoload/ale/handlers/haskell.vim
+++ b/autoload/ale/handlers/haskell.vim
@@ -31,10 +31,6 @@ function! ale#handlers#haskell#HandleGHCFormat(buffer, lines) abort
             continue
         endif
 
-        if !ale#path#IsBufferPath(a:buffer, l:match[1])
-            continue
-        endif
-
         let l:errors = matchlist(l:match[4], '\v([wW]arning|[eE]rror): ?(.*)')
 
         if len(l:errors) > 0
@@ -52,6 +48,7 @@ function! ale#handlers#haskell#HandleGHCFormat(buffer, lines) abort
         endif
 
         call add(l:output, {
+        \   'filename': ale#path#Simplify(getcwd() . '/' . l:match[1]),
         \   'lnum': l:match[2] + 0,
         \   'col': l:match[3] + 0,
         \   'text': l:text,
--- a/autoload/ale/handlers/haskell.vim
+++ b/autoload/ale/handlers/haskell.vim
@@ -1,7 +1,7 @@
 " Author: w0rp <devw0rp@gmail.com>
 " Description: Error handling for the format GHC outputs.
 
-function! ale#handlers#haskell#HandleGHCFormat(buffer, lines) abort
+function! ale#handlers#haskell#HandleGHCFormat(buffer, lines, wd) abort
     " Look for lines like the following.
     "
     "Appoint/Lib.hs:8:1: warning:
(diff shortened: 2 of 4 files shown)


autoload/ale/handlers/haskell.vim
```diff
@@ -52,6 +48,7 @@ function! ale#handlers#haskell#HandleGHCFormat(buffer, lines) abort
         endif
 
         call add(l:output, {
+        \   'filename': l:match[1],
```
w0rp: Is the filename here an absolute path to a file? Relative paths aren't supported. Linter callbacks should somehow resolve paths to absolute paths based on current directories, project directories, etc.

takano-akio: Oh I had missed that requirement. I have updated the patch so that it uses an absolute path.

autoload/ale/handlers/haskell.vim
```diff
@@ -52,6 +48,7 @@ function! ale#handlers#haskell#HandleGHCFormat(buffer, lines) abort
         endif
 
         call add(l:output, {
+        \   'filename': ale#path#Simplify(getcwd() . '/' . l:match[1]),
```
w0rp: We can't rely on the value of `getcwd()` here. This function could be run from any buffer, and the current directory might not be the directory the file is in.

Are all of the paths here relative to the file being checked? I don't know what the filenames look like. If they are, then you should use `expand('#' . a:buffer . ':p:h')` to get the directory that the file being checked is in. I recommend saving that value in a variable above.

Play around with something like this and test it.

takano-akio: > Are all of the paths here relative to the file being checked?

No, they are relative to the cwd of the `ghc` (or `ghc-mod` etc.) process. Is there an easy way to find this directory?

w0rp: In that case, you can do what was done for some other linters. You can use the commands for adding `cd ... &&` to the front of the commands so the directory `ghc` will run from will be predictable and then resolve the paths from that same directory in the callback for building the loclist items.

Here is an example from the tslint.vim file.

```vim
" The command always switches to the directory for the buffer explicitly.
    return ale#path#BufferCdString(a:buffer) . ...

... 

" The directory is computed once at the top of the handler.
function! ale_linters#typescript#tslint#Handle(buffer, lines) abort
    let l:dir = expand('#' . a:buffer . ':p:h')

...

" The absolute path is resolved using the working directory.
        \   'filename': ale#path#GetAbsPath(l:dir, l:error.name),
```

Hopefully that should help.

takano-akio: Hmm, I think it can be very surprising to the user if `ale` silently used a different working directory for running these commands, because it can mess up things like search paths. It will be an even bigger problem for `hdevtools`, one of the ghc-based linters, because it uses a unix domain socket in the current directory to communicate with the server.

Would it be acceptable to extend the `job_info` dictionary to include the cwd of the job, and pass it over to the linter callback upon completion?

w0rp: No, it should be fine. Other linters already do the same thing, and you can see the `cd` command in `ALEInfo`. Try it, and see what happens.

takano-akio: I just tried it and it did not work. The code can be found here: https://github.com/takano-akio/ale/commit/96a6e9720078ba577c1197de42d904d8f46ec4c5

Let me explain the problem: imagine a directory hierarchy like this:

```
src/
  main.hs # defines the module Main, which imports Lib.Foo
  Lib/
    Foo.hs # defines the module Lib.Foo, which imports Lib.Bar
    Bar.hs # defines the module Lib.Bar
```

I open vim with cwd = `src/`, and do `:e Lib/Foo.hs`. Then `ale` invokes `ghc` on this file. Without a `cd` this works fine. With a `cd` this doesn't work, because `ghc` fails to find the module `Lib.Bar`. According to the Haskell convention, the module `Lib.Bar` must be found in a file called `Lib/Bar.hs` relative to some source directory, and by default the only source directory is `.`. If `ale` issues a `cd` into the `Lib` directory before invoking `ghc`, `ghc` will look for `src/Lib/Lib/Bar.hs`, which doesn't exist.

w0rp: Okay cool. Change the directory to the project directory then.

takano-akio: There is no universal way to find the project directory. There sometimes is, depending on how the project is set up, but not always. Hence my suggestion of recording the cwd.

w0rp: Use whatever that directory is, but you can't use `getcwd()`. The function for handling errors can be run from other unrelated buffers, and users can set their working directory to any directory on their filesystem.

If there's no one way to find the project directory, then we should try and detect it and make it configurable, so we can find it most of the time and let users explicitly set it when we can't figure out out automatically.

autoload/ale/engine.vim
```diff
@@ -177,6 +177,7 @@ function! s:HandleExit(job_id, exit_code) abort
     let l:output = l:job_info.output
     let l:buffer = l:job_info.buffer
     let l:next_chain_index = l:job_info.next_chain_index
+    let l:wd = l:job_info.wd
```
w0rp: No, don't do this. This will mean we have to compute this information for linters even where it isn't relevant. Fix the problem in the GHC linter.

autoload/ale/engine.vim
```diff
@@ -552,6 +553,7 @@ function! s:RunJob(options) abort
         \   'buffer': l:buffer,
         \   'output': [],
         \   'next_chain_index': l:next_chain_index,
+        \   'wd': getcwd(),
```
w0rp: This is also possibly the wrong working directory, as this could run from some buffer other than the one you are checking.

w0rp:
I just pushed a commit to ban `getcwd()` from the codebase, with an explanation.

takano-akio:
Oops, sorry I didn't mean to push (yet). You said:

> If there's no one way to find the project directory, then we should try and detect it and make it configurable, so we can find it most of the time and let users explicitly set it when we can't figure out out automatically.

My reasoning for using `getcwd` is that the working directory of the window *is* the most obvious and least surprising guess for the "project directory". This is what you get if you use `hdevtools` with vim's builtin `:make` feature, for example.

w0rp:
You can get the directory the file is in, or some other directory, but `getcwd()` will return the wrong directory some of the time. The function which starts checking for errors or the function for handling errors can both be run from _any buffer at all_, and you don't know which one it will run from. So you have to use something like `expand('#' . a:buffer . ':p:h')` to get a directory name instead.

I pushed a commit to master which complains whenever `getcwd()` is used in the codebase, because it doesn't work.

w0rp:
Closed due to inactivity.

PR #884, opened 2017-08-29
Detect and use CM files for smlnj

jez:
SML/NJ uses `*.cm` files to manage compilation of multiple files.
This pull request detects when a `*.cm` file exists for this project, and uses
it to report errors.

Falls back to the old behavior when no `*.cm` file is found.

### DIFF
--- a/ale_linters/sml/smlnj.vim
+++ b/ale_linters/sml/smlnj.vim
@@ -3,7 +3,7 @@
 
 call ale#linter#Define('sml', {
 \   'name': 'smlnj',
-\   'executable': 'ale#handlers#sml#GetExecutableSmlnjFile',
+\   'executable_callback': 'ale#handlers#sml#GetExecutableSmlnjFile',
 \   'command': 'sml',
 \   'callback': 'ale#handlers#sml#Handle',
 \})
--- a/ale_linters/sml/smlnj.vim
+++ b/ale_linters/sml/smlnj.vim
@@ -1,77 +1,9 @@
 " Author: Paulo Alem <paulo.alem@gmail.com>, Jake Zimmerman <jake@zimmerman.io>
-" Description: Rudimentary SML checking with smlnj compiler
-
-" Let user manually set the CM file (in case our search for a CM file is
-" ambiguous and picks the wrong one)
-"
-" See :help ale-sml-smlnj for more information.
-call ale#Set('sml_smlnj_cm_file', '*.cm')
-
-function! ale_linters#sml#smlnj#Handle(buffer, lines) abort
-    " Try to match basic sml errors
-
-    let l:out = []
-    let l:pattern = '^.*\:\([0-9\.]\+\)\ \(\w\+\)\:\ \(.*\)'
-    let l:pattern2 = '^.*\:\([0-9]\+\)\.\?\([0-9]\+\).* \(\(Warning\|Error\): .*\)'
-
-    for l:line in a:lines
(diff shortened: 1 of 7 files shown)


ale_linters/sml/smlnj.vim
```diff
@@ -39,9 +45,33 @@ function! ale_linters#sml#smlnj#Handle(buffer, lines) abort
...
+    " Found a CM file; let's use it
+    return 'sml -m ' . l:cmfile . ' < /dev/null'
+  endif
+endfunction
+
 call ale#linter#Define('sml', {
 \   'name': 'smlnj',
 \   'executable': 'sml',
-\   'command': 'sml',
+\   'lint_file': 1,
```
w0rp: We're losing the ability to check for problems while you type here. Is there some way we can keep this behaviour, or is there some technical reason for why we have to check the file on disk?

jez: Is there a way to set `lint_file` dynamically from the `GetCommand` function? Or is it the kind of thing that once it's set in `Define` it's unchangable?

The way we invoke the project-level build tool is with `sml -m sources.cm`. The `sources.cm` file is going to declare all the dependencies for a file; CM (Compilation Manager) then finds all the files on disk and compiles the entire project. This is the only way to correctly resolve symbols that are defined in different files.

Unlike other languages, SML doesn't have "import" statements that say one files depends on another file. You declare all files that are used together up front in a list in the CM file, and CM computes the dependency graph for your project.

The way it works for a single file, SML need to use CM. Instead, it just pretends to interpret that one file; if you're interpreting a file which depends on external symbols, they're just going to be not found.

If there is a way to say "in single file mode, use `lint_file: 0`, otherwise `lint_file: 1`, I can add that.

w0rp: I see. I had considered adding something like `lint_file_callback` for dynamically computing the value. In this case, it seems like what you're doing here is just the right thing to do.

We could consider creating two different linters here which run the same program. If `sml` supports options similar to the options `gcc` has for only checking syntax, we could consider writing one linter which only checks for syntax errors in a buffer as you type, and another linter which looks for semantic errors when you open or save files, etc. That might be our best option.

w0rp: If we can't disable semantic analysis, we could also consider filtering out semantic errors when running `sml` against a buffer.

jez:
I took your advice of splitting it into two checkers. Instead having one for "syntax errors" and one for "semantic errors" I made one for "single file" and one for "CM project."

Only one of `smlnj` or `smlnj-cm` can be enabled for a given buffer. This works by abusing `executable_callback` to artificially say "available" for one and "not available" for the other.

Splitting things this way still doesn't catch "syntax" errors for CM projects. But I think it's nicer this way than mucking around with format strings for things like `syntax error` etc.

@w0rp Let me know if this approach looks good to you. I didn't write tests at first because I couldn't wrap my head around the testing framework. If this looks okay then I'll test the current implementation.

jez:
@w0rp Please take a look when you get a chance; tests are passing, and everything works for me locally.

jez:
(Also, an unrelated question while I have your attention, does ALE support showing multi-line error messages on multiple lines?)

w0rp:
Yeah, you can draw highlights for problems over many lines. Look at `:help ale-loclist-format` and the `end_col` and `end_lnum` keys.

w0rp:
Cheers! :beers:

PR #885, opened 2017-08-29
Add prettier_d fixer path

morhetz:
Add [prettier_d](https://github.com/josephfrazier/prettier_d) to prettier bin lookup


w0rp:
Could probably use a test, but this looks fine to me.

w0rp:
Cheers! :beers:

jonleopard:
Will ale check for prettier_d or eslint_d that are installed globally with Yarn? I can submit a quick PR if this isn't the case.

w0rp:
No, but you can change your settings to point to that executable with the `_executable` and `_use_global` options. I recommend installing everything locally in `node_modules`.

jonleopard:
Thanks for the info! I mistakenly thought that `'node_modules/.bin/prettier_d',` was the global path that ale looks at, so I was going to add yarn's global path to the prettier fixer file (since I use yarn for global package installations). 

Thanks for clearing that up for me :)

PR #886, opened 2017-08-29
Add support for prettier configuration file.

aliou:
As of version [1.6.0](https://github.com/prettier/prettier/releases/tag/1.6.0), `prettier` allows passing a `--config` argument with a path to a configuration file.

This PR adds support for this argument when a configuration file is present and some tests inspired by the rubocop fixer tests.

### DIFF
--- a/autoload/ale/fixers/prettier.vim
+++ b/autoload/ale/fixers/prettier.vim
@@ -4,6 +4,7 @@
 
 call ale#Set('javascript_prettier_executable', 'prettier')
 call ale#Set('javascript_prettier_use_global', 0)
+call ale#Set('javascript_prettier_use_local_config', 0)
 call ale#Set('javascript_prettier_options', '')
 
 function! s:FindConfig(buffer) abort
@@ -35,12 +36,13 @@ endfunction
 function! ale#fixers#prettier#Fix(buffer) abort
     let l:options = ale#Var(a:buffer, 'javascript_prettier_options')
     let l:config = s:FindConfig(a:buffer)
+    let l:use_config = ale#Var(a:buffer, 'javascript_prettier_use_local_config')
 
     return {
     \   'command': ale#Escape(ale#fixers#prettier#GetExecutable(a:buffer))
     \       . ' %t'
     \       . (!empty(l:options) ? ' ' . l:options : '')
-    \       . (!empty(l:config) ? ' --config ' . ale#Escape(l:config) : '')
+    \       . (l:use_config ? ' --config ' . ale#Escape(l:config) : '')
     \       . ' --write',
     \   'read_temporary_file': 1,
     \}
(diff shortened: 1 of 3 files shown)


aliou:
For some reason the test fail on the CI but pass on my machine. 🤔 Investigating.

aliou:
(Forgot to include the test prettier configuration file in the patch 😅)

aliou:
Using the argument with an older version doesn't seem to break, but prints a message to stderr.

When you say adding an option, do you mean adding something like a `g:javascript_prettier_use_config_file` or let the user set it through the `g:javascript_prettier_options` ?

w0rp:
Oh, does it ignore the additional argument? I'd expect that most programs would just stop working entirely if you use an invalid argument.

aliou:
Yup it ignores it on prettier 1.5.0, I haven't tried previous versions. But you're right it is probably safer to add an option in case previous versions don't ignore the argument.

What do you think about the `g:javascript_use_local_config` option, with `0` as a default value?

w0rp:
I think it would be best to add an option in ALE specific to prettier, like `g:ale_prettier_...`

aliou:
Oops, yeah that's what I meant 😅

aliou:
Added the `g:javascript_prettier_use_local_config` option and rebased to include the changes #885.

autoload/ale/fixers/prettier.vim
```diff
@@ -16,11 +35,14 @@ endfunction
...
     let l:options = ale#Var(a:buffer, 'javascript_prettier_options')
+    let l:config = s:FindConfig(a:buffer)
+    let l:use_config = ale#Var(a:buffer, 'javascript_prettier_use_local_config')
 
     return {
     \   'command': ale#Escape(ale#fixers#prettier#GetExecutable(a:buffer))
     \       . ' %t'
-    \       . ' ' . l:options
+    \       . (!empty(l:options) ? ' ' . l:options : '')
+    \       . (l:use_config ? ' --config ' . ale#Escape(l:config) : '')
```
w0rp: You also need to check `!empty(l:config)` here.

doc/ale-javascript.txt
```diff
@@ -86,6 +86,11 @@ g:ale_javascript_prettier_use_global     *g:ale_javascript_prettier_use_global*
 
   See |ale-integrations-local-executables|
 
+g:ale_javascript_prettier_use_local_config     *g:ale_javascript_prettier_use_local_config*
+                                               *b:ale_javascript_prettier_use_local_config*
+  Type: |Number|
+  Default: `0`
+
```
w0rp: Add a short description here explaining what the option does when set to `1`.

autoload/ale/fixers/prettier.vim
w0rp: You could use `&& !empty(l:config)` for the value of `l:user_config` above.

aliou:
Updated

w0rp:
Cheers! :beers:

PR #892, opened 2017-08-30
Add a linter for Apache Thrift IDL files

jparise:
This linter works by invoking the `thrift` compiler with the buffer
contents and reporting any parser and code generation issues.

The handler rolls its own output-matching loop because we have the
(unfortunate) requirement of handling error output that spans multiple
lines.

Unit tests cover both the command callback and handler, and there is
initial documentation for all of the option variables.


w0rp:
Cheers! :beers:

PR #902, opened 2017-09-01
Allow Perl::Critic issues to be displayed as warnings or errors

oalders:
This is the result of a little Twitter poll:

https://twitter.com/olafalders/status/898594830707679232

Basically, I don't want Perl::Critic violations to display as errors since they're generally not going to prevent my code from compiling.  If I'm whipping up a quick script, I just want to deal with the compilation errors etc first.

I tweaked the Vader tests slightly, but I didn't add a test for the new functionality because it wasn't clear to me how to do this.  I'm happy to add a test if you think it's necessary and can possibly point me at some prior art.

As far as the name of the new option, I can happily rename that as well.  Again, I wasn't sure if there's a different plugin that already does this and has a naming convention which I should follow.

### DIFF
--- a/ale_linters/perl/perlcritic.vim
+++ b/ale_linters/perl/perlcritic.vim
@@ -13,6 +13,9 @@ let g:ale_perl_perlcritic_options =
 let g:ale_perl_perlcritic_showrules =
 \   get(g:, 'ale_perl_perlcritic_showrules', 0)
 
+let g:ale_perl_perlcritic_display_mode =
+\   get(g:, 'ale_perl_perlcritic_display_mode', 'E')
+
 function! ale_linters#perl#perlcritic#GetExecutable(buffer) abort
     return ale#Var(a:buffer, 'perl_perlcritic_executable')
 endfunction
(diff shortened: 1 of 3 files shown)


w0rp:
You can already configure this with the `g:ale_type_map` option. See `:help g:ale_type_map`.

oalders:
Thanks for this.  As a follow-up, I'm having issues getting this to work for `perlcritic`.  I posted the question on StackOverflow, but I figured I'd attach it here for anyone else who comes along with a similar issue. https://stackoverflow.com/questions/46080490/forcing-ale-to-display-perlcritic-violations-as-warnings-rather-than-errors

w0rp:
Your example on stack overflow should work. Open an issue, share your vimrc file and anything else you can share, and paste the output of `:ALEInfo`. `let g:ale_type_map = {'perlcritic': ...}` is what you want, as that's the name of the linter. Bear in mind that you can overwrite this option by setting it later, so make sure you only use a single assignment. Make sure you're using the latest version of ALE.

PR #903, opened 2017-09-02
run slim-lint on real file and mark it as lint file only

doits:
This fixes slim-lint ignoring a `.rubocop.yml` in the file's or parent directory, since due to the way slim-lint calls rubocop, rubocop's working directory is the file's path.

### DIFF
--- a/ale_linters/slim/slimlint.vim
+++ b/ale_linters/slim/slimlint.vim
@@ -1,5 +1,25 @@
 " Author: Markus Doits - https://github.com/doits
-" Description: slim-lint for Slim files, based on hamllint.vim
+" Description: slim-lint for Slim files
+
+function! ale_linters#slim#slimlint#GetCommand(buffer) abort
+    let l:command = 'slim-lint %t'
+
+    let l:rubocop_config = ale#path#FindNearestFile(a:buffer, '.rubocop.yml')
+
+    " Set SLIM_LINT_RUBUCOP_CONF variable as it is needed for slim-lint to
+    " pick up the rubocop config.
+    "
+    " See https://github.com/sds/slim-lint/blob/master/lib/slim_lint/linter/README.md#rubocop
+    if !empty(l:rubocop_config)
+      if ale#Has('win32')
+        let l:command = 'set SLIM_LINT_RUBUCOP_CONF=' . ale#Escape(l:rubocop_config) . ' && ' . l:command
+      else
+        let l:command = 'SLIM_LINT_RUBUCOP_CONF=' . ale#Escape(l:rubocop_config) . ' ' . l:command
+      endif
+    endif
+
+    return l:command
+endfunction
 
 function! ale_linters#slim#slimlint#Handle(buffer, lines) abort
     " Matches patterns like the following:
@@ -21,6 +41,6 @@ endfunction
 call ale#linter#Define('slim', {
 \   'name': 'slimlint',
 \   'executable': 'slim-lint',
-\   'command': 'slim-lint %t',
+\   'command_callback': 'ale_linters#slim#slimlint#GetCommand',
 \   'callback': 'ale_linters#slim#slimlint#Handle'
 \})
--- a/test/command_callback/test_slimlint_command_callback.vader
+++ b/test/command_callback/test_slimlint_command_callback.vader
@@ -1,0 +1,35 @@
+Before:
+  runtime ale_linters/slim/slimlint.vim
+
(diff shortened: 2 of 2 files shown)


doits:
I changed it to search for the rubocop config and the the env var correctly. I didn't expect it to be ale's job to traverse up the path tree to search for config files (isn't that a performance penalty on each linting run?).

Anyway, do we need a test for this? If yes, can you give me a hint were something similar was already tested? Looks like jshint does not have a test for this.

If everything's OK I can rebase this into one single commit so we keep the history clean.

ale_linters/slim/slimlint.vim
```diff
@@ -1,5 +1,19 @@
...
+
+function! ale_linters#slim#slimlint#GetCommand(buffer) abort
+    let l:rubocop_config = ale#path#ResolveLocalPath(
+    \   a:buffer,
+    \   '.rubocop.yml',
+    \   ''
+    \)
+
+    if !empty(l:rubocop_config)
+        let $SLIM_LINT_RUBOCOP_CONF = l:rubocop_config
```
w0rp: You should use the `--config` option with the config path instead of changing the environment variable, which has more side-effects. Use `ale#Escape` on the path.

ale_linters/slim/slimlint.vim
```diff
@@ -1,5 +1,19 @@
 " Author: Markus Doits - https://github.com/doits
-" Description: slim-lint for Slim files, based on hamllint.vim
+" Description: slim-lint for Slim files
+
+function! ale_linters#slim#slimlint#GetCommand(buffer) abort
+    let l:rubocop_config = ale#path#ResolveLocalPath(
```
w0rp: Use `ale#path#FindNearestFile` instead.

ale_linters/slim/slimlint.vim
w0rp: It looks like `slim` has separate configuration files for itself, but it also uses rubocop configuration files? I don't know what's going on there.

The danger in setting the environment variable is that will will be applied to other invocations of `slim-lint` where no configuration file is found. I don't think the environment variable will persist after Vim is closed.

doits: > It looks like slim has separate configuration files for itself, but it also uses rubocop configuration files? I don't know what's going on there.

It is explained in the linter's readme:

https://github.com/sds/slim-lint/blob/5b258d4c922b47ee28e593992bbdcbdfd47d61fe/lib/slim_lint/linter/README.md#rubocop

I must admit it is a little bit complicated, but the env var seems exactly be the way how slim-lint expects linters to do it ...

w0rp: I wouldn't recommend setting environment variables like this, as you don't know if the environment variable is going to be changed again before the actual command is run. Plus there's the problem of the environment variable being used when you don't want it to be used, where you run `slim-lint` once with a configuration file available, and then again without it.

w0rp: I recommend prefixing the command string with the environment variable instead. Use `X=... <command>` for Unix, and `set X=... && <command>` for Windows. You can use `ale#Has('win32')` so you can test both versions when running the tests on Unix.

w0rp: Look at how `ale#Has` is used in existing Vader tests. There's a `Dictionary` which can be set for effectively mocking `has()` calls.

test/command_callback/test_slimlint_command_callback.vader
```diff
@@ -0,0 +1,18 @@
...
+After:
+  Restore
+  call ale#linter#Reset()
+  call ale#test#RestoreDirectory()
+
+Execute(The command should be correct without a .rubocop.yml):
+  AssertEqual 'slim-lint %t', ale_linters#slim#slimlint#GetCommand(bufnr(''))
+
+Execute(The .rubocop.yml should be detected correctly):
+  call ale#test#SetFilename('../slimlint-test-files/subdir/file.slim')
```
doits: @w0rp can you help me here? I'm really struggling to get this test to work, even after looking at other tests. How can I set the directory/file for `bufnr`?

doits: Hmm wait, it is failing on CI like expected, but here locally it does not fail ... weird. Looks like this is the right way then.

doits:
Will finish this later since I have to leave now, but looks like I'm on the right track!

doits:
OK, finished adding the tests. Is there anything else to change?

PR #909, opened 2017-09-05
Fix OCaml helptag name to match readme.

ode79:
The readme states for OCaml completion 'merlin see :help ale-integration-ocaml-merlin for configuration instructions'

This helptag didn't work, so I've fixed it.


ode79:
Done.

PR #910, opened 2017-09-06
Add 'prettier' fixer support to TypeScript, CSS, SCSS and JSON

renstrom:
Not sure this is the best approach, it might be a bit too magic-y or implicit.

### DIFF
--- a/autoload/ale/fixers/prettier.vim
+++ b/autoload/ale/fixers/prettier.vim
@@ -44,6 +44,8 @@ function! ale#fixers#prettier#Fix(buffer) abort
     if match(l:options, '--parser') == -1
         if &filetype is# 'typescript'
             let l:parser = 'typescript'
+        elseif &filetype =~# 'css\|scss'
+            let l:parser = 'postcss'
         else
             let l:parser = 'babylon'
         endif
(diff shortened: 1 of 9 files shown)


renstrom:
Not sure what's up with appveyor...

w0rp:
Ignore AppVeyor. I just started setting that up.

autoload/ale/fixers/prettier.vim
```diff
@@ -39,6 +39,21 @@ function! ale#fixers#prettier#Fix(buffer) abort
     let l:use_config = ale#Var(a:buffer, 'javascript_prettier_use_local_config')
                 \ && !empty(l:config)
 
+    " Append the --parser flag depending on the current filetype (unless it's
+    " already set in g:javascript_prettier_options).
+    if match(l:options, '--parser') == -1
+        if &filetype is# 'typescript'
```
w0rp: You should use `getbufvar` for `&filetype`. Because of the way async execution works, this code could be run from another buffer.

renstrom: Ah, right! This ought to do it 829e0442d191c711ed537c2c18a56b67d86a72e2

w0rp:
Cheers! :beers:

renstrom:
I guess there's also `prettier-standard` and `prettier-eslint`. But I don't really know what the difference between those and plain `prettier` is. So I just skipped that.

PR #915, opened 2017-09-08
Elm local install support

soywod:
Added here some support for local installation of Elm (for eg: via npm). All this code has been copied (and ajusted) from the Flow linter, so I don't think it is necessary to add some tests here.

### DIFF
--- a/ale_linters/elm/make.vim
+++ b/ale_linters/elm/make.vim
@@ -1,5 +1,14 @@
-" Author: buffalocoder - https://github.com/buffalocoder
+" Author: buffalocoder - https://github.com/buffalocoder, soywod - https://github.com/soywod
 " Description: Elm linting in Ale. Closely follows the Syntastic checker in https://github.com/ElmCast/elm-vim.
+
+call ale#Set('elm_make_executable', 'elm-make')
+call ale#Set('elm_make_use_global', 1)
+
+function! ale_linters#elm#make#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'elm_make', [
+    \   'node_modules/.bin/elm-make',
+    \])
+endfunction
 
 function! ale_linters#elm#make#Handle(buffer, lines) abort
     let l:output = []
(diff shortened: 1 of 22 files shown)


ale_linters/elm/make.vim
```diff
@@ -1,6 +1,15 @@
-" Author: buffalocoder - https://github.com/buffalocoder
+" Author: buffalocoder - https://github.com/buffalocoder, soywod - https://github.com/soywod
 " Description: Elm linting in Ale. Closely follows the Syntastic checker in https://github.com/ElmCast/elm-vim.
 
+call ale#Set('elm_make_executable', 'elm-make')
+call ale#Set('elm_make_use_global', 1)
```
w0rp: Change the default to `0`, to match the other linters. We should always prefer using a version more appropriate for a project by default.

ale_linters/elm/make.vim
```diff
@@ -63,14 +73,16 @@ function! ale_linters#elm#make#GetCommand(buffer) abort
     " a sort of flag to tell the compiler not to generate an output file,
     " which is why this is hard coded here.
     " Source: https://github.com/elm-lang/elm-make/blob/master/src/Flags.hs
-    let l:elm_cmd = 'elm-make --report=json --output='.ale#Escape('/dev/null')
+    let l:elm_cmd = ale#Escape(l:elm_exe)
+    \   . ' --report=json'
+    \   . ' --output=' . ale#Escape('/dev/null')
```
w0rp: Use `g:ale#util#nul_file` here instead, as `/dev/null` doesn't work on Windows. (Except for that one Go program...)

w0rp:
Make sure to document the new options too.

ale_linters/elm/make.vim
soywod: Done

ale_linters/elm/make.vim
soywod: Done

ale_linters/elm/make.vim
w0rp: I don't see the changes yet. You might need to push a commit.

soywod: It's just for me, to not forget any task. It's done in local, I'll push soon

w0rp:
There's a conflict to resolve here, from merging the other PR. I might fix this later.

soywod:
I resolve my conflict here ;)

w0rp:
Ah, could you move the `elm-make` after the `elm-format` section and make sure they are in the same order in the table of contents? They are supposed to be mostly sorted by name. I might add some Bash code for catching that at some point.

w0rp:
Don't forget the table of contents.

w0rp:
Cheers! :beers:

PR #916, opened 2017-09-09
Add elm-format as a new fixer

soywod:
Add a new fixer: elm-format

### DIFF
--- a/autoload/ale/fixers/format.vim
+++ b/autoload/ale/fixers/format.vim
@@ -1,0 +1,25 @@
+" Author: soywod <clement.douin@gmail.com>
+" Description: Integration of Elm with ALE.
+
+call ale#Set('elm_format_executable', 'elm-format')
+call ale#Set('elm_format_use_global', 1)
+call ale#Set('elm_format_options', '--yes')
+
+function! ale#fixers#format#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'elm_format', [
+    \   'node_modules/.bin/elm-format',
+    \   'node_modules/elm-format/bin/elm-format',
+    \   'node_modules/elm-format/index.js',
+    \])
+endfunction
+
+function! ale#fixers#format#Fix(buffer) abort
+    let l:options = ale#Var(a:buffer, 'elm_format_options')
+
+    return {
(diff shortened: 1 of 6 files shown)


autoload/ale/fixers/format.vim
```diff
@@ -0,0 +1,25 @@
...
+
+call ale#Set('elm_format_executable', 'elm-format')
+call ale#Set('elm_format_use_global', 1)
+call ale#Set('elm_format_options', '--yes')
+
+function! ale#fixers#format#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'elm_format', [
+    \   'node_modules/.bin/elm-format',
+    \   'node_modules/elm-format/bin/elm-format',
+    \   'node_modules/elm-format/index.js',
```
w0rp: For at least this file, you will have to execute it with `node` on Windows, or it could lead to the `.js` file being opened in a text editor or similar. Look at the code for the `eslint` fixer for an example.

autoload/ale/fixers/format.vim
```diff
@@ -0,0 +1,25 @@
+" Author: soywod <clement.douin@gmail.com>
+" Description: Integration of Elm with ALE.
+
+call ale#Set('elm_format_executable', 'elm-format')
+call ale#Set('elm_format_use_global', 1)
```
w0rp: Set this to `0` by default, to match the other options. Local versions should be tried first, then global versions, by default.

w0rp:
Oh yes, don't forget to document the new option too.

autoload/ale/fixers/format.vim
soywod: I made a mistake here, I just had a look on this file and it's like an installer, so I will remove it. Like `elm-make`, the `'node_modules/.bin/elm-format'` is enough

autoload/ale/fixers/format.vim
soywod: Done

autoload/ale/fixers/format.vim
soywod: Done

w0rp:
Cheers! :beers:

PR #933, opened 2017-09-14
[WIP] Begin to distinguish between Perl warnings and errors

oalders:
This is a work in progress, but I was interested in getting some feedback.

The biggest change here is, I think, not clobbering messages from previous line numbers.  I could also change this to concatenate the messages rather than letting the first one win.

### DIFF
--- a/ale_linters/perl/perl.vim
+++ b/ale_linters/perl/perl.vim
@@ -37,9 +37,10 @@ function! ale_linters#perl#perl#Handle(buffer, lines) abort
 
     for l:match in ale#util#GetMatches(a:lines, l:pattern)
         let l:line = l:match[3]
+        let l:file = l:match[2]
         let l:text = l:match[1]
 
-        if ( ale#path#IsBufferPath(a:buffer, l:match[2]) && !has_key(l:seen,l:line) )
+        if ( ale#path#IsBufferPath(a:buffer, l:file) && !has_key(l:seen,l:line) )
         \ && (
         \   l:text isnot# 'BEGIN failed--compilation aborted'
         \   || empty(l:output)
--- a/ale_linters/perl/perl.vim
+++ b/ale_linters/perl/perl.vim
@@ -28,8 +28,7 @@ function! ale_linters#perl#perl#Handle(buffer, lines) abort
     let l:basename = expand('#' . a:buffer . ':t')
 
     let l:type = 'E'
-    let l:syntax_ok = matchstr(a:lines[-1], 'syntax OK')
-    if l:syntax_ok isnot# ''
+    if a:lines[-1] =~# 'syntax OK'
         let l:type = 'W'
     endif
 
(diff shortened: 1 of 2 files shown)


ale_linters/perl/perl.vim
```diff
@@ -27,12 +27,20 @@ function! ale_linters#perl#perl#Handle(buffer, lines) abort
     let l:output = []
     let l:basename = expand('#' . a:buffer . ':t')
 
+    let type = 'E'
+    let syntax_ok = matchstr(a:lines[-1], 'syntax OK')
+    if syntax_ok != ''
+        let type = 'W'
+    endif
+
+    let seen = {}
```
w0rp: You should install `Vader` for checking your Vim code. Check out the output from the Travis CI build for the warnings it will give you. It will complain about your variable names.

ale_linters/perl/perl.vim
```diff
@@ -27,12 +27,20 @@ function! ale_linters#perl#perl#Handle(buffer, lines) abort
     let l:output = []
     let l:basename = expand('#' . a:buffer . ':t')
 
+    let type = 'E'
+    let syntax_ok = matchstr(a:lines[-1], 'syntax OK')
```
w0rp: If you just want to check if there is a match, you can use the operator for that. See `:help =~#`.

ale_linters/perl/perl.vim
```diff
@@ -27,12 +27,20 @@ function! ale_linters#perl#perl#Handle(buffer, lines) abort
...
+    let seen = {}
+
     for l:match in ale#util#GetMatches(a:lines, l:pattern)
         let l:line = l:match[3]
         let l:text = l:match[1]
-        let l:type = 'E'
+        let l:type = type
 
-        if ale#path#IsBufferPath(a:buffer, l:match[2])
+        if ( ale#path#IsBufferPath(a:buffer, l:match[2]) && !has_key(seen,l:line) )
```
w0rp: You don't need the parentheses here. There's already another condition on the line below, so you could maybe move your second check to a new line below.

ale_linters/perl/perl.vim
w0rp: I also prohibit the use of `==#` and `==?` from the codebase, as they behave poorly when you mix numbers and strings. `is#` and is?` are preferred instead.

oalders:
I think I've addressed all of the issues cited above.  I'd like to live with this for a few days before I remove the WIP label.  If this looks ok, would you like the commits squashed?

oalders:
BTW, I had no problems running `./run-tests` from my Mac.  Worked like a charm.

ale_linters/perl/perl.vim
oalders: Fixed in 43782df621315ad

ale_linters/perl/perl.vim
oalders: Fixed in f66d38bab7

w0rp:
Cheers! :beers:

oalders:
Thanks @w0rp!

PR #943, opened 2017-09-19
Improve LSP-specific docs

svenstaro:
The documentation was incomplete before. Fix that.

### DIFF
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -1421,17 +1421,34 @@ ale#linter#Define(filetype, linter)                       *ale#linter#Define()*
                          This argument may be omitted or `''` when a linter
                          does not represent an LSP linter.
 
+                         When this argument is set to `'socket'`, then the
+                         linter will be defined as an LSP linter which
+                         communicates with a running language server at
+                         an address provided by `address_callback`.
+                         With `socket`, none of `executable`,
+                         `executable_callback`, `command`, `command_callback`,
+                         `command_chain` may be provided.
+
                          When this argument is set to `'stdio'`, then the
                          linter will be defined as an LSP linter which keeps a
                          process for a language server runnning, and
                          communicates with it directly via a |channel|.
+                         In this case, `project_callback` and
+                         `language_callback` arguments must also be defined.
 
-                         When this argument is not empty, then the
-                         `project_callback` and `language_callback` arguments
-                         must also be defined.
+                         When this argument is set to `'tsserver'`, then the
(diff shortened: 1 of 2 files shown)


doc/ale.txt
```diff
@@ -1421,18 +1421,35 @@ ale#linter#Define(filetype, linter)                       *ale#linter#Define()*
                          This argument may be omitted or `''` when a linter
                          does not represent an LSP linter.
 
+                         When this argument is set to `'socket'`, then the
+                         linter will be defined as an LSP linter which
+                         communicates with a running language server at
+                         an address provided by `address_callback`.
+                         With `socket`, none of `executable`,
+                         `executable_callback`, `command`, `command_callback`,
+                         `command_chain` may be provided.
```
w0rp: I think let's hold off on documenting this for now, as it isn't yet fully implemented. I only want to document something if  it's tested and proven to work to some degree.

doc/ale.txt
```diff
@@ -1421,18 +1421,35 @@ ale#linter#Define(filetype, linter)                       *ale#linter#Define()*
...
+                         `language_callback` arguments must also be defined.
 
-                         When this argument is not empty, then the
-                         `project_callback` and `language_callback` arguments
-                         must also be defined.
+                         When this argument is set to `'tsserver'`, then the
+                         linter will be defined as an LSP linter which
+                         communicates with tsserver directly.
+                         In this case, `project_callback` and
+                         `language_callback` arguments must also be defined.
```
w0rp: I say don't bother documenting this either. The `'tsserver'` value is only for internal use, and it's only useful for specifically `tsserver`.

doc/ale.txt
```diff
@@ -1421,18 +1421,35 @@ ale#linter#Define(filetype, linter)                       *ale#linter#Define()*
...
 
                          LSP linters handle diagonstics automatically, so
                          the `callback` argument must not be defined.
 
+  `address_callback`       A |String| of |Funcref| for a callback function
+                         accepting a buffer number. A |String| should be
+                         returned representing the address to the language
+                         server.
```
w0rp: It's maybe okay to document this already. Replace "address to the language server," with "address for the language server," which is the correct grammar.

svenstaro:
I'll close this for now. Seems like this mostly addresses things that are not even supposed to be working yet.

w0rp:
Okay, I might write some of what you have here into the documentation later.

PR #949, opened 2017-09-22
Move dialect setting before user options (shellcheck)

TheLocehiliosan:
This PR simply moves the dialect option before the user defined options for the shellcheck linter.

I tried to override the dialect with:

    let g:ale_sh_shellcheck_options = '-s bash'

However, this doesn't work because the dialect discovered by ale is place _after_ the user defined options. This results is `-s bash -s sh`. I certainly think the user should be able to override if they want. With this change the new result is `-s sh -s bash` (shellcheck using the last defined option).

### DIFF
--- a/ale_linters/sh/shellcheck.vim
+++ b/ale_linters/sh/shellcheck.vim
@@ -44,9 +44,9 @@ function! ale_linters#sh#shellcheck#GetCommand(buffer) abort
     let l:dialect = ale_linters#sh#shellcheck#GetDialectArgument(a:buffer)
 
     return ale_linters#sh#shellcheck#GetExecutable(a:buffer)
+    \   . (!empty(l:dialect) ? ' -s ' . l:dialect : '')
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

TheLocehiliosan:
Thanks. You are correct that generally there is no need to specify the dialect. In my case, I author a few Bash-based utilities which need to be very portable. The location of `bash` is not uniform across platforms (nor is the location of `env`). However, the location of `sh` seems to be uniform. So there's an idiom that looks like this:

    #!/bin/sh
    if [ -z "$BASH_VERSION" ]; then
      exec bash "$0" "$@"
    fi

Which can improve portability. Of course, this messes with the ability to identify the dialect. ;)

Thanks for your fantastic plugin!

Cheers

w0rp:
Interesting. Thanks for the explanation. Makes sense to me.

PR #952, opened 2017-09-25
Advance C# linter based on mcs -t:module

hernot:
Added addtional c# linter called mcsc using the module target to not only catch syntax errors but also indicate compilation errors and possibly missing assemblies etc. the lint_file flag is set to 1 to avoid delays on large directory trees. 

Still missing is a possibility to define subdirectories which should be excluded from linting.

And suggestions for improvement wellcome.

Have added a ale-cs.txt file in doc describing the existing mcs and mcsc linter. as this is not included in the files on github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ formatting.
 | C | [cppcheck](http://cppcheck.sourceforge.net), [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint), [gcc](https://gcc.gnu.org/), [clang](http://clang.llvm.org/), [clangtidy](http://clang.llvm.org/extra/clang-tidy/) !!, [clang-format](https://clang.llvm.org/docs/ClangFormat.html)|
 | C++ (filetype cpp) | [clang](http://clang.llvm.org/), [clangcheck](http://clang.llvm.org/docs/ClangCheck.html) !!, [clangtidy](http://clang.llvm.org/extra/clang-tidy/) !!, [cppcheck](http://cppcheck.sourceforge.net), [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint) !!, [gcc](https://gcc.gnu.org/), [clang-format](https://clang.llvm.org/docs/ClangFormat.html)|
 | CUDA | [nvcc](http://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html) |
-| C# | [mcs](http://www.mono-project.com/docs/about-mono/languages/csharp/) |
+| C# | [mcs](http://www.mono-project.com/docs/about-mono/languages/csharp/) see:`help ale-cs-mcs` for details,[mcsc](http://www.mono-project.com/docs/about-mono/languages/csharp/) see:`help ale-cs-mcsc` for details and configuration|
 | Chef | [foodcritic](http://www.foodcritic.io/) |
 | CMake | [cmakelint](https://github.com/richq/cmake-lint) |
 | CoffeeScript | [coffee](http://coffeescript.org/), [coffeelint](https://www.npmjs.com/package/coffeelint) |
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ formatting.
 | C | [cppcheck](http://cppcheck.sourceforge.net), [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint), [gcc](https://gcc.gnu.org/), [clang](http://clang.llvm.org/), [clangtidy](http://clang.llvm.org/extra/clang-tidy/) !!, [clang-format](https://clang.llvm.org/docs/ClangFormat.html)|
 | C++ (filetype cpp) | [clang](http://clang.llvm.org/), [clangcheck](http://clang.llvm.org/docs/ClangCheck.html) !!, [clangtidy](http://clang.llvm.org/extra/clang-tidy/) !!, [cppcheck](http://cppcheck.sourceforge.net), [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint) !!, [gcc](https://gcc.gnu.org/), [clang-format](https://clang.llvm.org/docs/ClangFormat.html)|
 | CUDA | [nvcc](http://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html) |
-| C# | [mcs](http://www.mono-project.com/docs/about-mono/languages/csharp/) see:`help ale-cs-mcs` for details,[mcsc](http://www.mono-project.com/docs/about-mono/languages/csharp/) see:`help ale-cs-mcsc` for details and configuration|
+| C# | [mcs](http://www.mono-project.com/docs/about-mono/languages/csharp/) see:`help ale-cs-mcs` for details, [mcsc](http://www.mono-project.com/docs/about-mono/languages/csharp/) !! see:`help ale-cs-mcsc` for details and configuration|
 | Chef | [foodcritic](http://www.foodcritic.io/) |
 | CMake | [cmakelint](https://github.com/richq/cmake-lint) |
 | CoffeeScript | [coffee](http://coffeescript.org/), [coffeelint](https://www.npmjs.com/package/coffeelint) |
--- a/ale_linters/cs/mcsc.vim
+++ b/ale_linters/cs/mcsc.vim
@@ -1,71 +1,72 @@
+" general mcs options which are likely to stay constant across
+" source trees like -pkg:dotnet
 let g:ale_cs_mcsc_options = get(g:, 'ale_cs_mcsc_options', '')
-let g:ale_cs_mcsc_source = get(g:,'ale_cs_mcsc_source','')
-let g:ale_cs_mcsc_assembly_path = get(g:,'ale_cs_mcsc_assembly_path',[])
-let g:ale_cs_mcsc_assemblies = get(g:,'ale_cs_mcsc_assemblies',[])
+
+" path string pointing the linter to the base path of the
+" source tree to check
+let g:ale_cs_mcsc_source = get(g:, 'ale_cs_mcsc_source','.')
+
+" list of search paths for additional assemblies to consider
+let g:ale_cs_mcsc_assembly_path = get(g:, 'ale_cs_mcsc_assembly_path',[])
+
+" list of assemblies to consider
+let g:ale_cs_mcsc_assemblies = get(g:, 'ale_cs_mcsc_assemblies',[])
 function! ale_linters#cs#mcsc#GetCommand(buffer) abort
-	let l:path = ale#Var(a:buffer,'cs_mcsc_assembly_path')
-	if !empty(l:path)
-        if type(l:path) == type('')
-            let l:path = '-lib:' . l:path
-        elseif type(l:path) == type([])
-            let l:path = '-lib:' . join(l:path,',')
-        else
-            throw 'assembly path list must be string or list of path strings'
-        endif
-    elseif type(l:path) != type('') 
-        if type(l:path) != type([])
-            throw 'assembly path list must be string or list of path strings'
-        endif
+
+    " if list of assembly search paths is not empty convert it to
+    " appropriate -lib: parameter of mcs
+    let l:path = ale#Var(a:buffer, 'cs_mcsc_assembly_path')
+
+    if !empty(l:path)
+         let l:path = '-lib:"' . join(l:path, '","') .'"'
+    else
         let l:path =''
     endif
-    let l:assemblies = ale#Var(a:buffer,'cs_mcsc_assemblies')
-	if !empty(l:assemblies)
-        if type(l:assemblies) == type('')
-            let l:assemblies = '-r' . l:assemblies
-        elseif type(l:assemblies) == type([])
-            let l:assemblies = '-r:' . join(l:assemblies,',')
-        else
-            throw 'assembly list must be string or list of strings'
-        endif
-    elseif type(l:assemblies) != type('') 
-        if type(l:assemblies) != type([])
-            throw 'assembly list must be string or list of string'
-        endif
+
+    " if list of assemblies to link is not empty convert it to the
+    " appropriate -r: parameter of mcs
+    let l:assemblies = ale#Var(a:buffer, 'cs_mcsc_assemblies')
+
+    if !empty(l:assemblies)
+        let l:assemblies = '-r:"' . join(l:assemblies, '","') . '"'
+    else
         let l:assemblies =''
     endif
-	let l:base = ale#Var(a:buffer,'cs_mcsc_source')
-	let l:cwd = getcwd()
-	if isdirectory(l:base)
-		exe 'cd ' . l:base
-	elseif empty(l:base) && ( type(l:base) == type('') )
-		let l:base = '.'
-	else
-		throw 'ale_cs_mcs_source must point to an existing directory or empty string for current'
-	endif
-	let l:out = tempname()
-	call ale#engine#ManageFile(a:buffer,l:out)
-    let l:cmd = 'cd ' . l:base . ';'
-    \    . 'mcs -unsafe' 
+
+    " register temporary module target file with ale
+    let l:out = tempname()
+    call ale#engine#ManageFile(a:buffer, l:out)
+
+    " assemble linter command string to be executed by ale
+    " implicitly set -unsafe mcs flag set compilation
+    " target to module (-t:module), direct mcs output to
+    " temporary file (-out)
+    "
+    return 'cd "' . ale#Var(a:buffer, 'cs_mcsc_source') . '";'
+    \    . 'mcs -unsafe'
     \    . ' ' . ale#Var(a:buffer, 'cs_mcsc_options')
     \    . ' ' . l:path
     \    . ' ' . l:assemblies
-	\    . ' -out:' . l:out
+    \    . ' -out:' . l:out
     \    . ' -t:module'
-    \    . ' "' . join(glob('**/*.cs',v:false,v:true),'" "') . '"'
-	exe 'cd ' . l:cwd
-	return l:cmd
+    \    . ' -recurse:"*.cs"'
 endfunction
 
 function! ale_linters#cs#mcsc#Handle(buffer, lines) abort
     " Look for lines like the following.
     "
     " Tests.cs(12,29): error CSXXXX: ; expected
+    "
+    " NOTE: pattern also captures file name as linter compiles all
+    " files within the source tree rooted at the specified source
+    " path and not just the file loaded in the buffer
     let l:pattern = '^\(.\+\.cs\)(\(\d\+\),\(\d\+\)): \(.\+\): \(.\+\)'
     let l:output = []
+    let l:source = ale#Var(a:buffer, 'cs_mcsc_source')
 
     for l:match in ale#util#GetMatches(a:lines, l:pattern)
         call add(l:output, {
-		\   'filename': l:match[1],
+        \   'filename': fnamemodify(l:source . '/' . l:match[1], ':p'),
         \   'lnum': l:match[2] + 0,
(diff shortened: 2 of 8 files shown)


ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,86 @@
...
+    \    . 'mcs -unsafe' 
+    \    . ' ' . ale#Var(a:buffer, 'cs_mcsc_options')
+    \    . ' ' . l:path
+    \    . ' ' . l:assemblies
+	\    . ' -out:' . l:out
+    \    . ' -t:module'
+    \    . ' "' . join(glob('**/*.cs',v:false,v:true),'" "') . '"'
+	exe 'cd ' . l:cwd
+	return l:cmd
+endfunction
```
w0rp: This is some very poorly formatted code. Lose all of the type checking, and use blank lines and comments here or there.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,86 @@
...
+	let l:out = tempname()
+	call ale#engine#ManageFile(a:buffer,l:out)
+    let l:cmd = 'cd ' . l:base . ';'
+    \    . 'mcs -unsafe' 
+    \    . ' ' . ale#Var(a:buffer, 'cs_mcsc_options')
+    \    . ' ' . l:path
+    \    . ' ' . l:assemblies
+	\    . ' -out:' . l:out
+    \    . ' -t:module'
+    \    . ' "' . join(glob('**/*.cs',v:false,v:true),'" "') . '"'
```
w0rp: Don't use `glob` like this. It's painfully slow on many machines.

doc/ale-cs.txt
```diff
@@ -0,0 +1,75 @@
...
+	The mcsc linter uses the mono mcs compiler to generate a temporary module
+  target file (-t:module) including all '*.cs' files contained in the
+	directory by specified by |g:ale_cs_mcsc_source| or |b:ale_cs_mcsc_source|
+  variable and all sub directories. Currently none can be excluded from
+  linting. It uses the assembly directories as specified by 
+	|g:ale_cs_mcsc_assembly_path| or |b:ale_cs_mcsc_assembly_path| and selects
+  the assembly files specified by |g:ale_cs_mcsc_assemblies| or 
+  |b:ale_cs_mcsc_assemblies|. The mcs -unsafe option is set implicitly and has
+  not to be added using |g:ale_cs_mcsc_options| or |b:ale_cs_mcsc_options|
+  variable.
```
w0rp: The formatting of the text is very strange here. It looks like things are randomly indented. I find it hard to read.

doc/ale-cs.txt
```diff
@@ -0,0 +1,75 @@
...
+  Type: |List|
+  Default: `[]`
+
+  This variable defines a list of absolute or relative path strings pointing
+  to the location of the assembly files (*.dll) to be considered by mcsc
+  linter. If the list is not empty the list will be added to the mcsc command
+  line using the -lib: flag of mcs.
+
+g:ale_cs_mcsc_assemblies                            *g:ale_cs_mcsc_assemblies*
+                                                   *b:ale_cs_mcsc_assemblies*
```
w0rp: You need to align your help tags to the right margin.

doc/ale-cs.txt
```diff
@@ -0,0 +1,75 @@
...
+  Type: |List|
+  Default: `[]`
+
+  This variable defines a list of assembly files (*.dll) to be considered by
+  the mono mcs compiler when generating the temporary module. If the list is
+  not empty the list of assemblies will be added to the mcsc command
+  line using the -r: flag of mcs. To change the search path mcs uses to 
+  locate the specified assembly files use |g:ale_cs_mcsc_assembly_path| or
+  |b:ale_cs_mcsc_assembly_path| variables
```
w0rp: Try and explain things more succinctly, and use short sentences.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,86 @@
...
+    return l:output
+endfunction
+
+call ale#linter#Define('cs',{
+\   'name': 'mcsc',
+\   'output_stream': 'stderr',
+\   'executable': 'mcs',
+\   'command_callback': 'ale_linters#cs#mcsc#GetCommand',
+\   'callback': 'ale_linters#cs#mcsc#Handle',
+\	'lint_file': 1
```
w0rp: Bad indentation.

README.md
```diff
@@ -76,7 +76,7 @@ formatting.
 | C | [cppcheck](http://cppcheck.sourceforge.net), [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint), [gcc](https://gcc.gnu.org/), [clang](http://clang.llvm.org/), [clangtidy](http://clang.llvm.org/extra/clang-tidy/) !!, [clang-format](https://clang.llvm.org/docs/ClangFormat.html)|
 | C++ (filetype cpp) | [clang](http://clang.llvm.org/), [clangcheck](http://clang.llvm.org/docs/ClangCheck.html) !!, [clangtidy](http://clang.llvm.org/extra/clang-tidy/) !!, [cppcheck](http://cppcheck.sourceforge.net), [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint) !!, [gcc](https://gcc.gnu.org/), [clang-format](https://clang.llvm.org/docs/ClangFormat.html)|
 | CUDA | [nvcc](http://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html) |
-| C# | [mcs](http://www.mono-project.com/docs/about-mono/languages/csharp/) |
+| C# | [mcs](http://www.mono-project.com/docs/about-mono/languages/csharp/) see:`help ale-cs-mcs` for details,[mcsc](http://www.mono-project.com/docs/about-mono/languages/csharp/) see:`help ale-cs-mcsc` for details and configuration|
```
w0rp: You need to mark this with `!!`, as you've set `lint_file` to `1`.

You need to document this in `doc/ale.txt` too.

hernot:
Ok i have addressed all your recommendations. The only one i at max be able to address partially, if i get some instruction and support form your side is the writing of vader tests. I do not know vader and i do have no docker setup here. So even if able to write a test, i fear I would not be able to verify that it is really testing what it should test. Any advice? Any suggestions?

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,107 @@
...
+    \    . ' ' . ale#Var(a:buffer, 'cs_mcsc_options')
+    \    . ' ' . l:path
+    \    . ' ' . l:assemblies
+    \    . ' -out:' . l:out
+    \    . ' -t:module'
+    \    . ' "' . join(glob('**/*.cs',v:false,v:true),'" "') . '"'
+
+    " switch back to the working directory to the path
+    " stored above.
+    exe 'cd ' . l:cwd
```
w0rp: You shouldn't execute a command to change the working directory. Just include `cd ... &&` in the command you want ALE to run.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,107 @@
...
+    " target to module (-t:module), direct mcs output to
+    " temporary file (-out)
+    "
+    " TODO mcs would have a built in -recurse flag by which
+    " mcs would be able to recurse through the current directory
+    " tree by it self. In the setup using mcs version 5.2.0.215
+    " it seems not to work reliably, which is why vim built in
+    " glob function is used to collect the list of '*.cs'
+    " files.
+    let l:cmd = 'cd ' . l:base . ';'
```
w0rp: You can just use `return` here instead, with the advice above.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,107 @@
...
+    \    . 'mcs -unsafe'
+    \    . ' ' . ale#Var(a:buffer, 'cs_mcsc_options')
+    \    . ' ' . l:path
+    \    . ' ' . l:assemblies
```
w0rp: Try and include these additional arguments only if they aren't empty. Have a look at other linters for examples.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,107 @@
...
+    \    . ' -out:' . l:out
+    \    . ' -t:module'
+    \    . ' "' . join(glob('**/*.cs',v:false,v:true),'" "') . '"'
```
w0rp: `glob` will perform very badly on many machines. Can you pass just the one file instead, or something like that?

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,107 @@
...
+        let l:assemblies = '-r:' . join(l:assemblies,',')
+    else
+        let l:assemblies =''
+    endif
+
+    " remember the current working directory
+    let l:cwd = expand('#' . a:buffer . ':p:h')
+    " get the base path of the source tree to lint and if
+    " not empty and exists switch working directory to it
+    let l:base = ale#Var(a:buffer,'cs_mcsc_source')
```
w0rp: Add a space after commas, here and above.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,107 @@
...
+    if isdirectory(l:base)
```
w0rp: Add a blank line above if statements which do not start a new block.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,107 @@
...
+" source tree to check
+let g:ale_cs_mcsc_source = get(g:,'ale_cs_mcsc_source','')
+
+" list of search paths for additional assemblies to consider
+let g:ale_cs_mcsc_assembly_path = get(g:,'ale_cs_mcsc_assembly_path',[])
+
+" list of assemblies to consider
+let g:ale_cs_mcsc_assemblies = get(g:,'ale_cs_mcsc_assemblies',[])
+function! ale_linters#cs#mcsc#GetCommand(buffer) abort
+
```
w0rp: Delete the blank line here.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,107 @@
+" general mcs options which are likely to stay constant across
+" source trees like -pkg:dotnet
+let g:ale_cs_mcsc_options = get(g:, 'ale_cs_mcsc_options', '')
```
w0rp: You might like to use `ale#Set` for these variables. Look at the codebase for other examples.

hernot:
Hi 
I would appreciate if that workaround would not be necessary but with my installed version of mono mcs compiler the mcs compiler hang nine times out of 10 when trying to use the built in -recurse flag. If it would have worked 10 out of 10 times i would have used that instead of the glob function of vim without resorting to the cd solution if you have better idea for how to properly implement the workaround please let me know.

I decided against including the ale_cs_mcsc_source path inside the glob pattern  to keep command string as short as possible even if several hundred(* files have to be included in the build. 

(*)addmitted a rather poorly organized source tree and and likely indicating of design flaws, still if not all files are included in the linting command (see above) users would receive errors about missing assemblies for classes, types and namespaces defined by source files from the same source tree.

Currently i do have not the time and resources to investigating the reason for mcs hanging when using -recurse flag, in worst submitting a bugreport or even a bugfix to mono project and than adding version checks disabling linter if version with broken -recurse is installed.

hernot:
What the hack, now retrying the -recurse flag every thing works as expected. No idea why, what and where the problem was. Need to keep an eye on this. Never the less for the next commit i simplified to using -recurse flag of mcs compiler. Added a NOTE to the help that the case that mcsc does not indicate errors in the currently editor is related to mcs is a multipass comiler which stops after found errors at a specific comilation stage, not advancing to the next. 

Will have a look into the tests, hope i do not make things even worse than now

hernot:
a question does a test_mcs_handerl.vader not exist or was it not commited as the mcs linter was already present in the github master branch. If not present at all should i also add a vader test for the mcs like i did for the help files or just for the mcsc i have added?

hernot:
Ok added the tests for mcs.vim which was missing them in the master tree and for mcsc.vim
The  continuous-integration/travis-ci/pr test seems from what i can see fail cause i have inserted mcsc in ale-vim.txt file. not sure. please check your self and let me know if that would be the only reason.

hernot:
Finally, fixe some nasty behavior, when vim is not run from base directory of source tree.

test/command_callback/test_cs_mcsc_command_callbacks.vader
```diff
@@ -0,0 +1,86 @@
...
+  unlet! g:ale_cs_mcsc_source
+  unlet! g:ale_cs_mcsc_assembly_path
+  unlet! g:ale_cs_mcsc_assemblies
+  call ale#linter#Reset()
+
+Execute(Check for proper default command):
+
+  let b:command = ale_linters#cs#mcsc#GetCommand(bufnr(''))
+
+  Assert match( b:command, '^\s*cd\s\+"."\s*;\s*mcs\s\+-unsafe\s\+-out:\S\+\s\+-t:module\s\+-recurse:"\*\.cs"\s*$' ) == 0,
```
w0rp: The regular expressions here are very hard to read, so it's hard to make sense of what the tests are testing. I recommend writing tests for exact strings with `AssertEqual` instead. You will probably need to build certain parts of the strings base on the filename, etc.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,85 @@
...
+    return l:output
+endfunction
+
+call ale#linter#Define('cs',{
+\   'name': 'mcsc',
+\   'output_stream': 'stderr',
+\   'executable': 'mcs',
+\   'command_callback': 'ale_linters#cs#mcsc#GetCommand',
+\   'callback': 'ale_linters#cs#mcsc#Handle',
+\    'lint_file': 1
```
w0rp: The indentation is wrong here.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,85 @@
...
+    " Look for lines like the following.
+    "
+    " Tests.cs(12,29): error CSXXXX: ; expected
+    "
+    " NOTE: pattern also captures file name as linter compiles all
+    " files within the source tree rooted at the specified source
+    " path and not just the file loaded in the buffer
+    let l:pattern = '^\(.\+\.cs\)(\(\d\+\),\(\d\+\)): \(.\+\): \(.\+\)'
+    let l:output = []
+    let l:source = ale#Var(a:buffer,'cs_mcsc_source')
```
w0rp: Add a single space after commas.

ale_linters/cs/mcsc.vim
```diff
@@ -0,0 +1,85 @@
...
+    if !empty(l:path)
+         let l:path = '-lib:"' . join(l:path,'","') .'"'
+    else
+        let l:path =''
+    endif
+
+    " if list of assemblies to link is not empty convert it to the
+    " appropriate -r: parameter of mcs
+    let l:assemblies = ale#Var(a:buffer,'cs_mcsc_assemblies')
+    if !empty(l:assemblies)
```
w0rp: Put a single blank line before any if statement which doesn't start a new level of indentation.

test/command_callback/test_cs_mcsc_command_callbacks.vader
hernot: Is in this case not possible cause the GetCommand registers a temporary file with ale, which differs at every call so the assert is the only possibility to make test succeed, unless in your test machinery you have an AssertFunction or some other function which helps to handle exactly that case.

w0rp: I recommend replacing just the temporary file in the actual string with some other sequence of characters. Then you can use an exact string for the expected string.

w0rp: You could substitute `-out:[^ ]*` or something like that with `-out:TEMP` or just anything really.

hernot: you mean 

let b:command = subsitute(b:command,'/\+tmp/\+\S\+','/tmp/some_tmp_file.tmp')

and than use this modified string in the AssertEqual

w0rp: Yeah, something like that. But bear in mind that the directory will be different from `/tmp` on Mac and Windows, so you should probably search for some characters after `'-out:'`.

hernot: Sure i do. I finish the tests and the fixes commit and as soon as they are working properly i let you know

test/command_callback/test_cs_mcsc_command_callbacks.vader
```diff
@@ -0,0 +1,120 @@
...
+  unlet! g:temppath
+  unlet! g:temppathpattern
+  unlte! g:sometempfile
+  call ale#linter#Reset()
+
+Execute(Check for proper default command):
+
+  let b:command = ale_linters#cs#mcsc#GetCommand(bufnr(''))
+  let b:command = substitute(b:command, '-out:' . g:temppathpattern, '-out:' . g:sometempfile, '')
+  let b:command = substitute(b:command, '\s\+', ' ', 'g')
```
w0rp: Add `unlet! b:command` to your `After` block above. Otherwise this variable might end up being used in some other test.

test/command_callback/test_cs_mcsc_command_callbacks.vader
```diff
@@ -0,0 +1,120 @@
...
+  AssertEqual
+  \   b:command,
+  \   'cd ".";mcs -unsafe -out:' . g:sometempfile . ' -t:module -recurse:"*.cs"'
+
+Execute(The options should be be used in the command):
+
+  let g:ale_cs_mcsc_options = '-pkg:dotnet'
+
+  let b:command = ale_linters#cs#mcsc#GetCommand(bufnr(''))
+  let b:command = substitute(b:command, '-out:' . g:temppathpattern, '-out:' . g:sometempfile, '')
```
w0rp: `sometmpefile` will cause problems here with `\` in paths on Windows. I recommend using something simpler like `'-out:TEMP_FILE` and testing for that. You're replacing it anyway.

hernot:
Ok all fixed. 

Also removed Assert match in ale_mcs_handler.vader using a substitute to condense multiple whithespaces into a single space char.

w0rp:
Check out the results in Travis CI. `doc/ale.txt` and `README.md` are out of sync.

hernot:
The not ins sync problem is not that the tables are not in sync, it is just that the test script stumbles over the greediness of regular expressions in line

 53     | sed 's/see.*\(,\|$\)/\1/g' \

This removes in the following line everything starting at the first see till to the end of the line

| C# | [mcs](http://www.mono-project.com/docs/about-mono/languages/csharp/) see:`help ale-cs-mcs` for details, [mcsc](http://www.mono-project.com/docs/about-mono/languages/csharp/) !! see:`help ale-cs-mcsc` for details and configuration|

with the consequence that only mcs is reported.

I suggest to improve line 53 of the test script as follows

 53     | sed 's/see[^,]*\(,\|$\)/\1/g' \

With this change both tables appear in sync as are. Further down in the README.MD sombody solved the same problem by putting the see of the first entry between (). But is that the desired solution as the the () are already used to indicate the links to the compiler, linter project pages. Further it is not consistent with any other see following the last entry of the lists eg.

| Kotlin | [kotlinc](https://kotlinlang.org) !!, [ktlint](https://ktlint.github.io) !! see `:help ale-integration-kotlin` for configuration instructions

There is no () around the see.  What would be the conformant formatting of the table you suggest markdown would require. In any case i would recommend the fix for properly handling of the greediness of regular expressions.

w0rp:
Okay, makes sense. I'll fix it on master.

w0rp:
Cheers! :beers:

w0rp:
I implemented the regex you suggested. I think that'll do the trick for now.

PR #968, opened 2017-09-30
Luacheck: Respect warn_about_trailing_whitespace option

adriaanzon:
See http://luacheck.readthedocs.io/en/stable/warnings.html, warnings 611 to 614.


adriaanzon:
Sure thing, I updated the test to check if it leaves out the warning.

adriaanzon:
Sure thing, I updated the test to check if it leaves out the warning.

adriaanzon:
Good idea 🙂 

I finally found some time to sort this out.

w0rp:
Cheers! :beers:

PR #970, opened 2017-10-01
gofmt fixer for Go

aliou:
Hello, this PR adds `gofmt` as a fixer for Go.

### DIFF
--- a/autoload/ale/fixers/gofmt.vim
+++ b/autoload/ale/fixers/gofmt.vim
@@ -1,0 +1,18 @@
+" Author: aliou <code@aliou.me>
+" Description: Integration of gofmt with ALE.
+
+call ale#Set('go_gofmt_executable', 'gofmt')
+call ale#Set('go_gofmt_options', '')
+
+function! ale#fixers#gofmt#Fix(buffer) abort
+    let l:executable = ale#Var(a:buffer, 'go_gofmt_executable')
+    let l:options = ale#Var(a:buffer, 'go_gofmt_options')
+
+    return {
+    \   'command': ale#Escape(l:executable)
+    \       . '-l -w'
+    \       . (empty(l:options) ? '' : ' ' . l:options)
+    \       . ' %t',
+    \   'read_temporary_file': 1,
+    \}
+endfunction
--- a/autoload/ale/fixers/gofmt.vim
+++ b/autoload/ale/fixers/gofmt.vim
@@ -10,7 +10,7 @@ function! ale#fixers#gofmt#Fix(buffer) abort
 
     return {
     \   'command': ale#Escape(l:executable)
-    \       . '-l -w'
+    \       . ' -l -w'
     \       . (empty(l:options) ? '' : ' ' . l:options)
     \       . ' %t',
     \   'read_temporary_file': 1,
--- a/doc/ale-go.txt
+++ b/doc/ale-go.txt
@@ -50,4 +50,14 @@ g:ale_go_gometalinter_options                   *g:ale_go_gometalinter_options*
 
 
 ===============================================================================
+gofmt                                                            *ale-go-gofmt*
+
+g:ale_go_gofmt_options                                 *g:ale_go_gofmt_options*
+                                                       *b:ale_go_gofmt_options*
+  Type: |String|
+  Default: `''`
+
+  This variable can be set to pass additional options to the gofmt fixer.
+
+===============================================================================
   vim:tw=78:ts=2:sts=2:sw=2:ft=help:norl:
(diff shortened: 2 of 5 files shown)


autoload/ale/fixers/gofmt.vim
```diff
@@ -0,0 +1,18 @@
...
+call ale#Set('go_gofmt_executable', 'gofmt')
+call ale#Set('go_gofmt_options', '')
+
+function! ale#fixers#gofmt#Fix(buffer) abort
+    let l:executable = ale#Var(a:buffer, 'go_gofmt_executable')
+    let l:options = ale#Var(a:buffer, 'go_gofmt_options')
+
+    return {
+    \   'command': ale#Escape(l:executable)
+    \       . '-l -w'
```
w0rp: You're missing a space before `-l` here.

test/fixers/test_gofmt_fixer_callback.vader
```diff
@@ -0,0 +1,40 @@
...
+
+  call ale#test#RestoreDirectory()
+
+Execute(The gofmt callback should return the correct default values):
+  call ale#test#SetFilename('../prettier-test-files/testfile.js')
+
+  AssertEqual
+  \ {
+  \   'read_temporary_file': 1,
+  \   'command': ale#Escape(g:ale_go_gofmt_executable)
```
w0rp: Use a string literal for the executable here instead.

test/fixers/test_gofmt_fixer_callback.vader
```diff
@@ -0,0 +1,40 @@
...
+  \ ale#fixers#gofmt#Fix(bufnr(''))
+
+Execute(The gofmt callback should include custom gofmt options):
+  let g:ale_go_gofmt_options = "-r '(a) -> a'"
+  call ale#test#SetFilename('../prettier-test-files/testfile.js')
+
+  AssertEqual
+  \ {
+  \   'read_temporary_file': 1,
+  \   'command': ale#Escape(g:ale_go_gofmt_executable)
```
w0rp: Use a string literal for the executable here too.

test/fixers/test_gofmt_fixer_callback.vader
```diff
@@ -0,0 +1,40 @@
...
+  call ale#test#RestoreDirectory()
+
+Execute(The gofmt callback should return the correct default values):
+  call ale#test#SetFilename('../prettier-test-files/testfile.js')
+
+  AssertEqual
+  \ {
+  \   'read_temporary_file': 1,
+  \   'command': ale#Escape(g:ale_go_gofmt_executable)
+  \     . '-l -w'
```
w0rp: You missed the space in the test too.

aliou:
Updated with your review. I didn’t add gofmt to the readme or the help files because it is already there as a linter. Should I mention that it also works as a fixer?

w0rp:
Nah, that's okay. I didn't realise it was there.

doc/ale-go.txt
```diff
@@ -50,4 +50,14 @@ g:ale_go_gometalinter_options                   *g:ale_go_gometalinter_options*
 
 
 ===============================================================================
+gofmt                                                            *ale-go-gofmt*
```
w0rp: Check out the Travis CI results. It's complaining because there's no table of contents entry which references this. You should add `gofmt` to the table of contents.

w0rp:
Now Travis is complaining because `gofmt` has to come before `gometalinter` in the table of contents and also in the headings, because everything there is sorted lexicographically. (alphabetically)

aliou:
So that's what I did first and then again in the last commit, but Travis is still failing. Do you know something I might have missed on this?

 On a somewhat related note, I didn’t have these test fail locally because it seems like macOS’s version of `paste` (used in `test/script/check-tock`) doesn’t seem to work when passing data from STDIN and only seems to take a file. What do you think about also running this test in Docker?

Related line : 
https://github.com/w0rp/ale/blob/e0bd490ed9150c8a229f127dcacbcbf97c9f9861/test/script/check-toc#L30

w0rp:
The script is doing the right thing. You need to take your entire `gofmt` section that you added to the Go documentation, and then move that above the `gometalinter` section. One part of the script checks that the table of contents is sorted, and the other checks that the headings and table of contents are in the same order.

w0rp:
I wouldn't expect those scripts to run on Mac, because Macs use an ancient version of Bash and BSD userland stuff. I might be able to make them compatible some day, at the cost of some sanity, or just get them to run in Docker.

aliou:
Got it. Updated both the TOC and the order in the help files, and squash the relevant commits. LMK if something else is needed.

w0rp:
Cheers! :beers:

aliou:
Cool 🎉 thanks!

PR #974, opened 2017-10-03
Add all possible config files for prettier

odinuge:
Prettier uses cosmiconfig, and therefore it is possible to add different
extensions to the config file. More information can be found here:
https://github.com/prettier/prettier#configuration-file.

### DIFF
--- a/autoload/ale/fixers/prettier.vim
+++ b/autoload/ale/fixers/prettier.vim
@@ -10,6 +10,10 @@ call ale#Set('javascript_prettier_options', '')
 function! s:FindConfig(buffer) abort
     for l:filename in [
     \   '.prettierrc',
+    \   '.prettierrc.json',
+    \   '.prettierrc.yaml',
+    \   '.prettierrc.yml',
+    \   '.prettierrc.js',
     \   'prettier.config.js',
     \   'package.json',
     \ ]


autoload/ale/fixers/prettier.vim
```diff
@@ -10,6 +10,10 @@ call ale#Set('javascript_prettier_options', '')
 function! s:FindConfig(buffer) abort
     for l:filename in [
     \   '.prettierrc',
+    \   '.prettierrc.yaml',
+    \   '.prettierrc.yml',
+    \   '.prettierrc.json',
+    \   '.prettierrc.js',
     \   'prettier.config.js',
     \   'package.json',
     \ ]
```
w0rp: We should follow the same precedence that cosmiconfig uses, which is mentioned [here](https://github.com/davidtheclark/cosmiconfig#rcextensions) and implemented [here](https://github.com/davidtheclark/cosmiconfig/blob/master/src/loadRc.js).

I reckon the correct order is.

1. No extension
2. `.json`
3. `.yaml`
4. `.yml`
5. `.js`
6. `.config.js`
7. `package.json`

Rather confusingly, ESLint has a similar set of configuration files... with a completed different order of precedence. https://eslint.org/docs/user-guide/configuring#configuration-file-formats

odinuge: Ahh, that is true! Updated the commit to reflect this list.

It turns out that `package.json` is being matched first, but ONLY if it has a `"prettier": {}` element inside it. I think this is an acceptable soultion, even though the best would have been to use the command `prettier --find-config-path [filename]` to get the config directly from prettier.

w0rp:
Is that `--find-config-path` option a new one? We should use that at some point in the future.

I should perhaps implement command chains for fixing things, like I did for linters, so we can just check which version of `prettier` you are using and then use the appropriate option.

The `prettier` developers pretty much screwed up by not supporting configuration files from day one.

I'll merge what you have here for now.

w0rp:
Cheers! :beers:

PR #979, opened 2017-10-06
Add llc integration for LLVM IR

rhysd:
I added [llc](https://llvm.org/docs/CommandGuide/llc.html) support for linting LLVM IR code. I added implementation, documents and tests. I confirmed all tests I added were passing.
llc is an official compiler provided by LLVM. So I added it as a default linter when opening LLVM IR source file.


ale_linters/llvm/llc.vim
```diff
@@ -0,0 +1,31 @@
+" Author: rhysd <https://rhysd.github.io>
+" Description: Support for checking LLVM IR with llc
+
+call ale#Set('llvm_llc_command', 'llc')
+
+function! ale_linters#llvm#llc#GetCommand(buffer) abort
+    return ale#Var(a:buffer, 'llvm_llc_command')
+    \   . ' -filetype=null -o='
+    \   . ale#Escape(g:ale#util#nul_file)
```
w0rp: I just merged a pull request which used `/dev/null` on Windows for elm. Do you know of `nul` will actually work for `-o` on Windows? I wouldn't be surprised if yet another tool has special code for exactly the string `/dev/null`, but doesn't know what to do with `nul`.

autoload/ale/linter.vim
```diff
@@ -25,6 +25,7 @@ let s:default_ale_linters = {
 \   'csh': ['shell'],
 \   'go': ['gofmt', 'golint', 'go vet'],
 \   'help': [],
+\   'llvm': ['llc'],
```
w0rp: You can remove this line. Without any entries in the Dictionary here, ALE will run all available tools for linting. The only reason to specify something here is to _disable_ some linter by default.

doc/ale-llvm.txt
```diff
@@ -0,0 +1,19 @@
+===============================================================================
+ALE LLVM Integration                                           *ale-llvm-options*
```
w0rp: I can't tell if the tags are aligned to the right margin properly or not on GitHub, but they might not be. Make sure they are.

ale_linters/llvm/llc.vim
```diff
@@ -0,0 +1,31 @@
+" Author: rhysd <https://rhysd.github.io>
+" Description: Support for checking LLVM IR with llc
+
+call ale#Set('llvm_llc_command', 'llc')
```
w0rp: Name this variable `_executable` instead, so it matches the naming convention used for all other linters.

ale_linters/llvm/llc.vim
```diff
@@ -0,0 +1,31 @@
...
+    \   'lnum': str2nr(v:val[1]),
+    \   'col': str2nr(v:val[2]),
+    \   'text': v:val[3],
+    \   'type': 'E',
+    \}")
+endfunction
+
+call ale#linter#Define('llvm', {
+\   'name': 'llc',
+\   'executable': g:ale_llvm_llc_command,
```
w0rp: Write a `GetExecutable` function for this, use `exectuable_callback` instead, and use `ale#Var` for the variable. Then users will be able to configure the executable per-buffer, and while Vim is running. Call that function from `GetCommand`, and apply `ale#Escape` to the executable path in `GetCommand`, which could contain spaces and other troublesome characters.

rhysd:
Thank you for detailed reviews @w0rp! I'll fix the points.

ale_linters/llvm/llc.vim
rhysd: I'm usually not using Windows. But I dug LLVM source code a bit. Filesystem support is implemented here and llc seems to use this.

https://github.com/llvm-mirror/llvm/blob/master/include/llvm/Support/FileSystem.h

As long as I read the code, LLVM does not handle `/dev/null` specially. So I think `nul` is available but `/dev/null` is unavailable on Windows.

autoload/ale/linter.vim
rhysd: I'll remove the line, thanks.

doc/ale-llvm.txt
rhysd: I did not notice that `===` line is longer than title. I'll align titles.

ale_linters/llvm/llc.vim
rhysd: I'll rename the variable and add `GetCommand` callback.

rhysd:
@w0rp I believe I've fixed all points. Could you confirm?

ale_linters/llvm/llc.vim
```diff
@@ -0,0 +1,35 @@
...
+" Description: Support for checking LLVM IR with llc
+
+call ale#Set('llvm_llc_executable', 'llc')
+
+function! ale_linters#llvm#llc#GetExecutable(buffer) abort
+    return ale#Var(a:buffer, 'llvm_llc_executable')
+endfunction
+
+function! ale_linters#llvm#llc#GetCommand(buffer) abort
+    return ale_linters#llvm#llc#GetExecutable(a:buffer)
```
w0rp: Use `ale#Escape` on the executable here. Otherwise, paths with spaces and other characters won't work.

test/command_callback/test_llc_command_callback.vader
```diff
@@ -0,0 +1,24 @@
...
+  runtime ale_linters/llvm/llc.vim
+
+After:
+  unlet! g:ale_llvm_llc_executable
+  unlet! b:ale_llvm_llc_executable
+  Restore
+
+Execute(llc command is customizable):
+  let cmd = ale_linters#llvm#llc#GetCommand(bufnr(''))[:3]
+  AssertEqual 'llc ', cmd
```
w0rp: When testing the executable names, use `ale#Escape` on them here. This will make the tests pass in both Unix and Windows environments. (Windows escaping is very different, and may not add quotes at all in some cases.)

rhysd:
@w0rp Thank you for catching the point. I fixed implementation and tests using `ale#Escape`. And I added another test to check the path is escaped correctly.

test/command_callback/test_llc_command_callback.vader
```diff
@@ -0,0 +1,29 @@
...
+  AssertEqual stridx(cmd, ale#Escape('llc-5.0')), 0
+
+  let b:ale_llvm_llc_executable = 'llc-4.0'
+  let cmd = ale_linters#llvm#llc#GetCommand(bufnr(''))
+  AssertEqual stridx(cmd, ale#Escape('llc-4.0')), 0
+
+Execute(space can be contained in llc executable path):
+  let b:ale_llvm_llc_executable = '/path/space contained/llc'
+  let cmd = ale_linters#llvm#llc#GetCommand(bufnr(''))
+  AssertEqual stridx(cmd, ale#Escape('/path/space contained/llc')), 0
```
w0rp: I will probably end up changing these tests later, as the failure message will print `-1 != 0`, which doesn't tell you much of anything.

ale_linters/llvm/llc.vim
```diff
@@ -0,0 +1,35 @@
+" Author: rhysd <https://rhysd.github.io>
+" Description: Support for checking LLVM IR with llc
+
+call ale#Set('llvm_llc_executable', 'llc')
+
+function! ale_linters#llvm#llc#GetExecutable(buffer) abort
+    return ale#Escape(ale#Var(a:buffer, 'llvm_llc_executable'))
```
w0rp: Nope, don't escape this here. The return value of the callbacks for getting the executable must not be a escaped, but the executable should be escaped in the command callback.

test/command_callback/test_llc_command_callback.vader
```diff
@@ -0,0 +1,29 @@
...
+
+  runtime ale_linters/llvm/llc.vim
+
+After:
+  unlet! g:ale_llvm_llc_executable
+  unlet! b:ale_llvm_llc_executable
+  Restore
+
+Execute(llc command is customizable):
+  let cmd = ale_linters#llvm#llc#GetCommand(bufnr(''))
```
w0rp: Test the return value of `GetExecutable` too, and ensure it is not escaped.

test/command_callback/test_llc_command_callback.vader
rhysd: You're correct. But how can I write a test to check the string is prefixed? Is there any way to specify assertion message? I'm not familiar with vader.vim. So could you please help to improve these tests?

rhysd: Ah, I found that message can be specified. I'll improve message. https://github.com/junegunn/vader.vim#execute

ale_linters/llvm/llc.vim
rhysd: Ah, I got it.

test/command_callback/test_llc_command_callback.vader
rhysd: I'll do.

rhysd:
@w0rp I fixed the points in c7c1c8e and e179661. Could you check them?

test/command_callback/test_llc_command_callback.vader
```diff
@@ -0,0 +1,38 @@
+Before:
+  Save g:ale_llvm_llc_executable
+
+  unlet! g:ale_llvm_llc_executable
+  unlet! b:ale_llvm_llc_executable
+
+  runtime ale_linters/llvm/llc.vim
+
+  function! AssertHasPrefix(str, prefix) abort
```
w0rp: You'll need to delete this function in the `After` block, with `delfunction`. Otherwise it will be accessible from other tests.

rhysd: Fixed in 1e72204 :+1:

test/command_callback/test_llc_command_callback.vader
```diff
@@ -14,6 +14,7 @@ Before:
 After:
   unlet! g:ale_llvm_llc_executable
   unlet! b:ale_llvm_llc_executable
+  delfunction! AssertHasPrefix
```
w0rp: `delfunction!` doesn't work. The command doesn't accept a bang. (`!`)

rhysd: ah, it was my mistake. And `:Vader` did not report the error. I modified commit at 1e72204 and `push -f`ed again.

rhysd:
@w0rp Thank you for reviewing many points! :smile:

PR #982, opened 2017-10-08
Add g:ale_highlight_expand

laino:
I don't like having the sign column, so I only use highlighting.

Some linters however only return 1 character which will be highlighted -
easy to miss when you don't have a sign column:

</img>

This pull request adds a `g:ale_highlight_expand` configuration,
allowing you to specify a regular expression used to highlight additional characters.

In this example I used `let g:ale_highlight_expand = '\>'`

</img>

Much better.

The expression gets applied when only a single character would be
highlighted and is disabled by default.

I don't really know any vimscript, but hopefully I didn't mess it up too bad.

### DIFF
--- a/autoload/ale/highlight.vim
+++ b/autoload/ale/highlight.vim
@@ -31,10 +31,10 @@ function! ale#highlight#CreatePositions(line, col, end_line, end_col) abort
         " If only a single character would be highlighted, and there's an
         " expand expression, apply it to increase the amount of text that
         " would be highlighted.
-        if g:ale_highlight_expand != '0' && a:col == a:end_col
+        if g:ale_highlight_expand !=# '0' && a:col == a:end_col
             let l:line = getline(a:line)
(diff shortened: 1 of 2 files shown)


laino:
Travis tells me to use `!=#` instead of `!=`, only to then ask me to use `isnot#` instead of `!=#`.

Travis is an advanced troll.

w0rp:
Nah, this isn't the way to handle this problem.

The way to handle this is to find the linters which do not indicate where the end columns for problems are, and supply that information there. Many tools output lines, columns, end lines, and end columns. For others, the symbols in the error messages can be parsed instead.

laino:
Meh. Won't be coming from me.

This works well for me and I'll keep using it until someone has gone and invested
countless hours integrating every linter properly or implementing hacky stuff like 
parsing error messages (which is worse IMHO).

Maybe if it were configurable on a per-linter basis? 

</img>

PR #990, opened 2017-10-10
Use $VIRTUAL_ENV environment variable when present.

monokrome:
This modifies the virtualenv detection to use the $VIRTUAL_ENV variable when it is in the environment. I'm not sure what other cases exist, so I left the other code as a fallback.

I will totally write tests if this looks like a reasonable suggestion, but I don't know the test language so don't want to figure it out if this doesn't look useful.

Thanks for the dope plugin! ❤️

### DIFF
--- a/autoload/ale/python.vim
+++ b/autoload/ale/python.vim
@@ -71,7 +71,7 @@ function! ale#python#FindVirtualenv(buffer) abort
         endfor
     endfor
 
-    return ''
+    return $VIRTUAL_ENV
 endfunction
 
 " Run an executable check for Python scripts.
--- a/test/test_python_virtualenv.vader
+++ b/test/test_python_virtualenv.vader
@@ -1,0 +1,12 @@
+Before:
+  Save $VIRTUAL_ENV
+  let $VIRTUAL_ENV = "/opt/example/"
+
+After:
+  Restore
+
+Execute(ale#python#FindVirtualenv falls back to $VIRTUAL_ENV when no directories match):
+  AssertEqual
+  \  ale#python#FindVirtualenv(bufnr('%')),
+  \  '/opt/example/',
+  \  'Expected VIRTUAL_ENV environment variable to be used, but it was not'


w0rp:
This might be an idea. One thing we have to consider is what should take precedence, the environment variable, or the local directories. Consider starting Vim from one virtualenv, and then editing files which use another virtualenv. Will the code here result in running the wrong version of `flake8` in some cases?

Maybe ALE could search for directories first, and try the environment variable last. Then we could document the behavior and write tests for that. That's probably safe to do.

w0rp:
Maybe that's another matter of configuration, where you could want either behavior.

w0rp:
Another option would be to add support for absolute paths for the current `ale_virtualenv_dir_names` option, and then recommend setting `let g:ale_virtualenv_dir_names = [$VIRTUAL_ENV]` when you just want that.

monokrome:
@w0rp That makes sense, I'll change it and add tests. Do you have any specific suggestion on where the tests should go? 🙉

monokrome:
Oh, also I thought that it made sense here because it would solve the problem for anyone else who happens to not keep their virtualenv directories in their source directory. For instance, virtualenvwrapper users would need this in order for the integration to work 😺

w0rp:
Anywhere reasonable will do. There are a few tests which use virtualenv directory names already. Have a look through the current tests.

w0rp:
Yeah, the current solution only works for people who create virtualenv directories, ignored, inside their source trees. Your solution should help with people who put virtualenv in some other directory.

I wonder if it might be best to add support for absolute paths to the existing option, and then change the default value so it has `$VIRTUAL_ENV` at the end of it. Then it should pick up on virtualenv by default, without breaking anything people are using currently, apart from replacing global versions with virtualenv versions.

monokrome:
@w0rp I changed it to use $VIRTIUAL_ENV as a fallback, but your solution seems good! Is it safe to have an empty string in that list if $VIRTUAL_ENV is an empty string?

w0rp:
Then again, the option is used for a breadth-first search, so maybe just go with what you have no, where the environment variable is just used automatically on the end. You can always override it with the existing `use_global` options.

w0rp:
I think what you have now is best. Go with that.

monokrome:
Thanks @w0rp! I'll add tests later ❤️

monokrome:
@w0rp I added a test and *I think* I did the right thing. Let me know if it sucks ❤️

monokrome:
btw this unit testing framework is super cool and exciting :D

PR #1003, opened 2017-10-17
Modifications to location_list open behavior

JZL:
I put 2 related commits together:

[(1)](https://github.com/w0rp/ale/commit/40bf9a05763bbde4eb6d29b28413cab2b8bc6fcd): I use `ale` constantly but found the constant opening/closing of the quickfix/location_list to be annoying (which is, ironically, only because it's so fast async that it caught my semicolon errors before I could type them!). This modification along with the existing param to keeping the sign gutter open makes it a lot less jumpy.

[(2)](https://github.com/w0rp/ale/commit/ee4c53a3c5ed5b49c75da169cdb21b6ed1e20a5c): This makes the error menu change height to fit the number of errors, up to a max_height parameter (it uses the existing `list_window_size` which defaults to 10 max).

I thought a cool addition to # 2 would be to make the last item of the quickfix/location_list menu be "..." so the user knows there are more errors but since this is my first vimscript project, didn't know how to change that

I can make the necessary changes to the quickfix menu as well but thought I would get location_list feedback first

I really enjoy ale, it makes me 100x faster at catching bugs. Thanks so much for making it

----

I'll be happy to fix the tests but don't know how to best integrate it (if at all).

In my mind, there would be another parameter to enable the responsive-height. I don't know if anyone wants the always-open location_list, in which case, it just live on my fork

### DIFF
--- a/autoload/ale/list.vim
+++ b/autoload/ale/list.vim
@@ -90,33 +90,35 @@ function! s:SetListsImpl(timer_id, buffer, loclist) abort
     "
     " We'll check if the current buffer's List is not empty here, so the
     " window will only be opened if the current buffer has problems.
-    if s:ShouldOpen(a:buffer) && !empty(a:loclist)
-        let l:winnr = winnr()
-        let l:mode = mode()
-        let l:reset_visual_selection = l:mode is? 'v' || l:mode is# "\<c-v>"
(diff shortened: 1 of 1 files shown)


w0rp:
Nah, use the option for keeping the list window open. I won't support this.

Issue #852, opened 2017-08-16
Support formatting with error codes


w0rp:
For quickfix and loclist in Vim, you can set `nr` for an error code, and this error code will be displayed in windows, etc. However, strings are not accepted for `nr`, and most sensible tools use human-readable names for errors instead of numbers. Currently, most of the linters show the error code in the windows and in echo messages by placing the error codes inside of the `text` keys for the dictionaries like so:

```vim
{'text': 'some-error-code: Something is wrong !'}
```

The format of the dictionaries should be changed so the code is stored separately from the message.

```vim
{'code': 'some-error-code', 'text': 'Something is wrong!'}
```

Then a format function could be applied to both the eventual items which are used in loclist and quickfix, and the same format function could be applied to echo messages, like it is now, which would include the error codes by default.

w0rp:
`%...code...%` can be used for the error code now, where surrounding characters appear as written if there is an error code set. All of the linters need to be updated to set the `code` key for the items. The function for fixing the items needs to be updated to pass the `code` key on. The comparison functions for de-duplicating items need to be modified to include the codes.

w0rp:
For `pylint` I just threw away the `msgid` values like `E0103` and went with the names like `not-in-loop` which are actually human-readable. If someone complains I'll add an option for it.

w0rp:
`pycodestyle` errors and warnings are now style errors and warnings, except `E999`, which is the code for syntax errors.

w0rp:
I've modified just about every linter I could easily modify to capture the error codes. I'm going to call this good now. If anyone wants to configure error codes for other linters, they can open a pull request for it.

jorgehatccrma:
The README should be updated to include `%...code...%` as a possible formatting variable

w0rp:
I updated the README now. Good suggestion. :+1:

wxianxin:
I would like to have pylint show absolute `msg_id` instead of symbolic name(prefer `E0103` over `not-in-loop`), please add this option. Reason: It is much more easier to look up the exact error.

I tried to do this by let g:ale_python_pylint_options = '--msg-template="{msg_id}:{line:3d},{column}: {obj}: {msg}" --disable "C0103,W0621"'. I don't why it does not work.

If ale_python_pylint_options has only limited support for pylint. I think it is better to clarify in the doc.

Thanks.

mr-ubik:
@w0rp is there a way to show the old `msg-id`?

w0rp:
[...]

PR #1009, opened 2017-10-18
phpcs: include sniff code in message

sirbrillig:
When troubleshooting phpcs warnings/errors, it is often helpful to know the exact "sniff code" that represents that notice (eg: `Generic.Files.LowercasedFilename.NotFound`), but the default implementation of the ALE phpcs module hides that code, and just prints the text part of the message.

This PR includes the sniff code in the message instead.

## Screenshots

Current behavior:
phpcs-without-code

With this change
phpcs-with-code


sirbrillig:
That sounds good to me! I only added the option because I saw that the error code was being explicitly ignored, so I didn't want to break any intentional behavior. I'll make the change shortly.

sirbrillig:
@w0rp I removed the option. I retained the separation of the sniff code and the message text, in case we want to use that for something like #852 

I can squash these commits (or they can be squashed on merge) if this looks good to you.

sirbrillig:
@w0rp Sure thing (good opportunity to learn Vader... 😄 ). I added a test and squashed the old commits.

PR #1013, opened 2017-10-20
[Linter: hamllint] Add var for passing options to `hamllint`

thenoseman:
This pull request adds the possibility to pass options to `hamllint`.

For example when you set `let g:ale_haml_hamllint_options = '-c ~/use-my.haml-lint.yml'` the following command will be run:
```
haml-lint -c ~/use-my.haml-lint.yml /some/tmp/file
```

Cheers!


doc/ale-haml.txt
```diff
@@ -0,0 +1,32 @@
...
+Options
+-------------------------------------------------------------------------------
+
+g:ale_haml_hamllint_options                       *g:ale_haml_hamllint_options*
+
+  Type: |String|
+  Default: `''`
+
+  This variable can be set and will be passed to haml-lint as options
+  eg. let g:ale_haml_hamllint_options = '-c ~/use-my.haml-lint.yml'
```
w0rp: This could use some a bit of formatting for the example. I recommend ending your sentence on "options" with a full stop (or "period" in American English), and then write "For example:" and a little code block. Have a look at `:help g:ale_enabled`. That's one option which has some formatting like that.

ale_linters/haml/hamllint.vim
```diff
@@ -1,5 +1,17 @@
...
+" Options:
+" g:ale_haml_hamllint_options can be set and will be passed as options to
+" haml-lint
+" eg. let g:ale_haml_hamllint_options = '-c ~/use-my.haml-lint.yml'
+
+let g:ale_haml_hamllint_options =
+\   get(g:, 'ale_haml_hamllint_options', '')
+
+function! ale_linters#haml#hamllint#GetCommand(buffer) abort
+    return 'haml-lint ' . g:ale_haml_hamllint_options . ' %t'
```
w0rp: I recommend only adding the extra space on the left if the options aren't empty. That's better for tests. Have a look at some existing linters for examples.

thenoseman:
Implemented the requested changes.
Thanks for taking the time and building this plugin!

w0rp:
I just noticed that this pull request is related to issue #1035.

Is there an easy to recognize standard set of filenames for the `haml-lint` configuration file? If so, it would probably be a good idea to automatically search for configuration files, similar to what ALE does already for other linters like `JSHint`.

w0rp:
Another option that might would then would be prefixing the command with `cd`. Search the codebase for existing uses of `ale#path#BufferCdString`.

thenoseman:
If this PR is good to go I'll give #1035 a try if you like.

w0rp:
I think hold off on this until the configuration file issue is fixed. I wouldn't want to encourage users to set configuration files via the option implemented here if the configuration file can be used automatically.

thenoseman:
So I should remove the options altogether and replace them with a call to `ale#path#FindNearestFile` for `.haml-lint.yml` **and** `.rubocop.yml`?

That would also fix #1035 .

Totally fine with me.

w0rp:
@thenoseman Yeah, that sounds about right to me. That's what the slim-lint linter does. If there's another reason to have some options, it might be worth keeping them.

w0rp:
I'll close this pull request for now, and we could re-open it if we want options for some other reason.

PR #1018, opened 2017-10-21
Pretter fixer support for Less filetype

zkourouma:
adds support for prettier fixer for less filetype. includes test and doc file (but i don't know if the doc is necessary since its not a linter)

http://lesscss.org/


w0rp:
Cheers! :beers:

zkourouma:
thanks!

Issue #732, opened 2017-07-06
Can't get fixer to work


gregeinfrank:
ALEInfo:
```
 Current Filetype: vim
Available Linters: ['vint']
  Enabled Linters: ['vint']
 Linter Variables:

let g:ale_vim_vint_show_style_issues = 1
 Global Variables:

let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'E'
let g:ale_echo_msg_format = '[%linter%] %s [%severity%]'
let g:ale_echo_msg_warning_str = 'W'
let g:ale_enabled = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 0
let g:ale_lint_on_text_changed = 'always'
let g:ale_linter_aliases = {}
let g:ale_linters = {'jsx': ['eslint', 'flow'], 'hcl': [], 'ruby': ['ruby', 'rubocop'], 'javascript': ['eslint', 'flow'], 'python': ['flake8']}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_column_always = 1
let g:ale_sign_error = '🚫'
let g:ale_sign_offset = 1000000
let g:ale_sign_warning = '✋'
let g:ale_statusline_format = ['%d error(s)', '%d warning(s)', 'OK']
let g:ale_warn_about_trailing_whitespace = 1
  Command History:
```

Ale-fixer configs (not in ALEInfo, see #731 ):
```vim
let g:ale_fixers = {
\   'javascript': ['prettier'],
\   'ruby': ['rubocop'],
\}
let g:ale_enabled = 1
let g:ale_fix_on_save = 1
```

I can't seem to get this to do anything when I save - I also don't see the command in the command history from `ALEInfo`, both for rubocop and prettier. 

I'm super pumped to use the fix feature, but I just must be doing something wrong! Thanks :)

w0rp:
The command isn't currently added to the history for fixing files. Can you run either command from the command line?

w0rp:
You can also use the `:ALEFix` command to fix things. I personally use this instead of fixing files on save.

gregeinfrank:
@w0rp what do you mean by "Can you run either command from the command line"? 

[...]

PR #1029, opened 2017-10-22
Use Rubocop config also when linting

grddev:
This fixes the same problem as in #732 but for linters, and while doing so aligns the option generation for the linter with how the Rubocop fixer works.


grddev:
> Is this needed? I believe the `--stdin` option receives a filename, and the filename tells rubocop where to find the configuration file.

Weird, when I test it right now this indeed seems to be the case. I was convinced that it didn't work without it yesterday, but I guess it was caused by something else.

Sorry for the noise.

w0rp:
Okay, sounds good.

PR #1030, opened 2017-10-22
Update test scripts to work on OS X

grddev:
There were a couple of issues
 - `paste` requires a file argument
 - `mktemp` requires a pattern argument
 - `sort` doesn't support `-h`, but `-n` is enough for sorting on numbers, and `-s` was introduced to perform a stable sort instead.

The main issues were that BSD `sed` does not support:
 - Alternation (`\|`) - solved by splitting to multiple patterns
 - Bound shortcuts (`x\+`, `x\?`) - solved by replacing with `xx*` and `x\{0,1\}` respectively
 - Lower-casing (`\L`) - solved by piping through `tr` instead (this will lowercase everything and not only the integration names, but I assumed that wasn't too much of an issue, as a portable alternative for the selective downcasing would be much more involved).

### DIFF
--- a/test/script/check-supported-tools-tables
+++ b/test/script/check-supported-tools-tables
@@ -30,30 +30,30 @@ readme_section_size="$( \
 # shellcheck disable=SC2003
 readme_end_line="$(expr "$readme_start_line" + "$readme_section_size")"
 
-doc_file="$(mktemp)"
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

PR #1034, opened 2017-10-23
Handle multiple files localtion lists with gcc handler

### LINKED ISSUE #1023: Catch the filename in gcc handlers to have multi-file location list
Partly related to #985 (For the 'take care of the first matching group when appending to a location list" part)

As we see [here](https://github.com/w0rp/ale/blob/master/autoload/ale/handlers/gcc.vim#L103), the items added to location lists in ALE are not using the filename. The problem with this behaviour is that it becomes impossible to properly handle errors where the filename is actually different from the linted file. This is why we can't have a location list that will make us jump to the error/warning like Syntastic is able to do for example.

It would be nice to find a way to include the first match in gcc handlers in the ALE:items, so that the location list can be more precise and useful. To try and be more clear, I have created [basic tests](https://github.com/w0rp/ale/compare/master...Gagbo:loclist_multi-file_c_handling) to show what I mean and what I expect. They are not finished yet, and will be polished and completed for other linters once the main feature is done.

gagbo:
It will be easier to discuss code and features to fix #1023 in the Pull Request.

### The feature
As stated in the issue, the goal is to have a location list that populates with all the correct locations for errors/warnings, even for included files.

This is a basic implementation, I do not consider myself advanced enough in c/cpp to know what  other features we want from the handler. Some things we may want are : 
- Filtering out the warnings from included files (and keep only the errors) to limit clutter in output
- Add a gutter sign at the location where the included file is included in the linted code.

### The test
- Created a test in test_c_projects/included_error to see the behaviour that we want.
- There are 2 errors, each (theoretically) with a note; one is directly in the cpp file, and the other is hidden in the included file.
- The vader only covers tests for clangcheck checker, I'm not even sure yet this is a good way to test the feature. This is something to do better.

### The implementation
Using the 'filename' key of the l:items dictionary properly, since it's already captured by the handler's regex.

The gutter sign could be added by storing the line of inclusion like it was done before in gcc handler, I'm not sure yet of the best way to add a gutter sign "linked" to another error (i.e. with another filename,lnum and colnum to preserve jump accuracy) in the output from the linter.

### DIFF
--- a/autoload/ale/handlers/gcc.vim
+++ b/autoload/ale/handlers/gcc.vim
@@ -105,10 +105,17 @@ function! ale#handlers#gcc#HandleGCCFormat(buffer, lines) abort
                 continue
             endif
 
+            " If the 'error type' is a note, make it detail related to
+            " the previous error parsed in output
+            if l:match[4] is# 'note'
+                let l:output[-1]['detail'] = s:RemoveUnicodeQuotes(l:match[0])
+                continue
+            endif
+
             let l:item = {
             \   'filename': l:match[1],
             \   'lnum': str2nr(l:match[2]),
-            \   'type': l:match[4] =~# 'error' ? 'E' : (l:match[4] =~# 'note' ? 'I' : 'W'),
+            \   'type': l:match[4] =~# 'error' ? 'E' : 'W',
             \   'text': s:RemoveUnicodeQuotes(l:match[5]),
             \}
 
--- a/autoload/ale/handlers/gcc.vim
+++ b/autoload/ale/handlers/gcc.vim
@@ -88,6 +88,11 @@ function! ale#handlers#gcc#HandleGCCFormat(buffer, lines) abort
             " let l:included_filename = l:match[1]
             " call add(l:include_lines, l:line)
         else
+            " TODO(gagbo) : check if l:match[4] is note
+            " If it is, add the item as a detail to the previous item
+            " If it isn't, store the info of l:item in case we need to append
+            " detail later
+
             " If we hit a regular error again, then add the previously
             " collected lines as one error, and reset the include variables.
             call s:AddIncludedErrors(l:output, l:include_lnum, l:include_lines)
--- a/autoload/ale/handlers/gcc.vim
+++ b/autoload/ale/handlers/gcc.vim
@@ -87,8 +87,11 @@ function! ale#handlers#gcc#HandleGCCFormat(buffer, lines) abort
             " Hint : We probably only need to check l:include_lnum value since
             " it should be non-null only if the gutter is needed.
 
+            " The ternary operator in filename filters out the 'dummy'
+            " filenames like <nopath> or <stdin> and leave the location
+            " handling to engine#FixLocList
             let l:item = {
-            \   'filename': l:match[1],
+            \   'filename': (l:match[1][:0] is# '<') ? '' : l:match[1],
             \   'lnum': str2nr(l:match[2]),
             \   'type': l:match[4] is# 'error' ? 'E' : 'W',
             \   'text': s:RemoveUnicodeQuotes(l:match[5]),
--- a/autoload/ale/handlers/gcc.vim
+++ b/autoload/ale/handlers/gcc.vim
@@ -101,6 +101,7 @@ function! ale#handlers#gcc#HandleGCCFormat(buffer, lines) abort
             endif
 
             let l:item = {
+            \   'filename': l:match[1],
             \   'lnum': str2nr(l:match[2]),
             \   'type': l:match[4] =~# 'error' ? 'E' : 'W',
             \   'text': s:RemoveUnicodeQuotes(l:match[5]),
--- a/autoload/ale/handlers/gcc.vim
+++ b/autoload/ale/handlers/gcc.vim
@@ -43,6 +43,10 @@ endfunction
 
 function! ale#handlers#gcc#HandleGCCFormat(buffer, lines) abort
     let l:include_pattern = '\v^(In file included | *)from ([^:]*):(\d+)'
+    " Include pattern looks for lines like :
+    "
+    " In file included from test.h:1:0,
+    "                  from test.cpp:1:
     let l:include_lnum = 0
     let l:include_lines = []
     let l:included_filename = ''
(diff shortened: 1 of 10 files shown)


autoload/ale/handlers/gcc.vim
```diff
@@ -58,51 +58,42 @@ function! ale#handlers#gcc#HandleGCCFormat(buffer, lines) abort
...
             endif
 
+            " If the 'error type' is a note, make it detail related to
+            " the previous error parsed in output
+            if l:match[4] is# 'note'
+                if has_key(l:output[-1], 'detail')
+                    let l:output[-1]['detail'] .= s:RemoveUnicodeQuotes(l:match[0]) . ' --'
+                else
+                    let l:output[-1]['detail'] = s:RemoveUnicodeQuotes(l:match[0]) . ' --'
+                endif
```
w0rp: You can put this all in one line.

```vim
let l:output[-1]['detail'] = get(l:output[-1], 'detail', '') 
\    . s:RemoveUnicodeQuotes(l:match[0]) . ' --'
```

Gagbo:
I also changed the majority of the previous tests that started bugging now since the output dictionary now has a 'filename' key.

On a more important note, the tests that used to check that ALE added the gutter sign at the included_line with the "Errors in header" messages are now changed to just add the location of the error in the included file. The tested behaviour here does not have a gutter sign at the line of the included file (since now the pointed locations are directly at the error).

The last error in CI is related to the test I added, but correcting the other tests showed me how to properly setup the linter calls in vader.

test/handler/test_swiftlint_handler.vader
```diff
@@ -4,12 +4,14 @@ Execute(The swiftint handler should parse error messages correctly):
...
   \     'lnum': 1,
   \     'col': 7,
+  \     'filename': '<nopath>',
   \     'text': 'Operator Usage Whitespace Violation: Operators should be surrounded by a single whitespace when they are being used. (operator_usage_whitespace)',
   \     'type': 'W',
   \   },
   \   {
   \     'lnum': 1,
   \     'col': 11,
+  \     'filename': '<nopath>',
```
w0rp: These filenames will probably mess things up. What I recommend is checking for "filenames" starting with `'<'` in the callback, and don't set the filenames in those cases. Then ALE will see neither a `'filename'` nor a `'bufnr'` key, and assume that the problem is for the current buffer.

An easy startswith check is `l:x[:0] is# '<'`.

You can read `ale#engine#FixLocList` to see exactly what ALE does.

test/handler/test_vint_handler.vader
```diff
@@ -10,40 +10,46 @@ Execute(The vint handler should parse error messages correctly):
...
   \     'col': 8,
+  \     'filename': 'gcc.vim',
   \     'end_col': 15,
   \     'text': 'Make the scope explicit like ''l:filename'' (see Anti-pattern of vimrc (Scope of identifier))',
   \     'type': 'W',
   \   },
   \   {
   \     'lnum': 7,
   \     'col': 8,
+  \     'filename': 'gcc.vim',
```
w0rp: It's hard to tell with test data, but try to set absolute paths if you can. The real output might actually use absolute paths.

w0rp:
I wrote on the issue also, we should consider whether or not it's a good idea to keep a sign where the include is.

test/handler/test_vint_handler.vader
Gagbo: With clangcheck at least, the output contains the absolute path so there's no need to worry too much.

Maybe vint outputs relative paths, but then I don't know if we should expect the linters to have options to output absolute path, or if we should add another function in the gcc handler script to find the absolute path (maybe `expand(filename:p)` works ?)

test/handler/test_swiftlint_handler.vader
Gagbo: I agree, I just didn't know how to make this work properly/read it.

test/handler/test_vint_handler.vader
w0rp: I think Vint only works with a single file, I wouldn't worry about Vint.

w0rp: For most linters, if we needed to work out the absolute path ourselves, the way to do that is to determine what directory the paths are actually relative to, and then resolve the relative paths based on that path.

Tools which output absolute paths are better.

gagbo:
I think most of the work is done for the time being.

Proper handling of the gutter signs and the include lines can be finished with the C/C++ header features.

My main problem to wrap things up is that I just don't know how I broke this one test, and only for vim.

test/handler/test_vint_handler.vader
gagbo: The paths are given as relative to the location where the linter is launched, that is the current working directory of vim.

Even if I `:lcd` to another directory (and the loclist still shows the same relative path to the starting vim working directory), jumping to other files works as intended. I think it's safe to assume that vim handles these changes on its own

once again, the output is messy but please don't mind the errors, it's from clang_complete that I need to learn to configure properly https://asciinema.org/a/0DchjjNMt8xxuiuyEqOhS9oqP

w0rp: I wouldn't assume that.

The handlers for the errors can be executed from some buffer other than the buffer the command was run from. That's why `getcwd()` is blacklisted in the codebase. It can give the wrong output.

gagbo:
Crude method (`excutable --help | grep absol`) seems to point that clang can print absolute paths : 
```
-fdiagnostics-absolute-paths
                          Print absolute paths in diagnostics
```

But I had no luck with gcc.

w0rp:
It might be worth setting the option just for `clang` itself. We could try that later. I'd take something over nothing.

w0rp:
I think you can fix the failing test here, sort out the TODO issues or remove the TODO comments, and then I can merge this.

gagbo:
The exact commit that breaks the test (passed before, does not after) is [replacing a ==# with a is#](https://github.com/w0rp/ale/pull/1034/commits/9b9740098a7731bf81fa9e85412d3310b1d471e0) in the gcc handler. I don't know why it started triggering here (maybe a change in vim version used by CI ?) but : 

- I fixed the pattern in the test to set `magic` and to match through first whitespace characters (checking the output from the test I had a few of them in the lines that needed matching)
- And the test would not work because of localisation : my pattern shows french translation for `line` and `name`, so the pattern used in the test is not valid unless I change it to french names, or if I find the way to change my locale for the duration of the test (not sure the latter works)

w0rp:
The sign parsing issue is a common issue. The actual regex ALE uses for parsing signs is a little different from the test there.

w0rp:
Cheers! :beers:

w0rp:
I fixed that test for other locales now by copy and pasting what ALE uses for parsing signs for real, and by removing one of the capture groups.

PR #1036, opened 2017-10-23
Add `detail` key to haskell ghc handler

rob-b:
Add `detail` key to the dict returned from the haskell ghc handler


autoload/ale/handlers/haskell.vim
```diff
@@ -72,6 +72,7 @@ function! ale#handlers#haskell#HandleGHCFormat(buffer, lines) abort
         \   'col': l:match[3] + 0,
         \   'text': l:text,
         \   'type': l:type,
+        \   'detail': join(a:lines, "\n"),
```
w0rp: That's very excessive detail. That adds _all_ of the lines of output to every item.

What you should do instead is add just some specific lines of output, and then cover your changes here with tests.

autoload/ale/handlers/haskell.vim
```diff
@@ -66,12 +66,14 @@ function! ale#handlers#haskell#HandleGHCFormat(buffer, lines) abort
 
         " Replace temporary filenames in problem messages with the basename
         let l:text = substitute(l:text, l:temp_filename_regex, l:basename, 'g')
+        let l:detail = map(copy(a:lines), {key, val -> substitute(val, l:temp_filename_regex, l:basename, 'g')})
```
w0rp: You can't commit lambdas to the codebase, as they don't work in NeoVim 0.1.7. You could use strings with `v:val` instead. See the documentation for `map`.

You're copying every line of output here into the a detail list. That's too much information. If there's some extra information in lines nearby, you could parse and include just those lines instead.

Try starting with a Vader test with what you expect first, work backwards to the implementation from that. It's hard to see what you're trying to achieve here.

w0rp:
If you're going to create a pull request, you should at least write a short description explaining what it is you're trying to do.

rob-b:
I see your point about the number of lines in `ALEDetail` could get long and require pagination. To be honest, I just copied the way detail was defined in other linters.

There's two issues which would make it a bit hard to trim the output

1. I don't think there's a good heuristic for selecting which parts of the output are the most pertinent

2. The most recent version of GHC improved error formatting https://downloads.haskell.org/~ghc/8.2.1/docs/html/users_guide/8.2.1-notes.html#highlights
Overall that's a good thing but unfortunately neither of the linters I use (hdevtools + ghc-mod) work with ghc 8.2 right now and so it'd be difficult to be sure that any changes I make now to omit content selective won't break the handler in the future since I can't easily test the new output format with those linters.

Overall, if you'd prefer not to include any detail fields that could get too long then I'm happy to close the PR and handle this in some other way

autoload/ale/handlers/haskell.vim
rob-b: I double checked the docs for `ALEDetail` and they say [Show the full linter message for the current line](https://github.com/w0rp/ale/blob/960ae62aaa30d26fcfff57d6593e54a34a93ec68/doc/ale.txt#L1450-L1455). I realise that you wrote those docs and are of course free to change your mind about what `ALEDetail` should entail. As I mentioned in my other comment though, there's no real way to easily determine what part of a given error is most appropriate.

For example, given this error output

```
Entities.hs:49:23: error:
    • Couldn't match type ‘()’ with ‘b -> c’
      Expected type: a -> b -> c
        Actual type: a -> ()
    • Possible cause: ‘runSqlPool’ is applied to too many arguments
      In the second argument of ‘($)’, namely
        ‘runSqlPool (runMigration migrateAll) pool’
      In the expression: flip $ runSqlPool (runMigration migrateAll) pool
      In an equation for ‘migrate’:
          migrate pool = flip $ runSqlPool (runMigration migrateAll) pool
    • Relevant bindings include
        migrate :: Pool SqlBackend -> b -> a -> c
          (bound at Entities.hs:49:1)
Entities.hs:53:17: error:
    • Could not deduce: a ~ [IO a]
      from the context: (HasSpock m, SpockConn m ~ SqlBackend)
        bound by the type signature for:
                   runSQL :: (HasSpock m, SpockConn m ~ SqlBackend) =>
                             SqlPersistT (LoggingT IO) a -> m a
        at Entities.hs:52:1-86
      ‘a’ is a rigid type variable bound by
        the type signature for:
          runSQL :: forall (m :: * -> *) a.
                    (HasSpock m, SpockConn m ~ SqlBackend) =>
                    SqlPersistT (LoggingT IO) a -> m a
        at Entities.hs:52:11
      Expected type: m a
        Actual type: [SqlBackend] -> [IO a]
    • In the expression:
        map $ \ conn -> runStdoutLoggingT $ runSqlConn action conn
      In an equation for ‘runSQL’:
          runSQL action
            = map $ \ conn -> runStdoutLoggingT $ runSqlConn action conn
    • Relevant bindings include
        action :: SqlPersistT (LoggingT IO) a
          (bound at Entities.hs:53:8)
        runSQL :: SqlPersistT (LoggingT IO) a -> m a
          (bound at Entities.hs:53:1)
```

For the first example I would find the second bullet point most useful. For the second example I would find the first bullet point most useful.

As to the `map` call, the intention is apply the regex that removes the temp filename to each line of the output. I read the help page (it's how I discovered lamdas exist in vimscript) and I don't see how it would work using `v:val`. I am happy write it out in long form in a loop so that it's compatible with neovim.

Overall I'm cool with making the lambda change and of course fixing the failing tests (I actually wrote the tests for this handler in the first place) but just want to check where you stand on the detail level. From my perspective as a haskell user, I'd prefer the full detail level and having to page through the output if it gets longer than a page (still much preferable to switching to the term and running the linter by hand to get the output) but if that doesn't fit your vision of what you want `ALEDetail` to do then I'm fine with that too

w0rp: My point is that you are putting _the entire output_ in _every item_. You should put the details for _just that error_ in there.

You want this for the first error.

```
Entities.hs:49:23: error:
    • Couldn't match type ‘()’ with ‘b -> c’
      Expected type: a -> b -> c
        Actual type: a -> ()
    • Possible cause: ‘runSqlPool’ is applied to too many arguments
      In the second argument of ‘($)’, namely
        ‘runSqlPool (runMigration migrateAll) pool’
      In the expression: flip $ runSqlPool (runMigration migrateAll) pool
      In an equation for ‘migrate’:
          migrate pool = flip $ runSqlPool (runMigration migrateAll) pool
    • Relevant bindings include
        migrate :: Pool SqlBackend -> b -> a -> c
          (bound at Entities.hs:49:1)
```

Not the rest of the output.

rob-b: Ok I understand you now. To me, when you say "That's too much information. If there's some extra information in lines nearby, you could parse and include just those lines instead." it sounded more like you just had a distaste for long messages. But the issue is that each line reports _all_ errors in the file when it should really report only the errors for that line. 

Am I correct in thinking that you have no issue with how long the detail gets as long as it correctly reports only the issues for the current line?

w0rp: Yeah, the problem is that the code you had was taking the entire output and adding it to every item. You can find the lines between the one error and the next, or the end, and chuck that in instead.

w0rp: Remember to cover your changes with Vader tests.

w0rp:
I have handled this now. The lines of output starting from the error message are added to the `detail` key.

PR #1037, opened 2017-10-23
proselint for mail files

cswingle:
This is a very simple pull to add proselint to mail files.  I didn't see any tests for the other file types using proselint (asciidoc, rst, etc.), so I didn't add any here either.


w0rp:
Cheers! :beers:

Issue #866, opened 2017-08-22
Support linting Makefiles


iv597:
It'd be pretty useful to be able to lint Makefiles, perhaps through something like https://github.com/mrtazz/checkmake

If nobody else more qualified takes this up, if time allows any time soon, I may take a stab at it.

PR #1042, opened 2017-10-24
Add checkmake (#866)

aurieh:
Adds checkmake linter for `make` file type.
One annoying thing is that checkmake, for some reason, uses DOS line endings (`\r\n`), so vim displays `^M` in quickfix/loclist. Should I just substitute it?

### DIFF
--- a/ale_linters/make/checkmake.vim
+++ b/ale_linters/make/checkmake.vim
@@ -1,0 +1,24 @@
+" Author: aurieh - https://github.com/aurieh
+
+function! ale_linters#make#checkmake#Handle(buffer, lines) abort
+    let l:pattern = '\v^(\d+):(.+):(.+)$'
+    let l:output = []
+
+    for l:match in ale#util#GetMatches(a:lines, l:pattern)
+        let l:text = l:match[2] . ': ' . l:match[3]
+        call add(l:output, {
+        \    'bufnr': a:buffer,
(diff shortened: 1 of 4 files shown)


ale_linters/make/checkmake.vim
```diff
@@ -0,0 +1,24 @@
+" Author: aurieh - https://github.com/aurieh
+
+function! ale_linters#make#checkmake#Handle(buffer, lines) abort
```
w0rp: You might be able to use the standard Unix error handler here instead. `ale#handlers#unix#HandleAsError` might work here.

aurieh: Standard Unix error handler regex doesn' seem to match. I could probably modify checkmake's `--format` flag to make it work.

aurieh: Actually, I take that back: I've tried multiple format string combinations, still can't get the Unix handler to work.

w0rp: You could either make it match, or write add Vader tests for the new function. Either works for me.

w0rp: Maybe do the latter, keeping this in mind: #852

aurieh: I'll try to write a Vader test. Should I just pass a static input string or?

w0rp:
Yep, that works. Cheers! :beers:

PR #1043, opened 2017-10-24
example for 'rcfile' on ale_python_pylint_options

fengdasuk19:
Two hyphens instead of one will enable the option `rcfile`.


w0rp:
Cheers! :beers:

PR #1046, opened 2017-10-25
add tflint linter for Terraform files

### LINKED ISSUE #963: Support for tflint
Hi,

Could you add support for [tflint](https://github.com/wata727/tflint) when Terraform file are written?

Thanks

natw:
fixes #963 

https://github.com/wata727/tflint

### DIFF
--- a/ale_linters/terraform/tflint.vim
+++ b/ale_linters/terraform/tflint.vim
@@ -4,8 +4,8 @@
 " See: https://www.terraform.io/
 "      https://github.com/wata727/tflint
 
-let g:ale_terraform_tflint_options = get(g:, 'ale_terraform_tflint_options' ,'-f json')
-let g:ale_terraform_tflint_executable = get(g:, 'ale_terraform_tflint_executable', 'tflint')
+call ale#Set('terraform_tflint_options', '')
+call ale#Set('terraform_tflint_executable', 'tflint')
 
 function! ale_linters#terraform#tflint#Handle(buffer, lines) abort
     let l:output = []
@@ -30,10 +30,21 @@ function! ale_linters#terraform#tflint#Handle(buffer, lines) abort
 endfunction
 
 function! ale_linters#terraform#tflint#GetCommand(buffer) abort
-    return printf('%s %s %%t',
-    \   ale#Var(a:buffer, 'terraform_tflint_executable'),
-    \   escape(ale#Var(a:buffer, 'terraform_tflint_options'), '~')
-    \)
+    let l:cmd = ale#Escape(ale#Var(a:buffer, 'terraform_tflint_executable'))
+
+    let l:config_file = ale#path#FindNearestFile(a:buffer, '.tflint.hcl')
+    if !empty(l:config_file)
+        let l:cmd .= ' --config ' . ale#Escape(l:config_file)
+    endif
+
+    let l:opts = ale#Var(a:buffer, 'terraform_tflint_options')
+    if !empty(l:opts)
+        let l:cmd .= ' ' . l:opts
+    endif
+
+    let l:cmd .= ' -f json'
+
+    return l:cmd
 endfunction
 
 call ale#linter#Define('terraform', {
@@ -42,3 +53,5 @@ call ale#linter#Define('terraform', {
 \   'command_callback': 'ale_linters#terraform#tflint#GetCommand',
 \   'callback': 'ale_linters#terraform#tflint#Handle',
 \})
+
+" vim:sw=4
--- a/ale_linters/terraform/tflint.vim
+++ b/ale_linters/terraform/tflint.vim
@@ -29,6 +29,10 @@ function! ale_linters#terraform#tflint#Handle(buffer, lines) abort
     return l:output
 endfunction
 
+function! ale_linters#terraform#tflint#GetExecutable(buffer) abort
+    return ale#Var(a:buffer, 'terraform_tflint_executable')
+endfunction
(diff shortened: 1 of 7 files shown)


ale_linters/terraform/tflint.vim
```diff
@@ -0,0 +1,44 @@
...
+        \   'type': l:type,
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+function! ale_linters#terraform#tflint#GetCommand(buffer) abort
+    return printf('%s %s %%t',
+    \   ale#Var(a:buffer, 'terraform_tflint_executable'),
```
w0rp: If the executable is configurable, apply `ale#Escape` to it. Otherwise names with spaces and other things can cause problems.

ale_linters/terraform/tflint.vim
```diff
@@ -0,0 +1,44 @@
+" Author: Nat Williams <nat.williams@gmail.com>
+" Description: tflint for Terraform files
+"
+" See: https://www.terraform.io/
+"      https://github.com/wata727/tflint
+
+let g:ale_terraform_tflint_options = get(g:, 'ale_terraform_tflint_options' ,'-f json')
```
w0rp: You can use `ale#Set` for your variables here.

w0rp: `-f json` is essential to the linter, so don't make that part configurable. You could use an empty string by default.

ale_linters/terraform/tflint.vim
w0rp: You could only add a space before the options value if the options aren't empty. See many other linters for examples.

ale_linters/terraform/tflint.vim
```diff
@@ -0,0 +1,44 @@
...
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+function! ale_linters#terraform#tflint#GetCommand(buffer) abort
+    return printf('%s %s %%t',
+    \   ale#Var(a:buffer, 'terraform_tflint_executable'),
+    \   escape(ale#Var(a:buffer, 'terraform_tflint_options'), '~')
```
w0rp: The `escape()` call here puts a backslash before `~` characters. You can just not call `escape()` here. Users should handle escaping for options.

ale_linters/terraform/tflint.vim
```diff
@@ -0,0 +1,44 @@
...
+        \   'text': l:error.message,
+        \   'type': l:type,
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+function! ale_linters#terraform#tflint#GetCommand(buffer) abort
+    return printf('%s %s %%t',
```
w0rp: It will be easier to just concatenate things, instead of using `printf`, so you can only add spaces between things when they are set.

ale_linters/terraform/tflint.vim
```diff
@@ -0,0 +1,44 @@
...
+        \   'lnum': l:error.line,
+        \   'text': l:error.message,
+        \   'type': l:type,
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+function! ale_linters#terraform#tflint#GetCommand(buffer) abort
```
w0rp: You'll need to find the configuration file and set it in the command if one exists. You can use the `ale#path#FindNearestFile` function. Otherwise the configuration file won't be loaded in some cases.

ale_linters/terraform/tflint.vim
```diff
@@ -0,0 +1,44 @@
...
+    return printf('%s %s %%t',
+    \   ale#Var(a:buffer, 'terraform_tflint_executable'),
+    \   escape(ale#Var(a:buffer, 'terraform_tflint_options'), '~')
+    \)
+endfunction
+
+call ale#linter#Define('terraform', {
+\   'name': 'tflint',
+\   'executable': 'tflint',
```
w0rp: You'll need to set up an `executable_callback` here if the executable is configurable, and then add a Vader test for it.

test/command_callback/test_terraform_tflint_command_callback.vader
```diff
@@ -0,0 +1,24 @@
...
+Execute(The default command should be good):
+  AssertEqual
+  \ 'tflint -f json %t',
+  \ ale_linters#terraform#tflint#GetCommand(bufnr(''))
+
+Execute(Overriding things should work):
+  let g:ale_terraform_tflint_executable = 'fnord'
+  let g:ale_terraform_tflint_options = '--whatever'
+  AssertEqual
+  \ 'fnord --whatever %t',
```
w0rp: You can apply `ale#Escape` to the test here too. That should work on both Linux and Windows.

test/test_tflint_config_detection.vader
```diff
@@ -0,0 +1,18 @@
...
+  call ale#test#RestoreDirectory()
+  call ale#linter#Reset()
+
+Execute(adjacent config file should be found):
+  call ale#test#SetFilename('tflint-test-files/foo/bar.tf')
+  AssertEqual
+  \ (
+  \   ale#Escape('tflint')
+  \   . ' --config '
+  \   . ale#Escape(ale#path#Winify(g:dir . '/tflint-test-files/foo/.tflint.hcl'))
```
w0rp: You have to use `git add -f` for the file, as all dot files are ignored by default.

I'll do that.

w0rp:
I have merged this now.

Cheers! :beers:

PR #1049, opened 2017-10-26
Do not change 'runtimepath'

thinca:
A plugin must not change 'runtimepath' implicitly.
It is plugin manager's work.

### DIFF
--- a/plugin/ale.vim
+++ b/plugin/ale.vim
@@ -32,9 +32,6 @@ if !s:has_features
     finish
 endif
 
-" Add the after directory to the runtimepath
-let &runtimepath .= ',' . expand('<sfile>:p:h:h') . '/after'
-
 " Set this flag so that other plugins can use it, like airline.
 let g:loaded_ale = 1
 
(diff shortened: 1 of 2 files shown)


w0rp:
Ideally no, but there's no other way to get this to work for everyone.

w0rp:
I have pushed a commit here so you can stop ALE from modifying `runtimepath` if it bothers you: https://github.com/w0rp/ale/commit/f15c8f41279b8b8d9e397e55021d31ca983da3b0

The `after` plugin is only needed to warn users about incompatible plugins.

thinca:
NO!
The `after` directory is automatically added by plugin manager.  If not, it is a plugin manager's bug.
So the `after` directory duplicates in the `'runtimepath'` if you add it.
This occurs some problems in other plugins.
All plugins except plugin manager MUST not change the `'runtimepath'`.
Please revert before commit and merge this.

w0rp:
Sorry to disappoint, but ALE will continue to modify `runtimepath`. If you don't want that, set the option to disable that behaviour.

I'm more than willing to break from conventions if it means I have far fewer GitHub issues to deal with, because ALE is able to consistently warn users about installing incompatible plugins.

PR #1053, opened 2017-10-26
Add hackfmt fixer

SamHowie:
Adding hackfmt fixer for Hack

### DIFF
--- a/autoload/ale/fixers/hackfmt.vim
+++ b/autoload/ale/fixers/hackfmt.vim
@@ -1,0 +1,18 @@
+" Author: Sam Howie <samhowie@gmail.com>
+" Description: Integration of hackfmt with ALE.
+
+call ale#Set('php_hackfmt_executable', 'hackfmt')
+call ale#Set('php_hackfmt_options', '')
+
+function! ale#fixers#hackfmt#Fix(buffer) abort
+    let l:executable = ale#Var(a:buffer, 'php_hackfmt_executable')
+    let l:options = ale#Var(a:buffer, 'php_hackfmt_options')
+
+    return {
+    \   'command': ale#Escape(l:executable)
+    \       . ' -i'
+    \       . (empty(l:options) ? '' : ' ' . l:options)
+    \       . ' %t',
+    \   'read_temporary_file': 1,
+    \}
+endfunction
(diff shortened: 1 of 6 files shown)


mitermayer:
Awesome!!!

autoload/ale/fixers/hackfmt.vim
```diff
@@ -0,0 +1,17 @@
+" Author: Sam Howie <samhowie@gmail.com>
+" Description: Integration of hackfmt with ALE.
+
+call ale#Set('php_hackfmt_executable', 'hackfmt')
+call ale#Set('php_hackfmt_options', '-i')
```
w0rp: What is the meaning of `-i`? If it's essential to running the program, you might leave that out of configuration.

test/fixers/test_hackfmt_fixer_callback.vader
```diff
@@ -0,0 +1,38 @@
...
+
+Execute(The hackfmt callback should include custom hackfmt options):
+  let g:ale_php_hackfmt_options = "--skip-children"
+  call ale#test#SetFilename('../hack_files/testfile.php')
+
+  AssertEqual
+  \ {
+  \   'read_temporary_file': 1,
+  \   'command': ale#Escape('xxxinvalid')
+  \     . ' ' . g:ale_php_hackfmt_options
```
w0rp: Write a string literal here instead, rather than repeating the variable.

fredemmott:
Would it be practical to use `git diff | hackfmt --diff` instead when in a git repository, + the hg equivalent? This will make hackfmt only run on the modified lines rather than the whole file.

Otherwise, `-i`, is required - it's basically equivalent to the `-i` flag in sed

autoload/ale/fixers/hackfmt.vim
SamHowie: `-i` id the in-place option. It tells hackfmt to format the file in place.

It is essential, so I will move it out of the configuration.

test/fixers/test_hackfmt_fixer_callback.vader
SamHowie: Will do!

SamHowie:
@fredemmott, I like your idea!

I looked into implementing it and ran into a few difficulties which I need to figure out.

I propose that we keep this pull request simple and add the `hackfmt --diff` optimization afterwards.

PR #1054, opened 2017-10-27
Avoid duplicate ale/after entry in runtimepath with check

rhysd:
`ale/after` is duplicate in `runtimepath`. This is already reported in #1049. But as @w0rp explained, sometimes this hack is required to make ale work. 

So I implemented that it checks if `runtimepath` already contains `ale/after` in `runtimepath`, and if it does not exist, then adds it. I think this will fix the duplicate problem and does not break current behavior.

What do you think @w0rp and @thinca?


plugin/ale.vim
```diff
@@ -33,17 +33,20 @@ if !s:has_features
 endif
 
 " Add the after directory to the runtimepath
+let s:sep = has('win32') ? '\' : '/'
 let s:should_add_after_dir = 1
 for s:path in split(&runtimepath, ',')
-    if s:path =~# '\/ale\/after$'
+    if s:path =~# s:sep . 'after$' &&
```
mattn: This should check with `[\/]after$` instead since windows accepth both of the separators. For example, I often use slash to set runtimepath.

rhysd: ok, thank you. I'll fix it. And @thinca also gave me additional fix points.

plugin/ale.vim
```diff
@@ -32,13 +32,22 @@ if !s:has_features
...
+        let s:should_add_after_dir = 0
+        break
+    endif
+endfor
+unlet! s:sep
+unlet! s:path
+if s:should_add_after_dir
     let &runtimepath .= ',' . expand('<sfile>:p:h:h') . '/after'
 endif
+unlet! s:should_add_after_dir
```
w0rp: This is too complicated. Just search for something like `'ale[/\\]after'` in `runtimepath` instead.

w0rp: `match('ale[/\\]after$', &runtimepath) >= 0` should be enough.

w0rp:
```vim
if g:ale_emit_conflict_warnings && match('ale[/\\]after$', &runtimepath) < 0
```

w0rp:
I did it.

w0rp:
https://github.com/w0rp/ale/commit/4d29b0bc71fd02c07b28cf3022b7fe55991e53fa There's the commit.

rhysd:
@w0rp 

Thank you for the work. But I don't think the fix is sufficient. Please consider when ale is installed into `~/.vim` directly. The `after` directory is put as `~/.vim/after/plugin/ale.vim`. 4d29b0b cannot detect the case.

w0rp:
I don't think it matters. The script will run twice and finish on the `finish` line.

w0rp:
https://github.com/w0rp/ale/blob/master/after/plugin/ale.vim That's all that's in that directory, and that's all that will ever be there.

rhysd:
Um, ok. I think 4d29b0b must be better than previous. And I understood your policy. Thank you for your comments.

PR #1063, opened 2017-10-30
Add dart LSP (resolves #1006)

### LINKED ISSUE #1006: Add support for Dart LSP
Related: https://github.com/w0rp/ale/issues/1002 (dartanalyzer is too slow)

Non-official implementations of a langserver plugin for Dart have been built:
* https://github.com/natebosch/vim-lsc 
* https://github.com/prabirshrestha/vim-lsp

I don't know much about LSP, but I'm assuming the perfs are similar for any language implementation. Might also be easier to implement than https://github.com/w0rp/ale/issues/1005

aurieh:
Don't merge this yet: as discussed in #1006 comments, I can't get this to work on NeoVim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ formatting.
 | CSS | [csslint](http://csslint.net/), [stylelint](https://github.com/stylelint/stylelint), [prettier](https://github.com/prettier/prettier) |
 | Cython (pyrex filetype) | [cython](http://cython.org/) |
 | D | [dmd](https://dlang.org/dmd-linux.html) |
-| Dart | [dartanalyzer](https://github.com/dart-lang/sdk/tree/master/pkg/analyzer_cli) !! |
+| Dart | [dartanalyzer](https://github.com/dart-lang/sdk/tree/master/pkg/analyzer_cli) !!, [language_server](https://github.com/natebosch/dart_language_server) |
 | Dockerfile | [hadolint](https://github.com/lukasmartinelli/hadolint) |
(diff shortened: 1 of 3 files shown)


w0rp:
If you rebase your code now, LSP support for NeoVim should be fixed. The messages weren't being handled correctly.

I was also testing with the following patch, which might be worth considering with a `use_global` option.

```diff
diff --git a/ale_linters/dart/language_server.vim b/ale_linters/dart/language_server.vim
index 15c7701..6010c2d 100644
--- a/ale_linters/dart/language_server.vim
+++ b/ale_linters/dart/language_server.vim
@@ -4,6 +4,12 @@
 call ale#Set('dart_language_server_executable', 'dart_language_server')

 function! ale_linters#dart#language_server#GetExecutable(buffer) abort
+    let l:home_path = $HOME . '/.pub-cache/bin/dart_language_server'
+
+    if executable(l:home_path)
+        return l:home_path
+    endif
+
     return ale#Var(a:buffer, 'dart_language_server_executable')
 endfunction
```

natebosch:
We typically expect `pub global activate` packages to be in `$PATH` (and there is a message during that command if the directory is missing from the user's path) - I'd hope that we can avoid hardcoding `/.pub-cache/bin` since some users have multiple pub cache directories that they switch between.

aurieh:
Yeah, pub library binaries should be in the `PATH` at all times.
@w0rp I've rebased the code, seems to be working just fine 👍

w0rp:
Cheers! :beers:

PR #1068, opened 2017-10-31
Put info about lightline-ale in README.md (fixes #1065)

### LINKED ISSUE #1065: Wrapped recommended lightline config in a plugin
The [current section in FAQ](https://github.com/w0rp/ale#faq-lightline) has some minor issues and will pretty much be identical for everyone, so I decided to wrap it in a plugin: [maximbaz/lightline-ale](https://github.com/maximbaz/lightline-ale)

I propose to replace that FAQ section with just a reference to this plugin, and a note saying that if users want something more sophisticated, look at the sources of the plugin.

If you don't want to do this, feel free to close.

Links: #493, #1012.

-----

Fixed issues with the current example:

- Both `LightlineLinterWarnings` and `LightlineLinterErrors` return non-empty result based on `l:counts.total`, instead they should look at number of warnings and errors correspondingly.
- There's no need to compute `l:all_non_errors` in `LightlineLinterErrors`
- There's no need to compute `l:all_errors` and `l:all_non_errors` in `LightlineLinterOK`

maximbaz:
Fixes #1065

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,61 +418,9 @@ See `:help ale#statusline#Count()` for more information.
 ### 5.v. How can I show errors or warnings in my lightline?
 
 [lightline](https://github.com/itchyny/lightline.vim) does not have built-in
-support for ALE, nevertheless it's easy to do it yourself:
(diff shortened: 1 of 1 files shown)


w0rp:
Cheers! :beers:

PR #1071, opened 2017-10-31
windows compatible warning match pattern for erlc

selliera:
The pattern did search for lines starting with ([^:]+) as the filename,
however under windows, the file name erlc ouputs is like:
c:/Users/user/AppData/Local/Temp/XXX.tmp/file.erl
Simplifying the patters as it is done for the syntaxerl checker avoids
the issue.

I haven't found tests for this part, and I'm not proficient enough in vader to add a new one from scratch, but I can work on it provided some guidance.

### DIFF
--- a/ale_linters/erlang/erlc.vim
+++ b/ale_linters/erlang/erlc.vim
@@ -17,7 +17,7 @@ function! ale_linters#erlang#erlc#Handle(buffer, lines) abort
     " error.erl:4: variable 'B' is unbound
     " error.erl:3: Warning: function main/0 is unused
     " error.erl:4: Warning: variable 'A' is unused
-    let l:pattern = '\v^([^:]+):(\d+): (Warning: )?(.+)$'
+    let l:pattern = '\v\C:(\d+): (Warning: )?(.+)$'
 
     " parse_transforms are a special case. The error message does not indicate a location:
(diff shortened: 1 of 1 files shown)


ale_linters/erlang/erlc.vim
```diff
@@ -17,7 +17,7 @@ function! ale_linters#erlang#erlc#Handle(buffer, lines) abort
     " error.erl:4: variable 'B' is unbound
     " error.erl:3: Warning: function main/0 is unused
     " error.erl:4: Warning: variable 'A' is unused
-    let l:pattern = '\v^([^:]+):(\d+): (Warning: )?(.+)$'
+    let l:pattern = '\v\C:(\d+): (Warning: )?(.+)$'
```
w0rp: I wonder, can `erlc` return the names of other included files?

selliera: Indeed it is possible in theory.
Due to how erlc is used in the checker with a temp file, this is unlikely (and causes unwanted errors, such as printing warnings about undefined behaviors when using user defined behaviors).

We could modify the pattern to enable extracting the file name, for instance starting the pattern with ` \v^([a-zA-Z]?:?[^:]+)` instead of `\v^([^:]+)`.
This way we keep the ability to use the file name in the future.

PR #1072, opened 2017-11-01
Prevent logging of errors in the php linter.

drrcknlsn:
This change prevents the `php -l` linter from filling error logs with pointless transient errors caused by editing in environments where PHP is configured to log errors.


w0rp:
Cheers! :beers:

PR #1083, opened 2017-11-05
Add shfmt fixer for sh files

sbugert:
I added support for fixing sh files with [shftm](https://github.com/mvdan/sh).
shfmt can also lint files but I have not yet added this feature.

Also, this PR lacks tests. Are there already shell test files in the repo? I guess I would need some assistance for writing tests.


w0rp:
Cheers! :beers:

PR #1093, opened 2017-11-08
ale_linters: add 'dafny' linter

ttaylorr:
Here is a patch to add a linter for the Dafny programming language (https://rise4fun.org/Dafny).

This is my first contribution to ALE, so any feedback/criticism is appreciated 👍.

### DIFF
--- a/ale_linters/dafny/dafny.vim
+++ b/ale_linters/dafny/dafny.vim
@@ -1,0 +1,24 @@
+" Author: Taylor Blau <me@ttaylorr.com>
+
+function! ale_linters#dafny#dafny#Handle(buffer, lines) abort
+    let l:pattern = '\v(.*)\((\d+),(\d+)\): (.*): (.*)'
+    let l:output = []
+
+    for l:match in ale#util#GetMatches(a:lines, l:pattern)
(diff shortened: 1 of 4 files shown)


ale_linters/dafny/dafny.vim
```diff
@@ -0,0 +1,24 @@
...
+        \    'type': l:match[4] =~# '^Error' ? 'E' : 'W'
+        \ })
+    endfor
+    return l:output
+endfunction
+
+call ale#linter#Define('dafny', {
+\    'name': 'dafny',
+\    'executable': 'dafny',
+\    'command': 'dafny %s /compile:0',
```
w0rp: Does this look at the file on disk, or does it read from stdin? If it's the former, add `'lint_file': 1`, so you can only check for problems on enter or on save.

ttaylorr: @w0rp the `dafny` executable only reads files from disk, and I have updated this definition to match in da80129.

w0rp: Okay, cool.

w0rp:
Cheers! :beers:

PR #1094, opened 2017-11-08
remove --parser flag from prettier fixer

ahmedelgabri:
Prettier will infer the parser based on the file name https://github.com/prettier/prettier/issues/3190#issuecomment-342532165, so no need for the parser flag. 

And since prettier now supports Markdown & GraphQL this list will just grow & needs to be maintained manually which is not good thing.


w0rp:
This will take some consideration. Some other users added the `--parser` flag because it wasn't working otherwise. The `--parser` flag won't be added for unknown filetypes, so it shouldn't be a huge issue.

ahmedelgabri:
Maybe that was an issue was an old version or something, I'll leave this to you to decide.

w0rp:
I removed the `--parser` argument. If it's somehow needed for something, someone can create a pull request and explain why.

PR #1097, opened 2017-11-08
Add PyLS linter

aurieh:


ale_linters/python/pyls.vim
```diff
@@ -0,0 +1,33 @@
...
+    return ale#Var(a:buffer, 'python_pyls_executable')
+endfunction
+
+function! ale_linters#python#pyls#GetLanguage(buffer) abort
+    return 'python'
+endfunction
+
+function! ale_linters#python#pyls#GetProjectRoot(buffer) abort
+    " Start with more generic files
+    for l:possible_filename in ['setup.cfg', 'tox.ini', 'flake8.cfg', 'pycodestyle.cfg']
```
w0rp: I recommend using `ale#python#FindProjectRoot` here, and you could maybe modify that function so it will look for other configuration files. Then the behavior of searching for Python project roots will be consistent everywhere.

aurieh: PyLS only checks for that set of files

aurieh: I could probably use `python#FindProjectRoot`, but it wouldn't be accurate

w0rp: I recommend it. One important difference is that it uses a bread-first search for configuration files instead of depth-first search, so `foo/bar/baz/tox.ini` will be preferred over `foo/setup.cfg`.

aurieh: Yeah, that's a good point. Some people have multiple per-module configs.
I'll just add more filenames to `python#FindProjectRoot`.

ale_linters/python/pyls.vim
```diff
@@ -0,0 +1,21 @@
...
+function! ale_linters#python#pyls#GetLanguage(buffer) abort
+    return 'python'
+endfunction
+
+call ale#linter#Define('python', {
+\   'name': 'pyls',
+\   'lsp': 'stdio',
+\   'executable_callback': 'ale_linters#python#pyls#GetExecutable',
+\   'command_callback': 'ale_linters#python#pyls#GetExecutable',
+\   'language_callback': 'ale_linters#python#pyls#GetLanguage',
```
w0rp: I'll add support at some point for just writing `'language': 'python'`. I can see that coming up quite a bit.

ale_linters/python/pyls.vim
```diff
@@ -0,0 +1,21 @@
+" Author: aurieh <me@aurieh.me>
+" Description: A language server for Python
+
+call ale#Set('python_pyls_executable', 'pyls')
```
w0rp: I just noticed one thing missing, could you add a documentation entry for this new option?

ale_linters/python/pyls.vim
aurieh: Yeah, that would be nice. Is there anything else I should do?

ale_linters/python/pyls.vim
aurieh: Oh sure, one second.

ale_linters/python/pyls.vim
w0rp: Nah, it's fine.

doc/ale-python.txt
```diff
@@ -189,6 +189,15 @@ g:ale_python_pylint_use_global                 *g:ale_python_pylint_use_global*
...
 
+===============================================================================
+pyls                                                          *ale-python-pyls*
+
+g:ale_python_pyls_executable                     *g:ale_python_pyls_executable*
+                                                 *b:ale_python_pyls_executable*
+  Type: |String|
+  Default: `pyls`
+
+  See |ale-intergrations-local-executables|
```
aurieh: woops, that's a typo, Imma rebase this

ale_linters/python/pyls.vim
aurieh: On that note, is there a vim plugin that helps with doc file alignment? Doing it by hand is kinda confusing.

w0rp: There's a fixer for it. See `:ALEFixSuggest` when editing a help file.

w0rp:
Cheers! :beers:

ale_linters/python/pyls.vim
aurieh: Oh that's really cool, thanks for letting me know.

PR #1098, opened 2017-11-08
Added support for linting of proto files

deltaskelta:
- added docs

I can add tests if they are necessary. I am not too familiar with vimscript so if tests are necessary some advice would be useful. Thanks


ale_linters/proto/protoc-gen-lint.vim
```diff
@@ -0,0 +1,10 @@
+" Author: Jeff Willette <jrwillette88@gmail.com>
+" Description: run the protoc-gen-lint plugin for the protoc binary
+
+call ale#linter#Define('proto', {
+\   'name': 'protoc-gen-lint',
+\   'output_stream': 'stderr',
+\   'executable': 'protoc',
+\   'command': 'protoc --lint_out=. *.proto',
```
w0rp: It looks like `*.proto` is for checking files in the current directory. That will be expanded in the shell. Does this program read the file from stdin at all?

deltaskelta:
hmm there seems to be no options for reading from stdin. Is there any bash hackery that can be done to pipe stdin to it? I played around with some things but I couldn't get it to work

w0rp:
If the linter needs to look at files on disk, use `'lint_file': 1`, so it only checks files on disk. Instead of something like `*.proto`, give it `%s`, so it checks just the file you're working on.

deltaskelta:
ok, I am now getting an error because the `protoc` needs the parent directory of the `.proto` file. Is there a format directive to get that? I tried doing `%s/../` but it didn't like that :(

ale_linters/proto/protoc-gen-lint.vim
```diff
@@ -0,0 +1,11 @@
+" Author: Jeff Willette <jrwillette88@gmail.com>
+" Description: run the protoc-gen-lint plugin for the protoc binary
+
+call ale#linter#Define('proto', {
+\   'name': 'protoc-gen-lint',
+\   'output_stream': 'stderr',
+\   'lint_file': 1,
+\   'executable': 'protoc',
+\   'command': 'protoc -I $(dirname %s) --lint_out=. %s',
```
w0rp: You've got the right idea about using `'command_callback' here. What you should do is write a callback with `'command_callback'` and add `ale#Escape(expand('#' . a:buffer . ':p:h'))` into the command to add the escaped directory name. That will work everywhere. Then add a test for the command callback in the `test/command_callback` directory. There are a few similar tests.

deltaskelta:
I have what I think is a working function here (just from looking at others and copying) but I get an error when I open vim that says 

```
Error detected while processing function ale#Queue[9]..ale#CallWithCooldown[9]..<SNR>17_ALEQueueImpl[41]..ale#Lint[12]..ale#CallWithCooldown[9]..<SNR>17
_ALELintImpl[16]..ale#engine#RunLinters[12]..<SNR>102_RunLinter[7]..<SNR>102_InvokeChain[1]..ale#engine#ProcessChain[52]..ale#linter#GetCommand:
line    1:
E117: Unknown function: ale_linters#proto#protocgenlint#GetCommand
E15: Invalid expression: has_key(a:linter, 'command_callback')   ? ale#util#GetFunction(a:linter.command_callback)(a:buffer)   : a:linter.command
```

this is the file as it is when it gives the error. I have tried numerous things and nothing has gotten it to work

```
function! ale_linters#proto#protocgenlint#GetCommand(buffer) abort
    let l:dirname = expand('#' . a:buffer . ':p:h')

    "\   'command': 'protoc -I $(dirname %s) --lint_out=. %s',

    return 'protoc'
    \   . '-I' . ale#Escape(l:dirname)
    \   . '--lint_out=.' . ale#Escape(l:filename)
endfunction

call ale#linter#Define('proto', {
\   'name': 'protoc-gen-lint',
\   'output_stream': 'stderr',
\   'lint_file': 1,
\   'executable': 'protoc',
\   'command_callback': 'ale_linters#proto#protocgenlint#GetCommand',
\   'callback': 'ale#handlers#unix#HandleAsError',
\})
```

ale_linters/proto/protoc_gen_lint.vim
```diff
@@ -0,0 +1,22 @@
...
+
+function! ale_linters#proto#protoc_gen_lint#GetCommand(buffer) abort
+    let l:dirname = expand('#' . a:buffer . ':p:h')
+    let l:filename = expand('#' . a:buffer)
+
+    "\   'command': 'protoc -I $(dirname %s) --lint_out=. %s',
+
+    return 'protoc'
+    \   . ' -I ' . ale#Escape(l:dirname)
+    \   . ' --lint_out=. ' . ale#Escape(l:filename)
```
w0rp: You can use just `%s` on the end there. ALE will automatically replace that with the filename.

ale_linters/proto/protoc_gen_lint.vim
```diff
@@ -0,0 +1,22 @@
+" Author: Jeff Willette <jrwillette88@gmail.com>
+" Description: run the protoc-gen-lint plugin for the protoc binary
+
+function! ale_linters#proto#protoc_gen_lint#GetCommand(buffer) abort
+    let l:dirname = expand('#' . a:buffer . ':p:h')
+    let l:filename = expand('#' . a:buffer)
+
+    "\   'command': 'protoc -I $(dirname %s) --lint_out=. %s',
```
w0rp: You can remove this comment now.

doc/ale.txt
```diff
@@ -134,6 +134,9 @@ CONTENTS                                                         *ale-contents*
       phpstan.............................|ale-php-phpstan|
     pod...................................|ale-pod-options|
       write-good..........................|ale-pod-write-good|
+    proto.................................|ale-proto-options|
+      protoc..............................|ale-proto-protoc|
```
w0rp: Remove this entry and the documentation entry. There's no such linter with the name `protoc`.

deltaskelta:
@w0rp after the test failed I tried every which way I could think of to add `%s` on the end and nothing worked

```
. ' --lint_out=. ' . '%s'
. ' --lint_out=. ' . expand('%s')
. ' --lint_out=. '%s'
. ' --lint_out=. ' ale#Escape('%s')
. ' --lint_out=. ' . ale#Escape(expand('#' . '%s')
```

all failed for me

w0rp:
Put `%s` in the test too. `%s` is replaced later on by the engine.

PR #1105, opened 2017-11-09
add 'output_stream': 'stderr', let golint work

wuqiong4945:
have tested, without output_stream line, golint do not work


w0rp:
Cheers! :beers:

PR #1110, opened 2017-11-11
Add ocaml-language-server for OCaml and ReasonML

jungomi:
[ocaml-language-server](https://github.com/freebroccolo/ocaml-language-server) provides editor support for OCaml and Reason using the Language Server Protocol.

I had to remove the `id` field on the request object, because technically a notification must not have an `id`, which is different from `id: null` (present, but `null`).

From [JSON-RPC 2.0 Specification - Notification](http://www.jsonrpc.org/specification#notification)

> **4.1 Notification**
> A Notification is a Request object without an "id" member. [...]

And [JSON-RPC 2.0 Specification - Request object](http://www.jsonrpc.org/specification#request_object):

> **id**
> An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [1] and Numbers SHOULD NOT contain fractional parts [2]
>
> [...]
>
> [1] The use of Null as a value for the id member in a Request object is discouraged, because this specification uses a value of Null for Responses with an unknown id. Also, because JSON-RPC 1.0 uses an id value of Null for Notifications this could cause confusion in handling.

ocaml-language-server is built with [vscode-languageserver](https://github.com/Microsoft/vscode-languageserver-node), which is very strict about this and does not accept notifications with an `id`, even if it's `null`.

After removing `'id': v:null`, Vim reorders the fields to `{"method": "someMethod", "jsonrpc": "2.0"}` for whatever reason, so I swapped them in the source to avoid confusion. The order only affects the tests.

For the record:

```vim
echo {'id': v:null, 'jsonrpc': '2.0', 'method': 'someMethod'}
" => {'id': v:null, 'jsonrpc': '2.0', 'method': 'someMethod'}

echo {'jsonrpc': '2.0', 'method': 'someMethod'}
" => {'method': 'someMethod', 'jsonrpc': '2.0'}

echo {'jsonrpc': '2.0', 'method': 'someMethod', 'id': v:null}
" => {'method': 'someMethod', 'jsonrpc': '2.0', 'id': v:null}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ formatting.
 | nroff | [proselint](http://proselint.com/), [write-good](https://github.com/btford/write-good)|
 | Objective-C | [clang](http://clang.llvm.org/) |
 | Objective-C++ | [clang](http://clang.llvm.org/) |
-| OCaml | [merlin](https://github.com/the-lambda-church/merlin) see `:help ale-ocaml-merlin` for configuration instructions |
+| OCaml | [merlin](https://github.com/the-lambda-church/merlin) see `:help ale-ocaml-merlin` for configuration instructions, [ols](https://github.com/freebroccolo/ocaml-language-server) |
 | Perl | [perl -c](https://perl.org/), [perl-critic](https://metacpan.org/pod/Perl::Critic) |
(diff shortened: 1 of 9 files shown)


w0rp:
Cheers! :beers:

PR #1114, opened 2017-11-12
Strip Rails flavoring from eRuby when in a Rails project. Fixes #580.

### LINKED ISSUE #580: Basic erb file is generating syntax errors that I don't expect, but the site works.
Perhaps someone can point me in the right direction here.

I have this line in my vimrc: `let g:ale_echo_msg_format = '%linter% says %s'` which I tried to make sure I did to confirm that it's actually the correct linter throwing the errors.  The strangest part about the errors as the title says is that the site works.  I can go to the page with no problems and I don't get any syntax errors thrown by Rails.

https://github.com/justin-capalbo/grocerly/blob/site-layout/app/views/devise/sessions/new.html.erb

**5** - erubylint says syntax error, unexpected ')'
**26** - erubylint says syntax error, unexpected keyword_end, expecting ')'
**27** - erubylint says syntax error, unexpected end-of-input, expecting ')'

As I said, the site does work as expected and I don't think there's actually a problem with this .erb file.  If I need to provide anything else, please let me know.  The only command for ale I have in vimrc is that string display which I pasted above - other than that I just installed it using Pathogen with git clone.  I couldn't find any similar issues related to ERB.

elebow:
This adds a command callback for each of the two eRuby linters. It also renames `erubylint` to `erb`, which matches the name of the tool.

The command callback conditionally adds a filter (implemented as inline Ruby code in the command line) to transform some of the problematic Rails-specific eRuby syntax. Specifically, `<%=` tags are replaced with `<%`.

This does not reduce the effectiveness of the linter, because the transformed code is still evaluated.

This solution was suggested by @rgo at https://github.com/w0rp/ale/issues/580#issuecomment-337676607

### DIFF
--- a/ale_linters/eruby/erb.vim
+++ b/ale_linters/eruby/erb.vim
@@ -1,8 +1,8 @@
 " Author: Matthias Guenther - https://wikimatze.de
-" Description: erb-lint for eruby/erb files
+" Description: ERB from the Ruby standard library, for eruby/erb files
 
 call ale#linter#Define('eruby', {
-\   'name': 'erubylint',
+\   'name': 'erb',
 \   'executable': 'erb',
 \    'output_stream': 'stderr',
 \   'command': 'erb -P -x %t | ruby -c',
--- a/ale_linters/eruby/erb.vim
+++ b/ale_linters/eruby/erb.vim
@@ -1,11 +1,25 @@
-" Author: Matthias Guenther - https://wikimatze.de
+" Author: Matthias Guenther - https://wikimatze.de, Eddie Lebow https://github.com/elebow
 " Description: ERB from the Ruby standard library, for eruby/erb files
+
+function! ale_linters#eruby#erb#GetCommand(buffer) abort
+    let l:rails_root = ale#ruby#FindRailsRoot(a:buffer)
+
+    if empty(l:rails_root)
+        return 'erb -P -x %t | ruby -c'
+    endif
+
+    " Rails-flavored eRuby does not comply with the standard as understood by
+    " ERB, so we'll have to do some substitution. This does not reduce the
+    " effectiveness of the linter—the translated code is still evaluated.
+    return 'ruby -r erb -e ' . ale#Escape('puts ERB.new($stdin.read.gsub(%{<%=},%{<%}), nil, %{-}).src') . ' | ruby -c'
+endfunction
 
 call ale#linter#Define('eruby', {
 \   'name': 'erb',
+\   'aliases': ['erubylint'],
 \   'executable': 'erb',
-\    'output_stream': 'stderr',
-\   'command': 'erb -P -x %t | ruby -c',
+\   'output_stream': 'stderr',
+\   'command_callback': 'ale_linters#eruby#erb#GetCommand',
 \   'callback': 'ale#handlers#ruby#HandleSyntaxErrors',
 \})
 
--- a/test/command_callback/test_erb_command_callback.vader
+++ b/test/command_callback/test_erb_command_callback.vader
@@ -1,0 +1,21 @@
+Before:
+  runtime ale_linters/eruby/erb.vim
+  call ale#test#SetDirectory('/testplugin/test/command_callback')
+
+After:
+  call ale#linter#Reset()
+  call ale#test#RestoreDirectory()
+
+Execute(Executable should not contain any filter code by default):
+  call ale#test#SetFilename('../ruby_fixtures/not_a_rails_app/file.rb')
+
+  AssertEqual
+  \ 'erb -P -x %t | ruby -c',
+  \ ale_linters#eruby#erb#GetCommand(bufnr(''))
+
+Execute(Executable should filter invalid eRuby when inside a Rails project):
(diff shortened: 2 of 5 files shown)


ale_linters/eruby/erb.vim
```diff
@@ -0,0 +1,24 @@
...
+    return 'ruby -rerb -e "puts ERB.new(File.read(%t, encoding: \"BINARY\").gsub(\"<%=\",\"<%\"), nil, \"-\").src" | ruby -c'
+endfunction
+
+call ale#linter#Define('eruby', {
+\   'name': 'erb',
+\   'executable': 'erb',
+\   'output_stream': 'stderr',
+\   'command_callback': 'ale_linters#eruby#erb#GetCommand',
+\   'callback': 'ale#handlers#ruby#HandleSyntaxErrors',
+\})
```
w0rp: There's an undocumented feature for backwards compatibility you should use here. Add `'aliases': ['erubylint'],` to the dictionary here. Then the linter will continue to work for uses who have configured `g:ale_linters` with the old name. You can see how it works in `autoload/ale/linter.vim`.

ale_linters/eruby/erb.vim
```diff
@@ -0,0 +1,24 @@
+" Author: Matthias Guenther - https://wikimatze.de, Eddie Lebow https://github.com/elebow
+" Description: ERB from the Ruby standard library, for eruby/erb files
+
+function! ale_linters#eruby#erb#GetCommand(buffer) abort
+    let l:rails_root = ale#ruby#FindRailsRoot(a:buffer)
+
+    if l:rails_root is? ''
```
w0rp: You can use `if empty(l:rails_root)`. I think that's more commonly used.

ale_linters/eruby/erb.vim
```diff
@@ -0,0 +1,24 @@
...
+        return 'erb -P -x %t | ruby -c'
+    endif
+
+    " Rails-flavored eRuby does not comply with the standard as understood by
+    " ERB, so we'll have to do some substitution. This does not reduce the
+    " effectiveness of the linter—the translated code is still evaluated.
+    return 'ruby -rerb -e "puts ERB.new(File.read(%t, encoding: \"BINARY\").gsub(\"<%=\",\"<%\"), nil, \"-\").src" | ruby -c'
```
w0rp: The escaping of double quotes here won't work on Windows. I recommend removing the `\"` escapes, and use `ale#Escape()` for the parameter instead. That should work everywhere. Same for the file below.

test/command_callback/test_erb_command_callback.vader
```diff
@@ -0,0 +1,21 @@
...
+
+  AssertEqual
+  \ 'erb -P -x %t | ruby -c',
+  \ ale_linters#eruby#erb#GetCommand(bufnr(''))
+
+Execute(Executable should filter invalid eRuby when inside a Rails project):
+  call ale#test#SetFilename('../ruby_fixtures/valid_rails_app/app/views/my_great_view.html.erb')
+
+  AssertEqual
+  \ 'ruby -rerb -e "puts ERB.new(File.read(%t, encoding: \"BINARY\").gsub(\"<%=\",\"<%\"), nil, \"-\").src" | ruby -c',
```
w0rp: You'll have to use `ale#Escape` in the test for the parameter with the suggestion above. The string will differ between Linux and Windows.

ale_linters/eruby/erb.vim
w0rp: `ale#Escape('puts ERB.new(File.read(%t, encoding: "BINARY").gsub("<%=","<%"), nil, "-").src')` That might be what you want to add to the string there.

w0rp: Oh, and the percents will also be doubled on Windows. Windows is weird.

w0rp: I just realised, the `%t` inside of the escaped string might cause issues where the escaping doesn't work any more when the filename contains spaces. If you can pass in the filename with an argument and use `ARGV[0]` or something, then do that.

w0rp: You might also be able to do something like `'rbuy -rerb -e filename=%t' . ale#Escape(';puts ERB.new(File.read(filename, ...')`

elebow: I'm having trouble getting `ale#Escape()` to do what is needed here... it inserts quotes liberally, which is fine for filenames in a shell, but not in the middle of Ruby code.
But I was able to rewrite the line to not have any escaped double quotes:

```
 return 'ruby -r erb -e "puts ERB.new(\$stdin.read.gsub(%{<%=},%{<%}), nil, %{-}).src" | ruby -c'
```
And it works correctly on GNU/Linux. Is this okay for Windows? Do the `%` signs need to be escaped?

w0rp: Try this.

```vim
return 'ruby -r erb -e ' . ale#Escape('puts ERB.new($stdin.read.gsub(%{<%=},%{<%}), nil, %{-}).src') . ' | ruby -c'
```

The percent signs will probably need to be doubled on Windows.

elebow:
Review comments should now be addressed.

w0rp:
Cheers! :beers:

PR #1123, opened 2017-11-13
Added support for goimports fixer

deltaskelta:
I tried to follow the precedent that the gofmt fixer laid out, I couldn't find any tests in teh tests file so I wasn't sure if I should write tests or not.

Also, the docs seem to omit some of the options that gofmt has such as setting the executable so I wasn't sure what, if anything to add there.


deltaskelta:
is the executable check the `executable_callback` like in some of the other linters? I grepped through the linters and I can't exactly see how or if that is checking the executable

w0rp:
You can add...

```vim
if !executable(l:executable)
    return 0
endif
```

That's what the fixer for `isort` does.

PR #1125, opened 2017-11-13
Added filename keys for go linters

deltaskelta:
- reference #653 - add filename keys to existing linters (https://github.com/w0rp/ale/issues/653#issuecomment-323542811)

- added filename keys to gobuild and gometalinter

- removed --include `filename` option on gometalinter to get all errors from current
package

- added name to the authors lists

### DIFF
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -44,13 +44,6 @@ function! ale_linters#go#gobuild#Handler(buffer, lines) abort
     let l:output = []
 
     for l:match in ale_linters#go#gobuild#GetMatches(a:lines)
-        "
-        " Get the list of parent paths from the current filename
-        let l:paths = ale#path#Upwards(expand('#' . a:buffer . ':p'))
-        " Only see errors from this package (directory), ignore errors from imported packages
-        if !l:dir is# l:paths[1]
-            continue
-        endif
 
         echom l:match[1]
         call add(l:output, {
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -44,8 +44,6 @@ function! ale_linters#go#gobuild#Handler(buffer, lines) abort
     let l:output = []
 
     for l:match in ale_linters#go#gobuild#GetMatches(a:lines)
-
-        echom l:match[1]
         call add(l:output, {
         \   'filename': ale#path#GetAbsPath(l:dir, l:match[1]),
         \   'lnum': l:match[2] + 0,
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -1,4 +1,5 @@
-" Author: Joshua Rubin <joshua@rubixconsulting.com>, Ben Reedy <https://github.com/breed808>
+" Author: Joshua Rubin <joshua@rubixconsulting.com>, Ben Reedy <https://github.com/breed808>,
+" Jeff Willette <jrwillette88@gmail.com>
 " Description: go build for Go files
 
 " inspired by work from dzhou121 <dzhou121@gmail.com>
@@ -39,15 +40,21 @@ function! ale_linters#go#gobuild#GetMatches(lines) abort
 endfunction
 
 function! ale_linters#go#gobuild#Handler(buffer, lines) abort
+    let l:dir = expand('#' . a:buffer . ':p:h')
     let l:output = []
 
     for l:match in ale_linters#go#gobuild#GetMatches(a:lines)
-        " Omit errors from imported go packages
-        if !ale#path#IsBufferPath(a:buffer, l:match[1])
+        "
+        " Get the list of parent paths from the current filename
+        let l:paths = ale#path#Upwards(expand('#' . a:buffer . ':p'))
+        " Only see errors from this package (directory), ignore errors from imported packages
+        if !l:dir is# l:paths[1]
             continue
         endif
 
+        echom l:match[1]
         call add(l:output, {
+        \   'filename': ale#path#GetAbsPath(l:dir, l:match[1]),
         \   'lnum': l:match[2] + 0,
         \   'col': l:match[3] + 0,
         \   'text': l:match[4],
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -1,4 +1,4 @@
-" Author: Ben Reedy <https://github.com/breed808>
+" Author: Ben Reedy <https://github.com/breed808>, Jeff Willette <jrwillette88@gmail.com>
 " Description: Adds support for the gometalinter suite for Go files
 
 call ale#Set('go_gometalinter_options', '')
@@ -14,7 +14,6 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
 
     return ale#Escape(l:executable)
-    \   . ' --include=' . ale#Escape('^' . ale#util#EscapePCRE(l:filename))
     \   . (!empty(l:options) ? ' ' . l:options : '')
     \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
 endfunction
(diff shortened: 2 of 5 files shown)


deltaskelta:
Tests are failing  with expected errors due to the changes, I would like to know if the method and changes are valid and I can go through and fix the assertions in the current tests. Thanks

ale_linters/go/gobuild.vim
```diff
@@ -39,15 +40,21 @@ function! ale_linters#go#gobuild#GetMatches(lines) abort
...
     let l:output = []
 
     for l:match in ale_linters#go#gobuild#GetMatches(a:lines)
-        " Omit errors from imported go packages
-        if !ale#path#IsBufferPath(a:buffer, l:match[1])
+        "
+        " Get the list of parent paths from the current filename
+        let l:paths = ale#path#Upwards(expand('#' . a:buffer . ':p'))
+        " Only see errors from this package (directory), ignore errors from imported packages
+        if !l:dir is# l:paths[1]
```
w0rp: I'm confused. Above you're getting the directory for the filename. Then here you're getting the directory for the same filename again, but in a different and more expensive way, for every iteration?

ale_linters/go/gobuild.vim
```diff
@@ -39,15 +40,21 @@ function! ale_linters#go#gobuild#GetMatches(lines) abort
...
+    let l:dir = expand('#' . a:buffer . ':p:h')
     let l:output = []
 
     for l:match in ale_linters#go#gobuild#GetMatches(a:lines)
-        " Omit errors from imported go packages
-        if !ale#path#IsBufferPath(a:buffer, l:match[1])
+        "
+        " Get the list of parent paths from the current filename
+        let l:paths = ale#path#Upwards(expand('#' . a:buffer . ':p'))
+        " Only see errors from this package (directory), ignore errors from imported packages
```
w0rp: Why ignore them? You can include them in the loclist so you can go and fix errors in other packages. The only reason other files are ignored at the moment is because ALE didn't support adding those errors to the loclist.

ale_linters/go/gometalinter.vim
```diff
@@ -14,7 +14,6 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
 
     return ale#Escape(l:executable)
-    \   . ' --include=' . ale#Escape('^' . ale#util#EscapePCRE(l:filename))
```
w0rp: I believe this was added no only for filtering the errors, but because it executes much more quickly.

Could you make showing errors for other files optional?

ale_linters/go/gobuild.vim
deltaskelta: Ok

ale_linters/go/gobuild.vim
deltaskelta: It should have been outside of the loop...didn't catch that

As for why I was getting the directory again...I wanted to compare the error path against the current buffer path to only get things in the same directory. The error path should have been coming from `l:match` I guess but the whole line of thought becomes irrelevant due to the suggestion above..

ale_linters/go/gometalinter.vim
deltaskelta: I can make it optional, if it needs to be, but I ran a benchmark on the two executions and they came out just about the same on average...
averages are as follows
```
plain gometalinter call: 4346553250.25 ns/op
gometalinter with include: 4325110610.25 ns/op
```
```
BenchmarkPlain-4               1        4972665842 ns/op
BenchmarkPlain-4               1        4349992730 ns/op
BenchmarkPlain-4               1        4421776655 ns/op
BenchmarkPlain-4               1        4475188394 ns/op
BenchmarkPlain-4               1        4408009068 ns/op
BenchmarkPlain-4               1        4233487406 ns/op
BenchmarkPlain-4               1        4878184255 ns/op
BenchmarkPlain-4               1        4374577545 ns/op
BenchmarkPlain-4               1        4025479863 ns/op
BenchmarkPlain-4               1        4355798558 ns/op
BenchmarkPlain-4               1        4199595247 ns/op
BenchmarkPlain-4               1        4376063551 ns/op
BenchmarkPlain-4               1        4216004173 ns/op
BenchmarkPlain-4               1        3993578021 ns/op
BenchmarkPlain-4               1        4078005741 ns/op
BenchmarkPlain-4               1        4672426812 ns/op
BenchmarkPlain-4               1        4293836100 ns/op
BenchmarkPlain-4               1        4188911073 ns/op
BenchmarkPlain-4               1        4047435325 ns/op
BenchmarkPlain-4               1        4370048646 ns/op
BenchmarkInclude-4             1        4610757981 ns/op
BenchmarkInclude-4             1        4016014938 ns/op
BenchmarkInclude-4             1        4065846142 ns/op
BenchmarkInclude-4             1        4131692572 ns/op
BenchmarkInclude-4             1        4991080619 ns/op
BenchmarkInclude-4             1        4437625405 ns/op
BenchmarkInclude-4             1        4512699377 ns/op
BenchmarkInclude-4             1        4823607824 ns/op
BenchmarkInclude-4             1        4585688367 ns/op
BenchmarkInclude-4             1        4178835349 ns/op
BenchmarkInclude-4             1        3855721441 ns/op
BenchmarkInclude-4             1        3873105094 ns/op
BenchmarkInclude-4             1        4303892553 ns/op
BenchmarkInclude-4             1        4838156426 ns/op
BenchmarkInclude-4             1        3950300377 ns/op
BenchmarkInclude-4             1        3852298474 ns/op
BenchmarkInclude-4             1        4937783666 ns/op
BenchmarkInclude-4             1        4167373118 ns/op
BenchmarkInclude-4             1        4063617353 ns/op
BenchmarkInclude-4             1        4306115129 ns/op
```

```
func BenchmarkPlain(b *testing.B) {
	for n := 0; n < b.N; n++ {
		cmd := exec.Command("sh", "-c", "gometalinter")
		_, _ = cmd.CombinedOutput()
	}
}

func BenchmarkInclude(b *testing.B) {
	for n := 0; n < b.N; n++ {
		cmd := exec.Command("sh", "-c", "gometalinter", "--include main.go")
		_, _ = cmd.CombinedOutput()
	}
}
```

deltaskelta: @w0rp as far as I can tell, all that include does is run a regex on the results, so it shouldn't be affecting execution time much and maybe even slowing it down if there are tons of errors that need to be regexed

https://github.com/alecthomas/gometalinter/blob/9165748669edf1f7e77f96f1e01b01aa24a26ad3/execute.go#L269

w0rp: Okay sounds good. You can just remove it then. If someone complains about it, we can handle it later.

deltaskelta:
Before this gets merged, I have noticed a side effect that I have not seen before...

Only sometimes...and I have not been able to barrow down the reason why...when I get errors from `gometalinter` or `go build` in the quickfix window, they will not disappear when corrected (a subsequent save corrects the error and `:ALEInfo` shows there was no output from the linters

The window is only reset once I change to another buffer and then back

I don't think this would be directly caused by these changes, but I don't know enough of the codebase to say for sure

w0rp:
@deltaskelta That's because those linters can only be run against the files on disk, so you have to save the files to remove the problems.

ale_linters/go/gobuild.vim
```diff
@@ -39,15 +40,14 @@ function! ale_linters#go#gobuild#GetMatches(lines) abort
...
+    let l:dir = expand('#' . a:buffer . ':p:h')
     let l:output = []
 
     for l:match in ale_linters#go#gobuild#GetMatches(a:lines)
-        " Omit errors from imported go packages
-        if !ale#path#IsBufferPath(a:buffer, l:match[1])
-            continue
-        endif
 
+        echom l:match[1]
```
w0rp: You left an echo here.

w0rp: I just pushed a commit so TravisCI will complain about stray echo lines in future.

deltaskelta:
But I was already aware that they were `lint_file` linters and I was saving the buffer. When I said `:ALEInfo` shows no output I meant that I saved the file, the linter ran, found no errors, returned 0 and yet the error remained in the quickfix window until I switched to another buffer...

It might be an interaction with another plugin so I'll try to isolate what is going on because it only happens some of the time

I'll fix the echo as soon as I have a chance

w0rp:
Cheers! :beers:

svanharmelen:
@w0rp can we discuss this PR and why it got merged? I really don't see why the `--include` option is removed? I fail to understand the use case for running `gometalinter` on save (so on a buffer) and then letting it return all the errors for your whole package.

If you would like to `gometalint` your whole package it makes much more sense to do this from the command line instead of within your editor. This PR really messes up my workflow and user experience 😞 

Can we please revert the PR, or at least restore the `--include` part?

svanharmelen:
Additionally it seems this PR also breaks the statusline... I no longer see error/warning counts in there 😞

w0rp:
Okay, I reverted the changes.

svanharmelen:
Thank you so very much!! 🙏

w0rp:
The long filenames are probably just how the quickfix list and loclist look in Vim when your paths are very long. I don't know if there's a setting for making them shorter.

svanharmelen:
Yeah I understand that... But this https://github.com/w0rp/ale/pull/1125/files#diff-a80d8a5e414c003a0373a78f906ca0b5R28 changed it from relative to absolute in the output. 

That makes a quite a big difference and in Go it means that it will essentially prefix all lines with your `$GOPATH`. So it feels to me that doesn't add any value, just clutter.

But that is now also reverted, so all looks good and is workable again 👍

PR #1129, opened 2017-11-14
#852 - Capture error codes for Rubocop

aliou:
Hello!

This PR adds support to the new error code feature from #852 to the Rubocop linter. I used the the cop name as code, let me know if it should be something else.

Thanks!


w0rp:
Cheers! :beers:

aliou:
Yay! 🎉

Issue #1040, opened 2017-10-23
Clojure Support via Eastwood


bigDaddyS1oth:
Was wondering if would be possible to get Clojure linting support in Ale. 

There is a lint tool for Clojure called [Eastwood](https://github.com/jonase/eastwood). To make use of it, a user needs to simply have `{:user {:plugins [[jonase/eastwood "0.2.5"]] }}` in the `$HOME/.lein/profiles.clj` file, and call `lein eastwood` while in the directory that has the file they want to lint. 

Looking through the `ale_linters` directory, I have a _vague_ idea on how to create a proper `.vim` file to make use of Eastwood in Ale. But if it's more than just the common pattern I see in all the `linter.vim` files, then it might be a bit more difficult for me to do.

Issue #544, opened 2017-05-10
Clojure support


sirsean:
Please add support for Clojure & ClojureScript (ideally without needing an nREPL running).

w0rp:
Clojure linters exist by this point, so I'll close this.

PR #1133, opened 2017-11-15
add joker handler for clojure

### LINKED ISSUE #975: Support for joker
Joker has a mode where it lints for Clojure files: https://github.com/candid82/joker It seems to be best in class for Clojure right now.

I'm happy to start to do this myself, but I haven't the time right now.

nicwest:
Adds new linter for clojure using joker

https://github.com/candid82/joker

I haven't tested with clojurescript, edn, or joker (the dialect of clojure implemented by the joker interpreter), but the linter mode appears to support all three in addition to clojure (although it might need a option or something for handling files from stdin)

can I just add that this is the nicest vim plugin that I have ever worked on, really clear and easy to use API's, guidelines, tests, etc! 

fixes #975
ref #544 #1040


w0rp:
Cheers! :beers:

nicwest:
> Looks like a laugh and a half.

dad?

PR #1135, opened 2017-11-15
Capture error codes for checkmake

aurieh:


w0rp:
AppVeyor just failed randomly. That might happen sometimes.

PR #1138, opened 2017-11-16
add ale_asm_gcc_executable option

thvdburgt:


thvdburgt:
I added test, they are based on the c gcc linter command callbacks tests.

PR #1141, opened 2017-11-17
erb, erubis: Redirect file into first command.

elebow:
The previous version relied on a zsh-specific behavior where
`<filename` after a pipe could redirect to the first command. This
is the standard way to do it.


w0rp:
Cheers! :beers:

PR #1142, opened 2017-11-17
Redpen support for text, asciidoc, reST, LaTeX and Re:VIEW

rhysd:
Yesterday, I sent a PR for [redpen](http://redpen.cc/) support as markdown linter. And it seems to work well. Today, I added text, asciidoc, reST, LaTeX and Re:VIEW support by redpen.

### DIFF
--- a/autoload/ale/handlers/redpen.vim
+++ b/autoload/ale/handlers/redpen.vim
@@ -1,0 +1,29 @@
+" Author: rhysd https://rhysd.github.io
+" Description: Redpen, a proofreading tool (http://redpen.cc)
+
+function! ale#handlers#redpen#HandleRedpenOutput(buffer, lines) abort
+    " Only one file was passed to redpen. So response array has only one
+    " element.
+    let l:res = json_decode(join(a:lines))[0]
+    let l:output = []
+    for l:err in l:res.errors
+        let l:item = {
+        \   'text': l:err.message . ' (' . l:err.validator . ')',
+        \   'type': 'W',
+        \}
+        if has_key(l:err, 'startPosition')
+            let l:item.lnum = l:err.startPosition.lineNum
+            let l:item.col = l:err.startPosition.offset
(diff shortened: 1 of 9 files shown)


autoload/ale/handlers/redpen.vim
```diff
@@ -0,0 +1,29 @@
...
+        let l:item = {
+        \   'text': l:err.message . ' (' . l:err.validator . ')',
+        \   'type': 'W',
+        \}
+        if has_key(l:err, 'startPosition')
+            let l:item.lnum = l:err.startPosition.lineNum
+            let l:item.col = l:err.startPosition.offset + 1
+            if has_key(l:err, 'endPosition')
+                let l:item.end_lnum = l:err.endPosition.lineNum
+                let l:item.end_col = l:err.endPosition.offset
```
w0rp: Actually, won't you need to add `+ 1` here too? I imagine if the `startPosition` is off by 1, the `endPosition` will be too.

rhysd: No, we don't need. The range of error is `[start, end)`. So left edge is included but right edge is excluded. And it was revealed that the `offset` is 0-based. So we need to add 1 only to left edge.

w0rp: The `end_col` value has to be the column number the error ends on, so the range for `col` and `end_col` is `[start, end]`.

w0rp: The end offset does seem to be 1-based given the output, so maybe this is okay. For "plugin" it says the word is a spelling mistake from columns 10 to 15, so that checks out.

w0rp:
Cheers! :beers:

autoload/ale/handlers/redpen.vim
rhysd: oh, I see. I did not know that. Thank you.

PR #1143, opened 2017-11-17
Make options configurable in iverilog

RedBug312:
I've inserted the variable `verilog_iverilog_options` into command, which is set to empty by default.
And the test file is altered from `verilog_verilator_options.vader` in the same folder.

### DIFF
--- a/test/test_verilog_iverilog_options.vader
+++ b/test/test_verilog_iverilog_options.vader
@@ -14,11 +14,5 @@ Execute(Set Verilog iverilog linter additional options to `-y.`):
 
   call ale#Lint()
 
-  let g:run_cmd = ale_linters#verilog#iverilog#GetCommand(bufnr(''))
-  let g:matched = match(g:run_cmd, '\s' . g:ale_verilog_iverilog_options . '\s')
-
-  " match returns -1 if not found
-  AssertNotEqual
-  \   g:matched ,
-  \   -1 ,
-  \   'Additionnal arguments not found in the run command'
+  let g:cmd = ale_linters#verilog#iverilog#GetCommand(bufnr(''))
+  AssertEqual g:cmd, 'iverilog -t null -Wall -y. %t',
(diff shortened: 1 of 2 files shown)


test/test_verilog_iverilog_options.vader
```diff
@@ -0,0 +1,24 @@
...
+
+  call ale#Lint()
+
+  let g:run_cmd = ale_linters#verilog#iverilog#GetCommand(bufnr(''))
+  let g:matched = match(g:run_cmd, '\s' . g:ale_verilog_iverilog_options . '\s')
+
+  " match returns -1 if not found
+  AssertNotEqual
+  \   g:matched ,
+  \   -1 ,
```
w0rp: Use `AssertEqual` with the full command string instead. That's more reliable.

RedBug312: Got it. I've changed the assertion to match the whole desired string.

test/test_verilog_iverilog_options.vader
```diff
@@ -0,0 +1,18 @@
...
+Execute(Set Verilog iverilog linter additional options to `-y.`):
+  runtime! ale_linters/verilog/iverilog.vim
+
+  " Additional args for the linter
+  let g:ale_verilog_iverilog_options = '-y.'
+
+  call ale#Lint()
+
+  let g:cmd = ale_linters#verilog#iverilog#GetCommand(bufnr(''))
+  AssertEqual g:cmd, 'iverilog -t null -Wall -y. %t'
```
w0rp: The order of the arguments for Vader is `expected, actual`. I'll just swap them around.

test/test_verilog_iverilog_options.vader
```diff
@@ -0,0 +1,18 @@
...
+  Restore
+  call ale#linter#Reset()
+
+Execute(Set Verilog iverilog linter additional options to `-y.`):
+  runtime! ale_linters/verilog/iverilog.vim
+
+  " Additional args for the linter
+  let g:ale_verilog_iverilog_options = '-y.'
+
+  call ale#Lint()
```
w0rp: This isn't needed.

Issue #1115, opened 2017-11-12
Wrapper for all external commands


vorner:
Hello

Would it be possible to add a configuration option for a general wrapper of external commands? Some of the linters are a bit resource hungry and I'd like to run everything inside `nice`, or maybe even inside something that limits them globally (cgroups possibly).

w0rp:
Interesting idea. Feel free to play around and create a pull request.

PR #1147, opened 2017-11-18
WIP: A command wrapper config

vorner:
This is my first attempt at #1115.

I know I'm missing tests (that's the reason for the WIP). Before I spend the time learning Vader, I'd like some feedback if I'm doing it correctly and if I'm putting it to the right place. If so, I'll add them.

Thanks

### DIFF
--- a/autoload/ale/job.vim
+++ b/autoload/ale/job.vim
@@ -174,14 +174,20 @@ function! ale#job#PrepareCommand(command) abort
     " but we'll do this explicitly, so we use thes same exact command for both
     " versions.
     if ale#Has('win32')
-        return 'cmd /c ' . a:command
+        if g:ale_cmd_wrapper isnot# ''
+            return g:ale_cmd_wrapper . ' cmd /c ' . a:command
+        else
+            return 'cmd /c ' . a:command
+        endif
     endif
+
+    let l:precmd = split(g:ale_cmd_wrapper)
 
     if &shell =~? 'fish$'
-        return ['/bin/sh', '-c', a:command]
+        return l:precmd + ['/bin/sh', '-c', a:command]
     endif
 
-    return split(&shell) + split(&shellcmdflag) + [a:command]
+    return l:precmd + split(&shell) + split(&shellcmdflag) + [a:command]
 endfunction
 
 " Start a job with options which are agnostic to Vim and NeoVim.
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -573,7 +573,7 @@ g:ale_change_sign_column_color                 *g:ale_change_sign_column_color*
   windows.
 
 
-g:ale_cmd_wrapper                                           g:ale_cmd_wrapper
+g:ale_cmd_wrapper                                           *g:ale_cmd_wrapper*
 
   Type: String
   Default: ''
@@ -585,16 +585,17 @@ g:ale_cmd_wrapper                                           g:ale_cmd_wrapper
 
   If creating your own wrapper, be sure to handle a command passed as multiple
   arguments and handle arguments with spaces and special characters in them.
-  Prefer this:
+  Prefer this: >
 
     #/bin/sh
     exec nice -n15 ionice -c idle "$@"
+<
 
-  Over this:
+  Over this: >
 
     #/bin/sh
     exec nice -n15 ionice -c idle $1
-
+<
 
 g:ale_completion_delay                                 *g:ale_completion_delay*
 
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -583,9 +583,9 @@ g:ale_cmd_wrapper                                           *g:ale_cmd_wrapper*
   to 'nice -n15' to have lower priority than other processes, including the
   editor itself.
 
-  If creating your own wrapper, be sure to handle a command passed as multiple
-  arguments and handle arguments with spaces and special characters in them.
-  Prefer this: >
+  The wrapper must be able to accept the command as multiple arguments,
+  possibly with special characters and spaces in them. If creating your own,
+  prefer this: >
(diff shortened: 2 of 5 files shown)


vorner:
Hmm, I probably can use the failing tests as a basis for my own ones. I'll have a look some time soon.

vorner:
Hello again.

I have another, more complete, attempt now. It should handle the cases you describe, it has some tests and the documentation was extended.

However, I have no idea what the right thing to do it on windows is and the test there fails. Can I ask for some advice? Google Driven Development with check through appveyor fails me :-(.

doc/ale.txt
```diff
@@ -573,6 +573,29 @@ g:ale_change_sign_column_color                 *g:ale_change_sign_column_color*
   windows.
 
 
+g:ale_cmd_wrapper                                           g:ale_cmd_wrapper
```
w0rp: You want to put `*g:ale_cmd_wrapper*` on the right margin, and make sure it lines up with everything else. That creates a help tag for looking up `:help g:ale_cmd_wrapper`.

doc/ale.txt
```diff
@@ -573,6 +573,29 @@ g:ale_change_sign_column_color                 *g:ale_change_sign_column_color*
...
+  Default: ''
+
+  All commands executed will be run prefixed with this wrapper. This allows
+  restricting resource usage of background linters. You can set it for example
+  to 'nice -n15' to have lower priority than other processes, including the
+  editor itself.
+
+  If creating your own wrapper, be sure to handle a command passed as multiple
+  arguments and handle arguments with spaces and special characters in them.
+  Prefer this:
```
w0rp: You can use `>` and `<` for  blocks of code. Look at existing example blocks of code. The characters are concealed. See `:help conceal`.

autoload/ale/job.vim
```diff
@@ -174,14 +174,20 @@ function! ale#job#PrepareCommand(command) abort
...
     if ale#Has('win32')
-        return 'cmd /c ' . a:command
+        if g:ale_cmd_wrapper isnot# ''
+            return g:ale_cmd_wrapper . ' cmd /c ' . a:command
+        else
+            return 'cmd /c ' . a:command
+        endif
     endif
 
+    let l:precmd = split(g:ale_cmd_wrapper)
```
w0rp: What happens when you have a command like `'"/path to script/foo" bar'`? It can be split weirdly.

How should wrappers which want the entire command to be passed as one argument be handled?

doc/ale.txt
vorner: I see… my configuration had the characters concealed by default, so I didn't knob about the ones in other commands. Should be OK now.

autoload/ale/job.vim
vorner: Yes, that'd split them weirdly and it wouldn't work. But I think the same limitation applies to the `&shell` variable. Is that a problem? Other option is not to split it and support just a command, without arguments. I really don't want to go the whole way of shell-aware splitting. Or maybe the cmd wrapper could be a list, but that feels a bit inconsistent with the rest of ale's settings.

I don't think such wrappers that wants it as a single argument need to be supported. I mean, this is just a tiny feature and probably doesn't need a whole framework of handling of commands, so considering how many (or few) people will use it and that such command can be handled by writing a shell script to convert it is easy, it would be OK to support just a single mode. Or, in other words, postponing the other mode until one actually needs it 😇. Does that sound OK, or did you have some specific (simple) solution in mind?

w0rp: Okay, could you document all of this? Document that the command must not contain spaces, and the command must eat all of the following arguments.

vorner: I reworded the thing with multiple arguments to be hopefully clearer and added the one with spaces.

test/test_history_saving.vader
```diff
@@ -19,6 +20,7 @@ Before:
   let g:ale_buffer_info = {}
   let g:ale_max_buffer_history_size = 20
   let g:ale_history_log_output = 0
+  let g:ale_cmd_wrapper = ''
```
w0rp: Add this in the `After` block too. That might be why the other test is failing.

vorner: I did add that now. But, shouldn't the `Save` in `Before` and `Restore` in `After` do just that (I'm just guessing)?

Anyway, the error seems to be related to the `start` thing I found online, because it doesn't provide `exit_code`. I don't know how to handle that, I have no experience with Windows and none at hand.

vorner:
```
    (2/7) [EXECUTE] (X) ['command', 'job_id', 'status'] should be equal to ['command', 'exit_code', 'job_id', 'status']
```

vorner: Any tips on how to move this thing forward? Unless someone knows a working `nice` equivalent for windows, my proposals would be:

* Don't test actual execution of the command on windows.
* Use some other dummy command there (eg. another layer of `cmd`?).

I believe this is the last thing blocking it, isn't it?

w0rp: I recommend using `if has('unix')` and skipping the tests on Windows.

w0rp:
I wrote my own implementation of this. Here are some notes.

* The option is called `g:ale_command_wrapper`, and can be configured per buffer too.
* By default all commands are spread on the end of the wrapped command.
* A `%*` marker can be used to spread the wrapped command elsewhere inside of the wrapped command.
* A `%@` marker can be used to wrap a command as a single argument.
* For commands with `&&` or `;`, only the last command is wrapped. So you can end up with `cd some_dir && nice -n5 foo bar` for linters which change directory, and so on.

vorner:
Thank you :-)

PR #1153, opened 2017-11-20
Fix #1145 added `gometalinter_lint_dir` option

### LINKED ISSUE #1145: gometalinter - Lint only opened files
Is it possible to revert https://github.com/w0rp/ale/commit/e721f851b41b8f6f31067ae2a137019e1cb5546c and reintroduce this with a config option? 

I ask because this changes the previously expected behavior drastically. Previously, we linted open files/buffers. Now, we lint the entire directory.

I also believe the path provided to the quickfix window from the linting error is incorrect. /foo/bar/file.go turns into /foo/bar/foo/bar/file.go - to reproduce, try navigating to file failing the lint from quickfix (using gometalinter).

deltaskelta:
- added an option to lint only the current file with gometalinter
- added command_callback test to test the option

EDIT: I will add docs for the new option, but just wanted to get some input on the changes first

### DIFF
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -3,6 +3,7 @@
 
 call ale#Set('go_gometalinter_options', '')
 call ale#Set('go_gometalinter_executable', 'gometalinter')
+call ale#Set('go_gometalinter_lint_dir', 1)
 
 function! ale_linters#go#gometalinter#GetExecutable(buffer) abort
     return ale#Var(a:buffer, 'go_gometalinter_executable')
@@ -12,8 +13,16 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:executable = ale_linters#go#gometalinter#GetExecutable(a:buffer)
     let l:filename = expand('#' . a:buffer)
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
+    let l:lint_dir = ale#Var(a:buffer, 'go_gometalinter_lint_dir')
+
+    if l:lint_dir
+        return ale#Escape(l:executable)
+        \   . (!empty(l:options) ? ' ' . l:options : '')
+        \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
+    endif
 
     return ale#Escape(l:executable)
+    \   . ' --include=' . ale#Escape('^' . ale#util#EscapePCRE(l:filename))
     \   . (!empty(l:options) ? ' ' . l:options : '')
     \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
 endfunction
--- a/test/command_callback/test_gometalinter_command_callback.vader
+++ b/test/command_callback/test_gometalinter_command_callback.vader
@@ -1,9 +1,11 @@
 Before:
   Save b:ale_go_gometalinter_executable
   Save b:ale_go_gometalinter_options
+  Save b:ale_go_gometalinter_lint_dir
 
   let b:ale_go_gometalinter_executable = 'gometalinter'
   let b:ale_go_gometalinter_options = ''
+  let b:ale_go_gometalinter_lint_dir = 1
 
   runtime ale_linters/go/gometalinter.vim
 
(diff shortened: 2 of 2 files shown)


svanharmelen:
I would also really ask for restoring the previous behaviour. And if really wanted/needed then add an option to lint the whole package instead of the file being worked on.

Personally I fail to understand why such a change (which impacts user experience) was merged...

ale_linters/go/gometalinter.vim
```diff
@@ -3,6 +3,7 @@
 
 call ale#Set('go_gometalinter_options', '')
 call ale#Set('go_gometalinter_executable', 'gometalinter')
+call ale#Set('go_gometalinter_lint_dir', 1)
```
svanharmelen: Maybe a better name would be `go_gometalinter_package`? In then in the doc's you can describe how settings this option will lint the whole package instead of the file your currently working on.

ale_linters/go/gometalinter.vim
```diff
@@ -12,8 +13,16 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:executable = ale_linters#go#gometalinter#GetExecutable(a:buffer)
     let l:filename = expand('#' . a:buffer)
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
+    let l:lint_dir = ale#Var(a:buffer, 'go_gometalinter_lint_dir')
```
svanharmelen: And of course this should then be `l:package` :wink:

ale_linters/go/gometalinter.vim
```diff
@@ -12,8 +13,16 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
...
+
+    if l:lint_dir
+        return ale#Escape(l:executable)
+        \   . (!empty(l:options) ? ' ' . l:options : '')
+        \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
+    endif
```
svanharmelen: Maybe an `if..else` is a bit cleaner/clearer here? I know in Go `if...else` is not done, but in vimscript it's a standard practice and personally I think that in this piece of code it's a better fit...

svanharmelen:
@w0rp this PR does fix the `--include` issue, but it still leaves the problem with the extreme long filenames we now have... So instead of merging this PR I still think reverting the PR that introduced these issues makes much more sense.

w0rp:
Okay, I reverted that merge. I recommend working on the `gobuild` and `gometalinter` linters separately, so it will be easy to revert the changes for only one of them. I have reverted both for now. You can open a new pull request with something better later.

deltaskelta:
I really don't see why it needs to be reverted completely. `gometalinter` already lints the whole package and just runs a regex when the `--include` option is present. This happens because most of the toolchain in go is meant to run on packages and not files.

`go build` needs to look at the whole package, `go vet`s help text specifies to run it on packages, `gometalinter` lints the whole package by default and only optionally runs a regex on the results to select for certain filenames if you choose to. This means by making `--include` present by default we are wasting CPU on every save by linting the whole package anyway and throwing away the results. 

If I understand you correctly, then given that ale has the option to support both package and file linting, the only issue would be the filenames, can't we just work on fixing that without throwing out all these changes?

svanharmelen:
@deltaskelta I totally understand that all these tools (Go linters) run on the package and not on a file (mainly because they use the AST to do the linting). And when running these tools from the commandline against your code it makes perfect sense to target the package and not a single file inside a package.

But when you're working on a specific file in your editor, it doesn't make much sense to give all the errors and warnings from your complete package. In that case I only want to see the errors and warnings related to the file I'm working on so that I can see the impact of my code changes.

If I would get all the errors and warnings then the output in my location list and the status line would be extremely noisy and annoying to use as I would either have to scroll through all of them to see if any are applicable on the file I'm currently working on, or I would constantly have to jump around from file to file.

I think most (if not all) editors and IDE's work like this... They only show you the errors and warnings of the file you are working in (unless you do a build or compile or something alike, but not when saving or editing a single file).

Of course you can have a different preference and way of working, so maybe we should also support the behaviour you wanted to add. But in that case I think it should not impact/change the current behaviour, but the new behaviour should come with a config option to activate the new/different behaviour...

w0rp:
I just reverted the commit because it was generating complaints. I recommend opening a new pull request where the new behavior is off by default. That should be pretty safe.

I reckon the long filename issue can probably be solved by removing the filename key from the loclist when the buffer number is valid.

PR #1154, opened 2017-11-20
Added filename key for `go build` linter

deltaskelta:
- Re: f224ce8a377bbb3a0deb78b98fdc6c43555791e2

- The issues that prompted the above commit which reverted changes made to `go build` and
`gometalinter` seemed to suggest that the main issue was with gometalinter and that
changes should be put into different commits so they are independent of each other

- This commit reinstates the changes to the `go build` linter which seem to be uncontested
and it also seems absolutely necessary to show errors from all files in the package which
may have caused a build failure.


w0rp:
Okay, this is the right idea. You can open a pull request for the gometalinter changes and tag the other people in there.

w0rp:
One idea I had is removing the `filename` key automatically from items where the buffer can be found. That will probably fix the long filename issue.

PR #1155, opened 2017-11-20
add option to override TMPDIR from Vim

bennycornelissen:
Hi,

this is my first attempt ever at contributing to a Vim plugin, so please bear with me 😉 

I ran into the issue that running linters inside Docker containers (using Docker Machine on MacOS) breaks ALE. I traced the issue to the `$TMPDIR` location, which on a Mac is a random dir inside `/var/folders`. However, if you are using Docker Machine, this directory by default can not be made available to a Docker container, so any linter running in a container would fail because of this.

To solve this problem, I introduced a new setting (unset by default), that allows a user to override `TMPDIR` specifically for ALE.

```vim
let g:ale_override_tmpdir = $HOME.'/.vimtmp'
```

If I now run `ALELint` for my Terraform code, it will run TFlint in a container using `$HOME/.vimtmp` as the temporary dir, which is passed to the container, and everything is fine: 

```
(started) ['/usr/local/bin/bash', '-lc', '''tflint'' -f json ''/Users/benny/.vimtmp/vHyo9Sb/35/bastion.tf''']
(finished - exit code 0) ['/usr/local/bin/bash', '-lc', '''tflint'' -f json ''/Users/benny/.vimtmp/vHyo9Sb/36/bastion.tf''']
<<<OUTPUT STARTS>>>
[{"detector":"aws_instance_default_standard_volume","type":"WARNING","message":"\"volume_type\" is not specified. Default standard volume type is not recommended. You can use \"gp2\
", \"io1\", etc instead.","line":27,"file":"/Users/benny/.vimtmp/vHyo9Sb/36/bastion.tf","link":"https://github.com/wata727/tflint/blob/master/docs/aws_instance_default_standard_v
olume.md"}]
```

I have absolutely no experience with Vader testing either so I'm still figuring out how to test this. Any help in that regard is appreciated. I've also updated the docs, but again, first time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -678,3 +678,27 @@ let g:ale_pattern_options_enabled = 1
 ```
 
 Buffer-local variables for settings always override the global settings.
+
+### 5.xvi. How can I override my TMPDIR setting?
+
+By default, ALE will use whatever your `$TMPDIR` environment variable is set
+to. If it is unset, ALE will default to `/tmp`. However, when you want to
+specifically override your `$TMPDIR` setting for ALE, you can do so in the
+following way:
+
+```vim
+" Override TMPDIR for ALE
+let g:ale_override_tmpdir = '/path/to/tmp'
+```
+
+One of the reasons to do this, is when you run linters in containers, using
+Docker Machine. By default only your home directory is shared with the Docker
+VM, which means `/tmp` or `/var/folders` (MacOS) are never available to a 
+linter inside a container. But if we were to set `$TMPDIR` to something
+inside our home directory, suddenly everything works fine:
+
+```vim
+" Set TMPDIR to subdir of homedir, just for ALE
+let g:ale_override_tmpdir = $HOME.'/.vimtmp'
+```
+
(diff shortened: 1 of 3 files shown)


bennycornelissen:
@w0rp can you elaborate on how I would set `TMPDIR` in my terminal to be different for just Vim? Because that's what I'm trying to achieve here. I don't want to completely change where my `TMPDIR` is, I just want ALE to copy my files to a different location (of my choosing).

bennycornelissen:
I agree that, in theory, I could have also just set `let TMPDIR = '/path/of/my/choosing'` in my `.vimrc` but that doesn't change that you cannot specify the tmp location for ALE. Maybe my implementation isn't the best (I think it would be better to stay away from reassigning `TMPDIR`). I think the default behaviour of ALE is just fine btw, I just think it would be better if I had the option to override that behaviour in a fine-tuned way.

w0rp:
You have a way, you set `TMPDIR` to something, and ALE won't change it.

bennycornelissen:
That would still mean reassigning `TMPDIR` at least for all of Vim, instead of being able to define for just ALE where temporary files should be kept. So there's no fine-tuned way. There's just the workaround of reassigning `TMPDIR` so ALE (and everything else) will use a different temporary storage. 

But anyway, if you don't think such fine-grained settings are needed, then let's agree to disagree and close the PR.

w0rp:
Okay.

PR #1156, opened 2017-11-20
Added option for `gometalinter` to lint package

deltaskelta:
- added tests for the `gometalinter` command

@notjrbauer @svanharmelen this is a WIP to get `gometalinter` to lint the current package

### DIFF
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -20,15 +20,13 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     if l:lint_package
         return ale#path#BufferCdString(a:buffer)
         \   . ale#Escape(l:executable)
-        \   . (!empty(l:options) ? ' ' . l:options : '')
-        \   . ' .' 
+        \   . (!empty(l:options) ? ' ' . l:options : '') . ' .'
     endif
 
     return ale#path#BufferCdString(a:buffer)
     \   . ale#Escape(l:executable)
     \   . ' --include=' . ale#Escape(ale#util#EscapePCRE(l:filename))
-    \   . (!empty(l:options) ? ' ' . l:options : '')
-    \   . ' .' 
+    \   . (!empty(l:options) ? ' ' . l:options : '') . ' .'
 endfunction
 
 function! ale_linters#go#gometalinter#GetMatches(lines) abort
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -34,9 +34,12 @@ function! ale_linters#go#gometalinter#GetMatches(lines) abort
 endfunction
 
 function! ale_linters#go#gometalinter#Handler(buffer, lines) abort
-    let l:dir = expand('#' . a:buffer . ':p:h')
+    let l:dir = getcwd()
     let l:output = []
 
+    " gometalinter always gives the output in filenames with a path that is relative to
+    " the current directory, so passing `l:dir` and the match name to `GetAbsPath` should
+    " compute the correct absolute filepath
     for l:match in ale_linters#go#gometalinter#GetMatches(a:lines)
         call add(l:output, {
         \   'filename': ale#path#GetAbsPath(l:dir, l:match[1]),
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -1,8 +1,9 @@
-" Author: Ben Reedy <https://github.com/breed808>
+" Author: Ben Reedy <https://github.com/breed808>, Jeff Willette <jrwillette88@gmail.com>
 " Description: Adds support for the gometalinter suite for Go files
 
 call ale#Set('go_gometalinter_options', '')
 call ale#Set('go_gometalinter_executable', 'gometalinter')
+call ale#Set('go_gometalinter_lint_package', 0)
 
 function! ale_linters#go#gometalinter#GetExecutable(buffer) abort
     return ale#Var(a:buffer, 'go_gometalinter_executable')
@@ -12,6 +13,13 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:executable = ale_linters#go#gometalinter#GetExecutable(a:buffer)
     let l:filename = expand('#' . a:buffer)
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
+    let l:lint_package = ale#Var(a:buffer, 'go_gometalinter_lint_package')
+
+    if l:lint_package
+        return ale#Escape(l:executable)
+        \   . (!empty(l:options) ? ' ' . l:options : '')
+        \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
+    endif
 
     return ale#Escape(l:executable)
     \   . ' --include=' . ale#Escape('^' . ale#util#EscapePCRE(l:filename))
@@ -26,10 +34,12 @@ function! ale_linters#go#gometalinter#GetMatches(lines) abort
 endfunction
 
 function! ale_linters#go#gometalinter#Handler(buffer, lines) abort
+    let l:dir = expand('#' . a:buffer . ':p:h')
     let l:output = []
 
     for l:match in ale_linters#go#gometalinter#GetMatches(a:lines)
         call add(l:output, {
+        \   'filename': ale#path#GetAbsPath(l:dir, l:match[1]),
         \   'lnum': l:match[2] + 0,
         \   'col': l:match[3] + 0,
         \   'type': tolower(l:match[4]) is# 'warning' ? 'W' : 'E',
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -15,20 +15,20 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
     let l:lint_package = ale#Var(a:buffer, 'go_gometalinter_lint_package')
 
-    " standard format string given by gometalinter, but with absolute path to file
-    let l:fmt = '{{.Path.Abs}}:{{.Line}}:{{if .Col}}{{.Col}}{{end}}:{{.Severity}}: {{.Message}} ({{.Linter}}'
+    " BufferCdString is used so that we can be sure the paths output from gometalinter can
+    " be calculated to absolute paths in the Handler
     if l:lint_package
-        return ale#Escape(l:executable)
-        \   . ' --format=' . ale#Escape(l:fmt)
+        return ale#path#BufferCdString(a:buffer)
+        \   . ale#Escape(l:executable)
         \   . (!empty(l:options) ? ' ' . l:options : '')
-        \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
+        \   . ' .' 
     endif
 
-    return ale#Escape(l:executable)
-    \   . ' --format=' . ale#Escape(l:fmt)
+    return ale#path#BufferCdString(a:buffer)
+    \   . ale#Escape(l:executable)
     \   . ' --include=' . ale#Escape(ale#util#EscapePCRE(l:filename))
     \   . (!empty(l:options) ? ' ' . l:options : '')
-    \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
+    \   . ' .' 
 endfunction
 
 function! ale_linters#go#gometalinter#GetMatches(lines) abort
@@ -38,12 +38,13 @@ function! ale_linters#go#gometalinter#GetMatches(lines) abort
 endfunction
 
 function! ale_linters#go#gometalinter#Handler(buffer, lines) abort
+    let l:dir = expand('#' . a:buffer . ':p:h')
     let l:output = []
 
     for l:match in ale_linters#go#gometalinter#GetMatches(a:lines)
         " l:match[1] will already be an absolute path, output from gometalinter
         call add(l:output, {
-        \   'filename': l:match[1],
+        \   'filename': ale#path#GetAbsPath(l:dir, l:match[1]),
         \   'lnum': l:match[2] + 0,
         \   'col': l:match[3] + 0,
         \   'type': tolower(l:match[4]) is# 'warning' ? 'W' : 'E',
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -11,7 +11,7 @@ endfunction
 
 function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:executable = ale_linters#go#gometalinter#GetExecutable(a:buffer)
-    let l:filename = expand('#' . a:buffer)
+    let l:filename = expand('#' . a:buffer . ':t')
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
     let l:lint_package = ale#Var(a:buffer, 'go_gometalinter_lint_package')
 
--- a/ale_linters/go/gometalinter.vim
+++ b/ale_linters/go/gometalinter.vim
@@ -15,14 +15,18 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
     let l:lint_package = ale#Var(a:buffer, 'go_gometalinter_lint_package')
 
+    " standard format string given by gometalinter, but with absolute path to file
+    let l:fmt = '{{.Path.Abs}}:{{.Line}}:{{if .Col}}{{.Col}}{{end}}:{{.Severity}}: {{.Message}} ({{.Linter}}'
     if l:lint_package
         return ale#Escape(l:executable)
+        \   . ' --format=' . ale#Escape(l:fmt)
         \   . (!empty(l:options) ? ' ' . l:options : '')
         \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
     endif
 
     return ale#Escape(l:executable)
-    \   . ' --include=' . ale#Escape('^' . ale#util#EscapePCRE(l:filename))
+    \   . ' --format=' . ale#Escape(l:fmt)
+    \   . ' --include=' . ale#Escape(ale#util#EscapePCRE(l:filename))
     \   . (!empty(l:options) ? ' ' . l:options : '')
     \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
 endfunction
@@ -34,15 +38,12 @@ function! ale_linters#go#gometalinter#GetMatches(lines) abort
 endfunction
 
 function! ale_linters#go#gometalinter#Handler(buffer, lines) abort
-    let l:dir = getcwd()
     let l:output = []
 
-    " gometalinter always gives the output in filenames with a path that is relative to
-    " the current directory, so passing `l:dir` and the match name to `GetAbsPath` should
-    " compute the correct absolute filepath
     for l:match in ale_linters#go#gometalinter#GetMatches(a:lines)
+        " l:match[1] will already be an absolute path, output from gometalinter
         call add(l:output, {
-        \   'filename': ale#path#GetAbsPath(l:dir, l:match[1]),
+        \   'filename': l:match[1],
         \   'lnum': l:match[2] + 0,
         \   'col': l:match[3] + 0,
         \   'type': tolower(l:match[4]) is# 'warning' ? 'W' : 'E',
--- a/test/command_callback/test_gometalinter_command_callback.vader
+++ b/test/command_callback/test_gometalinter_command_callback.vader
@@ -25,7 +25,7 @@ Execute(The gometalinter callback should return the right defaults):
   AssertEqual
   \ 'cd ' . ale#Escape(expand('%:p:h')) . ' && '
   \   . ale#Escape('gometalinter')
-  \   . ' --include=' . ale#Escape(ale#util#EscapePCRE(expand('%')))
+  \   . ' --include=' . ale#Escape(ale#util#EscapePCRE(expand('%' . ':t')))
   \   . ' .',
   \ ale_linters#go#gometalinter#GetCommand(bufnr(''))
 
@@ -38,7 +38,7 @@ Execute(The gometalinter callback should use a configured executable):
   AssertEqual
   \ 'cd ' . ale#Escape(expand('%:p:h')) . ' && '
   \   . ale#Escape('something else')
-  \   . ' --include=' . ale#Escape(ale#util#EscapePCRE(expand('%')))
+  \   . ' --include=' . ale#Escape(ale#util#EscapePCRE(expand('%' . ':t')))
   \   . ' .', 
   \ ale_linters#go#gometalinter#GetCommand(bufnr(''))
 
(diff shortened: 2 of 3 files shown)


deltaskelta:
https://github.com/w0rp/ale/pull/1154#issuecomment-345829947, it seems all that has to be solved is the path issue for this to be doable. 

I don't really know the consequences of changing the path or where to go from here so suggestions are welcome

ale_linters/go/gometalinter.vim
```diff
@@ -1,8 +1,9 @@
-" Author: Ben Reedy <https://github.com/breed808>
+" Author: Ben Reedy <https://github.com/breed808>, Jeff Willette <jrwillette88@gmail.com>
```
svanharmelen: Personally I would not change this line. Ben Reedy was the original author who added/created this file. After that several people (including myself) made changes to the file, but that doesn't change the fact that Ben is the original author and we all just made small changes.

Think it will become pretty messy if everyone who touches a file starts adding him/herself as an author...

ale_linters/go/gometalinter.vim
```diff
@@ -1,8 +1,9 @@
...
 " Description: Adds support for the gometalinter suite for Go files
 
 call ale#Set('go_gometalinter_options', '')
 call ale#Set('go_gometalinter_executable', 'gometalinter')
+call ale#Set('go_gometalinter_lint_package', 0)
```
svanharmelen: Maybe call this `go_metalinter_package` to make it just a bit less stuttery?

ale_linters/go/gometalinter.vim
```diff
@@ -12,6 +13,13 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
     let l:executable = ale_linters#go#gometalinter#GetExecutable(a:buffer)
     let l:filename = expand('#' . a:buffer)
     let l:options = ale#Var(a:buffer, 'go_gometalinter_options')
+    let l:lint_package = ale#Var(a:buffer, 'go_gometalinter_lint_package')
```
svanharmelen: Same here... Maybe just call it `l:package`?

ale_linters/go/gometalinter.vim
```diff
@@ -12,6 +13,13 @@ function! ale_linters#go#gometalinter#GetCommand(buffer) abort
...
+
+    if l:lint_package
+        return ale#Escape(l:executable)
+        \   . (!empty(l:options) ? ' ' . l:options : '')
+        \   . ' ' . ale#Escape(fnamemodify(l:filename, ':h'))
+    endif
```
svanharmelen: Maybe do an `if...else` here instead? I know that in Go that is "not done", but this is vimscript so seems like a better pattern in here...

ale_linters/go/gometalinter.vim
```diff
@@ -26,10 +34,12 @@ function! ale_linters#go#gometalinter#GetMatches(lines) abort
 endfunction
 
 function! ale_linters#go#gometalinter#Handler(buffer, lines) abort
+    let l:dir = expand('#' . a:buffer . ':p:h')
     let l:output = []
 
     for l:match in ale_linters#go#gometalinter#GetMatches(a:lines)
         call add(l:output, {
+        \   'filename': ale#path#GetAbsPath(l:dir, l:match[1]),
```
svanharmelen: The changes to this function not only cause much longer filenames, but is also causes my statusline to stop showing me the errors/warnings counts and it stops the error and warning signs to show up in the signcolumn.

So this part actually causes 3 different (yet maybe related) issues that should be fixed before we merge this PR.

ale_linters/go/gometalinter.vim
deltaskelta: IDK, I'll do whatever @w0rp wants, but I think it becomes pretty clear that he is the original author and others have touched it so that someone may think twice and check the git history before reaching out to him in the future about code that he didn't write.

ale_linters/go/gometalinter.vim
deltaskelta: I disagree, the precedent that has been set is clearly `[lang]_[tool]_[option]`. There is no tool called `metalinter` and it would be very confusing seeing the pattern of the other options broken with this rename

ale_linters/go/gometalinter.vim
deltaskelta: without the verb `lint` it becomes unclear what the variable is used for.

ale_linters/go/gometalinter.vim
deltaskelta: This seems largely objective and I find examples within `ale` which use this pattern as it is concise and clean agnostic of the language used. Look at the `scala` linters in `ale`

ale_linters/go/gometalinter.vim
deltaskelta: I agree this part might need changing, but I have too many questions about how the quickfix window handles the absolute filenames. 

Something changes in the way they are interpreted when there is a buffer change. I don't know enough of the internals to know what is happening, but I have seen it change from an absolute filename to a "relative to the current buffer" filename when a lint even happens and then the buffer is changed. 

It is possible that ale needs to change somewhere else to only display the relative path to the user in the quickfix window even though ale itself has received the absolute path. It is also possible that the linter needs to give a relative path with something like `ale#path#GetRelPath` I am not sure, @w0rp do you have any input on how to handle this?

ale_linters/go/gometalinter.vim
svanharmelen: Uuh... `gometalinter` is definately a [tool](https://github.com/alecthomas/gometalinter), so that should indeed be `[lang] => go`, `[tool] => gometalinter` and `[option] => package` Or am I missing anything?

Seems to be perfectly inline with all the other namings in this file:

`go_gometalinter_options`
`go_gometalinter_executable`

As a matter of fact, following your own reasoning it seems you have added a new pattern yourself `[lang]_[tool]_[option]_[???]`.

But hey, naming is hard right? :wink: It just caught my eye, so I thought to make a suggestion.

deltaskelta: > Maybe call this go_metalinter_package to make it just a bit less stuttery?

according to your response here the lang is `go` and the tool is `metalinter` which does not exist and could be a name of another tool. `gometalinter` is the name of the tool. IMO the option requires two words in order to be clear so `lint_package` is the option

svanharmelen: Oh, I made a type there... I intended to write `go_gometalinter_package` 😏

ale_linters/go/gometalinter.vim
w0rp: I opened an issue for doing something about long filenames here https://github.com/w0rp/ale/issues/1160

You must use absolute paths for the filename key to know which files errors are for, for other files.

@svanharmelen What is your actual problem? Is the problem that you aren't seeing errors in a file where you did before? Maybe we could just not include the `filename` key if the path is the same as the buffer's path.

svanharmelen: @w0rp well, it seems that using this code gives me 3 problems:

1. Long filenames that fill up the loclist, leaving little room for the actual error/warning. So this means every finding uses multiple lines which makes it very annoying to read;
2. I don't get the error/warning count in the statusline anymore;
3. I don't get the error/warnings signs in the signcolumn.

w0rp: It sounds like the filename key is just wrong then. If the filename matched the buffer's filename, you'd see the errors and warnings still. That means that the paths are relative to some other directory than the current file's directory. To fix that, you'll have to figure out which directory the paths are actually relative to. You might have to prefix the command with `cd file_directory &&` if the paths are just relative to whatever the working directory is.

deltaskelta:
just to clarify, it sounds like you are saying that ale is working as expected and that the clash is coming from somewhere else in the `vimrc`?

@svanharmelen can you checkout this commit and see if your problem persists even with the `gometalinter_lint_package` setting in its default state of `0`?

ale_linters/go/gometalinter.vim
deltaskelta: just to clarify, it sounds like you are saying that ale is working as expected and that the clash is coming from somewhere else in the `vimrc`?

@svanharmelen can you checkout this commit and see if your problem persists even with the `gometalinter_lint_package` setting in its default state of `0`?

w0rp: I'm saying the filename isn't correct on @svanharmelen's machine. If he sees valid errors for a file, uses this patch, and the errors disappear, then the filename has to be wrong. That means that the paths from the output aren't relative to the directory we are using.

svanharmelen: What I did earlier today is checkout this commit and test if things kept working as expected. So I didn't change my `init.vim` (I'm using NeoVim) as the default for `go_gometalinter_lint_package` is already set to `0` and so it should not change anything, right?

As for the problem(s) we are discussing, I do use [vim-rooter](https://github.com/airblade/vim-rooter) to always set my `pwd` to the project root. So I guess that could cause the differences in paths you are refering to @w0rp?

w0rp: You might need to prefix the command with `ale#path#BufferCdString(a:buffer)` to get it to work everywhere, which will make the working directory for the command predictable. You might need to do something else.

svanharmelen: Just out of curiosity... What are we trying to fix or achieve by adding the line:

```
  \ 'filename': ale#path#GetAbsPath(l:dir, l:match[1]),
```

to the output? As apparently I'm missing a use case and/or a purpose of why we are trying to change this... 🤓

w0rp: It's for showing errors for _other files_.

If you have a file `foo.go` which imports `bar.go`, you can show errors for `foo.go` and also errors in the imported `bar.go` file, which might be the cause of problems. As long as the correct filenames are used, the only difference you should notice is that some more information appears in the quickfix window or the loclist window.

deltaskelta:
I'm a bit lost on the state of things, do you need anything else from me on this?

w0rp:
@svanharmelen Could you provide the output of `:ALEInfo` when the errors disappear for you with this patch, and the filename you are checking? We can look at those and try and figure out what the problem is. I suspect the command needs to be prefixed with `cd ... &&`, or something like that.

svanharmelen:
See the output below:

```
 Current Filetype: go
Available Linters: ['go build', 'gofmt', 'golint', 'gometalinter', 'gosimple', 'go vet', 'staticcheck']
  Enabled Linters: ['gometalinter']
 Linter Variables:

let g:ale_go_gometalinter_executable = 'gometalinter'
let g:ale_go_gometalinter_options = ' --aggregate --fast --sort=line --vendor --vendored-linters --disable=gas --disable=goconst --disable=gocyclo '
 Global Variables:

let g:ale_echo_cursor = 1
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_linter_aliases = {}
let g:ale_linters = {'go': ['gometalinter'], 'html': [], 'javascript': ['eslint']}
let g:ale_open_list = 0
let g:ale_set_highlights = 0
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_error = '✖'
let g:ale_sign_warning = '⚠'
let g:ale_warn_about_trailing_whitespace = 0
  Command History:

(executable check - success) gometalinter
(started) ['/usr/local/bin/zsh', '-c', '''gometalinter'' --include=''^cmd/dsfwd/tfapi/providers/azure/config\.go''  --aggregate --fast --sort=line --vendor --vendored-linters --disable=gas --disable=goconst --disable=gocyclo  ''cmd/dsfwd/tfapi/providers/azure''']
(finished - exit code 2) ['/usr/local/bin/zsh', '-c', '''gometalinter'' --include=''^cmd/dsfwd/tfapi/providers/azure/config\.go''  --aggregate --fast --sort=line --vendor --vendored-linters --disable=gas --disable=goconst --disable=gocyclo  ''cmd/dsfwd/tfapi/providers/azure''']

<<<NO OUTPUT RETURNED>>>

(finished - exit code 3) ['/usr/local/bin/zsh', '-c', '''gometalinter'' --include=''^cmd/dsfwd/tfapi/providers/azure/config\.go''  --aggregate --fast --sort=line --vendor --vendored-linters --disable=gas --disable=goconst --disable=gocyclo  ''cmd/dsfwd/tfapi/providers/azure''']

<<<OUTPUT STARTS>>>
cmd/dsfwd/tfapi/providers/azure/config.go:98:2:error: bla (invalid operand) is not used (gotype)
cmd/dsfwd/tfapi/providers/azure/config.go:98:2:error: undeclared name: bla (gotype)
<<<OUTPUT ENDS>>>
```

And the output in my loclist looks like this:
```
/Users/svanharmelen/Documents/GoCode/src/sbp.gitlab.schubergphilis.com/toolkit/dsfw/cmd/dsfwd/tfapi/providers/azure/cmd/dsfwd/tfapi/providers/azure/config.go|98 col 2 error| bla (invalid operand) is not used (gotype)
/Users/svanharmelen/Documents/GoCode/src/sbp.gitlab.schubergphilis.com/toolkit/dsfw/cmd/dsfwd/tfapi/providers/azure/cmd/dsfwd/tfapi/providers/azure/config.go|98 col 2 error| undeclared name: bla (gotype)
```

I notice 2 things here... 

1. The path contains `cmd/dsfwd/tfapi/providers/azure` twice!
2. These paths are really long. Even when the duplicate part is fixed, it's still much much longer as before. It would be really nice if we could strip the common path (which is this whole part: `/Users/svanharmelen/Documents/GoCode/src/sbp.gitlab.schubergphilis.com/toolkit/dsfw/` in this case) so it only shows the paths related to the root folder I'm working in...

w0rp:
Aha, that's what I thought. That's the bug. The path is just wrong. What the handler is doing is resolving the path relative to say `/Users/svanharmelen/Documents/GoCode/src/sbp.gitlab.schubergphilis.com/toolkit/dsfw/cmd/dsfwd/tfapi/providers/azure`. It needs to resolve it with `/Users/svanharmelen/Documents/GoCode/src/sbp.gitlab.schubergphilis.com/toolkit/dsfw` to get the actual absolute path.

Is there some way to figure out which path to use? If there's a filename or a list of filenames we could look for, we could discover the project root and resolve the path with the path to the project root. The long filenames shouldn't be an issue for you. You won't see the long filename in the quickfix list as long as the path is correct, which is currently isn't. The only reason you saw the long filename in the list was because it was the wrong one.

w0rp:
@deltaskelta What you need to do to fix this is replace this.

```vim
let l:dir = expand('#' . a:buffer . ':p:h')
```

With this.

```vim
let l:dir = s:GetProjectRoot(a:buffer)
```

Define that function somehow. I don't know how go project roots can be detected. If you need to run a command to get the project root, then you should use a `command_chain` to get the project root. Look at say the `flake8` linter for an example of chaining commands, or perhaps the `dmd` linter for D. If you return an empty string for a callback, you can skip running a command, so you could avoid running a command to get the project root if the option for checking other files is disabled. You might be able to detect the project root by finding a file, like many other linters do.

deltaskelta:
what about just making it relative to the `$GOPATH` instead of the project root? that should be an env variable set on every machine that runs go I believe. (It would be relative to the `GoCode` directory in his output

w0rp:
If that's the way to do it, go for it. I'm not a Go user, so I don't know an awful lot about Go.

svanharmelen:
Using this $GOPATH will not work. But how is this a problem with Go only? Isn’t this a problem with Python as well?

Or is Go the only one were you want to do this multiple file/package checking which require the full path?

deltaskelta:
I was playing around with gometalinter and it always gives the relative path to the file based on the current directory no matter if a relative or absolute path is passed to it. This means that if I change `l:dir` to `getcwd()` then `GetAbsPath` should properly compute the absolute path.

svanharmelen:
I'll give it a go (no pun intended :wink:)

svanharmelen:
Using `getcwd()` seems to work perfectly 😀

deltaskelta:
it does in this case but it seems the it is causing a failure in tests because there was a check specifically added to prevent the use of `getcwd()` so we need some input from @w0rp I think

w0rp:
Never use `getcwd()` in this project. You cannot rely on it. The function can be called from a random buffer, so it could return the working directory for some other random file.

w0rp:
The following is possible.

1. Start a Go linter from a Go file.
2. Open a Python file, where the working directory is changed for whatever reason.
3. The handler for handling the messages from the Go linter runs from the Python file.

All because the functions are executed asynchronously.

svanharmelen:
That makes sense and would indeed cause other issues over time...

deltaskelta:
then since `gometalinter` always outputs relative paths, nothing can be done with that part of the filepath... which leaves two options as far as I can see...

1. use `let l:dir = expand('#' . a:buffer . ':p:h')` which gives the absolute path up to the dir of the buffer, and then yank only the filename from the gometalinter output. If this is the case then we would probably need to exclude errors coming from other packages as they would have the incorrect path.

2. get an PR pulled merged into `gometalinter` which would output absolute filepaths

Do you guys see any other options?

svanharmelen:
About the second option. They moved aways from absolute filepaths for a good reason: https://github.com/alecthomas/gometalinter/pull/319 So I very much doubt if we should want to change that back...

deltaskelta:
as far as I can tell, that PR is talking about giving absolute paths as input, and I was speaking of adding a CLI option to output absolute paths from the errors.

w0rp:
You have to figure out what the paths are relative to. The paths are relative to _something_, so figure out which path they are relative to. Then use the existing function to resolve the paths to absolute paths, which are needed here.

deltaskelta:
the paths output from `gometalinter` are relative to the current working directory. I previously had an `autocmd` that change my current directory to the directory that the active buffer is in which is why @svanharmelen didn't get the error (he must not have that in his vimrc) If I open vim in some random directory, the `gometalinter` output will be relative to that directory

deltaskelta:
I managed to squeeze out a PR in `gometalinter` that added an option to output absolute paths, so that might solve the problem simply if it gets accepted

deltaskelta:
My PR finally got merged into `gometalinter` which can now handle the absolute paths itself..

I have edited the tests to reflect the new command and output as well.

w0rp:
Can you make the code in ALE also handle the relative paths for older versions of `gometalinter`? You'll need to find what the base of the paths is, and resolve them based on that.

deltaskelta:
But how can I go about doing that?

`getcwd()` is unreliable and I need the directory from which the gometalinter binary ran in (which would be `getcwd()` the majority of the time). This means that the buffer directory is irrelevant as someone may have opened vim in some random directory far away from the file. 

Should I just get rid of the `--format` option for the standard `lint_package = 0` setting? or do you see any other way forward?

w0rp:
If the directory is just the directory the command ran from then that's easy. Just prefix the command with `cd ... &&` and the file's directory. Then use the same directory for resolving the paths. Look at how other linters handle this with the `BufferCdString` function, or whatever it's called.

deltaskelta:
should I then just ditch the whole `--format` option?

w0rp:
Yeah, you shouldn't need that.

w0rp:
Travis is just complaining because of some trailing whitespace.

w0rp:
@svanharmelen Do you want to give that a try?

svanharmelen:
@w0rp yes, I'll test this in a few minutes...

svanharmelen:
I'm afraid this still doesn't work 😞  It now goes one directory up and uses that as the path. But in a lot of cases that is not from where I execute the cmd... Let me know what info/data you need to troubleshoot.

deltaskelta:
so what does the output look like for you? it will always execute the command from that spot no matter what you do

svanharmelen:
My `pwd` is at `/Users/svanharmelen/Documents/GoCode/src/gitlab.com/toolkit/dsi` and I'm working on a file located at `/Users/svanharmelen/Documents/GoCode/src/gitlab.com/toolkit/dsi/cmd/dsi/main.go`. And now the ALE tries to `cd` to `/Users/svanharmelen/Documents/GoCode/src/gitlab.com/toolkit/dsi/dsi` and then includes (in the `gometalinter` cmd) `dsi/main.go`...

deltaskelta:
I think I see where its wrong

deltaskelta:
It should be fixed now, @svanharmelen can you try that out?

svanharmelen:
It looks like I get errors now, but it still doesn't look good:

```
 Current Filetype: go
Available Linters: ['go build', 'gofmt', 'golint', 'gometalinter', 'gosimple', 'go vet', 'staticcheck']
  Enabled Linters: ['gometalinter']
 Linter Variables:
let g:ale_go_gometalinter_executable = 'gometalinter'
let g:ale_go_gometalinter_lint_package = 0
let g:ale_go_gometalinter_options = ' --aggregate --fast --sort=line --vendor --vendored-linters --disable=gas --disable=goconst --disable=gocyclo '
 Global Variables:
let g:ale_echo_cursor = 1
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_linter_aliases = {}
let g:ale_linters = {'go': ['gometalinter'], 'html': [], 'javascript': ['eslint']}
let g:ale_open_list = 0
let g:ale_set_highlights = 0
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_error = '✖'
let g:ale_sign_warning = '⚠'
let g:ale_warn_about_trailing_whitespace = 0
  Command History:
(started) ['/usr/local/bin/zsh', '-c', 'cd ''/Users/svanharmelen/Documents/GoCode/src/gitlab.com/toolkit/dsi/dsi'' && ''gometalinter'' --include=''main\.go''  --aggregate --fast --sort=line --vendor --vendored-linters --disable=gas --disable=goconst --disable=gocyclo  .']
(finished - exit code 2) ['/usr/local/bin/zsh', '-c', 'cd ''/Users/svanharmelen/Documents/GoCode/src/gitlab.com/toolkit/dsi/dsi'' && ''gometalinter'' --include=''main\.go''  --aggregate --fast --sort=line --vendor --vendored-linters --disable=gas --disable=goconst --disable=gocyclo  .']
<<<NO OUTPUT RETURNED>>>
(finished - exit code 3) ['/usr/local/bin/zsh', '-c', 'cd ''/Users/svanharmelen/Documents/GoCode/src/gitlab.com/toolkit/dsi/dsi'' && ''gometalinter'' --include=''main\.go''  --aggregate --fast --sort=line --vendor --vendored-linters --disable=gas --disable=goconst --disable=gocyclo  .']
<<<OUTPUT STARTS>>>
main.go:54:3:error: bla (invalid operand) is not used (gotype)
main.go:54:3:error: undeclared name: bla (gotype)
<<<OUTPUT ENDS>>>
```

So it tries to `cd` to `/Users/svanharmelen/Documents/GoCode/src/gitlab.com/toolkit/dsi/dsi` which doesn't exists??

svanharmelen:
Ow, and it looks like the output is actually relative again:

```
<<<OUTPUT STARTS>>>
main.go:54:3:error: bla (invalid operand) is not used (gotype)
main.go:54:3:error: undeclared name: bla (gotype)
<<<OUTPUT ENDS>>>
```

So that is of course why I do get the errors... But you can test this yourself as well right?

deltaskelta:
The output should be relative (to your eyes) because ale takes the absolute path and truncates it based on where you are. I have no idea why you get a double cd into `dsi` but it must be a valid path because it is working, right?

On my machine everything works and I do not get a double `cd`

deltaskelta:
If it was an invalid `cd` path it would be exiting with status 1 as far as I know

svanharmelen:
Yes, hold on and let me check something...

svanharmelen:
I checked it again, and indeed it seems to be working now 👍

w0rp:
Okay, sounds good to me.

@deltaskelta Thank you for all of your work on this. :+1:
@svanharmelen deltaskelta Thank you for testing this.

PR #1157, opened 2017-11-21
[eruby] Add erubi linter

elebow:
Erubi is yet another parser for eRuby. This is the default parser in
Rails as of version 5.1. It supports some additional syntax with similar
behavior to Rails' extensions to the language, though incompatible.
Rails currently still recommends their own syntax, so GetCommand still
has to do the translation introduced in
https://github.com/w0rp/ale/pull/1114 .

Erubi does not supply an executable—It is intended to be invoked only
from within a Ruby program. Consequently, GetExecutable has to do some
trickery to inform ALE when the linter is available.

### DIFF
--- a/ale_linters/eruby/erubi.vim
+++ b/ale_linters/eruby/erubi.vim
@@ -1,0 +1,35 @@
+" Author: Eddie Lebow https://github.com/elebow
+" Description: eruby checker using `erubi`
+
+function! ale_linters#eruby#erubi#CheckErubi(buffer) abort
+    return 'ruby -r erubi/capture_end -e ' . ale#Escape('""')
+endfunction
+
+function! ale_linters#eruby#erubi#GetCommand(buffer, check_erubi_output) abort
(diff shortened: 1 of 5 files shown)


ale_linters/eruby/erubi.vim
```diff
@@ -0,0 +1,33 @@
+" Author: Eddie Lebow https://github.com/elebow
+" Description: eruby checker using `erubi`
+
+function! ale_linters#eruby#erubi#GetExecutable(buffer) abort
+    call system('ruby -r erubi/capture_end -e ' . ale#Escape('""'))
```
w0rp: This system call blocks `vim`. There's a mechanism for handling these kinds of problems with more background jobs in ALE. See `:help ale-command-chain`, and you can see existing linters which use the `command_chain` option. You can return an empty string for any command callback for skipping them, including the last callback in the chain.

elebow:
Refactored to use `command_chain`.

w0rp:
I just had a look, and it looks fine to me. Cheers! :beers:

PR #1169, opened 2017-11-27
Add support for linting GraphQL with ESLint

benjie:
Using https://github.com/apollographql/eslint-plugin-graphql, ESLint can be used to lint GraphQL files - this is good because it means the same lint rules can apply to GraphQL queries inline within your code as well as in dedicated files.

This PR basically just copies the `javascript/eslint.vim` linter and changes `javascript` for `graphql`.


doc/ale-graphql.txt
```diff
@@ -2,6 +2,9 @@
 ALE GraphQL Integration                                   *ale-graphql-options*
 
 
+===============================================================================
+eslint                                                     *ale-graphql-eslint*
```
w0rp: You should document that the JavaScript options are also used for GraphQL here. See `:help ale-typescript-eslint`.

benjie: Sorry, I'm not familiar at all with writing vim docs - could you point me to an analogous location I can copy?

w0rp: Yes, that help tag points to it. See `ale-typescript.txt`.

w0rp: `*foo*` creates a help tag for `foo`, and `|foo|` references a help tag.

benjie:
I wasn't sure quite how to word it so I just went with something simple. I also fixed a typo in the typescript docs; you're welcome to drop that commit if you want to keep things separate 👍

benjie:
(Thanks so much for this plugin by the way, it's super useful 👍 )

w0rp:
Cheers! :beers:

PR #1182, opened 2017-11-30
Puppet: allow for when parser doesn't supply column for errors

davewongillies:
Sometimes (mostly with older versions) puppet doesn't supply the column for where an error is, eg:
```
Error: Could not parse for environment production: Syntax error at '='; expected '}' at /root/puppetcode/modules/statsd/manifests/init.pp:5
```

Updated `l:pattern` to make the column optional.

### DIFF
--- a/ale_linters/puppet/puppet.vim
+++ b/ale_linters/puppet/puppet.vim
@@ -5,7 +5,7 @@ function! ale_linters#puppet#puppet#Handle(buffer, lines) abort
     " Error: Could not parse for environment production: Syntax error at ':' at /root/puppetcode/modules/nginx/manifests/init.pp:43:12
     " Error: Could not parse for environment production: Syntax error at '='; expected '}' at /root/puppetcode/modules/pancakes/manifests/init.pp:5"
 
-    let l:pattern = '^Error: .*: \(.\+\) at \/.\+\.pp:\(\d\+\):\=\(\d*\)'
+    let l:pattern = '^Error: .*: \(.\+\) at .\+\.pp:\(\d\+\):\=\(\d*\)'
     let l:output = []
 
     for l:match in ale#util#GetMatches(a:lines, l:pattern)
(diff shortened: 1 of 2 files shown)


davewongillies:
@w0rp I've never written test suites for vim before. Can you point me where to start? Or an existing test suite I could base mine off of?

w0rp:
Have a look at the tests in the `test/handler` directory. You can follow the structure of the tests there for checking how the Puppet messages are parsed.

ale_linters/puppet/puppet.vim
```diff
@@ -4,7 +4,7 @@ function! ale_linters#puppet#puppet#Handle(buffer, lines) abort
     " Matches patterns like the following:
     " Error: Could not parse for environment production: Syntax error at ':' at /root/puppetcode/modules/nginx/manifests/init.pp:43:12
 
-    let l:pattern = '^Error: .*: \(.\+\) at .\+:\(\d\+\):\(\d\+\)$'
+    let l:pattern = '^Error: .*: \(.\+\) at .\+:\(\d\+\):\=\(\d*\)$'
```
w0rp: If you look at the test result now, you'll see that it didn't parse the column number, so you can tackle that.

davewongillies: ¯\_(ツ)_/¯

I'm baffled. The weird part is that when I added the test into commit 8e3fa4a without my changes, this passes on Travis but fails on my local machine in the same way that the current tests are failing (I'm running vim8 though and I noticed Travis is testing against neovim).

davewongillies: I made some slight changes to the regex and its now passing the tests

ale_linters/puppet/puppet.vim
```diff
@@ -3,8 +3,9 @@
 function! ale_linters#puppet#puppet#Handle(buffer, lines) abort
     " Matches patterns like the following:
     " Error: Could not parse for environment production: Syntax error at ':' at /root/puppetcode/modules/nginx/manifests/init.pp:43:12
+    " Error: Could not parse for environment production: Syntax error at '='; expected '}' at /root/puppetcode/modules/pancakes/manifests/init.pp:5"
 
-    let l:pattern = '^Error: .*: \(.\+\) at .\+:\(\d\+\):\(\d\+\)$'
+    let l:pattern = '^Error: .*: \(.\+\) at \/.\+\.pp:\(\d\+\):\=\(\d*\)'
```
w0rp: The `/` at the start here won't work for Windows paths like `C:\`.

w0rp:
Cheers! :beers:

PR #1183, opened 2017-11-30
Support prospector

### LINKED ISSUE #565: Prospector Python linter support
[Prospector](http://prospector.readthedocs.io/en/master/) appears to nicely combine some of the best Python linters with sane defaults. We should support it.

chocoelho:
Resolves #565.

`prospector` is a tool that allows a unified interface for multiple linters for python code. It's a really good tools and I felt it was missing in ale.


ale_linters/python/prospector.vim
```diff
@@ -0,0 +1,80 @@
...
+        elseif l:error.source =~# '\v\[%(dodgy|pyroma|vulture)\]$'
+            let l:type = 'W'
+        else
+            let l:type = 'E'
+        endif
+
+        let l:item = {
+        \   'lnum': l:error.location.line,
+        \   'col': l:error.location.character + 1,
+        \   'text': empty(l:error.code) ? l:error.message : printf('(%s) %s: %s', l:error.source, l:error.code, l:error.message),
```
w0rp: Don't include the error code in the message. Otherwise you can't remove the error code via the formatting options. The default formatting options put the code at the front of the message.

ale_linters/python/prospector.vim
```diff
@@ -0,0 +1,80 @@
...
+        \   'code': l:error.code,
```
w0rp: Maybe the `source` should be part of the `code` here.

ale_linters/python/prospector.vim
chocoelho: Alright. What about the source? Should it be removed as well?

ale_linters/python/prospector.vim
chocoelho: In this case, what would be preferable: having the source being displayed after or before the code?

w0rp:
Cheers! :beers:

PR #1188, opened 2017-12-02
Fixed Command String for phpcbf Fixer

ejsexton82:
In order to fix STDIN instead of local files and directories for the phpcbf fixer, the command must include a single hyphen (-) as a parameter. Otherwise, phpcbf reports the error "You must supply at least one file or directory to process."


w0rp:
Cheers! :beers:

PR #1192, opened 2017-12-04
Add solhint support

fvictorio:
Add support for [solhint](https://github.com/protofire/solhint), a linter for the Solidity language.


test/handler/test_solhint_handler.vader
```diff
@@ -0,0 +1,54 @@
...
+After:
+  call ale#linter#Reset()
+
+Execute(The vint handler should parse error messages correctly):
+  AssertEqual
+  \ [
+  \   {
+  \     'lnum': 1,
+  \     'col': 17,
+  \     'text': 'Compiler version must be fixed (compiler-fixed)',
```
w0rp: Can you capture the error code at the end of the output and put it into the `code` key? Then the formatting for the error code will be configurable in ALE. Have a look at how this is handled for flake8 and ESLint.

fvictorio: Updated, thanks for the feedback.

fvictorio:
@w0rp Is the last commit what you suggested?

PR #1194, opened 2017-12-05
cppcheck should not use compilation database, it's a wrong behavior

bbchung:
ALE is a awsome vim plugin, however there is a small bug in cppcheck integration. 
ALE should check current buffer, but if it invokes cppcheck with --project=compile_commands, cppcheck will check the whole project, and it's extremely slow and make no sense

### DIFF
--- a/ale_linters/c/cppcheck.vim
+++ b/ale_linters/c/cppcheck.vim
@@ -9,23 +9,9 @@ function! ale_linters#c#cppcheck#GetExecutable(buffer) abort
 endfunction
 
 function! ale_linters#c#cppcheck#GetCommand(buffer) abort
-    " Search upwards from the file for compile_commands.json.
-    "
-    " If we find it, we'll `cd` to where the compile_commands.json file is,
-    " then use the file to set up import paths, etc.
-    let l:compile_commmands_path = ale#path#FindNearestFile(a:buffer, 'compile_commands.json')
-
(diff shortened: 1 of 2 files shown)


w0rp:
The `--project` argument was used so `cppcheck` will automatically detect compliation flags to use. Do you have another suggestion for automatically determining compilation flags? This is the suggested way to load compilation flags in the manual. http://cppcheck.sourceforge.net/manual.pdf See chapter 4.

w0rp:
The behaviour isn't "wrong," it's just different from what you want personally. Maybe you could change your pull request so it adds an option to turn this off, and leave the behavior on by default for those who are using it.

bbchung:
Does cppcheck need compilation flags? It's just linter but not come with compiler. I think cppcheck just need file names from a compilation database to check the whole project.

w0rp:
It needs to know where the headers are to do its job.

bbchung:
ok, I see, thanks for your reply!

bbchung:
However, in current implementation with compilation database, ALE is extremely slow if the project is large and unable to use. It maybe a todo job

w0rp:
I reckon you could add some options for turning this behavior off, and update the documentation explaining that you'll have to set other flags for setting headers yourself.

w0rp:
Closed due to inactivity.

PR #1198, opened 2017-12-06
goimports fixer doesn't work for vendored libraries

Carpetsmoker:
In Go you can "vendor" packages by putting them in the `vendor/`
directory for a project. `goimports` picks up these packages, in
addition to what you have in GOPATH.

The `goimports` fixer works by copying the Go file to a temporary
directory (in e.g. `/tmp/`) and then running `goimports` on it. This
means that it won't see the vendored packages, since that is outside of
the project's directory.

To be useful, `goimports` should be run on the real file, and not a
temporary file.

### DIFF
--- a/autoload/ale/fixers/goimports.vim
+++ b/autoload/ale/fixers/goimports.vim
@@ -14,7 +14,7 @@ function! ale#fixers#goimports#Fix(buffer) abort
 
     return {
     \   'command': ale#Escape(l:executable)
-    \       . ' -l -w'
+    \       . ' -l -w -srcdir %s'
     \       . (empty(l:options) ? '' : ' ' . l:options)
     \       . ' %t',
     \   'read_temporary_file': 1,
--- a/test/fixers/test_goimports_fixer_callback.vader
+++ b/test/fixers/test_goimports_fixer_callback.vader
@@ -25,7 +25,7 @@ Execute(The goimports callback should the command when the executable test passe
   AssertEqual
   \ {
   \   'read_temporary_file': 1,
-  \   'command': ale#Escape(g:ale_go_goimports_executable) . ' -l -w %t'
+  \   'command': ale#Escape(g:ale_go_goimports_executable) . ' -l -w -srcdir %s %t'
   \ },
   \ ale#fixers#goimports#Fix(bufnr(''))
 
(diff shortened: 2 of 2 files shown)


w0rp:
`:ALEFix` can't modify any of your files. It needs to pipeline the buffer data from one fixer to another. Is there any way to run `goimports` on a temporary file and also tell it where the libraries are?

autoload/ale/command.vim
```diff
@@ -35,6 +36,11 @@ function! ale#command#FormatCommand(buffer, command, pipe_file_if_needed) abort
         let l:command = substitute(l:command, '%s', '\=ale#Escape(l:filename)', 'g')
     endif
 
+    if l:command =~# '%d'
+        let l:dir = fnamemodify(bufname(a:buffer), ':p:h')
+        let l:command = substitute(l:command, '%d', '\=ale#Escape(l:dir)', 'g')
+    endif
```
w0rp: Adding this will slow down processing for every command, and it probably won't be used by too many of them. I recommend just adding the directory into the command directly instead.

autoload/ale/fixers/goimports.vim
```diff
@@ -14,7 +14,7 @@ function! ale#fixers#goimports#Fix(buffer) abort
 
     return {
     \   'command': ale#Escape(l:executable)
-    \       . ' -l -w'
+    \       . ' -l -w -srcdir %d '
     \       . (empty(l:options) ? '' : ' ' . l:options)
```
w0rp: Apart from the comment above, that looks good. :ok_hand:

autoload/ale/command.vim
Carpetsmoker: Actually we don't need this at all. -srcdir is documented as being able to work on the file directly, but when I tested that yesterday it was very slow for some reason. That was probably due to me importing a Windows VM at the same time which made /tmp/ access very slow for some reason :-/ By the time I figured that out I was practically falling asleep as I was rather tired for some reason (too much VimScript?) >_<

Anyway, PR updated and should be good to go now. Sorry for making it more complicated than it had to be.

w0rp:
Cheers! :beers:

Issue #1193, opened 2017-12-04
Ability to show if ALE is running jobs


Carpetsmoker:
I have some modules that take 10+ seconds to run all linters for. This is okay since I configured ALE to only run on write, but the annoying thing is that there is, as far as I can find, no straightforward way to see if ALE is still running a job.

This is especially annoying if you fix something and want to to wait to verify that it's indeed fixed. Because you never know if the linter stopped running, it can be a bit hard to see if you actually fixed it, or if the linter just needs a bit more time.

I created a quick patch below which fixed this; I added `let &stl .= '%{ale#statusline#IsRunning() ? "[linting]" : "[done]"}'` to my vimrc and that seems to work well.

I don't mind finishing this up and opening a proper PR, but I'm not sure if this is also the *best* way to fix it, though. Perhaps I missed some feature in ALE that already does this? Or perhaps it might be better to add this to `ale#statusline#Count()`? Or maybe something else?

---

```
diff --git i/autoload/ale/statusline.vim w/autoload/ale/statusline.vim
index a073f7a..0b06801 100644
--- i/autoload/ale/statusline.vim
+++ w/autoload/ale/statusline.vim
@@ -105,3 +105,21 @@ function! ale#statusline#Status() abort
 
     return ''
 endfunction
+
+" IsRunning reports if ALE is currently running a background job.
+function! ale#statusline#IsRunning() abort
+    let l:job_list = []
+
+    " Gather all of the jobs from every buffer.
+    for l:info in values(g:ale_buffer_info)
+        call extend(l:job_list, get(l:info, 'job_list', []))
+    endfor
+
+    for l:job_id in l:job_list
+        if ale#job#IsRunning(l:job_id)
+            return 1
+        endif
+    endfor
+
+    return 0
+endfunction
```

w0rp:
This could be a good idea.

w0rp:
You might as well make the function return the number of running jobs, and then that could be used for either printing a number or just telling you if anything is running.

Carpetsmoker:
Do you think it would make more sense to create a new `ale#statusline#IsRunning()` function, or to add a new `jobs` key to the existing `ale#statusline#Count()` function?

Currently, this gets jobs for all buffers; which could be considered both a good or bad thing, depending on what you want. Something to be said for both, I think.

w0rp:
[...]

PR #1203, opened 2017-12-07
Add ALEStartLint autocmd

Carpetsmoker:
This grew out of my work in #1193; to ensure the statusline was being
updated I had to add:

    fun! s:redraw(timer)
        redrawstatus
    endfun

    augroup ALEProgress
        autocmd!
        autocmd BufWritePost * call timer_start(100, function('s:redraw'))
        autocmd User ALELint redrawstatus
    augroup end

Which kind of works, but is ugly. With this, I can replace the
`BufWritePost` with:

    autocmd User ALEStartLint redrawstatus

Which is much better, IMHO.

Actually, this patch actually replaces adding a function, since you can
do:

    augroup ALEProgress
        autocmd!
        autocmd User ALEStartLint hi Statusline ctermfg=darkgrey
        autocmd User ALELint      hi Statusline ctermfg=NONE
    augroup end

or:

    let s:ale_running = 0
    let l:stl .= '%{s:ale_running ? "[linting]" : ""}'
    augroup ALEProgress
        autocmd!
        autocmd User ALEStartLint let s:ale_running = 1 | redrawstatus
        autocmd User ALELint      let s:ale_running = 0 | redrawstatus
    augroup end

Both seem to work very well in my testing.

No need to `ale#Statusline#IsRunning()` anymore, I think?

### DIFF
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -556,6 +556,8 @@ function! s:RunJob(options) abort
         \   'output': [],
         \   'next_chain_index': l:next_chain_index,
         \}
+
+        silent doautocmd <nomodeline> User ALEStartLint
     endif
 
     if g:ale_history_enabled
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -321,6 +321,8 @@ function! ale#engine#SetResults(buffer, loclist) abort
         call ale#engine#RemoveManagedFiles(a:buffer)
 
         " Call user autocommands. This allows users to hook into ALE's lint cycle.
+        silent doautocmd <nomodeline> User ALELintPost
+        " Old DEPRECATED name; call it for backwards compatibility.
         silent doautocmd <nomodeline> User ALELint
     endif
 endfunction
@@ -556,8 +558,6 @@ function! s:RunJob(options) abort
         \   'output': [],
         \   'next_chain_index': l:next_chain_index,
         \}
-
-        silent doautocmd <nomodeline> User ALEStartLint
     endif
 
     if g:ale_history_enabled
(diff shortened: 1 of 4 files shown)


Carpetsmoker:
Maybe it would be nice to document some of those autocmds in the help page too, btw?

autoload/ale/engine.vim
```diff
@@ -556,6 +556,8 @@ function! s:RunJob(options) abort
         \   'output': [],
         \   'next_chain_index': l:next_chain_index,
         \}
+
+        silent doautocmd <nomodeline> User ALEStartLint
```
w0rp: `ALEStartLint` isn't consistent with the `ALELint` event here. `ALELint` is only executed after all of the jobs finish. I recommend moving this to the `RunLinters` function instead.

The question is, should the event run before jobs are run, or slightly after they are run? What if no jobs end up being run? It might be best to just run the event before any linters are run if and only if at least one linter is in the list. Then you'll be able to change options in the command before they are used.

w0rp:
This feature is described in this existing issue here. https://github.com/w0rp/ale/issues/623 I recommend using the name `ALELintPre` instead, which is consistent with Vim events names. Unfortunately the current event name will have to say as `ALELint` for now, but we could also run `ALELintPost` and encourage people to use that instead.

autoload/ale/engine.vim
Carpetsmoker: I changed it so that it always gets run before all jobs unless there are no defined linters. That's consistent with how `ALELint` currently works as far as I can see.

w0rp:
Cheers! :beers:

unphased:
@Carpetsmoker @w0rp you guys must be using airline, right? Is there no integration? I'd like to have a section in here that uses these new autocommands to indicate the status of async jobs. Preferably also be able to count the # of currently running jobs.

w0rp:
Open an issue for `vim-airline`, or create a plugin for it. That kind of stuff isn't handled by this codebase, just the support for writing that kind of stuff.

unphased:
Do you know if counting # of running ALE jobs is possible?

w0rp:
There's nothing for getting the number of running jobs, but you can check if a buffer is currently being checked. See `:help ale#engine#IsCheckingBuffer()`

Have a look through `:help ale-api` for everything else the API offers.

PR #1206, opened 2017-12-08
Add support for go_gobuild_options

hasufell:
Use case: `let g:ale_go_gobuild_options = "-gcflags='-e'"`

### DIFF
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -4,6 +4,8 @@
 
 " inspired by work from dzhou121 <dzhou121@gmail.com>
 
+call ale#Set('go_gobuild_options', '')
+
 function! ale_linters#go#gobuild#GoEnv(buffer) abort
     if exists('s:go_env')
         return ''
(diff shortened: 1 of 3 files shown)


Carpetsmoker:
Would it make sense to set the default to `"-gcflags='-e'"` (`-e` limits the number of reported errors; to 10 I believe).
Limiting errors makes sense when building something with `go install` or `go test`, but in the context of ALE it probably makes a bit less sense?

hasufell:
@Carpetsmoker I agree, it's unlikely that you only want the first 10 errors and counter-intuitive.

wrt tests, I have no idea how they work and I am not sure I have the time to look into that right now

Carpetsmoker:
I don't mind picking up from this PR and finishing up with tests etc. if you don't have time. May be a few days though.

hasufell:
I lost interest and use gometalinter now.

w0rp:
I just picked this up and added support for the option now with some tests. I also fixed the setting of `GOPATH` so it should work on Windows, and escaped the paths for `GOPATH` and `cd`.

PR #1220, opened 2017-12-13
Add a linter for alex

languitar:
https://github.com/wooorm/alex

Enabled for text-like file formats and documented in README and doc.

Tests for the message handler exist.

Things have been constructed along the write-good example, apart from the variable command and option handling.


w0rp:
Cheers! :beers:

PR #1230, opened 2017-12-17
Enable the vale linter also for further markup languages

languitar:
vale also works quite well on LaTeX and rst files.


languitar:
The added files need to be updated after #1232 or the other way around.

w0rp:
Cheers! :beers:

PR #1232, opened 2017-12-17
Use JSON output with vale

languitar:
Switches all vale instances to JSON output and provides an appropriate handler for that. Without JSON, no end_col is provided and text highlighting only catches the first character of every result.

Tests are included.


w0rp:
Looks good to me.

w0rp:
Cheers! :beers:

PR #1233, opened 2017-12-17
Add support for linting git commit message files

nickdiego:
- This makes possible to lint commit messages
- Using `gitlint` project: https://github.com/jorisroovers/gitlint

### DIFF
--- a/ale_linters/gitcommit/gitlint.vim
+++ b/ale_linters/gitcommit/gitlint.vim
@@ -8,12 +8,13 @@ let g:ale_gitcommit_gitlint_use_global = get(g:, 'ale_gitcommit_gitlint_use_glob
 
 
 function! ale_linters#gitcommit#gitlint#GetExecutable(buffer) abort
-    return ale#Var(a:buffer, 'gitcommit_gitlint_executable')
+    return ale#python#FindExecutable(a:buffer, 'gitcommit_gitlint', ['gitlint'])
 endfunction
 
 function! ale_linters#gitcommit#gitlint#GetCommand(buffer) abort
     let l:options = ale#Var(a:buffer, 'gitcommit_gitlint_options')
-    return ale#Escape(ale_linters#gitcommit#gitlint#GetExecutable(a:buffer))
+    let l:executable = ale_linters#gitcommit#gitlint#GetExecutable(a:buffer)
+    return ale#Escape(l:executable)
     \   . (!empty(l:options) ? ' ' . l:options : '')
     \   . ' lint'
 endfunction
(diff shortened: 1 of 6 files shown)


nickdiego:
Hello,
As I was not familiar to the ALE codebase neither to Vim script, I implemented gitlint integration based on flake8 linter, so maybe there is still some unnecessary code in this patches.
Let me know if there's something I should remove or rearrange..
Thanks

nickdiego:
> I think you can probably remove the parts of the code relating to running the linter with -m. That's mostly useful for flake8 for running it with different Python versions, which accept different syntax.
> 
> You can remove the version checking parts. Replace command_chain with command_callback and just the one callback instead.
> 
> You can remove the part which parses Python exceptions probably. That's needed for flake8 as it sometimes throws exceptions for syntax errors and such.
> 
> You can remove the old _args fallback, as this is a new linter.

Ok, done. :)

Thanks

ale_linters/gitcommit/gitlint.vim
```diff
@@ -0,0 +1,50 @@
+" Author: Nick Yamane <nick.diego@gmail.com>
+" Description: gitlint for git commit message files
+
+let g:ale_gitcommit_gitlint_executable =
+\   get(g:, 'ale_gitcommit_gitlint_executable', 'gitlint')
+let g:ale_gitcommit_gitlint_options = get(g:, 'ale_gitcommit_gitlint_options', '')
+let g:ale_gitcommit_gitlint_use_global = get(g:, 'ale_gitcommit_gitlint_use_global', 0)
```
w0rp: This flag isn't used. You should either remove it from here and the documentation, or use it to find the path to the executable in virtualenv with the python function. You might want to do the latter.

nickdiego: Oops, done.

doc/ale.txt
```diff
@@ -68,6 +68,8 @@ CONTENTS                                                         *ale-contents*
       gcc.................................|ale-fortran-gcc|
     fusionscript..........................|ale-fuse-options|
       fusion-lint.........................|ale-fuse-fusionlint|
+    git commit.............................|ale-gitcommit-options|
```
w0rp: The helptag is misaligned here.

doc/ale-gitcommit.txt
```diff
@@ -0,0 +1,41 @@
...
+
+  This variable can be changed to add command-line arguments to the gitlint
+  invocation.
+
+  For example, to dinamically set the gitlint configuration file path, you
+  may want to set >
+
+  let g:ale_gitcommit_gitlint_options = '-C /home/user/.config/gitlint.ini'
+<
+g:ale_gitcommit_gitlint_use_global         *g:ale_gitcommit_gitlint_use_global*
```
w0rp: Put two blank lines after option sections. The line with `<` can be considered blank, as the character is concealed.

doc/ale.txt
nickdiego: Done.

doc/ale-gitcommit.txt
nickdiego: Ok. Done.

PR #1238, opened 2017-12-18
Make alex a file linter

languitar:
alex does not find its configuration file when using temporary files for input.


w0rp:
Cheers! :beers:

Issue #1163, opened 2017-11-23
ALE/linter ignores specified include directories (even with compile_commands.json and ctags files)


maskedbyte:
```Current Filetype: c
Available Linters: ['clang', 'clangtidy', 'cppcheck', 'gcc']
  Enabled Linters: ['clang', 'clangtidy', 'cppcheck', 'gcc']
 Linter Variables:

let g:ale_c_clang_executable = 'clang'
let g:ale_c_clang_options = '-std=c11 -Wall'
let g:ale_c_clangtidy_checks = ['*']
let g:ale_c_clangtidy_executable = 'clang-tidy'
let g:ale_c_clangtidy_options = ''
let g:ale_c_cppcheck_executable = 'cppcheck'
let g:ale_c_cppcheck_options = '--enable=style'
let g:ale_c_gcc_executable = 'gcc'
let g:ale_c_gcc_options = '-std=c11 -Wall'
 Global Variables:

let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'always'
let g:ale_linter_aliases = {}
let g:ale_linters = {}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_column_always = 0
let g:ale_sign_error = '!'
let g:ale_sign_offset = 1000000
let g:ale_sign_warning = '*'
let g:ale_statusline_format = ['%d error(s)', '%d warning(s)', 'OK']
let g:ale_warn_about_trailing_whitespace = 1
  Command History:

(executable check - success) clang
(started) ['/usr/local/bin/zsh', '-c', '''clang'' -S -x c -fsyntax-only -iquote ''/usr/home/omega/code/motherkit/src'' -std=c11 -Wall - < ''/tmp/nvimGZO0Dd/4/main.c''']
(executable check - failure) clang-tidy
(executable check - failure) cppcheck
(executable check - success) gcc
[...]

PR #1242, opened 2017-12-20
use compile_commands.json for clang linter

tex:
This parses compile_commands.json and uses specific compile command for clang. Works fine for me.

### DIFF
--- a/ale_linters/cpp/clang.vim
+++ b/ale_linters/cpp/clang.vim
@@ -26,7 +26,11 @@ function! ale_linters#cpp#clang#GetCommand(buffer) abort
         let l:b = fnamemodify(bufname(a:buffer), ':p')
         for l:d in l:l
             if l:d["file"] == l:b
-                return l:d["command"] . ' -fsyntax-only'
+                " Key 'command' contains -o with path relative to key 'directory'
+                " Make it absolute path. This way this linter generates valid
+                " .o which makes subsequent build of whole project faster...
+                let l:s = substitute(l:d["command"], "-o ", "-o ".l:d["directory"]."/","")
+                return l:s
             endif
         endfor
         echom "ALE: Please refresh your compile_commands.json!"
--- a/ale_linters/cpp/clang.vim
+++ b/ale_linters/cpp/clang.vim
@@ -1,4 +1,5 @@
 " Author: Tomota Nakamura <https://github.com/tomotanakamura>
+" Author: Milan Svoboda <https://github.com/tex>
 " Description: clang linter for cpp files
 
 call ale#Set('cpp_clang_executable', 'clang++')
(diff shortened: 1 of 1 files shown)


tex:
also fixes #1163

tex:
also fixes #1163

tex:
I did not notice the #725 though.

tex:
Also this works - but it is dependency on roxma/ncm-clang:
    autocmd BufEnter *.cpp,*.h,*.hpp,*.hxx let g:ale_cpp_clang_options = join(ncm_clang#compilation_info()['args'], ' ')

esran:
I created #725 a while ago in order to use compile_commands with the various c/c++ linters. Unfortunately I've never had time to go back and finish it off properly. I've actually moved on; switched back to master and put the compile_commands stuff in my vimrc. Not sure when or if I'd be able to find the time to go back and finish stuff off (some edge cases and vader testing).

ale_linters/cpp/clang.vim
```diff
@@ -9,15 +10,31 @@ function! ale_linters#cpp#clang#GetExecutable(buffer) abort
...
+        let l:paths = ale#c#FindLocalHeaderPaths(a:buffer)
+        " -iquote with the directory the file is in makes #include work for
+        "  headers in the same directory.
+        return ale#Escape(ale_linters#cpp#clang#GetExecutable(a:buffer))
+        \   . ' -S -x c++ -fsyntax-only '
+        \   . '-iquote ' . ale#Escape(fnamemodify(bufname(a:buffer), ':p:h')) . ' '
+        \   . ale#c#IncludeOptions(l:paths)
+        \   . ale#Var(a:buffer, 'cpp_clang_options') . ' -'
+    else
+        let l:c = readfile(l:build_dir . '/compile_commands.json')
```
w0rp: For implementing this, you ought to make the `readfile` part a command to execute, so the file can be read in the background. `cat` might work for Unix. For windows, I dunno, maybe `type`? The command should be part of a command chain.

tex: Why?

w0rp: `readfile` calls are blocking and you can read the file in the background instead, so it won't interrupt editing the file.

tex: Well, do you have something like that already used anywhere else so I can look how to do it?
Besides, readfile is native function, it should be pretty fast...This code gets executed only on some even (save, or something), does it makes any hit into user's perception?

w0rp:
Closed due to inactivity.

ale_linters/cpp/clang.vim
tbodt: `readfile` is very fast, that is true, but `json_decode` is not so fast.

ale_linters/cpp/clang.vim
```diff
@@ -9,15 +10,31 @@ function! ale_linters#cpp#clang#GetExecutable(buffer) abort
...
+        let l:l = json_decode(l:c)
+        let l:b = fnamemodify(bufname(a:buffer), ':p')
+        for l:d in l:l
+            if l:d["file"] == l:b
```
tbodt: The filenames in the compile_commands.json files generated by ninja (and therefore meson) are relative paths, so l:d["file"] should be expanded to an absolute path as well to compare.

PR #1243, opened 2017-12-21
Add fountain proselint support

JansenMitchell:
This is the result of optimizing my Vim setup for screenwriting. [Fountain](https://fountain.io/) is essentially a screenplay markdown and I wanted to lint it with proselint. proselint works on the file type and I decided to integrate it to work with ALE in the same manner that it works for markdown.

The integration works as it should and I feel it could be useful for anyone desiring the same functionality. If there is anything else that needs to be done I'm all ears.


w0rp:
I merged this via the commandline after fixing some bad indentation. Cheers! :beers:

PR #1248, opened 2017-12-23
Add ALEFixPre and ALEFixPost events

### LINKED ISSUE #623: Add autocmd events for performing actions before linting or fixing is done
Add `ALELintPre` and `ALEFixPre` for running code before ALE lints or fixes a file.

For completion's sake, add an `ALEFix` event to match the existing `ALELint` event.

Carpetsmoker:
To run autocmd before and after every fix cycle.

Fixes #623 (`ALELintPre` was added in #1203).

### DIFF
--- a/autoload/ale/fix.vim
+++ b/autoload/ale/fix.vim
@@ -54,6 +54,8 @@ function! ale#fix#ApplyQueuedFixes() abort
         let l:should_lint = l:data.changes_made
     endif
 
+    silent doautocmd <nomodeline> User ALEFixPost
+
     " If ALE linting is enabled, check for problems with the file again after
     " fixing problems.
     if g:ale_enabled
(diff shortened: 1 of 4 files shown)


autoload/ale/fix.vim
```diff
@@ -457,13 +457,17 @@ function! ale#fix#Fix(...) abort
...
+    silent doautocmd <nomodeline> User ALEFixPre
+
     call s:RunFixer({
     \   'buffer': l:buffer,
     \   'input': g:ale_fix_buffer_data[l:buffer].lines_before,
     \   'callback_index': 0,
     \   'callback_list': l:callback_list,
     \})
 
+    silent doautocmd <nomodeline> User ALEFixPost
```
w0rp: This line should be moved to `ApplyFixes` or at the end of `ApplyQueuedFixes`, but before linting is queued again. The former would execute when ALE is ready to replace a buffer's lines. The latter would execute after ALE has actually replaced the lines. Vim buffers can't be edited unless you're looking at them, so the the changes aren't applied until the buffer is opened again, which isn't often a problem as things tend to get fixed while you're still looking at the buffer. I'd recommend the latter option.

Sorry about the delay. I was on holiday.

Carpetsmoker: Thanks for the feedback :+1:

I moved the line and tested again. Seems to work well. Also rebased the branch on master as it's been a while.

w0rp:
Cheers! :beers:

PR #1252, opened 2017-12-26
Fish shell linter

nthapaliya:
Simple linter for fish files, using output from `fish -n`

The output for `fish -n` is multiline, so required some hacky parsing.

Tests pass. Docs and readme updated.

I rarely write vim script, so if there are better and more readable ways to achieve what my code does, please let me know.


w0rp:
Cheers! :beers:

PR #1254, opened 2017-12-27
Don't use temporary file for rustfmt fixer

bkhl:
rustfmt normally acts on a file in place, and applies configuration from rustfmt.toml files according to the path of the file.

Using a temporary file in `/tmp/` for rustfmt breaks this functionality, so removing the '%t' from the rustfmt command.


w0rp:
Cheers! :beers:

PR #1256, opened 2018-01-01
Fix #1255: Move extra clang-check args to before user options

### LINKED ISSUE #1255: clang-check unused arguments
## Information
When executing clang-check, it warns about unused arguments, introduced in #703 
Removing the "-extra-arg" removes the warning and does not generate plist files.

This seems like it should have been caught in #703 if it weren't version dependent however..?

**VIM version**
NVIM v0.2.2
Build type: Release

Operating System: Linux

LLVM version 5.0.1

## What went wrong
```
warning: argument unused during compilation: '-e xtra-arg' [-Wunused-command-line-argument]
warning: argument unused during compilation: '-analyzer-output=text' [-Wunused-command-line-argument]
error: unknown argument: '-extra-arg'
```

## Reproducing the bug

Steps for repeating the bug:

1. Run clang-check
2. Profit

jonatanolofsson:
Fix issue #1255

### DIFF
--- a/ale_linters/cpp/clangcheck.vim
+++ b/ale_linters/cpp/clangcheck.vim
@@ -24,9 +24,9 @@ function! ale_linters#cpp#clangcheck#GetCommand(buffer) abort
     " detected.
     return ale#Escape(ale_linters#cpp#clangcheck#GetExecutable(a:buffer))
     \   . ' -analyze %s'
+    \   . (empty(l:build_dir) ? ' -extra-arg -Xclang -extra-arg -analyzer-output=text' : '')
     \   . (!empty(l:user_options) ? ' ' . l:user_options : '')
     \   . (!empty(l:build_dir) ? ' -p ' . ale#Escape(l:build_dir) : '')
-    \   . (empty(l:build_dir) ? ' -extra-arg -Xanalyzer -extra-arg -analyzer-output=text' : '')
 endfunction
 
 call ale#linter#Define('cpp', {
--- a/test/command_callback/test_cpp_clangcheck_command_callbacks.vader
+++ b/test/command_callback/test_cpp_clangcheck_command_callbacks.vader
@@ -47,8 +47,8 @@ Execute(The options should be configurable):
   AssertEqual
   \ ale#Escape('clang-check')
   \   . ' -analyze %s'
-  \   . ' -extra-arg -Xclang -extra-arg -analyzer-output=text',
-  \   . ' --something'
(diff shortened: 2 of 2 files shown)


w0rp:
@mkarpoff Will this work? I don't use C or C++ much myself.

jonatanolofsson:
I did some more digging, and it turned out the error was not the extra args, but that they were inserted after the user options, which breaks the command if they contain the '--' separator, which is used e.g. when setting include-directories and defines. Instead of removing them, I moved them to before the user_options.

ale_linters/cpp/clangcheck.vim
```diff
@@ -24,9 +24,9 @@ function! ale_linters#cpp#clangcheck#GetCommand(buffer) abort
     " detected.
     return ale#Escape(ale_linters#cpp#clangcheck#GetExecutable(a:buffer))
     \   . ' -analyze %s'
+    \   . (empty(l:build_dir) ? ' -extra-arg -Xclang -extra-arg -analyzer-output=text' : '')
```
w0rp: Should this be `-Xanalyzer` instead of `-Xclang`?

jonatanolofsson: For me, -Xanalyzer yields
> warning: argument unused during compilation: '-Xanalyzer -analyzer-output=text' [-Wunused-command-line-argument]

whereas -Xclang yields no warnings.

Source: https://stackoverflow.com/questions/22371789/how-to-make-the-clang-static-analyzer-output-its-working-from-command-line.

Both seem to report errors on the output though, but -Xclang doesn't throw a warning itself doing so.

w0rp: Okay!

test/command_callback/test_cpp_clangcheck_command_callbacks.vader
```diff
@@ -38,16 +38,17 @@ Execute(The executable should be used in the command):
...
 Execute(The options should be configurable):
   let b:ale_cpp_clangcheck_options = '--something'
 
   AssertEqual
   \ ale#Escape('clang-check')
-  \   . ' -analyze %s --something'
-  \   . ' -extra-arg -Xanalyzer -extra-arg -analyzer-output=text',
+  \   . ' -analyze %s'
+  \   . ' -extra-arg -Xclang -extra-arg -analyzer-output=text',
+  \   . ' --something'
```
w0rp: This test broke because of the comma on the previous line.

PR #1262, opened 2018-01-02
html: Avoid old tidy on macOS

rhysd:
## Problem

On macOS, Apple's command line toolchain installs very old `tidy` command (It was released on 31 Oct 2006) to `usr/bin`. It does not consider new specs such as HTML5 so we should avoid it.

## Solution

If `tidy` command is resolved to `/usr/bin/tidy`, let's say to ignore the executable to avoid above problem. I added `exists('*exepath')` check because `exepath()` was added to Vim recently.

### DIFF
--- a/ale_linters/html/tidy.vim
+++ b/ale_linters/html/tidy.vim
@@ -25,8 +25,16 @@ function! ale_linters#html#tidy#GetCommand(buffer) abort
     \   'utf-8':        '-utf8',
     \ }, &fileencoding, '-utf8')
 
+    " On macOS, old tidy (released on 31 Oct 2006) is installed. It does not
+    " consider HTML5 so we should avoid it.
+    let l:executable = ale#Var(a:buffer, 'html_tidy_executable')
+    if has('mac') && l:executable is# 'tidy' && exists('*exepath')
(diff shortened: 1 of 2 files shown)


rhysd:
Umm... I cannot say there is no user who want to use old `/usr/bin/tidy`. I thought that `/usr/bin/tidy` raises some outdated warnings (i.e. it claims that `charset` should accompany with `content` attribute, but actually `content` should NOT be added on HTML5 (deprecated)) and it is not useful.

Is there any way to show a warning? Then I think we can raise a warning that `/usr/bin/tidy` is old and ignored.

> We should probably update the documentation to explain why the old version isn't run, and how to install newer versions.

Definitely.

w0rp:
I think just add a note in the documentation that the old version of `tidy` is disabled on Mac OSX because it is misleading.

rhysd:
Sure. I'll update documents tomorrow.

rhysd:
@w0rp I update the document for `tidy`. Could you review 7a1a534?

w0rp:
Cheers! :beers:

rhysd:
I’m not a native English user. So I’d appreciate correcting my language. Thank you.

PR #1266, opened 2018-01-06
Work around hot-reloading issue

sharils:
See https://github.com/phoenixframework/phoenix/issues/1165 for more detail

---
This change is [Reviewable](https://reviewable.io/reviews/w0rp/ale/1266)

### DIFF
--- a/ale_linters/elixir/credo.vim
+++ b/ale_linters/elixir/credo.vim
@@ -32,6 +32,6 @@ endfunction
 call ale#linter#Define('elixir', {
 \   'name': 'credo',
 \   'executable': 'mix',
-\   'command': 'mix credo suggest --format=flycheck --read-from-stdin %s',
+\   'command': 'mix help credo && mix credo suggest --format=flycheck --read-from-stdin %s',
 \   'callback': 'ale_linters#elixir#credo#Handle',
 \})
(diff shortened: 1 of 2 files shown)


ale_linters/elixir/credo.vim
```diff
@@ -32,6 +32,6 @@ endfunction
 call ale#linter#Define('elixir', {
 \   'name': 'credo',
 \   'executable': 'mix',
-\   'command': 'mix credo suggest --format=flycheck --read-from-stdin %s',
+\   'command': 'MIX_ENV=lint mix credo suggest --format=flycheck --read-from-stdin %s',
```
w0rp: For windows, this will need to be `set MIX_ENV=lint && ...` Look at say `ale_linters/slim/slimlint.vim` for an example of setting environment variables for all supported platforms.

sharils: Thanks @w0rp . But I'll need more time to incorporate that and test it on windows. Feel free to close this PR if you'd like to. I can open another one when it's ready.

sharils:
Hi @w0rp - Would you mind reviewing again? I've just rebased and merged my latest commit because the first commit caused other issue, and I have to take another approach and it turns out that approach doesn't have to rely on setting environment variables. Since the intermediate approach doesn't work, I just removed that commit. Hope you don't mind. Thanks a million.

w0rp:
Cheers! :beers:

PR #1268, opened 2018-01-06
Make it possible to inject flags of protoc invocation.

bbannier:
Typically proto files depend on and make use of proto definitions in
other files. When invoking protoc user can supply paths to inspect for
dependencies.

This patch makes it possible to configure flags passed to protoc. This
makes it e.g., possible to change include paths of the linter's protoc
invocation.


bbannier:
Good catch on the default include path @w0rp. It could indeed be pretty cumbersome to restore the original behavior for arbitrary file paths if all the user wanted was to add an additional path. I left the original include flag in place. I also updated the documentation to call this out.

I also added a dedicated test checking the behavior of the conf var.

w0rp:
Cheers! :beers:

PR #1271, opened 2018-01-08
set missing Credo column to 1

tap349:
Credo column might be missing in Credo output:

```
lib/filename.ex:19: F: Pipe chain should start with a raw value.
```

in this case it's set to 0 in _credo.vim_:

```vim
\   'col': l:match[2] + 0,
```

but 0 is invalid Vim column - it causes a bug when navigating to previous issue with `ALEPrevious` or `ALEPreviousWrap` commands.

### DIFF
--- a/ale_linters/elixir/credo.vim
+++ b/ale_linters/elixir/credo.vim
@@ -20,7 +20,7 @@ function! ale_linters#elixir#credo#Handle(buffer, lines) abort
         " column can be missing:
         "
         " lib/filename.ex:19: F: Pipe chain should start with a raw value.
-        let l:col = l:match[2] || 1
+        let l:col = empty(l:match[2]) ? 1 : l:match[2]
 
         call add(l:output, {
         \   'bufnr': a:buffer,
--- a/ale_linters/elixir/credo.vim
+++ b/ale_linters/elixir/credo.vim
@@ -17,10 +17,15 @@ function! ale_linters#elixir#credo#Handle(buffer, lines) abort
             let l:type = 'W'
         endif
 
+        " column can be missing:
+        "
+        " lib/filename.ex:19: F: Pipe chain should start with a raw value.
+        let l:col = l:match[2] || 1
+
         call add(l:output, {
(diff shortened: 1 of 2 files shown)


tap349:
@w0rp I fixed existing test for Credo but I think it might be an issue for other linters as well (AFAIU) since for many linters `'col': 0` is considered to be a valid value in their respective tests.

consider this scenario:

1) say, file has 2 errors found by ALE. say, their line and column numbers are [1, 10] and [20, 0]. note: column number of the 2nd error is 0 since it has no information about column (just like for Credo - or else it might be linter specific)
2) I want to navigate to the 2nd error and call `ALENext` (or `ALENextWrap`)
3) `cursor(l:nearest)` is called in `loclist_jumping.vim` when navigating to the 2nd error - `l:nearest` at this point is a list - [20, 0]
4) when `cursor` function is passed column 0 it jumps to specified line number (20) but keeps current column number (10 - column number of the 1st error) - since it cannot jump to column 0. so new cursor position is [20, 10] - not [20, 0] or [20, 1]
5) then I want to get back to the 1st error and call `ALEPrevious` (or `ALEPreviousWrap`)
6) but `ale#loclist_jumping#FindNearest` function finds not the 1st error but the 2nd error again since current cursor position is [20, 10] and cursor position of the 2nd error is [20, 0] so the 2nd error is still behind current cursor position - but when `cursor(l:nearest)` is called it doesn't jump to [20, 0] but keeps cursor at the same position (because of column 0)

=> nothing happens when I call `ALEPrevious` (or `ALEPreviousWrap`) - cursor remains at position [20, 10] forever.

sorry if I couldn't manage to express thoughts clearly. also I can be wrong about described scenario since my knowledge of how ALE works is very limited.

w0rp:
I think you might be right there. A `0` column value is accepted for when there's no column index for a problem. There might be a bug with jumping through errors so jumping doesn't work when the column is 0.

w0rp:
I opened an issue for the column jumping bug.

PR #1276, opened 2018-01-11
Change Pylint code to be the message ID

pmacosta:
The message ID is the information needed to disable error or warning
types (if desired) via "#pylint: disable={MSD-ID}" so it is useful to have
it be available (g:ale_echo_msg_format needs to include %code%)


w0rp:
You can also use `#pylint: disable=friendly-name` too. You should use that, as it's human-readable.

pmacosta:
Thank you, I was unaware of this option

PR #1282, opened 2018-01-13
 fixed #1038:  htmlhint should be use current filename

### LINKED ISSUE #1038: Set --config for htmlhint
I'm trying to get htmlhint working with my htmlhint config file. I have it created and named `.htmlhintrc` and it is in the root of my project. While htmlhint is working in Ale it seem to be using the default config for the lint rules. No matter what I change in my .htmlhintrc file it doesn't take. According to https://github.com/yaniswang/HTMLHint/wiki/Usage .htmlhintrc is the correct name and it should be in the root dir. Running htmlhint from the command line works just fine.

Doing this approach works great with stylelint. I was hoping the same would be for htmlhint. Is this a real issue, is it something wrong with htmlhint, or is it just me doing something dumb? Thanks

Current Filetype: html
Available Linters: ['htmlhint', 'proselint', 'tidy']
  Enabled Linters: ['htmlhint']
 Linter Variables:

let g:ale_html_htmlhint_executable = 'htmlhint'
let g:ale_html_htmlhint_options = '--format=unix'
let g:ale_html_htmlhint_use_global = 0
 Global Variables:

let g:ale_echo_cursor = 1
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_linter_aliases = {}
let g:ale_linters = {'html': ['htmlhint'], 'javascript': ['eslint'], 'css': ['stylelint']}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 1
let g:ale_set_signs = 1
let g:ale_sign_error = '>>'
let g:ale_sign_warning = '--'
  Command History:

(executable check - success) htmlhint
(started) ['/bin/zsh', '-c', 'htmlhint --format=unix ''/var/folders/7m/5rzsq7915692mlnd1p2m4ny40000gn/T/nvimLAOmxp/4/acuity.html''']
(finished - exit code 1) ['/bin/zsh', '-c', 'htmlhint --format=unix ''/var/folders/7m/5rzsq7915692mlnd1p2m4ny40000gn/T/nvimLAOmxp/7/acuity.html''']

<<<OUTPUT STARTS>>>
/var/folders/7m/5rzsq7915692mlnd1p2m4ny40000gn/T/nvimLAOmxp/7/acuity.html:13:45: The attribute name of [ viewBox ] must be in lowercase. [error/attr-lowercase]

^[[31m1 problems^[[39m
<<<OUTPUT ENDS>>>

iamdhj:

### DIFF
--- a/ale_linters/html/htmlhint.vim
+++ b/ale_linters/html/htmlhint.vim
@@ -14,7 +14,7 @@ endfunction
 function! ale_linters#html#htmlhint#GetCommand(buffer) abort
     return ale_linters#html#htmlhint#GetExecutable(a:buffer)
     \   . ' ' . ale#Var(a:buffer, 'html_htmlhint_options')
-    \   . ' %t'
+    \   . ' %s'
 endfunction
 
 call ale#linter#Define('html', {
--- a/test/test_htmlhint_htmlhintrc_detection.vader
+++ b/test/test_htmlhint_htmlhintrc_detection.vader
@@ -1,0 +1,24 @@
+Before:
+  call ale#test#SetDirectory('/testplugin/test')
+
+  runtime ale_linters/html/htmlhint.vim
+
+After:
+  call ale#test#RestoreDirectory()
+  call ale#linter#Reset()
+
+Execute(support .htmlhintrc should be use current filename):
+  call ale#test#SetFilename('/project/test.html')
+
+  let b:buffer = bufnr('')
+
+  let b:temporary_file = ''
+  let b:command = ale_linters#html#htmlhint#GetCommand(b:buffer)
+  let [b:temporary_file, b:command] = ale#command#FormatCommand(b:buffer, b:command, 0)
+
+  let b:filename = fnamemodify(bufname(b:buffer), ':p')
+
+  AssertNotEqual
+  \ stridx(b:command, shellescape(b:filename)),
+  \ -1,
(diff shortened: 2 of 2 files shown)


lovesora:
it works! thank you @iamdhj

w0rp:
I'm checking this out now. What does this fix? Is the issue that the configuration file can't be found because of the temporary file name? Either stdin or a temporary file need to be used so problems can be found while you type. We should probably replace `%t` with `stdin` as mentioned in the [README](https://github.com/yaniswang/HTMLHint). Then we could search for the configuration file and pass it explicitly if `--config` isn't already in the options. See the JSHint linter for an example.

iamdhj:
The HTMLHint tool will search configuration file <code>.htmlhintrc</code> in current directory and all parent directories, temporary directory is not in search range, so it should replace <code>%t</code> to <code>%s</code>.

w0rp:
I fix this now by doing what I mentioned before.

lovesora:
@iamdhj as the author said, change `%t` to `%s` isn't a perfect solution. It will not check for problems when we type. 65fc5d11c77527bbf19a94da919fb37b5a02924b has solved the problem.
Thank you very much! @w0rp

PR #1284, opened 2018-01-14
Support for fixjson as JSON code fixer/formatter

rhysd:
I added JSON fixer support of [fixjson](https://github.com/rhysd/fixjson); a JSON file fixer/formatter for humans using (relaxed) JSON5.

Here is one of fix examples:

modify keys


w0rp:
I have merged this via the commandline now, after fixing the conflicts.

Cheers! :beers:

w0rp:
A next step for this would be to support local `node_modules` installs for `fixjson` too, just like most of the other Node tools.

rhysd:
Thank you! ok, I will add support for local installation. Although this tool does not aim to be used with local installation, it should work.

Issue #1288, opened 2018-01-15
LSP-based linters should send 'textDocument/didSave'


rlerm:
Currently, when the file content changes or `:ALELint` is called, ale sends the `'textDocument/didChange'` message to the server. There is a `ale#lsp#message#DidSave` function in `autoload/ale/lsp/message.vim` to generate the message, but it's never called.

Some LSP servers might also need to be notified when the file is saved `'textDocument/didSave'`, e.g. if they want to kick off more expensive checks.

This might be hard to do since by the time `s:CheckWithLSP` is called there not enough information to determine if we're responding to a save or edit event (unless I'm missing something).

rlerm:
Forget my last comment, it appears that `ale_save_event_fired` can be used to determine if we're coming from a save event or not.

PR #1295, opened 2018-01-17
Send didSave message to LSP-based linters

rlerm:
#1288: We should also notify the LSP server of saves.

One situation where this might be important is if an LSP would handle saves differently, e.g. starting more expensive checks. Currently, the server is only notified on changes, and all that `:ALELint` does is to send a new `didChange` event.

With this patch, if the file has just been saved we'll send both a `didChange` and a `didSave` event (the `didChange` being at most redundant and probably harmless). 

Tests were heavily based on test/completion/test_lsp_completion_messages.vader, which is the only test I found that mocked communication with the server, although I don't think this test would fit in that file since it's not about completion.


w0rp:
Cheers! :beers:

rlerm:
Sorry about that, thanks for fixing!

PR #1303, opened 2018-01-20
Add `jq` as JSON fixer

rafaelrinaldi:
This PR adds [`jq`](https://stedolan.github.io/jq) as a JSON fixer. It has been working great for me for a while and decided to attempt a PR.

If something is missing, please do let me know so I can make all changes necessary.


w0rp:
It looks like you're merging code in a really weird way. I'll see if I can pull the relevant code out of this.

w0rp:
I merged this via the commandline now. I renamed the options to `ale_json_jq_` to match the other option names, and set the `use_global` option to `0` by default, to match the other options. I might add a test for this later.

Cheers! :beers:

rafaelrinaldi:
For some reason rebasing my commits indeed introduced weird logs. If I had noticed before I would have fixed it myself.

Glad I could contribute anyway. Thanks for ale.

rafaelrinaldi:
@w0rp FYI it seems like ale is trying to resolve the binary as a node module, but this is not a node binary. Is that expected as a default behavior? I had to do the following to get it working:

```fish
let g:ale_jq_executable = '/usr/local/bin/jq'
let g:ale_jq_use_global = 1
```

w0rp:
That's what these lines of code you wrote do.

```
+    return ale#node#FindExecutable(a:buffer, 'jq', [
+    \   'jq',
+    \])
```

I removed that and the `use_global` option now, so that shouldn't be a problem anymore.

PR #1308, opened 2018-01-23
Improving hadolint checker

lorenzo:
- Displaying dockerfile parse errors instead of silently failing
- Adding error detail with link to the docs

### DIFF
--- a/ale_linters/dockerfile/hadolint.vim
+++ b/ale_linters/dockerfile/hadolint.vim
@@ -7,26 +7,46 @@ call ale#Set('dockerfile_hadolint_docker_image', 'lukasmartinelli/hadolint')
 function! ale_linters#dockerfile#hadolint#Handle(buffer, lines) abort
     " Matches patterns line the following:
     "
-    " stdin:19: F: Pipe chain should start with a raw value.
-    let l:pattern = '\v^/dev/stdin:?(\d+)? (\S+) (.+)$'
+    " /dev/stdin:19 DL3001 Pipe chain should start with a raw value.
+    " /dev/stdin:19:3 unexpected thing
+    let l:pattern = '\v^/dev/stdin:(\d+):?(\d+)? ((DL|SC)(\d+) )?(.+)$'
     let l:output = []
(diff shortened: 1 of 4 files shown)


ale_linters/dockerfile/hadolint.vim
```diff
@@ -7,26 +7,46 @@ call ale#Set('dockerfile_hadolint_docker_image', 'lukasmartinelli/hadolint')
...
+
         let l:type = 'W'
-        let l:text = l:match[3]
+        let l:text = l:match[6]
+        let l:detail = l:match[6]
+        let l:domain = 'https://github.com/hadolint/hadolint/wiki/'
+
+        if l:match[4] is# 'SC'
+            let l:domain = 'https://github.com/koalaman/shellcheck/wiki/'
+        endif
```
w0rp: Could you add an option for this, and leave it off by default? This will make the messages too long for some users.

lorenzo:
No idea how to write tests for vim, want to guide me through it?

ale_linters/dockerfile/hadolint.vim
lorenzo: This is only shown in the error detail, I think that full error messages are more often than not longer than this, aren't they?

lorenzo:
In particular, I would like to know how to run tests only for the things I change.

ale_linters/dockerfile/hadolint.vim
w0rp: Ah, I see. That's okay then.

w0rp:
You can run the tests with the `./run-tests` script. You can pass a path to one of the Vader test files in the test directory to run that test, or you can use the `:Vader` command with the test file open to run a test directly if you install [Vader](https://github.com/junegunn/vader.vim).

lorenzo:
@w0rp thanks for the hints, I added some tests as you requested.

I also updated some urls that were pointing to now deprecated resources.

w0rp:
Cheers! :beers:

PR #1313, opened 2018-01-24
fixjson: Add support for local installation

rhysd:
As of https://github.com/w0rp/ale/pull/1284#issuecomment-360090290, I added support the case where `fixjson` is installed locally via `npm`.


w0rp:
Cheers! :beers:

PR #1321, opened 2018-01-27
add po support with proselint, writegood, msgfmt and alex

butlerx:
add support for [PO](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html) files with multiple linters


w0rp:
Cheers! :beers:

PR #1327, opened 2018-01-31
Use filename in unix matcher

JelteF:
I ran into the issue that the `gotype` matcher matched all lines to the current file, even though some were for other files.


GinjaNinja32:
#1433

w0rp:
The issue should be fixed for gotype now.

PR #1329, opened 2018-01-31
Add prettier fixer support to vue

blyoa:
I added `vue` filetype to the `suggested_filetypes` of prettier to run the fixer on Vue single file components.


w0rp:
Cheers! :beers: 

You can run any fixer on any filetype, but the suggestions are helpful.

PR #1346, opened 2018-02-13
Respect buffer-level ale_lint_on_* variables

KeithScheiwiller:
This makes ale check for buffer-level (i.e. `b:..`) versions of the `ale_lint_on_*` variables before linting.

I implemented this because I wanted to be able to e.g. not lint on text change for very large files using `ale_pattern_options`, but still lint on save, etc. However I noticed it was written to only check `g:ale_lint_on_*` and not `b:ale_lint_on_*`.

### DIFF
--- a/autoload/ale/events.vim
+++ b/autoload/ale/events.vim
@@ -11,8 +11,27 @@ function! ale#events#QuitRecently(buffer) abort
     return l:time && ale#util#ClockMilliseconds() - l:time < 1000
 endfunction
 
+function! ale#events#TextChangedEvent(buffer) abort
+    " This value used to be a Boolean as a Number, and is now a String.
+    let l:text_changed = '' . ale#Var(a:buffer, 'lint_on_text_changed')
+
+    if l:text_changed is? 'always' || l:text_changed is# '1'
+        call ale#Queue(g:ale_lint_delay)
+    elseif l:text_changed is? 'normal'
+        call ale#Queue(g:ale_lint_delay)
+    elseif l:text_changed is? 'insert'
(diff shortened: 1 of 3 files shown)


KeithScheiwiller:
Poor choice of words, sorry. Just meant it doesn't look outside of the global scope.
Re-adding the `g:ale_lint_on_*` conditionals before binding the `autocmd`s should be simple enough, and I'd be glad to do that.
I don't think I'd have the time to implement the 'if-enabled-locally' option personally though, especially as that would require handling the change from integer options to string. My knowledge of vimscript is also limited at best, but I figured the overhead from running an `autocmd` that immediately exits (if disabled locally) would be minimal, especially compared to trying to run pylint on a 17k+ line Python file on each keystroke.

So I guess the question is, do you think this functionality is as beneficial as I do? Or is it not worth merging in?

w0rp:
The important thing is to not enable `autocmd` events for people who currently have them turned off, without any breaking changes. The only way forward is to make `1` mean that an option is enabled and the autocmd event is set, and a new value can be used to disable the option globally, but turn the autocmd commands on.

w0rp:
I recommend that someone with more VimL knowledge pick this up and support this without the breaking changes to the options. I'd change the options from numbers to strings and use `'never'` and `'always'` instead of `0` and `1` to be consistent with `ale_lint_on_text_changed`, but treat `0` as `'never'` and `1` as `'always'` like that option. Then make `'if-enabled-locally'` enable the autocmd events, but don't run linters unless the buffer-local options are enabled.

w0rp:
Closing due to inactivity.

PR #1351, opened 2018-02-15
This fixes issue #936 by linting the whole package

svanharmelen:
You need to set `let g:ale_go_staticcheck_lint_package = 1` in order to enable the new logic.

PR #1356, opened 2018-02-18
fix issue with go1.10 govet

eikenb:
In go1.10 they have extended 'go vet' with access to more information.
As a side effect it seems you can no longer run 'go vet' against a
single file. You now need to specify all the files or the package.

Similar to w0rp/ale#936, without this it fails to find anything
(types, functions, etc) the other files in the package and reports a
bunch of spurious errors.

This change runs 'go vet' over the package containing the file loaded in
the current buffer.

### DIFF
--- a/ale_linters/go/govet.vim
+++ b/ale_linters/go/govet.vim
@@ -1,10 +1,42 @@
 " Author: neersighted <bjorn@neersighted.com>
 " Description: go vet for Go files
+"
+" Author: John Eikenberry <jae@zhar.net>
+" Description: updated to work with go1.10
+
+" set 'b:ale_go_govet_lint_package = 1' to enable
+call ale#Set('go_govet_lint_package', 0)
+
+function! ale_linters#go#govet#GetCommand(buffer) abort
+    let l:lint_package = ale#Var(a:buffer, 'go_govet_lint_package')
+    if l:lint_package
+        return ale#path#BufferCdString(a:buffer) . ' go vet .'
+    endif
+    return 'go tool vet %t'
+endfunction
+
+function! ale_linters#go#govet#Handler(buffer, lines) abort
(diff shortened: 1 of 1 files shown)


eikenb:
I ran `vint -s` over it and it was clean. I added my name at the top as recommended.

There was no existing test for go-vet and I am unfamiliar with vader tests. If tests are required
that will take some time and pointers to documentation.

Thanks.

eikenb:
> Does this take a while to execute?

I'm not exactly sure what a good benchmark is for this, but I tried it on a couple of the larger packages I had around and it takes longer the first run, then is fast. The longest I could found was on the aws-sdk's request package which took 1.2 seconds the first run, then 200 milliseconds after.

> You might want to consider solving this like the way staticcheck...

You mean with the (off by default) flag for the package scanning with the fallback of scanning the individual file? The problem with that is the results are kind of useless if you just scan the individual file as you will get errors for any references to other things in the same package. So unless you worked on standalone, single file packages a lot it just seems like if it were slow you'd disable it.

eikenb:
Reading through issue #1358, maybe the fallback would be to use 'go tool vet' as the fallback if the whole-package-scanning variable is not set. I'll look into making that change and see how it works out.

eikenb:
Just pushed a new version that uses a variable like #1351, falling back to 'go tool vet' when not set.
It is not set by default.

apriendeau:
#1369 is a side effect of this, I believe.

eikenb:
I'd be happy to change it to have go_govet_lint_package be true by default if that is preferred. Just let me know.

w0rp:
Hmm, let's just do it all of the time. That `lint_file` option has to be `1` all the time or `0` all of the time. Let's just make this check the files on save

w0rp:
I merged this via the commandline now, except with just linting the package all of the time, and with tests.

eikenb:
@w0rp Thanks for writing the tests.

Issue #1306, opened 2018-01-22
Auto-close the error list


chauncey-garrett:
Is there anything similar to https://github.com/vim-syntastic/syntastic/pull/126?

That is, a way to automatically close the error window when closing buffers.

What I currently experience is that I can open a file that contains lint errors, the error window appears, then I close that file but am left with the error window of the now closed file.

w0rp:
Feel free to create a PR for this if you can figure it out.

Christian-Gibbons:
It looks like this can be accomplished, assuming you are using the loclist rather than quickfix, with an autocommand like this:

`autocmd QuitPre * if empty(&bt) | lclose | endif`

I'll put together a pull request that piggybacks on the QuitPre autocmd that ale already has.

w0rp:
I updated the documentation to tell people how to write an `autocmd` command to do this, so you can copy and paste that into vimrc or wherever to do this.

PR #1367, opened 2018-02-22
Close loclist when associated buffer is closed

Christian-Gibbons:
This should automatically close the loclist when you quit the associated buffer.  Related to #1306

### DIFF
--- a/autoload/ale/events.vim
+++ b/autoload/ale/events.vim
@@ -3,7 +3,7 @@
 function! ale#events#QuitEvent(buffer) abort
     " Remember when ALE is quitting for BufWrite, etc.
     call setbufvar(a:buffer, 'ale_quitting', ale#util#ClockMilliseconds())
-    if empty(&bt) | lclose | endif
+    if empty(&buftype) | lclose | endif
 endfunction
 
 function! ale#events#QuitRecently(buffer) abort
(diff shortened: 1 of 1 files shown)


w0rp:
We'd need to use an option for turning this on, with it off by default, as some users won't want this. You can set this up yourself just as easily with your own `autocmd`, so I'll recommend setting that in the documentation instead. Then you won't pay for what you don't use.

wxianxin:
Why would a user doesn't want to close the loclist window when he/she exit vim or closes the buffer?
I would say **less than 1%** of people would need this bizarre behavior at some point of their life.


I agree with your "Don't pay for what you don't use". But I don't think this case  it.

People use vim plugins to ease my life, why would people want to write command for a plugin?
ale has many functions, 20% of the functions will not be used by 50% of users. These functions are still here.
loclist will be used by >20% people, the majority of them will prefer it on rather than not. It is not here.
I know I am guessing around. It a reasonable guess without personal preference.

w0rp:
The loclist is a Vim feature, and it's just something that ALE uses. If you want to close it when your buffer closes, write `if empty(&buftype) | lclose | endif` into an autocmd function. It requires only one line in vimrc for you to use it, and not everyone wants this. This ain't being merged.

Issue #1226, opened 2017-12-15
Including Unity3d Engine Dll's for ALE linting.


pieromarini:
**VIM version**
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Nov 14 2017 20:05:55)
Included patches: 1-1297

Operating System: ArchLinux

### :ALEInfo
Using mcs, mcsc linters
Set g:ale_cs_mcsc_assembly_path = ['/opt/UnityBeta/Editor/']  # This folders/subfolders contain the dll's needed. 

Basically a lot of lines with:

> ./FileName.cs: error The type or namespace name 'UnityEngine' could not be found. Are you missing 
   an assembly reference

This repeats for various types/namespaces related to Unity.

Hope that's clear enough...
Am I doing something wrong? 

Thanks for the help

w0rp:
I'm not sure. Maybe the authors of the linters might know.

pieromarini:
What I noticed is that when using the "ale_cs_mcsc_assembly_path", the linter searches for "*.cs" files, not "*.dlls" as specified in the help docs. When using "ale_cs_mcsc_assemblies_src" pointed towards the specific dll files it works as expected.

w0rp:
I have updated the documentation to explain this better now, with a reference to Unity specifically.

PR #1371, opened 2018-02-23
Add Unity project path to Unity example in docs

idbrii:
For Ale to recognize the symbols in your project, you need to point it
at where Unity builds your assemblies. Add this path to the example.

On Windows, the dll path is this (by default):

    " or on Windows:
    let g:ale_cs_mcsc_assemblies = [
    \ expand('$ProgramFiles/Unity/Editor/Data/Managed/UnityEngine.dll'), 
    \ 'path-to-unityproject/obj/Debug',
    \]

I didn't add since it doesn't contribute much (just need to figure out the right prefix).

(I found #1226 before I found the help docs so I'll mention it so any others can find their way here.)


w0rp:

PR #1375, opened 2018-02-25
doc/c: fix duplicate ale_c_flawfinder_executable help tag

zchee:
Fix duplicate `ale_c_flawfinder_executable` help tag which both of `g:` prefix.


zchee:
@w0rp thanks quickly

PR #1381, opened 2018-02-26
Option to open lists vertically

### LINKED ISSUE #1352: Allow option to open vertical list
Currently if you have list option set the list will always open horizontally. I would like to be able to open it vertically.

I imagine something like:

```vim
" list.vim

if g:ale_set_quickfix
    if !ale#list#IsQuickfixOpen()
        if g:ale_list_vertical
            silent! execute 'copen ' . str2nr(ale#Var(a:buffer, 'list_window_size'))
        else
            silent! execute 'vert copen ' . str2nr(ale#Var(a:buffer, 'list_window_size'))
        endif
    elseif g:ale_set_loclist
        if g:ale_list_vertical
            silent! execute 'lopen ' . str2nr(ale#Var(a:buffer, 'list_window_size'))
        else
            silent! execute 'vert lopen ' . str2nr(ale#Var(a:buffer, 'list_window_size'))
        endif
    endif
endif
```

ajcrites:
Introduces a new option, `ale_list_vertical` that can be set globally or for a buffer to open the quickfix window vertically instead of the default horizontally.

Closes #1352

### DIFF
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -1222,6 +1222,8 @@ g:ale_open_list                                               *g:ale_open_list*
 
   The window size can be configured with |g:ale_list_window_size|.
 
+  Windows can be opened vertically with |g:ale_list_vertical|.
+
   If you want to close the loclist window automatically when the buffer is
   closed, you can set up the following |autocmd| command: >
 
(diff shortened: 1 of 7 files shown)


doc/ale.txt
```diff
@@ -1231,6 +1233,15 @@ g:ale_open_list                                               *g:ale_open_list*
   augroup END
 <
 
+g:ale_list_vertical                                       *g:ale_list_vertical*
```
w0rp: The options should be sorted by name, so move this further up.

test/test_list_opening.vader
```diff
@@ -33,16 +35,29 @@ Before:
     return 0
   endfunction
 
+  " If the window is vertical, window size should match column size/width
+  function GetQuickfixIsVertical(cols) abort
+    for l:win in range(1, winnr('$'))
+        if getwinvar(l:win, '&buftype') ==# 'quickfix'
```
w0rp: `is#` is preferred in this codebase when checking strings because `==` behaves weirdly with strings and numbers.

w0rp:
Cheers! :beers: I'll see if that Windows failure happens on master and fix it, if it does.

PR #1383, opened 2018-02-27
add textlint support

tokida:
I added support for [textlint](https://textlint.github.io/), a proofreading tool for writers and programmers. It supports document formats markdown. I usually use textlint for markdown so I added it as a markdown linter at first.

### DIFF
--- a/ale_linters/markdown/textlint.vim
+++ b/ale_linters/markdown/textlint.vim
@@ -1,0 +1,9 @@
+" Author: tokida https://rouger.info
+" Description: textlint, a proofreading tool (https://textlint.github.io/)
+
+call ale#linter#Define('markdown', {
+\   'name': 'textlint',
+\   'executable': 'textlint',
+\   'command': 'if [ -f .textlintrc ]; then textlint -f json %t ; else echo "[{\"filePath\":\"%t\",\"messages\":[]}]"; fi',
+\   'callback': 'ale#handlers#textlint#HandleTextlintOutput',
+\})
--- a/ale_linters/markdown/textlint.vim
+++ b/ale_linters/markdown/textlint.vim
@@ -1,9 +1,23 @@
 " Author: tokida https://rouger.info
 " Description: textlint, a proofreading tool (https://textlint.github.io/)
 
+function! ale_linters#markdown#textlint#GetCommand(buffer) abort
+	let l:cmd_path = ale#path#FindNearestFile(a:buffer, '.textlintrc')
+	if !empty(l:cmd_path)
+		return 'textlint'
+		\	. ' -c '
+		\	. l:cmd_path
+		\ 	. ' -f json %t'
+	else
+		return "[{\'filePath\':\'%t\',\'messages\':[]}]"
+	endif
+endfunction
+
+
 call ale#linter#Define('markdown', {
 \   'name': 'textlint',
 \   'executable': 'textlint',
-\   'command': 'if [ -f .textlintrc ]; then textlint -f json %t ; else echo "[{\"filePath\":\"%t\",\"messages\":[]}]"; fi',
+\   'command_callback': 'ale_linters#markdown#textlint#GetCommand',
 \   'callback': 'ale#handlers#textlint#HandleTextlintOutput',
 \})
+
--- a/ale_linters/markdown/textlint.vim
+++ b/ale_linters/markdown/textlint.vim
@@ -2,15 +2,15 @@
 " Description: textlint, a proofreading tool (https://textlint.github.io/)
 
 function! ale_linters#markdown#textlint#GetCommand(buffer) abort
-	let l:cmd_path = ale#path#FindNearestFile(a:buffer, '.textlintrc')
-	if !empty(l:cmd_path)
-		return 'textlint'
-		\	. ' -c '
-		\	. l:cmd_path
-		\ 	. ' -f json %t'
-	else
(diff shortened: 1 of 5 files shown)


ale_linters/markdown/textlint.vim
```diff
@@ -0,0 +1,9 @@
+" Author: tokida https://rouger.info
+" Description: textlint, a proofreading tool (https://textlint.github.io/)
+
+call ale#linter#Define('markdown', {
+\   'name': 'textlint',
+\   'executable': 'textlint',
+\   'command': 'if [ -f .textlintrc ]; then textlint -f json %t ; else echo "[{\"filePath\":\"%t\",\"messages\":[]}]"; fi',
```
w0rp: There are two problems here.

1. This code will only work on some Unix variants.
2. You're only looking for a file in what _might_ be the current working directory, or which could be somewhere else entirely.

Try using `ale#path#FindNearestFile` in a command callback instead. See the `cppcheck.vim` file for an example, which is the first one I found.

w0rp: You don't need the part where you echo some default JSON. Return `''` from the callback instead to skip running the linter.

autoload/ale/handlers/textlint.vim
```diff
@@ -0,0 +1,18 @@
+" Author: tokida https://rouger.info
+" Description: Redpen, a proofreading tool (http://redpen.cc)
+
+function! ale#handlers#textlint#HandleTextlintOutput(buffer, lines) abort
+    let l:res = json_decode(join(a:lines))[0]
```
w0rp: Move this callback into the `ale_linters` file, like other linters. Use `ale#util#FuzzyJSONDecode(a:lines, [])` to get the list of results and then `get(l:whatever, 0, {'messages': []})` on that. Add a test with some empty output.

autoload/ale/handlers/textlint.vim
```diff
@@ -0,0 +1,18 @@
...
+    let l:output = []
+    for l:err in l:res.messages
+        let l:item = {
+        \   'text': l:err.message,
+        \   'type': 'W',
+        \   'code': l:err.ruleId,
+        \}
+        let l:item.lnum = l:err.line
+        let l:item.col = l:err.column
+        call add(l:output, l:item)
```
w0rp: You can write `call add(l:output, {...})` with everything in the Dictionary here.

autoload/ale/handlers/textlint.vim
```diff
@@ -0,0 +1,18 @@
...
+    endfor
+    return l:output
```
w0rp: Write blank lines before `if`, `return`, `for`, and `while` if they don't start a block.

ale_linters/markdown/textlint.vim
```diff
@@ -0,0 +1,23 @@
...
+
+function! ale_linters#markdown#textlint#GetCommand(buffer) abort
+    let l:cmd_path = ale#path#FindNearestFile(a:buffer, '.textlintrc')
+    if !empty(l:cmd_path)
+        return 'textlint'
+        \    . ' -c '
+        \    . l:cmd_path
+        \     . ' -f json %t'
+    else
+        return "[{\'filePath\':\'%t\',\'messages\':[]}]"
```
w0rp: You need to return a command to run here, not the output of it. You should return `''` instead, which will skip running the command. Write a return statement at the end of the function, instead of putting it in an `else` block.

ale_linters/markdown/textlint.vim
```diff
@@ -0,0 +1,23 @@
+" Author: tokida https://rouger.info
+" Description: textlint, a proofreading tool (https://textlint.github.io/)
+
+function! ale_linters#markdown#textlint#GetCommand(buffer) abort
+    let l:cmd_path = ale#path#FindNearestFile(a:buffer, '.textlintrc')
+    if !empty(l:cmd_path)
```
w0rp: Write a blank line before `if`. Can `textlint` be run without a configuration file? If it can, maybe you can run it without passing the `-c` option, if no configuration file is found.

autoload/ale/handlers/textlint.vim
```diff
@@ -0,0 +1,18 @@
+" Author: tokida https://rouger.info
+" Description: Redpen, a proofreading tool (http://redpen.cc)
+
+function! ale#handlers#textlint#HandleTextlintOutput(buffer, lines) abort
+    let l:res = get(ale#util#FuzzyJSONDecode(a:lines, []), 0, {'messages': []})
+    let l:output = []
+    for l:err in l:res.messages
```
w0rp: Add a blank line before `for`.

ale_linters/markdown/textlint.vim
```diff
@@ -0,0 +1,23 @@
...
+endfunction
+
+
+call ale#linter#Define('markdown', {
+\   'name': 'textlint',
+\   'executable': 'textlint',
+\   'command_callback': 'ale_linters#markdown#textlint#GetCommand',
+\   'callback': 'ale#handlers#textlint#HandleTextlintOutput',
+\})
+
```
w0rp: Remove the extra blank line here.

ale_linters/markdown/textlint.vim
tokida: I tried run with out a configuration file . result, textlint was necessary -c option.

w0rp:
Cheers! :beers:

PR #1387, opened 2018-02-28
Explain in more detail why perl checks are disabled by default

oalders:
This is a purely a doc patch.  I thought it would be helpful to explain in a little more detail the new defaults around Perl linting.


w0rp:
Sure.

w0rp:
Cheers! :beers:

PR #1395, opened 2018-03-04
Fix elm-format fixer name

gfontenot:
Support for elm-format as a fixer has existed since Sept 2017, but it's not
easy to discover because the fixer was named `format`. This breaks the
convention of the other fixers that use the full name in the registry.

I've gone ahead and fixed this discrepancy, but I left the existing registry
entry in place. We should move people towards using `elm-format` as the fixer
name, but I'd hate to break existing setups.

### DIFF
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -23,6 +23,12 @@ let s:default_registry = {
 \       'description': 'Apply prettier-standard to a file.',
 \       'aliases': ['prettier-standard'],
 \   },
+\   'elm-format': {
+\       'function': 'ale#fixers#elm_format#Fix',
+\       'suggested_filetypes': ['elm'],
+\       'description': 'Apply elm-format to a file.',
+\       'aliases': ['format'],
+\   },
 \   'eslint': {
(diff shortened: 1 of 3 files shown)


autoload/ale/fix/registry.vim
```diff
@@ -23,6 +23,11 @@ let s:default_registry = {
 \       'description': 'Apply prettier-standard to a file.',
 \       'aliases': ['prettier-standard'],
 \   },
+\   'elm-format': {
+\       'function': 'ale#fixers#elm_format#Fix',
+\       'suggested_filetypes': ['elm'],
+\       'description': 'Apply elm-format to a file.',
```
w0rp: What you should do here is remove the `format` key below, and add `'aliases': ['format'],` here to run the fixer with the old name.

w0rp:
You've got the right idea here. Use a new name, but don't break existing configs.

autoload/ale/fix/registry.vim
gfontenot: Ah, cool! I should have guessed that there was an existing mechanism for defining aliases. Fixed this now.

w0rp:
Cheers! :beers:

Issue #1385, opened 2018-02-27
Flawfinder: Handle severity level


Christian-Gibbons:
Flawfinder piggybacks off of the GCC handler which expects problems to be either an Error or a Warning.  Flawfinder has a severity range of 0-5 inclusive.  Instead of displaying just 'Warning' for anything that doesn't match 'error', consider displaying severity level.  Alternatively, or additionally, allow users to set a minimum severity level to be considered error level while lesser severity issues be considered warnings.

yamsu:
I have just installed flawfinder and it is working but I get the following error on save and load of a simple cpp file.

> E716: Key not present in Dictionary: ale_c_flawfinder_error_severity
> E116: Invalid arguments for function get(l:vars, l:full_name, g:[l:full_name])
> E15: Invalid expression: get(l:vars, l:full_name, g:[l:full_name])
> Error detected while processing function <SNR>162_VimCloseCallback:

Christian-Gibbons:
I have a fix for this locally that I had apparently failed to push upstream.  I'll put in a pull request and get it fixed.

PR #1361, opened 2018-02-21
New linter: Flawfinder

Christian-Gibbons:
This pull request adds support for the Flawfinder linter for C and C++.   Resolves issue: #1161

PR #1400, opened 2018-03-05
Handle flawfinder severity level

Christian-Gibbons:
This should allow a user to specify a severity level to be considered an error while lesser issues would be treated as warnings.  It also adds a little more information to issue.  Intended to solve issue #1385 

Not sure if I handled the Vader test properly for this one.  Most of the changes dealt with the GCC handler rather than the flawfinder linter itself.

### DIFF
--- a/ale_linters/c/flawfinder.vim
+++ b/ale_linters/c/flawfinder.vim
@@ -4,6 +4,7 @@
 call ale#Set('c_flawfinder_executable', 'flawfinder')
 call ale#Set('c_flawfinder_options', '')
 call ale#Set('c_flawfinder_minlevel', 1)
+call ale#Set('flawfinder_error_severity', 6)
 
 function! ale_linters#c#flawfinder#GetExecutable(buffer) abort
    return ale#Var(a:buffer, 'c_flawfinder_executable')
--- a/ale_linters/c/flawfinder.vim
+++ b/ale_linters/c/flawfinder.vim
@@ -4,7 +4,7 @@
 call ale#Set('c_flawfinder_executable', 'flawfinder')
 call ale#Set('c_flawfinder_options', '')
 call ale#Set('c_flawfinder_minlevel', 1)
-call ale#Set('flawfinder_error_severity', 6)
+call ale#Set('c_flawfinder_error_severity', 6)
 
 function! ale_linters#c#flawfinder#GetExecutable(buffer) abort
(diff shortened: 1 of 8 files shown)


ale_linters/c/flawfinder.vim
```diff
@@ -4,6 +4,7 @@
 call ale#Set('c_flawfinder_executable', 'flawfinder')
 call ale#Set('c_flawfinder_options', '')
 call ale#Set('c_flawfinder_minlevel', 1)
+call ale#Set('flawfinder_error_severity', 6)
```
w0rp: This should be named `c_flawfinder_error_severity`.

Christian-Gibbons: I was unsure about this one.  This variable affects both C and C++.

autoload/ale/handlers/gcc.vim
```diff
@@ -24,7 +24,7 @@ function! ale#handlers#gcc#HandleGCCFormat(buffer, lines) abort
     " <stdin>:8:5: warning: conversion lacks type at end of format [-Wformat=]
     " <stdin>:10:27: error: invalid operands to binary - (have ‘int’ and ‘char *’)
     " -:189:7: note: $/${} is unnecessary on arithmetic variables. [SC2004]
-    let l:pattern = '\v^([a-zA-Z]?:?[^:]+):(\d+):(\d+)?:? ([^:]+): ?(.+)$'
+    let l:pattern = '\v^([a-zA-Z]?:?[^:]+):(\d+):(\d+)?:? ([^:]+): (.+)$'
```
w0rp: Don't forget to undo the changes to this file too.

Christian-Gibbons: That change was intentional.  I added the question mark to that regular expression when I added the initial Flawfinder support that piggybacked off of the GCC handler in #1361 .   Now that Flawfinder has its own handler, I figured I would revert that back to what it had been before I commandeered the GCC handler for Flawfinder.  I can add it back in if you like, but there shouldn't be anything relying on that question mark anymore.

w0rp: Okay, what about the other changes below?

Christian-Gibbons: Hmm, not sure how I missed all that.  I'll try to get a fix in later today.

PR #1410, opened 2018-03-08
Allow passing options to mix_format

fribmendes:
Allow setting `g:ale_elixir_mix_format_options` and `b:ale_elixir_mix_format_options` to pass options down to the `mix format` invocation.

This is particularly useful for umbrella apps, since you can specify the root project `.formatter.exs` by doing something like:

```vim
if filereadable("../../.formatter.exs")
  let g:ale_fixers['elixir'] = ['mix_format']
  let g:ale_elixir_mix_format_options = "--dot-formatter ../../.formatter.exs"
endif
```


w0rp:
Cheers! :beers:

PR #1416, opened 2018-03-11
handle multibyte string when linting text with redpen

glabra:
This PR fixes highlight misalignment that occurs when highlighting multibyte string.

When highlighting multibyte string, current implementation highlights slightly misaligned place.
For example, when highlighting the word "そう言って", only "そう" is highlighted.

This PR is workaround to the problem.


w0rp:
Cheers! :beers:

Issue #1362, opened 2018-02-21
Feature request: qmllint integration


pylipp:
QML is a language for front-end design using the QtQuick framework. With the Qt installation comes the command line utility `qmllint` that warns you about errors, like in the following:
```
// example.qml
Rectangle {
    width: 100

    Button {
    }
```
it reports `example.qml:6 : Expected token \`}'`.

It would be nice to have such warnings displayed by ale. I'm happy to implement a solution myself. However some people might have useful pointers. What do you think?

fenuks:
This issue can be closed now, I suppose. :-)

PR #1419, opened 2018-03-13
Add basic qmllint support

pylipp:
Addresses #1362

I could not find a matching existing handler.

Can I add a test?

### DIFF
--- a/ale_linters/qml/qmllint.vim
+++ b/ale_linters/qml/qmllint.vim
@@ -1,0 +1,30 @@
+" Author: pylipp
+" Description: qmllint for QML files
+
+" Find lines like
+" /home/foo_user42/code-base/qml/Screen.qml:11 : Expected token `}'
+function! ale_linters#qml#qmllint#Handle(buffer, lines) abort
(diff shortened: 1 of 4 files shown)


ale_linters/qml/qmllint.vim
```diff
@@ -0,0 +1,30 @@
...
+    endfor
+
+    return l:output
+endfunction
+
+call ale#linter#Define('qml', {
+\   'name': 'qmllint',
+\   'output_stream': 'stderr',
+\   'executable': 'qmllint',
+\   'command': 'qmllint %s',
```
w0rp: `%s` will be replaced with the filename. Could you use `%t` here to use a temporary file instead? You can't check the file itself while you type. If checking a temporary file doesn't work, you should add `'lint_file': 1`, on the end, so ALE will only check for problems when you open a file or save it, etc. Then you'd need to edit the documentation to contain a `!!` marker next to the linter, so people know it can't be used for linting while you type.

w0rp:
Cheers! :beers:

Issue #1167, opened 2017-11-26
Automatically determine build flags by parsing `make -n` output


w0rp:
Setting build flags for C and C++ projects is a complicated process. One way to automatically determine build flags which would work for many projects would be to parse `make -n` output. The process would work like so.

1. Look for a `Makefile` in a parent directory if an option for automatically using `make -n` parsing is on.
2. If the `Makefile` exists, then run `make -n` in that directory in the background to get the dry run output.
3. Search for a line for the file being edited, and take the relevant flags from the output, minus flags for producing output, etc.

We could fall back on other options if parsing fails. We could also use the same flags headers which lie next to source files.

korken89:
Hi, repurposing https://github.com/rdnetto/YCM-Generator could be an option, or just support the YCM config files (.ycm_extra_conf.py).
It is used with good results for generating the YCM config.

roel0:
My unit tests fail because the dockerfile doesn't have the make executable.
@w0rp Why is the dockerfile not part of the repository?
This is blocking my pull request.

w0rp:
https://github.com/w0rp/ale/blob/master/Dockerfile You mean this file?

w0rp:
See `:help ale-command-chain` for the documentation for that.

roel0:
Haha oeps, completely looked over that file. 

Oke thanks, will finish this soon.

w0rp:
Thanks to @roel0, support for this is in now for some linters. I'll document the option, try to support this with more linters, and leave the issue open until we've fixed a few bugs with it which will surely come up.

w0rp:
Support for this has existed for a while now.

PR #1434, opened 2018-03-20
Automatically determine build flags by parsing `make -n` output #1167

roel0:
- Automatically determine build flags by parsing `make -n` output for C parsers
- Reads makefile in project root 
- Needs to be enabed with  g:ale_c_parse_makefile
- Unit tests added

### DIFF
--- a/ale_linters/c/clang.vim
+++ b/ale_linters/c/clang.vim
@@ -9,15 +9,7 @@ function! ale_linters#c#clang#GetExecutable(buffer) abort
 endfunction
 
 function! ale_linters#c#clang#GetCommand(buffer, output) abort
-    let l:cflags = []
-    if !empty(a:output)
-        let l:cflags = join(ale#c#ParseMakefile(a:buffer, join(a:output, '\n')), ' ')
-    endif
-    if empty(l:cflags)
-        let l:cflags = ale#c#IncludeOptions(ale#c#FindLocalHeaderPaths(a:buffer))
-    else
-        let l:cflags .= ' '
-    endif
+    let l:cflags = ale#c#GetCFlags(a:buffer, a:output)
 
     " -iquote with the directory the file is in makes #include work for
     "  headers in the same directory.
@@ -33,7 +25,7 @@ call ale#linter#Define('c', {
 \   'output_stream': 'stderr',
 \   'executable_callback': 'ale_linters#c#clang#GetExecutable',
 \   'command_chain': [
-\       {'callback': 'ale#c#ParseMakefile', 'output_stream': 'stdout'},
+\       {'callback': 'ale#c#GetMakeCommand', 'output_stream': 'stdout'},
 \       {'callback': 'ale_linters#c#clang#GetCommand'}
 \   ],
 \   'callback': 'ale#handlers#gcc#HandleGCCFormat',
--- a/ale_linters/c/clang.vim
+++ b/ale_linters/c/clang.vim
@@ -3,7 +3,7 @@
 
 call ale#Set('c_clang_executable', 'clang')
 call ale#Set('c_clang_options', '-std=c11 -Wall')
-call ale#Set('c_gcc_parse_makefile', 0)
+call ale#Set('c_clang_parse_makefile', 0)
 
 function! ale_linters#c#clang#GetExecutable(buffer) abort
     return ale#Var(a:buffer, 'c_clang_executable')
@@ -11,11 +11,13 @@ endfunction
 
 function! ale_linters#c#clang#GetCommand(buffer) abort
 let l:cflags = []
-    if g:ale_c_parse_makefile
-        let l:cflags = join(ale#c#ParseMakefile(a:buffer), ' ')
+    if g:ale_c_clang_parse_makefile
+        let l:cflags = join(ale#c#ParseMakefile(a:buffer), ' ') . ' '
     endif
     if empty(l:cflags)
         let l:cflags = ale#c#IncludeOptions(ale#c#FindLocalHeaderPaths(a:buffer))
+    else
+        let l:cflags .= ' '
     endif
 
     " -iquote with the directory the file is in makes #include work for
@@ -23,7 +25,7 @@ let l:cflags = []
     return ale#Escape(ale_linters#c#clang#GetExecutable(a:buffer))
     \   . ' -S -x c -fsyntax-only '
     \   . '-iquote ' . ale#Escape(fnamemodify(bufname(a:buffer), ':p:h')) . ' '
-    \   . l:cflags . ' '
+    \   . l:cflags
     \   . ale#Var(a:buffer, 'c_clang_options') . ' -'
 endfunction
 
--- a/ale_linters/c/clang.vim
+++ b/ale_linters/c/clang.vim
@@ -3,20 +3,27 @@
 
 call ale#Set('c_clang_executable', 'clang')
 call ale#Set('c_clang_options', '-std=c11 -Wall')
+call ale#Set('c_gcc_parse_makefile', 0)
 
 function! ale_linters#c#clang#GetExecutable(buffer) abort
     return ale#Var(a:buffer, 'c_clang_executable')
 endfunction
 
 function! ale_linters#c#clang#GetCommand(buffer) abort
-    let l:paths = ale#c#FindLocalHeaderPaths(a:buffer)
+let l:cflags = []
+    if g:ale_c_parse_makefile
+        let l:cflags = join(ale#c#ParseMakefile(a:buffer), ' ')
+    endif
+    if empty(l:cflags)
+        let l:cflags = ale#c#IncludeOptions(ale#c#FindLocalHeaderPaths(a:buffer))
+    endif
 
     " -iquote with the directory the file is in makes #include work for
     "  headers in the same directory.
     return ale#Escape(ale_linters#c#clang#GetExecutable(a:buffer))
     \   . ' -S -x c -fsyntax-only '
     \   . '-iquote ' . ale#Escape(fnamemodify(bufname(a:buffer), ':p:h')) . ' '
-    \   . ale#c#IncludeOptions(l:paths)
+    \   . l:cflags . ' '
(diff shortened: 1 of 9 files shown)


roel0:
New unit tests are not windows proof, will fix tomorrow

ale_linters/c/clang.vim
```diff
@@ -8,22 +8,33 @@ function! ale_linters#c#clang#GetExecutable(buffer) abort
     return ale#Var(a:buffer, 'c_clang_executable')
 endfunction
 
-function! ale_linters#c#clang#GetCommand(buffer) abort
-    let l:paths = ale#c#FindLocalHeaderPaths(a:buffer)
+function! ale_linters#c#clang#GetCommand(buffer, output) abort
+    let l:cflags = []
+    if !empty(a:output)
+        let l:cflags = join(ale#c#ParseMakefile(a:buffer, join(a:output, '\n')), ' ')
```
w0rp: I think you meant to use `ParseCFlags` here.

These lines for parsing the flags are all the same, so you could write one function in the `autoload/ale/c.vim` which does both and just write `let l:cflags = ale#c#GetFlags(a:buffer, a:output)` here.

ale_linters/c/clang.vim
```diff
@@ -8,22 +8,33 @@ function! ale_linters#c#clang#GetExecutable(buffer) abort
...
+    endif
+    if empty(l:cflags)
+        let l:cflags = ale#c#IncludeOptions(ale#c#FindLocalHeaderPaths(a:buffer))
+    else
+        let l:cflags .= ' '
```
w0rp: You can't use `.= ' '` with a List. The variable should be either a String or a List, but not both.

autoload/ale/c.vim
```diff
@@ -22,6 +23,84 @@ function! ale#c#FindProjectRoot(buffer) abort
...
+            endif
+        endfor
+        if !empty(l:cflags)
+            return ale#c#ParseCFlagsToList(a:buffer, l:cflags)
+       endif
+    endif
+    retur []
+endfunction
+
+function! ale#c#ParseMakefile(buffer) abort
```
w0rp: You should name this function `GetMakeCommand` or something like that instead. This function doesn't parse a Makefile.

autoload/ale/c.vim
```diff
@@ -22,6 +23,84 @@ function! ale#c#FindProjectRoot(buffer) abort
...
+    if g:ale_c_parse_makefile
+        let l:project_root = ale#c#FindProjectRoot(a:buffer)
+        let l:project_cflags = []
+
+        if !empty(l:project_root)
+            if !empty(globpath(l:project_root, 'Makefile', 0))
```
w0rp: Instead of using `FindProjectRoot` and then `globpath` to find the file, use `ale#path#FindNearestFile` to find `Makefile` instead. `globpath` can run kind of slowly for large directories, and you should find the nearest `Makefile`, instead of the one at the project root. I believe the Linux kernel and a few other projects used nested Makefiles so you can make subsets of the project faster.

autoload/ale/c.vim
```diff
@@ -22,6 +23,84 @@ function! ale#c#FindProjectRoot(buffer) abort
...
+                return 'cd '. l:project_root . ' && make -n'
+            endif
+        endif
+    endif
+    return ''
```
w0rp: Leave a blank line before `if`, `while`, `for`, `return`, or `try` when those lines don't start a new block.

autoload/ale/c.vim
```diff
@@ -22,6 +23,84 @@ function! ale#c#FindProjectRoot(buffer) abort
...
+endfunction
+
+function! ale#c#ParseMakefile(buffer) abort
+    if g:ale_c_parse_makefile
+        let l:project_root = ale#c#FindProjectRoot(a:buffer)
+        let l:project_cflags = []
+
+        if !empty(l:project_root)
+            if !empty(globpath(l:project_root, 'Makefile', 0))
+                return 'cd '. l:project_root . ' && make -n'
```
w0rp: You can change the `cd` path to the directory of wherever the `Makefile` was found with `fnamemodify`. You should apply `ale#Escape` to the project path, or this won't work with paths that contain spaces, etc.

autoload/ale/c.vim
```diff
@@ -22,6 +23,84 @@ function! ale#c#FindProjectRoot(buffer) abort
...
+            if index(l:cflags_list, l:option) < 0
+                call add(l:cflags_list, l:option)
+            endif
+        endif
+    endfor
+    return l:cflags_list
+endfunction
+
+function! ale#c#ParseCFlags(buffer, stdout_make) abort
+    if g:ale_c_parse_makefile
```
w0rp: I'd prefer an early return instead. Check the inverse of that and return `[]` early on.

autoload/ale/c.vim
```diff
@@ -22,6 +23,84 @@ function! ale#c#FindProjectRoot(buffer) abort
...
+            if stridx(l:cflags, expand('#' . a:buffer . '...'))
+                let l:cflags = split(l:cflags)
+                break
+            endif
+        endfor
+        if !empty(l:cflags)
+            return ale#c#ParseCFlagsToList(a:buffer, l:cflags)
+       endif
+    endif
+    retur []
```
w0rp: You missed an `n` here.

autoload/ale/c.vim
```diff
@@ -22,6 +23,84 @@ function! ale#c#FindProjectRoot(buffer) abort
...
+
+function! ale#c#ParseCFlags(buffer, stdout_make) abort
+    if g:ale_c_parse_makefile
+        for l:cflags in split(a:stdout_make, '\n')
+            if stridx(l:cflags, expand('#' . a:buffer . '...'))
+                let l:cflags = split(l:cflags)
+                break
+            endif
+        endfor
+        if !empty(l:cflags)
```
w0rp: You can probably remove the `empty()` check here. Just pass the empty list to the next function.

autoload/ale/c.vim
```diff
@@ -22,6 +23,84 @@ function! ale#c#FindProjectRoot(buffer) abort
...
+            endif
+        endif
+    endfor
+    return l:cflags_list
+endfunction
+
+function! ale#c#ParseCFlags(buffer, stdout_make) abort
+    if g:ale_c_parse_makefile
+        for l:cflags in split(a:stdout_make, '\n')
+            if stridx(l:cflags, expand('#' . a:buffer . '...'))
```
w0rp: `expand('#' . a:buffer . '...')` will be the same for every iteration, so you can do that once outside of the loop.

autoload/ale/c.vim
```diff
@@ -22,6 +23,84 @@ function! ale#c#FindProjectRoot(buffer) abort
...
+        endfor
+        if !empty(l:cflags)
+            return ale#c#ParseCFlagsToList(a:buffer, l:cflags)
+       endif
+    endif
+    retur []
+endfunction
+
+function! ale#c#ParseMakefile(buffer) abort
+    if g:ale_c_parse_makefile
```
w0rp: I'd prefer an early return here. Invert the condition and return `''`.

test/test_c_projects/makefile_project/Makefile
```diff
@@ -0,0 +1,3 @@
+file.o : subdir/file.c
```
w0rp: You can delete this file now.

w0rp:
One thing I'd like to consider in future is replacing the `clang` and `gcc` linters with a backwards-compatible `cc` linter, and determine which command to run based on what `make -n` returns, or the same for other formats like `compile_commands.json`.

roel0:
Thanks for the review, I (think) have adapted them all

autoload/ale/c.vim
```diff
@@ -22,6 +23,102 @@ function! ale#c#FindProjectRoot(buffer) abort
...
+
+        if l:previous_option isnot? ''
+            let l:option = l:previous_option
+            let l:previous_option = ''
+        endif
+
+        " Fix relative paths if needed
+        if stridx(l:option, '-I') >= 0
+            if stridx(l:option, '-I' . s:sep) < 0
+                let l:option = '-I' . a:path_prefix . s:sep . l:option[2:]
```
w0rp: There is a bug here with escaping the path I think you should fix. I think `l:option[2;]` might already be escaped when returned by `make -n`, but your path prefix here won't be. You should probably write `ale#Escape(a:path_prefix) . ...`, in case the path contains spaces, etc.

w0rp:
Cheers! :beers: This should hopefully be useful for a few people.

ranebrown:
I'm seeing issues where all the flags are not being correctly captured. I took a look `c.vim` and it appears only `-I` and `-D` flags get added to the list of cflags ([see here](https://github.com/w0rp/ale/blob/master/autoload/ale/c.vim#L78-L83)) Is this desired behavior? As far as I can tell it looks like the same thing happens using `compile_commands.json`

It seems like things like `-std=gnu99` and other flags should be pulled from the Makefile or compilation database as well.

w0rp:
Okay. Could you share some example output from the Makefile you are using or `compile_commands.json`? The only way to parse almost everything is to just keep adding more test cases until everything is covered.

ranebrown:
Here is the output of `make -n`:

```
gcc -Wall -Wextra -g -I../inc -std=gnu99 -MMD -MP -c ../src/command_args.c -o command_args.o
gcc -Wall -Wextra -g -I../inc -std=gnu99 -MMD -MP -c ../src/main.c -o main.o
gcc -Wall -Wextra -g -I../inc -std=gnu99   -o phoenix command_args.o main.o
```

Here are some common flags that I think should be included.
```
-Wall
-Wextra
-std=xxxx
-Wxxxx
-Wno-xxxx
-pedantic
-pedantic-errors
```

Thanks!

ranebrown:
Just remembered something else too. I think directories such as build should be checked when searching upwards for a Makefile in the same way as done for `compile_commands.json`. I have this directory structure:
```
build
src
inc
```

Currently, I'm using a symlink to my Makefile in the src directory which works but isn't ideal. I would guess the behavior could just be copied over.

PR #1442, opened 2018-03-22
Added the word 'find' as it was missing in docs/ale-cs.txt

Crysikrend:
I haven't changed anything else but the wording inside ale-cs.txt file, so no tests should fail


w0rp:
The column size will be off, but I'll fix that.

w0rp:
Cheers! :beers:

PR #1449, opened 2018-03-25
Fix linter for textlint

januswel:
- Fix not working
- Use a nearest executable
- Add tests for command_callback


w0rp:
Both what was on master and what you have here aren't correct. The local executable wasn't being used on master, and here you removed the fallback path for executing the script which will be needed on Windows. Neither version used `node.exe` to run the `.js` file, which will be needed for Windows. Your version uses `%s`, which means linting while you type wouldn't work any more, and the master version didn't have tests for the command. Both built the executable string in ugly ways.

I pushed a commit now to address these issues: https://github.com/w0rp/ale/commit/8b34a4bf6b837e1ce7e060b053ec8abedd20cdb1

januswel:
Thank you for your comment and modifications.
I am so happy to be able to use textlint with ALE finally.

PR #1451, opened 2018-03-26
Added support for Python black fixer

Jackevansevo:
I've added support for [black](https://github.com/ambv/black).

Code is essentially just the autopep8 fixer code copied across with variables changed. Everything seems to work fine.

One thing this PR is missing is tests. I tried to duplicate the tests from autopep8 (since the two fixers are practically identical) but the tests failed.

If someone could point me in the right direction in regards to the tests, I'd be happy to try fix.

Cheers, Jack


Jackevansevo:
Not sure if I took the correct approach here, but looking briefly I couldn't find any location to automatically install the test deps.

So I copied a Python black executable into: `ale/test/command_callback/python_paths/with_virtualenv/env/bin` in order to test the Travis CI tests passing.

The Windows appveyor test is failing as I'm unsure how to get an windows .exe file in the Scripts directory.

test/command_callback/python_paths/with_virtualenv/env/bin/black
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
w0rp: Instead of copying the script into this repository, just create an empty file with the execute bit set with `chmod`. We don't need to actually run the script, we just need to see if it's there so ALE will try to run it later.

test/fixers/test_black_fixer_callback.vader
```diff
@@ -0,0 +1,39 @@
+Before:
```
w0rp: This test looks fine to me. What you need to do for Windows is create another empty executable file in the `Scripts` directory, in addition to the file for Linux in the `bin` directory, and I think you need to name it `.exe` for it to work. Have a look at some of the other dummy files for Windows testing.

w0rp:
Cheers! :beers:

PR #1452, opened 2018-03-26
Add fsc linter for Scala

PawkyPenguin:
Hey,
Scala has `fsc`, the "fast scala compiler", that can also be used for linting. `fsc` has less overhead than `scalac` because it starts a daemon that can be reused across multiple executions. I figured I'd add a linting option for it.

I pulled up most of the functionality in `ale_linters/scala/scalac.vim` into `autoload/ale/handlers/scala.vim` because `fsc` and `scalac` are basically identical for programming purposes.

By the way, I wasn't quite sure what the correct directories for my new vader tests are. I put them all into the `test/handler` directory because I figured just putting them under `test/` wouldn't be very tidy. However, I have three unit tests and only one of them tests the handler functions. Should I move `test_scalac_executable_and_command` and `test_fsc_executable_and_command` somewhere else?


autoload/ale/scala.vim
```diff
@@ -0,0 +1,23 @@
...
+
+function! ale#scala#GetCommandForLinter(buffer, linter) abort
+    let l:executable = ale#scala#GetExecutableForLinter(a:buffer, a:linter)
+
+    if empty(l:executable)
+        return ''
+    endif
+
+    return ale#Escape(l:executable) . ' -Ystop-after:parser %t'
+endfunction
```
w0rp: I'd rather you just copy and paste the `scalac` code into the new `fsc` file, and leave the old code alone. They are similar at the moment, but both using the same functions like this will make it harder to modify one without breaking the other one. `scalac` and `fsc` probably won't support _exactly_ the same options at some point, say if one program is updated before another.

autoload/ale/handlers/scala.vim
```diff
@@ -0,0 +1,38 @@
+" Author: Nils Leuzinger - https://github.com/PawkyPenguin
+"         PawkyPenguin <nilsl@student.ethz.ch>
+" Description: Scala linting handlers for scalac-like compilers.
+
+function! ale#handlers#scala#HandleScalacLintFormat(buffer, lines) abort
```
w0rp: This part is good. You can use the same error handler for both linters safely.

test/handler/test_fsc_executable_and_command.vader
```diff
@@ -0,0 +1,18 @@
...
+Execute(The default executable and command should be correct):
+  AssertEqual 'fsc', ale_linters#scala#fsc#GetExecutable(bufnr(''))
+  AssertEqual
+  \ ale#Escape('fsc') . ' -Ystop-after:parser %t',
+  \ ale_linters#scala#fsc#GetCommand(bufnr(''))
+
+Given scala.sbt(An empty SBT file):
+Execute(fsc should not be run for sbt files):
+  AssertEqual '', ale_linters#scala#fsc#GetExecutable(bufnr(''))
+  AssertEqual '', ale_linters#scala#fsc#GetCommand(bufnr(''))
```
w0rp: You can move the tests for `GetExecutable` or `GetCommand` to `test/command_callback`.

PawkyPenguin:
Thanks for the feedback!
I implemented the changes you suggested. I also noticed that `test_scala_handler` wasn't actually testing the handler, which was probably an oversight of the person before me (and of me as well). It tests the handler properly now.

w0rp:
Cheers! :beers:

PR #1462, opened 2018-03-29
Integration of qmlfmt linting tool

pylipp:
qmlfmt has several advantages over qmllint as a QML linting tool. It
gives information about error types and the precise position of the
error. It can also be used to format QML code (which might be a separate
PR).

The commits, new files and new documentation should speak for themselves
;)


w0rp:
Cheers! :beers:

PR #1465, opened 2018-03-31
pylint: Detect pylintrc file

rhysd:
I'm using ALE for running [pylint](https://github.com/PyCQA/pylint). But I have a trouble with it; `pylint` is invoked with default configuration.
pylint can be configured with [pylintrc](https://docs.pylint.org/en/latest/user_guide/run.html#command-line-options) file in repository. In my project, `.pylintrc` is put at the top of the repository and I want `pylint` invoked by ALE to use it.

By this PR, ALE can detect the pylintrc config file in parent directory and use it for linting.
When a user specified config file with `--rcfile` option  in `g:ale_python_pylint_options`, ALE checks it and the detection is not performed.


rhysd:
I was likely wrong. Current implementation can handle `pylintrc` file if `$CWD` is correct (=starting Vim within the respository). `pylint` automatically can detect local config file. So this implementation is not needed.

w0rp:
I think you're on to something, though. I've seen this kind of problem before. The current working directory could be anything, and it ought to be deterministic. For other linters, this problem has been solved by setting the working directory in the command that is run, so you always get the right results.

I'll either create an issue for this or add a comment to an existing one explaining what the problem is and how to fix it.

w0rp:
I opened an issue here: #1472

rhysd:
@w0rp Sure. Thank you for creating a new issue. Let's discuss this on the thread.

Issue #1424, opened 2018-03-15
Haskell fixer brittany is not working


ifyouseewendy:
## Information

**VIM version**

```
NVIM v0.2.2
Build type: Release
```

Operating System: macOS

### :ALEInfo

```
  Command History:
(finished - exit code 0) ['/usr/local/bin/bash', '-c', '''brittany'' ''/var/folders/4w/yh_7w05d29l9w6n5pgy0mcnm0000gn/T/nvimLcS8ld/3/brittany.hs''']
```

## What went wrong

I'm trying to use [brittany](https://github.com/lspitzner/brittany) (0.9.0.0) to auto fix haskell file, which doesn't work.

## Reproducing the bug

Steps for repeating the bug:

1. Create a new file with this demo script [provided by brittany](https://hexagoxel.de/brittany/).

```hs
-- brittany.hs
input   =  if properly   formatted then
     leave document 
         as is
     else
  return [reformat decl | 
    decl <- moduledecls, isTypeSig decl
         || isBinding decl ]
```

2. `:ALEFix`

## Fix

When I use brittany in my terminal, I notice that it writes to `STDOUT` by default. According to the manual, we can append a `--write-mode inplace` to override respective input file.

So I made a patch on `ale/autoload/ale/fixers/brittany.vim` to make it support passed in options and set `let g:ale_haskell_brittany_options = "--write-mode inplace"` in my `.vimrc`. It works well.

```vim
function! ale#fixers#brittany#Fix(buffer) abort
    let l:executable = ale#Var(a:buffer, 'haskell_brittany_executable')
    let l:options = ale#Var(a:buffer, 'haskell_brittany_options')

    return {
    \   'command': ale#Escape(l:executable)
    \       . (!empty(l:options) ? ' ' . l:options : '')
    \       . ' %t',
    \   'read_temporary_file': 1,
    \}
endfunction
```

By checking the file  https://github.com/w0rp/ale/blob/7a71186d62d1964e191dc723d8b65ab7786e0063/autoload/ale/fixers/brittany.vim , I see @eborden created this fixer last year. I'm wondering would you mind helping me confirm the issue when you have time? If you think this is a good fix, I'd love to make a PR for that.

w0rp:
[...]

PR #1480, opened 2018-04-08
Implemented fix from @ifyouseewendy

aspidites:
# Description
This basically implements the fix from #1424. The only deviation from the proposed fix is to imply the '--write-mode inplace' option as I cannot imagine a case where you don't want that, and I don't think it's idiomatic for a user to have to specify that in there config before the fixer even works. 

I have to admit that I've never written VimL, so don't know how to use vader. Looking in `test/fixers`, though, it appears that we are only testing that the command is called correctly, not that a particular fixer executes as expected, so I think I'm OK? I've also never used vint, but after installing `vint -s brittany.vim` doesn't produce output, so I assume I'm OK here as well. 

Happy to take feedback in either case, so long as you realize that responses will be a bit slow, as I only really have weekends to play around with this.

### DIFF
--- a/autoload/ale/fixers/brittany.vim
+++ b/autoload/ale/fixers/brittany.vim
@@ -1,4 +1,4 @@
-" Author: eborden <evan@evan-borden.com>
+" Author: eborden <evan@evan-borden.com>, ifyouseewendy <ifyouseewendy@gmail.com>, aspidiets <emarshall85@gmail.com>
 " Description: Integration of brittany with ALE.
 
 call ale#Set('haskell_brittany_executable', 'brittany')
--- a/autoload/ale/fixers/brittany.vim
+++ b/autoload/ale/fixers/brittany.vim
@@ -2,16 +2,13 @@
 " Description: Integration of brittany with ALE.
 
 call ale#Set('haskell_brittany_executable', 'brittany')
-call ale#Set('haskell_brittany_options', '--write-mode inplace')
 
 function! ale#fixers#brittany#Fix(buffer) abort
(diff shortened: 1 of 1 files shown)


autoload/ale/fixers/brittany.vim
```diff
@@ -1,13 +1,17 @@
-" Author: eborden <evan@evan-borden.com>
+" Author: eborden <evan@evan-borden.com>, ifyouseewendy <ifyouseewendy@gmail.com>, aspidiets <emarshall85@gmail.com>
 " Description: Integration of brittany with ALE.
 
 call ale#Set('haskell_brittany_executable', 'brittany')
+call ale#Set('haskell_brittany_options', '--write-mode inplace')
```
w0rp: Does brittany need this option to be set to work? If so, take this out of the options and write it into the command string below. Was this always needed, and did the brittany fixer never work, or did the behavior change in a recent version? If the behavior did change, then we probably detect which version of brittany is being run, and only add this option if needed.

w0rp:
The tests need to be updated here to reflect your changes. I'll sort that out.

w0rp:
Also, there was a space missing before the argument.

w0rp:
I have included this and fixed it now.

aspidites:
@w0rp , I learned that the hard way. Didn't realize '.' was concatenation until 2 minutes ago. Went ahead and squashed changes as well to prevent things from looking absurd in the commit history. I really ought to learn how to test things locally.

aspidites:
@w0rp, you're way too fast :laughing:. I was hoping that you wouldn't notice my fumbling about vimscript.

PR #1487, opened 2018-04-09
pylint: Move to the buffer's directory before running pylint command

### LINKED ISSUE #1472: pylintrc isn't detected if the current working directory isn't correct
`pylint` will search for a configuration file by following the current working directory. The current working directory cannot be relied upon, as it could be set to anything when a linter is run. This problem can be fixed by using either `ale#path#BufferCdString` or `ale#path#CdString` with whatever directory `pylint` actually needs, which could be detected. Generally, `pylint` ought to work no matter what current working directory you are using.

rhysd:
Fixes #1472 

I confirmed that this also covers #1465 because `pylint` detects `pylintrc` file in parent directories of the current working directory.

I fixed tests for `ale_linters/python/pylint.vim` also. I confirmed that I could run pylint via ALE from out of the repository.

### DIFF
--- a/ale_linters/python/pylint.vim
+++ b/ale_linters/python/pylint.vim
@@ -14,7 +14,8 @@ function! ale_linters#python#pylint#GetExecutable(buffer) abort
 endfunction
 
 function! ale_linters#python#pylint#GetCommand(buffer) abort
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

PR #1489, opened 2018-04-10
Add clangcheck as a C linter

tbodt:
Currently clangcheck can only be used for c++, which is a little annoying for me as a C programmer, since I have to use clangtidy instead which has some serious problems relating to non-absolute filenames. I could rant about those for days, but for now clangcheck does what I need just fine.


w0rp:
Have a look at the existing `clangcheck` test and documentation.

w0rp:
Closing due to inactivity.

PR #1491, opened 2018-04-10
correct cpp linter name from `g++` to `gcc`

skywind3000:
fixed: https://github.com/w0rp/ale/issues/1490


w0rp:
This will break existing configurations. What we should do instead is use the alias option for making either work.

w0rp:
I have changed the name and I added an alias for the old name now.

PR #1493, opened 2018-04-10
fixed: pylint doesn't work on windows:

skywind3000:
This patch solved: https://github.com/w0rp/ale/issues/1492

On windows, pylint will output file's full path name which contains colon and backslash in it. and it will break the regexp in pylint.vim.
So, all the pylint output will be discarded by pylint.vim.

Since filename is ignored in pylint.vim, so I change the pylint's output template {path} to {module}, and it will never contain any colon and backslash again.

This patch works great on windos and linux.


w0rp:
Instead of doing this, we should fix the handler so it handles Windows files names, similar to other handlers. Later on, we could consider gathering errors for other files, if there's a plugin which reports errors from multiple files or something.

skywind3000:
Thanks for the fast response and the update.

PR #1496, opened 2018-04-12
Parse compile_commands.json to get flags for the clang linters

tbodt:
Spiritual successor to #1242. A few things have been fixed and improved, and there are now tests.

### DIFF
--- a/autoload/ale/c.vim
+++ b/autoload/ale/c.vim
@@ -1,4 +1,6 @@
-" Author: gagbo <gagbobada@gmail.com>, w0rp <devw0rp@gmail.com>, roel0 <postelmansroel@gmail.com>
+" Author: gagbo <gagbobada@gmail.com>, w0rp <devw0rp@gmail.com>
+" Author: roel0 <postelmansroel@gmail.com>
+" Author: tbodt <https://github.com/tbodt>
 " Description: Functions for integrating with C-family linters.
 
 call ale#Set('c_parse_makefile', 0)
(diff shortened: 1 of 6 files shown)


w0rp:
Awesome, I'm glad someone is picking this up. I'll have a read through this later. Try to address the merge conflicts for now, if you can.

autoload/ale/c.vim
```diff
@@ -91,3 +92,27 @@ function! ale#c#FindCompileCommands(buffer) abort
...
 endfunction
+
+" Given a buffer number, return the relevant command from
+" compile_commands.json. Return 0 if none could be found.
+function! ale#c#GetCompileCommand(buffer) abort
+    let l:build_dir = ale#c#FindCompileCommands(a:buffer)
+    if empty(l:build_dir)
+        return 0
+    endif
+    let l:commands = json_decode(readfile(l:build_dir . '/compile_commands.json'))
```
w0rp: As mentioned in the previous pull request, we should try and replace `readfile()` with a command in the command chain which can be executed to read the file. `cat` will work on Unix machines, and `type` will probably work on Windows. `json_decode` is probably quite a bit slower, but there's no easy way around that for now.

Have a look at the linter code which can execute `make -n` now. What I think we should do is read `compile_commands.json` if the option is on, and fall back on parsing `make -n` if no `compile_commands.json` file can be found. The command callback should return the command to get the contents of `compile_commands.json`, and the other function which parses the output should parse JSON files. It's probably enough to look at the first line of output and see if it looks like the start of some JSON to tell the difference between the two types out output.

autoload/ale/c.vim
```diff
@@ -91,3 +92,27 @@ function! ale#c#FindCompileCommands(buffer) abort
...
+    let l:file = fnamemodify(bufname(a:buffer), ':p')
+    " save current directory, so we can restore it before returning
+    let l:cwd = getcwd()
```
w0rp: `getcwd()` is banned in the codebase because it doesn't work in asynchronous callbacks, because you could switch to another buffer with a different working directory. Use `expand('#' . a:buffer . ':p')` instead. The Bash script for checking VimL code will mention that in TravisCI.

autoload/ale/c.vim
```diff
@@ -91,3 +92,27 @@ function! ale#c#FindCompileCommands(buffer) abort
...
+    for l:command in l:commands
+        " temporarily change dir to run fnamemodify
+        execute 'cd' l:command.directory
```
w0rp: Don't change the working directory, as it could break other plugins that are being run. If you want want to get an absolute path for a relative path, you can use `ale#path#GetAbsPath(l:base_directory, l:filename)`.

autoload/ale/c.vim
```diff
@@ -91,3 +92,27 @@ function! ale#c#FindCompileCommands(buffer) abort
...
+        let l:command.file = fnamemodify(l:command.file, ':p')
+        if l:file ==# l:command.file
```
w0rp: For strings, I recommend using `is#` instead, as `'x' ==# 0`, which can be source of bugs.

autoload/ale/c.vim
```diff
@@ -91,3 +92,27 @@ function! ale#c#FindCompileCommands(buffer) abort
 
     return ''
 endfunction
+
+" Given a buffer number, return the relevant command from
+" compile_commands.json. Return 0 if none could be found.
+function! ale#c#GetCompileCommand(buffer) abort
```
w0rp: I recommend writing a function which doesn't do any I/O and returns the commands that are parsed, given a base directory and a List of lines in the file. Then it will be easy to write a series of unit tests against that function, and a small number of integration tests.

w0rp: Could, I thought you'd get relative paths from the file, and have to resolve them using a base directory.

w0rp:
Closing due to inactivity.

PR #1498, opened 2018-04-12
Fix rust rls linter argument

Eyenseo:
As far as I can tell [1] there is no +nighly (or similar) option leading to
the termination of the server.

Not sure how to test this ...

[1] https://github.com/rust-lang-nursery/rls/blob/master/src/main.rs#L87


Eyenseo:
Looks good?

w0rp:
Cheers! :beers:

PR #1510, opened 2018-04-18
Support passing fixers names to ALEFix as arguments 

### LINKED ISSUE #1471: Support passing fixers names to ALEFix as arguments
The `:ALEFix` command currently accepts no arguments and runs fixers defined with `g:ale_fixers` or `b:ale_fixers`. The command could be extended so it accepts a number of arguments for fixers to run, which will override either setting. The fixers ought to be run in the order given to the command, so `:ALEFix one two three` would run fixers `one`, `two`, and `three` in order. Given no arguments, `:ALEFix` would behave as it does now.

fenuks:
Fixes #1471.

### DIFF
--- a/autoload/ale/fix.vim
+++ b/autoload/ale/fix.vim
@@ -356,8 +356,11 @@ function! s:RunFixer(options) abort
     call ale#fix#ApplyFixes(l:buffer, l:input)
 endfunction
 
+function! s:GetCallbacks(linters) abort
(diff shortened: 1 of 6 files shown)


w0rp:
Bonus points for adding completion support too. :+1:  Cheers! :beers:

fenuks:
:grin: Thank you for merging and for bonus points.

PR #1511, opened 2018-04-19
Add `cucumber` checker for Cucumber files (closes #1374)

### LINKED ISSUE #1374: Cucumber support?
Hi there! I'd very much like to switch to Ale, but one feature is holding me back. I'd love the ability to support linting `.feature` files for undefined step definitions when writing cucumber.

https://github.com/vim-syntastic/syntastic/blob/master/syntax_checkers/cucumber/cucumber.vim

I'm not sure how hard it is to convert a syntastic checker, but the above works well (minus the wait time). Any pointers would be awesome. Thanks!

elebow:
This adds `cucumber` as a checker for Cucumber files (https://cucumber.io/).

For speed purposes, it is only feasible to use the `--dry-run` option, which means the only meaningful check that can be done is for undefined steps. Luckily, that is exactly the thing a user would most want to check.

Tested only with Cucumber for Ruby, but it should work for any flavor of Cucumber that follows a substantially similar directory structure, with tests in a directory named `features/` and step definitions somewhere below that. Each flavor of Cucumber has a slightly different interface.


elebow:
The command callback test is failing on Windows due to the path syntax. I'm not sure how to handle that.

w0rp:
The test is probably just failing due to different slashes, and it can probably be fixed by using the `ale#path#Simplify` function.

test/cucumber_fixtures/features/cuke.feature
```diff
@@ -0,0 +1,14 @@
+Feature: A user can view another user's profile
```
w0rp: You can just use blank files for these files. We won't run the files ourselves.

test/command_callback/test_cucumber_command_callback.vader
```diff
@@ -0,0 +1,25 @@
...
+
+  call ale#linter#Reset()
+  call ale#test#RestoreDirectory()
+
+Execute(Should require the nearest features dir, if one is found):
+  call ale#test#SetFilename('cucumber_fixtures/features/cuke.feature')
+
+  AssertEqual
+  \   'cucumber --dry-run --quiet --strict --format=json '
+  \   . '-r ''' . ale#path#Simplify('/testplugin/test/cucumber_fixtures/features/') . ''' %t',
```
w0rp: You can fix the problem with Windows tests here by using `g:dir . '/rest-of-path'`. You can use `../` if you need to. That will start the path with the drive letter. Failing that, you could replace `^/` with `C:\` if `has('win32')` is true.

elebow:
Thanks; changes made.

PR #1513, opened 2018-04-19
Add Elixir linter for mix

franciscoj:
Hi!

This adds a linter for `mix` the build tool from Elixir https://hexdocs.pm/mix/Mix.html

The tests are still missing (They're coming) but I would appreciate some feedback on it. I had to preprocess the warning messages into a single line (warnings in Elixir come on 2 lines).

Also I separated the matches for errors since they have a completely different format that the one for warnings.

I'm not sure those 2 things are a bit of "smelly" so came to ask for some feedback before continuing.

Thanks!

PS: Some notes about it

* It is a work in progress, I still need to fix a couple of things like having linting on `*.exs` files (which arn't compiled)
* It lags a bit on big projects.
* Elixir's compilation can run arbitrary code, which might not be appropriate, I would disable it by default.

### DIFF
--- a/doc/ale-elixir.txt
+++ b/doc/ale-elixir.txt
@@ -3,6 +3,21 @@ ALE Elixir Integration                                     *ale-elixir-options*
 
 
 ===============================================================================
+dialyxir                                                  *ale-elixir-dialyxir*
+
+Dialyzer, a DIscrepancy AnaLYZer for ERlang programs.
+http://erlang.org/doc/man/dialyzer.html
+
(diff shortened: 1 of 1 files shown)


franciscoj:
It does. I understand that being a concern but I'm not sure there is an alternative.

I'm also researching how to do it with elixir-ls, but it would still compile the code and generate binaries... Although on a different folder.

If you think that's a problem, let's close this one and I will open another once I find another solution.

PR #1533, opened 2018-04-26
flake8: Move to the buffer's directory before running flake8 command

inducer:
This is analogous to #1487, but for `flake8`, which looks at `setup.cfg` [and a number of other files](http://flake8.pycqa.org/en/latest/user/configuration.html?highlight=setup.cfg) relative to the buffer directory.

### DIFF
--- a/ale_linters/python/flake8.vim
+++ b/ale_linters/python/flake8.vim
@@ -39,6 +39,7 @@ function! ale_linters#python#flake8#VersionCheck(buffer) abort
 endfunction
 
 function! ale_linters#python#flake8#GetCommand(buffer, version_output) abort
+    let l:cd_string = ale#path#BufferCdString(a:buffer)
     let l:executable = ale_linters#python#flake8#GetExecutable(a:buffer)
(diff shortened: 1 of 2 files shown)


w0rp:
There shouldn't be a need to do this for `flake8`, it ought to figure out where the configuration files are based on the filename parameter.

inducer:
> There shouldn't be a need to do this for flake8, it ought to figure out where the configuration files are based on the filename parameter.

I have found this to be not quite true. If I open a file using `fzf.vim` (my default mode of opening a file at this point), the `flake8` config from the current working directory (if any) appears to be used.

inducer:
Also, thanks for taking a look!

reshleman:
This appears to prevent the usage of `b:ale_python_flake8_options` to specify an explicit `--config` which is relative to the buffer directory.

I think this would affect pylint, too, except `g:ale_python_pylint_change_directory` allows this behavior to be overridden for pylint. I don't see anything comparable for flake8 in the docs.

Thoughts?

w0rp:
Open an issue for adding that option.

PR #1539, opened 2018-04-27
set marks to add to jump list

JonnyRa:
looks plausible!

### DIFF
--- a/autoload/ale/loclist_jumping.vim
+++ b/autoload/ale/loclist_jumping.vim
@@ -66,6 +66,7 @@ function! ale#loclist_jumping#Jump(direction, wrap) abort
     let l:nearest = ale#loclist_jumping#FindNearest(a:direction, a:wrap)
 
     if !empty(l:nearest)
+        normal! m`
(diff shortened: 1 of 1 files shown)


JonnyRa:
Cool, ok,

Could you point me in the right direction?  Are there already some tests for the jump stuff?  How do I run them?  I've not done any vim testing stuff before.

Cheers

Jonny

w0rp:
Have a look at the `run-tests` script in the main ale directory. That script will run all of the tests via a Docker image. You'll need to have Docker installed to run that. You should hopefully be able to just run the script if you're on Mac OSX or Linux. There's a Batch file for running tests on Windows too, though that can be harder to set up.

Consult the documentation for [Vader](https://github.com/junegunn/vader.vim) the testing tool, and look at the tests in the `test` directory. `test/test_loclist_jumping.vader` has tests for jumping already, and you can write some VimL code in there for checking if jumps are set correctly.

JonnyRa:
Thanks!  Sorry I've been really busy and not had chance to do anything

PR #1543, opened 2018-04-28
Add ALEJobStarted User autocommand event

vancluever:
The `ALELintPre` and `ALELintPost` autocommand events are currently being
used by [lightline-ale](https://github.com/maximbaz/lightline-ale) to refresh the status line and check the linter
status for a current buffer. One of the plugin's checks looks to see if
linters are currently running, via `ale#engine#IsCheckingBuffer()`. This
currently only works partially in certain situations. In my particular
case, working with Go files, this only seems to function properly when a
file is initially opened. Saving a file does not correctly update the
status.

This seems to be due to the fact that `ALELintPre` actually runs before
any jobs are carried out, making it plausible that hooking into
`ALELintPre` for the purpose of checking to see if there are any currently
running linters for a buffer is unreliable as it would be prone to
pretty obvious race conditions.

This adds a new User autocommand, `ALEJobStarted`, that gets fired at the
start of every new job that is successfully run. This allows a better
point to hook into checking the linter status of a buffer using
`ale#engine#IsCheckingBuffer()` by ensuring that at least one job has
started by the time `IsCheckingBuffer` is run.

-----

I've included tests for most of the above explained things here, namely for the new event and _also_ to confirm the timing issues above surrounding using `IsCheckingBuffer` at `ALELintPre`. Let me know if there's anything else that needs fixing - or if the autocommand should be named something else or placed at a different location, etc.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,17 +524,21 @@ Will give you:
 ### 5.viii. How can I execute some code when ALE starts or stops linting?
 
 ALE runs its own [autocmd](http://vimdoc.sourceforge.net/htmldoc/autocmd.html)
-events when a lint or fix cycle are started and stopped. These events can be
-used to call arbitrary functions before and after ALE stops linting.
+events when a lint or fix cycle are started and stopped. There is also an event
+that runs when a linter job has been successfully started. These events can be
+used to call arbitrary functions during these respective parts of the ALE's
+operation.
 
 ```vim
 augroup YourGroup
     autocmd!
-    autocmd User ALELintPre  call YourFunction()
-    autocmd User ALELintPost call YourFunction()
+    autocmd User ALELintPre    call YourFunction()
+    autocmd User ALELintPost   call YourFunction()
 
-    autocmd User ALEFixPre   call YourFunction()
-    autocmd User ALEFixPost  call YourFunction()
+    autocmd User ALEJobStarted call YourFunction()
+
+    autocmd User ALEFixPre     call YourFunction()
+    autocmd User ALEFixPost    call YourFunction()
 augroup END
 ```
 
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -2461,6 +2461,7 @@ ALEJobStarted                                           *ALEJobStarted-autocmd*
   |ale#engine#IsCheckingBuffer()| over |ALELintPre-autocmd|, which is actually
   triggered before any linters are executed.
 
+
 ===============================================================================
 10. Special Thanks                                         *ale-special-thanks*
 


maximbaz:
Interesting. My notes:

1. Looks like `lightline-ale` is a bit obsolete, it subscribes to `ALELint` autocmd, which as I see now is deprecated according to https://github.com/w0rp/ale/pull/1203. And `ALELint` is only an old name for `ALELintPost`, so `lightline-ale` does not currently subscribe to `ALELintPre` at all (we should fix that).
2. Instead of adding new autocmd events, can we maybe request some guarantees from ALE itself, that by the time `ALELintPre` is being fired, `ale#engine#IsCheckingBuffer` is guaranteed to return `true`, and by the time `ALELintPost` is being fired, it is guaranteed to return `false`? Basically I want to solve the race conditions within ALE itself, not on the `lightline-ale` side, and not with adding more events that would make `lightline-ale` refresh more often than necessary.

What do you think?

vancluever:
Hey @maximbaz,

Yeah, it totally slipped my mind in all of my troubleshooting that `lightline-ale` is subscribing to the deprecated event!

I'm all for less things, for sure, but right now the problem is with the placement of the other events. `ALELintPre` fires [here](https://github.com/w0rp/ale/blob/91d7e81ebc08581fc0339c939c0adecaa590f2f7/autoload/ale/engine.vim#L825), which I think is _technically_ correct as it's designed to be a pre-lint hook, so that make sense. The core of the issue is that the current implementation of [`IsCheckingBuffer`](https://github.com/w0rp/ale/blob/91d7e81ebc08581fc0339c939c0adecaa590f2f7/autoload/ale/engine.vim#L87) is not based on a opaque started/stopped flag, but rather determines its status from a more real-time measure of whether or not any linters are literally running for the buffer at that specific point in time (via checking the active linter list for a buffer). Hence more often than not, `IsCheckingBuffer` will not return true at the time that event fires.

Considering the above, I think it would be correct in concluding that the `ALELintPre` event should _not_ be used in conjunction with `IsCheckingBuffer`, nor do I think it should be moved, as other plugins of course may depend on its current location.

Thinking about it, I almost wonder if it would be a good idea to just not even bother with `ALELintPre` (considering that it would not really be useful to `lightline-ale` at all) and maybe hook into something like `ALEJobStarted` (or whatever it ends up being). This event fires at the successful start of any job and pretty much guarantees that `IsCheckingBuffer` will be true, as shown by the supplied tests. I think this is close to what you are looking for and should only result in a marginal performance hit (assuming someone does not have dozens of buffers open with lots of expensive linters, although in that case I think the performance demands of a refreshing statusline will be trivial in comparison).

Thoughts?

maximbaz:
I see, thanks for explaining the details!

I'm not opposed to your proposal, but more as a curiosity I quickly drafted https://github.com/maximbaz/lightline-ale/pull/9, could you test whether it solves your issue, and tell me what you think about that approach in general? 🙂

That code simply doesn't use `IsCheckingBuffer` which as you said is not really linked to `ALELintPre` and `ALELintPost`, instead it tracks the lint status in a buffer variable.

vancluever:
Hey @maximbaz, I won't have a chance to test extensively right now, but I don't think this will work 100% as expected as if I understand correctly, the autocommand events fire independent of a specific buffer. See https://github.com/w0rp/ale/issues/1363#issuecomment-368307402 for an explanation (which also seemed to be the genesis of `IsCheckingBuffer` being public).

maximbaz:
Makes sense, then I think your approach is the best we can get 😉

vancluever:
Nice! Thanks @w0rp!

Issue #1524, opened 2018-04-24
Improve start up performance


w0rp:
ALE starts pretty quickly, but it could be better. We should measure start up performance and make a few tweaks to improve start up times.

## Ideas

- [x] Move code for printing `:ALEInfo` to another file, loaded only when the command is first issued.
- [x] Try replacing `ale#Set` calls in the main plugin file with `let` commands using `get(g:, ...)`.

w0rp:
The `:ALEInfo` idea I mentioned was already how it was, but I made sure not to mess that up when adding the new LSP error debugging.

w0rp:
This should be good enough now.

PR #1547, opened 2018-04-30
Performance pass: reduce startup time by 60%

neersighted:
This is a work-in-progress refactoring based on performance profiling.

Some major changes have occured, including removing deprecated code and
combining as many augroups as possible.

This should not be merged until changes are discussed and the test suite
is updated to pass these changes.

Addresses #1524.

### DIFF
--- a/after/plugin/ale.vim
+++ b/after/plugin/ale.vim
@@ -9,9 +9,8 @@ endif
 " Set the flag so this file is not run more than one time.
 let g:loaded_ale_after = 1
 
-" Check if the flag is available and set to 0 to disable checking for and
-" emitting conflicting plugin warnings.
-if exists('g:ale_emit_conflict_warnings') && !g:ale_emit_conflict_warnings
+" Check if the conflict warnings are disabled.
+if !get(g:, 'ale_emit_conflict_warnings', 1)
     finish
 endif
 
--- a/autoload/ale/autocmd.vim
+++ b/autoload/ale/autocmd.vim
@@ -1,15 +1,9 @@
 function! ale#autocmd#InitAuGroups() abort
-    " This value used to be a Boolean as a Number, and is now a String.
-    let l:text_changed = '' . g:ale_lint_on_text_changed
+    if g:ale_enabled
+        " This value used to be a Boolean as a Number, and is now a String.
+        let l:text_changed = '' . g:ale_lint_on_text_changed
 
-    augroup ALEPatternOptionsGroup
-        autocmd!
-        autocmd BufEnter,BufRead * call ale#pattern_options#SetOptions(str2nr(expand('<abuf>')))
-    augroup END
-
-    augroup ALERunOnTextChangedGroup
-        autocmd!
-        if g:ale_enabled
+        augroup ALETriggerGroup
             if l:text_changed is? 'always' || l:text_changed is# '1'
                 autocmd TextChanged,TextChangedI * call ale#Queue(g:ale_lint_delay)
             elseif l:text_changed is? 'normal'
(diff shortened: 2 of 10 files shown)


after/plugin/ale.vim
```diff
@@ -9,9 +9,8 @@ endif
 " Set the flag so this file is not run more than one time.
 let g:loaded_ale_after = 1
 
-" Check if the flag is available and set to 0 to disable checking for and
-" emitting conflicting plugin warnings.
-if exists('g:ale_emit_conflict_warnings') && !g:ale_emit_conflict_warnings
+" Check if the conflict warnings are disabled.
+if !get(g:, 'ale_emit_conflict_warnings', 1)
```
w0rp: I'm considering removing the `after` file for conflicting plugin warnings and the `runtimepath` part in the main plugin file now. I think enough people understand that Syntastic and ALE conflict with one another, and Neomake could be installed too if you don't use it for diagnostics where ALE is used too.

autoload/ale/autocmd.vim
```diff
@@ -1,77 +1,62 @@
...
+            endif
 
-    augroup ALERunOnInsertLeave
-        autocmd!
-        if g:ale_enabled && g:ale_lint_on_insert_leave
-            autocmd InsertLeave * call ale#Queue(0)
-        endif
-    augroup END
+            if g:ale_lint_on_save
+                autocmd BufWritePost * call ale#events#SaveEvent(str2nr(expand('<abuf>')))
```
w0rp: This event will still need to be there if ALE fixes something on save, when `g:ale_enabled` is `0`. The `fix_on_save` setting could only be on for a particular buffer, so we probably have to always have this event set up.

autoload/ale/autocmd.vim
```diff
@@ -1,77 +1,62 @@
...
+                " Look for a warning to echo as soon as we leave Insert mode.
+                " The script's position variable used when moving the cursor will
+                " not be changed here.
+                autocmd InsertLeave * call ale#cursor#EchoCursorWarning()
+            endif
+        augroup END
+
+        augroup ALEPatternOptionsGroup
+            autocmd BufEnter,BufRead * call ale#pattern_options#SetOptions(str2nr(expand('<abuf>')))
+        augroup END
```
w0rp: We'll need the pattern options to work even when linting is turned off, for fixing code. That's so you can enable fixing code only for specific buffers, even when linting is off.

autoload/ale/autocmd.vim
```diff
@@ -1,77 +1,62 @@
...
 
-    augroup ALEPatternOptionsGroup
-        autocmd!
-        autocmd BufEnter,BufRead * call ale#pattern_options#SetOptions(str2nr(expand('<abuf>')))
-    augroup END
-
-    augroup ALERunOnTextChangedGroup
-        autocmd!
-        if g:ale_enabled
+        augroup ALETriggerGroup
```
w0rp: It might be best to leave the `autocmd!` lines in, if they don't cost too much to execute, in case a script is reloaded which runs this function again.

autoload/ale/events.vim
```diff
@@ -11,8 +11,32 @@ function! ale#events#QuitRecently(buffer) abort
...
+function! ale#events#EnterEvent(buffer) abort
+    " When entering a buffer, we are no longer quitting it.
+    call setbufvar(a:buffer, 'ale_quitting', 0)
+
+    " Save the original filetype for later comparison.
+    let l:filetype = getbufvar(a:buffer, '&filetype')
+    call setbufvar(a:buffer, 'ale_original_filetype', l:filetype)
+
+    " Don't run costly, pointless operations on startup.
+    if v:vim_did_enter
```
w0rp: One thing you'll have to check is, do you get results for linting a file if you open it from the command line? Say if you run `vim whatever.py`, does it still run `pylint` and give you results right away, even if the option for linting when the filetype changes is off?

autoload/ale/events.vim
```diff
@@ -28,42 +52,25 @@ function! ale#events#SaveEvent(buffer) abort
     endif
 endfunction
 
-function! s:LintOnEnter(buffer) abort
-    if ale#Var(a:buffer, 'enabled')
-    \&& g:ale_lint_on_enter
-    \&& has_key(b:, 'ale_file_changed')
-        call remove(b:, 'ale_file_changed')
+function! ale#events#FileChangedEvent(buffer) abort
+    " Lint if we are in the foreground, otherwise defer to EnterEvent.
```
w0rp: :+1: Good optimization.

autoload/ale/events.vim
```diff
@@ -11,8 +11,32 @@ function! ale#events#QuitRecently(buffer) abort
...
+    let l:filetype = getbufvar(a:buffer, '&filetype')
+    call setbufvar(a:buffer, 'ale_original_filetype', l:filetype)
+
+    " Don't run costly, pointless operations on startup.
+    if v:vim_did_enter
+        " Update highlights.
+        call ale#highlight#UpdateHighlights()
+
+        " Apply fix if one was run in the background.
+        call ale#fix#ApplyQueuedFixes(a:buffer)
```
w0rp: Calling this function will cause `autoload/ale/fix.vim` to be loaded, even if you don't use ALE for fixing things. You could wrap this call in an `exists` check like we do for `ale#statusline#Update` in `autoload/ale/engine.vim`.

autoload/ale/highlight.vim
```diff
@@ -58,7 +58,7 @@ function! ale#highlight#RemoveHighlights() abort
 endfunction
 
 function! ale#highlight#UpdateHighlights() abort
-    let l:item_list = get(b:, 'ale_enabled', 1) && g:ale_enabled
+    let l:item_list = get(b:, 'ale_enabled', g:ale_enabled)
```
w0rp: Leave the `&& g:ale_enabled` check in here. ALE can't be enabled locally when it's disabled globally, so that we know we can remove the events needed for running locally when ALE is disabled globally.

autoload/ale/autocmd.vim
```diff
@@ -1,77 +1,62 @@
...
+                autocmd FileType * call ale#events#FileTypeEvent(str2nr(expand('<abuf>')), expand('<amatch>'))
+            endif
 
-    if !g:ale_enabled
-        augroup! ALERunOnTextChangedGroup
-        augroup! ALERunOnEnterGroup
-        augroup! ALERunOnInsertLeave
-        augroup! ALECursorGroup
+            " Clean up highlights when buffers are hidden.
+            autocmd BufHidden * call ale#highlight#RemoveHighlights()
```
w0rp: You should probably move these events for the highlights back to the file they are in. It's tricky. They should run if and only if `g:ale_set_highlights` is `1`, but they should also run if the setting is initially off and later turned on while Vim is running.

autoload/ale/events.vim
w0rp: After reading code from different files, I think you should probably move this call back to where it was in the `fix.vim` file. We will still need to run this function if `g:ale_enabled` is `0`, and if you run `:ALEFix` and switch buffers quickly enough.

autoload/ale/statusline.vim
```diff
@@ -60,53 +53,6 @@ function! s:GetCounts(buffer) abort
...
-    endif
-
-    return l:res
-endfunction
-
-" Returns a formatted string that can be integrated in the statusline.
-"
-" This function is deprecated, and should not be used. Use the airline plugin
-" instead, or write your own status function with ale#statusline#Count()
-function! ale#statusline#Status() abort
```
w0rp: These functions are deprecated, but we have to keep them for now. After we release version 1.9, we can start removing deprecated features for version 2.0. We should mark some more code as deprecated first, like any remaining uses of `_args` options, and emit warnings for each.

autoload/ale/statusline.vim
```diff
@@ -42,16 +38,13 @@ function! ale#statusline#Update(buffer, loclist) abort
...
     let g:ale_buffer_info[a:buffer].count = l:count
 endfunction
 
-" Get the counts for the buffer, and update the counts if needed.
-function! s:GetCounts(buffer) abort
-    if !exists('g:ale_buffer_info') || !has_key(g:ale_buffer_info, a:buffer)
+" Returns a Dictionary with counts for use in third party integrations.
+function! ale#statusline#Count(buffer) abort
+    " Check if ALE has not run (and return a dummy if so).
+    if !has_key(get(g:, 'ale_buffer_info', {}), a:buffer)
```
w0rp: I'd probably revert your changes to this file for now, but you can keep this `has_key` part.

autoload/ale/toggle.vim
```diff
@@ -40,10 +40,10 @@ function! ale#toggle#Toggle() abort
             call ale#balloon#Enable()
         endif
     else
-        call s:CleanupEveryBuffer()
         call s:DisablePostamble()
+        call s:CleanupEveryBuffer()
 
-        if has('balloon_eval')
+        if exists('*ale#balloon#Disable')
```
w0rp: :+1:

autoload/ale/toggle.vim
```diff
@@ -89,10 +84,10 @@ function! ale#toggle#ToggleBuffer(buffer) abort
     if l:enabled
         call s:EnablePreamble()
     else
+        call s:DisablePostamble()
```
w0rp: I think this is probably safe to move up here, but I'm not sure.

plugin/ale.vim
```diff
@@ -14,7 +14,7 @@ let g:loaded_ale_dont_use_this_in_other_plugins_please = 1
 
 " A flag for detecting if the required features are set.
 if has('nvim')
-    let s:has_features = has('timers')
+    let s:has_features = has('nvim-0.2.0')
```
w0rp: We can't change this just yet. First we'll need to get version 1.9 out and emit a few more deprecation warnings, and then we can require NeoVim 0.2.x or above. NeoVim versions below that are deprecated for now, but still supported.

plugin/ale.vim
```diff
@@ -23,25 +23,21 @@ endif
...
 let g:ale_emit_conflict_warnings = get(g:, 'ale_emit_conflict_warnings', 1)
 
-if g:ale_emit_conflict_warnings
-\&& match(&runtimepath, '[/\\]ale[/\\]after') < 0
+if g:ale_emit_conflict_warnings &&
+\ matchstr(&runtimepath, '/ale/after') is# ''
+\ && matchstr(&runtimepath, '\\ale\\after') is# ''
     " Add the after directory to the runtimepath
     " This is only done if the after directory isn't already in runtimepath
     let &runtimepath .= ',' . expand('<sfile>:p:h:h') . '/after'
```
w0rp: I think we can just remove the conflict warnings stuff now. I don't think we need it any more. If we get issues about conflicts, we can just document them and leave some comments.

plugin/ale.vim
```diff
@@ -198,27 +195,27 @@ let g:ale_history_log_output = get(g:, 'ale_history_log_output', 1)
...
-call ale#Set('completion_enabled', 0)
-call ale#Set('completion_delay', 100)
-call ale#Set('completion_max_suggestions', 50)
+let g:ale_completion_enabled = get(g:, 'ale_completion_enabled', 0)
+let g:ale_completion_delay = get(g:, 'ale_completion_delay', 100)
+let g:ale_completion_max_suggestions = get(g:, 'ale_completion_max_suggestions', 50)
 
 " A setting for wrapping commands.
-call ale#Set('command_wrapper', '')
+let g:ale_command_wrapper = get(g:, 'ale_command_wrapper', '')
```
w0rp: Removing `ale#Set` everywhere here is good.

plugin/ale.vim
```diff
@@ -299,35 +299,3 @@ nnoremap <silent> <Plug>(ale_go_to_definition) :ALEGoToDefinition<Return>
 nnoremap <silent> <Plug>(ale_go_to_definition_in_tab) :ALEGoToDefinitionInTab<Return>
 nnoremap <silent> <Plug>(ale_find_references) :ALEFindReferences<Return>
 nnoremap <silent> <Plug>(ale_hover) :ALEHover<Return>
-
-" Set up autocmd groups now.
-call ale#autocmd#InitAuGroups()
-
-" Housekeeping
-
-augroup ALECleanupGroup
```
w0rp: Leave the deprecated functions in for now. We can remove them after we get version 1.9 out. I'd leave a bit of a grace period after 1.9 is out. We can probably release 1.9 this month, leave these deprecated things in for another month, and push a commit to remove them in July.

neersighted:
I should have time to rebase this and fix up the tests soon.

w0rp:
I've incorporated not checking files with empty filetypes, and removed the plugin conflict warning code. The currently deprecated functionality will have to stay for now. The part where `ale#ShouldDoNothing` is moved around breaks stopping ALE from running for a little while if errors are thrown, so those calls will have to stay where they are. The parts which change the `autocmd` lines a lot will probably break something or other, and need some thinking.

I'll replace `ale#Set` in the main plugin file with `let ... get` around about now, and close this pull request. You can open some more pull requests for different performance tweaks.

PR #1550, opened 2018-05-01
Add bundle option to mdl

deivid-rodriguez:
Just like [rubocop](https://github.com/bbatsov/rubocop), [mdl](https://github.com/markdownlint/markdownlint) is distributed as a ruby gem. So like with `rubocop`, I think it makes sense to provide a `bundler` option, so that a well defined version of the library is used.

This PR is an adaptation of https://github.com/w0rp/ale/pull/561 to `mdl`.


deivid-rodriguez:
Thanks!

PR #1552, opened 2018-05-02
Add XO fixer

a-marquez:
Small patches to add XO fixer definitions, the XO linter has been previously defined  and XO uses eslint under the hood.

### DIFF
--- a/autoload/ale/fixers/xo.vim
+++ b/autoload/ale/fixers/xo.vim
@@ -1,0 +1,23 @@
+" Author: Albert Marquez - https://github.com/a-marquez
+" Description: Fixing files with XO.
+
+call ale#Set('javascript_xo_executable', 'xo')
+call ale#Set('javascript_xo_use_global', get(g:, 'ale_use_global_executables', 0))
+call ale#Set('javascript_xo_options', '')
+
+function! ale#fixers#xo#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'javascript_xo', [
+    \   'node_modules/xo/cli.js',
+    \   'node_modules/.bin/xo',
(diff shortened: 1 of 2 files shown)


autoload/ale/fixers/xo.vim
```diff
@@ -0,0 +1,23 @@
...
+    \   'node_modules/.bin/xo',
+    \])
+endfunction
+
+function! ale#fixers#xo#Fix(buffer) abort
+    let l:executable = ale#fixers#xo#GetExecutable(a:buffer)
+
+    return {
+    \   'command': ale#node#Executable(a:buffer, l:executable)
+    \       . ' --fix %t',
```
w0rp: There's a bug here. This command won't read any `xo` configuration files. Does `xo` offer some options similar to `eslint` for fixing this? If you look at the `eslint` fixer, it uses `--stdin-filename %s --stdin --fix-dry-run --format=json` and then pulls the changes out of the JSON data, for the newest version of eslint. `--stdin-filename %s --stdin --fix-to-stdout` is used for `eslint_d`.

a-marquez: Thanks for reviewing. `xo` does expose the options `--stdin-filename` and `--stdin` but not `--fix-dry-run` or `--format=json`. `xo` reads configuration from `package.json` implicitly and I do see `:ALEFix` respecting my configuration including when I edit it between fixes. Though I could still edit the command to:
```
\       . ' --stdin-filename %s --stdin'
```
I referenced the `standard` fixer to do this as `xo` is meant to be used in similar fashion, as various opinionated rules with little to no additional configuration.

w0rp:
Cheers! :beers:

PR #1555, opened 2018-05-04
Added NASM assembly linter

BobuSumisu:
Added linter for filetype/asmsyntax `nasm` (a file linter as NASM doesn't support input from stdin).

Was uncertain if this should have been added as a NASM linter for `asm`, but considering `asm` and `nasm` being different filetypes I guess this is the correct organization. 

Tested with NASM 2.12.01 (latest in Debian stable) and NASM 2.13.03 (latest).


w0rp:
Cheers! :beers:

PR #1556, opened 2018-05-05
[WIP] TUI / GUI tooltip with content from ALEHover

gagbo:
This is a RFC / small fix maybe.

I would like to have `:ALEHover` content displayed in a tooltip like `termdebug` plugin so beautifully do in [vim](https://asciinema.org/a/BQiwffN0QVS9K54XLydo49s3e)

I found out that ale already uses the balloon feature that termdebug uses under the hood, therefore I would like to find the correct way to access the string that `ale#hover#Show()` ultimately shows in the message bar. I discovered that it's not that easy. What do you think the best way is ?

Should I reimplement all functions that autoload/hover.vim uses ?
- Should I make more functions public so we can have easier access (ShowDetails ?) ?
- Should I change the signatures of some functions so they actually return strings (namely the callback handlers ?) ?

Ultimately, the goal is to allow users to choose a source for the `balloonexpr`, from loclist like it's currently done, or from `:ALEHover` content (typically a setting that would be buffer dependent, so it's possible to activate the Hover source only for filetypes where LSP linters are plugged in)

Also, since I was there, I added some guards around the `set ballooneval`, so neovim or old vim versions won't complain, and also added `balloonevalterm` because in vim the GUI and TUI tooltips are separated options (but I assumed that we would want them in all possible cases if the user activated the feature).

### DIFF
--- a/autoload/ale/balloon.vim
+++ b/autoload/ale/balloon.vim
@@ -20,9 +20,33 @@ function! ale#balloon#Expr() abort
 endfunction
 
 function! ale#balloon#Disable() abort
-    set noballooneval balloonexpr=
+    if !has('balloon_eval') && !has('balloon_eval_term')
+        finish
+    endif
+
+    if has('balloon_eval')
+        set noballooneval
+    endif
+
+    if has('balloon_eval_term')
+        set noballoonevalterm
+    endif
+
+    set balloonexpr=
 endfunction
 
 function! ale#balloon#Enable() abort
-    set ballooneval balloonexpr=ale#balloon#Expr()
+    if !has('balloon_eval') && !has('balloon_eval_term')
+        finish
+    endif
+
+    if has('balloon_eval')
+        set ballooneval
+    endif
+
+    if has('balloon_eval_term')
+        set balloonevalterm
+    endif
+
+    set balloonexpr=ale#balloon#Expr()
 endfunction
--- a/autoload/ale/balloon.vim
+++ b/autoload/ale/balloon.vim
@@ -19,6 +19,13 @@ function! ale#balloon#Expr() abort
     return ale#balloon#MessageForPos(v:beval_bufnr, v:beval_lnum, v:beval_col)
 endfunction
 
+function! ale#balloon#HoverExpr() abort
+    " TODO : Find a way with ale#hover API to just extract the message, given
+    " beval_bufnr, beval_lnum, beval_col.
+    " Note : Termdebug uses beval_text
+    return 'Test'
+endfunction
+
 function! ale#balloon#Disable() abort
     if !has('balloon_eval') && !has('balloon_eval_term')
         finish


gagbo:
I added a bunch of `" XXX:` comments to show the basic changed I had to make to make the feature work as I want (= make the tooltip appear with some relevant text).

If we agree on the basic mechanism, we can start discussing about how making this clean (options for the user, avoid writing information at 2 positions, avoid code duplication...)

gagbo:
I made a few changes accordingly : 
- there's only one ale#hover#Show() function now taking all the parameters
- There was a check on cursor position in `ale#hover#HandleLSPResponse()` function that prevented handling the answer correctly since mouse and cursor are often in different positions
- I'll do the vader corrections when I know I can finish the features.

There's only 1 thing I'm still wondering, but it's more 'vim' than 'ale'. When I run in TUI, the tooltip always disappear after a short delay, whereas in GUI, the tooltip stays longer. Maybe it's because my terminal emulator (kitty) is using openGL and there are redraws implied, I'll have to investigate.

Here are some gifs of what it currently looks like (in TUI (dark) and GUI (light)) : 
peek_vim_c41e062
peek_gvim_c41e062

gagbo:
Note : the tests currently fail because vim versions that have `+balloon_eval` and/or `+balloon_eval_term` do not necessarily have `balloon_show()` (which came in [patch 0396](https://github.com/vim/vim/commit/59716a27bd5c6c64def6c3ca430ff1246deae749)).

I'm assuming tests are failing because vim versions in test instances are in this gap.

gagbo:
The last (I think) 'pure design' decision remaining is knowing if we should print Hover output to both messages and balloon or just balloon when enabled and available. The issue with having both output is for multiline output (that trigger a new split and take the cursor), because not having the cursor in the same window anymore breaks Hover (see the GIFs).

- I suppose we should also pass `v:beval_winnr` in some way to `ale#hover#Show` in order to send the proper message to lsp, but I'll check that later.

- And then at some point I'll test the feature with diagnostics to check priority.

- And then at some point I'll try to find a way to test this in Vader (I don't know yet how to send arbitrary LSP messages, nor how to send mouse events, nor how to check for a function call instead of a return value)

gagbo:
Starting Vader: C:\testplugin\test\test_history_saving.vader
    (1/6) [  GIVEN] Some imaginary filetype
    (1/6) [EXECUTE] History should be set when commands are run
    (1/6) [EXECUTE] (X) ['command', 'job_id', 'status'] should be equal to ['command', 'exit_code', 'job_id', 'status']

Something happened in the merge

gagbo:
The last commit raised the timeout on one failing appveyor test, since the output from the tests seemed to show that the linter was not finished when the check was called.

Feel free to revert it.

autoload/ale/hover.vim
```diff
@@ -95,34 +108,45 @@ function! s:ShowDetails(linter) abort
...
-    for l:linter in ale#linter#Get(&filetype)
+" Obtain Hover information for the specified position
+" Only 1 optional argument is expected :
+"   - called_from_balloonexpr, this flag marks if we want the result from this
+"     ale#hover#Show to display in a balloon if possible
+"
+" Currently, the callbacks displays the info
+" - in the balloon if requested from balloonexpr and balloon_show is detected
+" - as status message otherwise
+function! ale#hover#Show(buffer, line, col, ...) abort
```
w0rp: Keeping track of the arguments will probably get a bit tedious as we go on. I think it's probably best to just write a single argument here instead, which contains all of the arguments in on `Dictionary`. That's probably the most future-proof option.

autoload/ale/hover.vim
```diff
@@ -34,15 +40,19 @@ function! ale#hover#HandleLSPResponse(conn_id, response) abort
...
-        let l:buffer = bufnr('')
-        let [l:line, l:column] = getcurpos()[1:2]
-        let l:end = len(getline(l:line))
 
-        if l:buffer isnot l:options.buffer
-        \|| l:line isnot l:options.line
-        \|| min([l:column, l:end]) isnot min([l:options.column, l:end])
-            " Cancel display the message if the cursor has moved.
-            return
+        if !get(l:options, 'hover_from_balloonexpr', 0)  " If the call did __not__
```
w0rp: Move the comments on the RHS to lines above instead. That's my preferred style.

autoload/ale/hover.vim
```diff
@@ -24,7 +24,13 @@ function! ale#hover#HandleTSServerResponse(conn_id, response) abort
 
         if get(a:response, 'success', v:false) is v:true
         \&& get(a:response, 'body', v:null) isnot v:null
-            call ale#util#ShowMessage(a:response.body.displayString)
+            if get(l:options, 'hover_from_balloonexpr', 0)
+            \&& exists('*balloon_show')
+            \&& ale#Var(l:options.buffer, 'set_balloons')
```
w0rp: We'll need to think about the meaning of the `set_balloons` setting. I guess it makes sense how you've done it here. I doubt many people really look at the balloons for the error messages at the moment.

I say we should do it how you've done it here, and update the documentation to say that the option also enables hover information for LSP and tsserver.

I might update the tsserver stuff myself, or you can give it a go if you want. If you install TypeScript into `node_modules` for a project, it should be pretty easy to get it running `tsserver` comes with TypeScript.

gagbo: I assumed that the `ale#hover` callbacks were only called by `ale#hover#Show`, and therefore the value of set_ballons in this context is only relevant for the Hover information to begin with. I think this specific part of the code is not related to the actual design decision, it's really just a sanity check on what the user asked for relatively to balloons before we make it appear.

The part that makes the meaning of `set_balloons` matter in a "hover vs diagnostics" way is in `ale#balloon#MessageForPos` (where I decided that if diagnostics are found, show them and don't even try to get hover information for the position).

autoload/ale/hover.vim
gagbo: Optional arguments is my (lazy) way of making changes without having to delve too much in changing the tests to make them work, this is definitely something I want to make better. I'll implement the dictionary, and call it opt (for optional arguments).

The `get()` function should help with default values anyway.

gagbo:
> It sounds like you hit a similar issue to what I hit with Vim. Vim doesn't like you showing messages via :echo from an async context, but NeoVim is okay with it. It would show the message, then immediately vanish. The messages for diagnostics that are echoed work because the message echoing gets truncated, and some things can happen in a synchronous context, which came from data from an async context. Balloons are probably similar, knowing how Vim works.

I think the "nice" way to make this work if to call the whole `ale#balloon#MessageForPos()` asynchronously from `balloonexpr` function (this is what termdebug does). Currently, the balloonexpr either return directly a string (when there's a diagnostic) or calls the hover method that will create an async job which will eventually call balloon_show *after* balloonexpr returned `''` in `ale#balloon#MessageForPos()`.

But as far as I understand, it's not possible to create a job for vim functions, so I think the easy way is to leave it this way, and hope that async support for vim functions will happen some day (neovim can already create children processes can't it ?)

PR #1565, opened 2018-05-09
Improve g:ale_set_balloons default value

rhysd:
Vim help says:

```
m  *+balloon_eval*	|balloon-eval| support in the GUI. Included when
			compiling with supported GUI (Motif, GTK, GUI) and
			either Netbeans/Sun Workshop integration or |+eval|
			feature.
H  *+balloon_eval_term*	|balloon-eval| support in the terminal,
			'balloonevalterm'
```

It means `balloon_eval` is for GUI, and `balloon_eval_term` should be referred on CUI instead. I changed the default value of `g:ale_set_balloons` considering that.

### DIFF
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -1350,7 +1350,8 @@ g:ale_set_balloons                                         *g:ale_set_balloons*
                                                            *b:ale_set_balloons*
 
   Type: |Number|
-  Default: `has('balloon_eval')`
+  Default: `has('balloon_eval') && has('gui_running') ||`
+           `has('balloon_eval_term') && !has('gui_running')`
 
   When this option is set to `1`, balloon messages will be displayed for
   problems. Problems nearest to the cursor on the line the cursor is over will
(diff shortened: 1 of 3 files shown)


w0rp:
I'll have to try running the tests for this myself. It might break some tests on Travis.

rhysd:
Build seemed not started. Could you restart the job?

https://travis-ci.org/w0rp/ale/builds/376853433

rhysd:
I force pushed 7f74db8 as 57f0454 for restarting CI job. Nothing was changed (only ammending).

rhysd:
I updated test cases also.

w0rp:
Cheers! :beers:

w0rp:
I unfortunately must disable balloon support for terminals by default, due to a Vim bug which causes very strange behavior when using `balloonexpr` in combination with `ttymouse` being set to the wrong setting. We might be able to enable terminal balloon support by default later in newer versions of Vim by checking the patch number with `has()`.

w0rp:
See #1631

PR #1573, opened 2018-05-11
Run sass-lint from the target's directory

zed0:
Running sass-lint from the directory of the target file causes the local linter config to be used (if it exists) rather than the global config.

> * If you add or modify a function for computing a command line string for
>   running a command, please add Vader tests for that.

Can someone point me in the direction of a Vader test that does this already?  I'm not familiar with vimscript so I'll have a very hard time without something to follow.

### DIFF
--- a/ale_linters/sass/sasslint.vim
+++ b/ale_linters/sass/sasslint.vim
@@ -1,8 +1,18 @@
-" Author: KabbAmine - https://github.com/KabbAmine
+" Author: KabbAmine - https://github.com/KabbAmine, Ben Falconer
+" <ben@falconers.me.uk>
+
+function! ale_linters#sass#sasslint#GetCommand(buffer) abort
+    return ale#path#BufferCdString(a:buffer)
+    \   . ale#Escape('sass-lint')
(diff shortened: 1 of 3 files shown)


ale_linters/sass/sasslint.vim
```diff
@@ -3,6 +3,6 @@
 call ale#linter#Define('sass', {
 \   'name': 'sasslint',
 \   'executable': 'sass-lint',
-\   'command': 'sass-lint -v -q -f compact %t',
```
w0rp: That won't work on Windows. Look at how other linters use `BufferCdString`. `grep` for it.

w0rp:
I just added this sentence to the pull request template, which should help you.

>Look at existing tests in the test/command_callback directory, etc.

zed0:
I've added a basic test and changed how the directory is changed as suggested.
I'm not sure how to have tests for both versions of this linter and maintain the naming scheme for the tests given they are both called `sasslint`.

w0rp:
Cheers! :beers:

PR #1574, opened 2018-05-11
Support Elm 0.19

hecrj:
The Elm 0.19 alpha period started yesterday. Here are the main relevant changes:

  * The compiler is significantly faster.
  * There is now just one elm binary at the command line. `elm-make` is `elm make` now.
  * The output when using `--report=json` has changed. [Here](https://github.com/elm-lang/project-metadata-utils/blob/master/src/Elm/Error.elm) is a specification of the new format.

I have modified the current linter to support these changes, while also adding some tests. I hope this PR can help to get ALE ready for Elm 0.19 once the alpha period finishes :)

There is a main concern: these changes break support for Elm 0.18. How should we deal with that? Different linters? Configurable options? Use `elm version` to support both?

Finally, I want to thank you all for this amazing plugin.

### DIFF
--- a/ale_linters/elm/make.vim
+++ b/ale_linters/elm/make.vim
@@ -22,6 +22,10 @@ function! ale_linters#elm#make#Handle(buffer, lines) abort
                 " General problem
                 let l:details = ale_linters#elm#make#ParseMessage(l:report.message)
 
+                if empty(l:report.path)
+                    let l:report.path = 'Elm.Kernel'
+                endif
+
                 if ale_linters#elm#make#FileIsBuffer(l:report.path)
                     call add(l:output, {
                                 \    'lnum': 1,
@@ -36,7 +40,6 @@ function! ale_linters#elm#make#Handle(buffer, lines) abort
                                 \    'detail': l:report.path ." ----------\n\n". l:details,
                                 \})
                 endif
-
             else
                 " Compilation errors
                 for l:error in l:report.errors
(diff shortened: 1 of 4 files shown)


hecrj:
I have added support for Elm 0.18. The linter is now backwards compatible!

ale_linters/elm/make.vim
```diff
@@ -57,23 +38,142 @@ function! ale_linters#elm#make#Handle(buffer, lines) abort
...
+function! ale_linters#elm#make#HandleElm018Line(line, output) abort
+    let l:errors = json_decode(a:line)
+
+    for l:error in l:errors
+        let l:file_is_buffer = ale_linters#elm#make#FileIsBuffer(l:error.file)
+
+        if l:file_is_buffer
+            " Current buffer has problems
+            call add(a:output, {
+                        \    'lnum': l:error.region.start.line,
```
w0rp: To make lines shorter, put `\` at the same level of indentation, and just up-indent extra lines with one level of indentation. See the rest of the codebase for examples. This applies throughout all files.

w0rp:
I've been away for a little while. Sorry I didn't have the chance to check this until now.

hecrj:
> There's one part here which isn't quite backwards compatible, and that's changing elm-make to elm make.

The `elm` executable is also provided in Elm 0.18 when installed either globally or locally. `elm make` works just fine. You can see how they test its presence here: https://github.com/elm-lang/elm-platform/blob/0.18.0-exp/.travis.yml

I have been using this branch to work on different 0.18 and 0.19 projects without any issue so far.

Should we still do what you proposed anyways?

w0rp:
Cheers! :beers:

PR #1593, opened 2018-05-18
Add psalm linter for PHP

twitchard:
This adds the PHP linter psalm. Probably I should customize it to add specification of additional variables, but at least it's a start and would help anybody who uses this linter.


w0rp:
Closing due to inactivity.

PR #1599, opened 2018-05-21
Convert cygwin unix path to windows for eslint

kevinthedestroyr:
If cygpath in user path, assume on cygwin and use it to convert to windows path before passing to eslint. This is necessary because eslint runs in cmd outside of cygwin and doesn't understand unix-like cygwin path.


kevinthedestroyr:
This was hasty -- going back to fix + add tests.

w0rp:
Just so you're aware, pull requests with `system` calls won't be accepted, as they block Vim. If ALE is to support cygwin, the support would have to be pretty comprehensive. I tend to think that cygwin isn't worth supporting, and supporting the "Windows Subsystem for Linux" is a better option.

PR #1600, opened 2018-05-21
added vimwiki as filetype using markdown linters

gtheys:
I read the contribution.md and documentation for Ale. Saw the section how to add Linters, but couldn't figure out and easy way to add another filetype.

I copied the markdown linters and setup so it will support vimwiki file format. I kept the original linters author as I didn't really write anything.

If there is a better way to add another filetype to the markdown linters please let me know.

### DIFF
--- a/ale_linters/vimwiki/alex.vim
+++ b/ale_linters/vimwiki/alex.vim
@@ -1,0 +1,11 @@
+" Author: Johannes Wienke <languitar@semipol.de>
+" Description: alex for markdown files
+
+call ale#linter#Define('vimwiki', {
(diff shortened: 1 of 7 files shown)


w0rp:
Yep, there is a better way to handle this particular type of issue. Filetypes can be aliased to other filetypes, including a range of filetypes, including the original filetype itself. I'll update the default aliases.

w0rp:
See the commit I added now, so you can see how I handled this.

gtheys:
Ok seems I missed that. Thanks for giving this feedback!

PR #1617, opened 2018-05-29
Fix for puppet version >= 5.4.0

luitzifa:
I just have 5.4.0 installed, so i cannot really tell since when the output has changed. But i can tell this change works with older versions as well.


w0rp:
This looks okay to me. I'll see if I can track down which version this was changed in first.

w0rp:
Cheers! :beers:

Issue #340, opened 2017-02-13
`mix compile` does not provide messages for syntax errors


Wexcode:
[credo](https://github.com/rrrene/credo) only provides style warnings, but no syntax errors. Syntax errors can be generated through running `mix compile`.

A quick side note on the credo linter - `mix` tasks only work if you call them from the project root. As a result, this linter only works if you open vim from the project root.

w0rp:
Is there some way to detect the location of the project root? If that can be done, then ALE could always run the command from that directory.

w0rp:
Do you have an example of what a syntax error looks like, and some of the normal errors? Some Vader tests could be added to cover that, and that could be fixed.

Wexcode:
> Is there some way to detect the location of the project root? If that can be done, then ALE could always run the command from that directory.

Yeah, traverse up the structure until you find the file `mix.exs`.

> Do you have an example of what a syntax error looks like, and some of the normal errors? Some Vader tests could be added to cover that, and that could be fixed.

```
== Compilation error on file {relative_path_to_file.ex} ==
** (SyntaxError) {relative_path_to_file.ex}:{line_no}: {msg}
    {stackTrace}
```

w0rp:
I'll take a look later. Could you provide some example output for errors and warnings? I'm after some real output I can add to the Vader test suite. That's the best way to fix issues like this and check for regressions.

anhari:
Some example output (in code blocks) for errors:

Example error: missing an `end` statement
```
mix: missing terminator: end (for "do" starting at line 1)
```

Example error: unused variable
```
mix: variable users is unused
```

Example error: typing `Repo.inset` instead of `Repo.insert` (using phoenix/ecto)
```
mix: function Rumbl.Repo.inset/1 is undefined or private
```

w0rp:
Thanks for the example output. :+1: If someone wants to give this a go, go for it. I realised this is actually not a bug in the current linter, but a request for another one.

Changing to the project root for running the linter shouldn't be hard. There is a utility function which will help with this.

```vim
let l:project_dir = fnamemodify(ale#util#FindNearestFile(a:buffer, 'mix.exs'), ':h')

return 'cd ' . fnameescape(l:project_dir) . ' && mix compile ...'
```

I'm not sure about the other details. Implementing #333 might be required for this, maybe not.

archSeer:
[...]

PR #1618, opened 2018-05-29
[new linter] Add mix linter for elixir

colbydehart:
This deals with #340 by adding a mix linter for elixir to get syntax errors. Used @evnu 's example in that issue, but using mix instead of elixirc so that dependencies won't crash the linter.

### DIFF
--- a/ale_linters/elixir/mix.vim
+++ b/ale_linters/elixir/mix.vim
@@ -30,8 +30,16 @@ function! ale_linters#elixir#mix#Handle(buffer, lines) abort
     return l:output
 endfunction
 
+function! ale_linters#elixir#mix#FindProjectRoot(buffer) abort
+    let l:project_root = ale#path#FindNearestFile(a:buffer, 'mix.exs')
+    if !empty(l:project_root)
+      return fnamemodify(l:project_root, ':h')
+    endif
+    return ''
+endfunction
+
 function! ale_linters#elixir#mix#GetCommand(buffer) abort
-    let l:project_root = fnamemodify(ale#path#FindNearestFile(a:buffer, 'mix.exs'), ':h')
+    let l:project_root = ale_linters#elixir#mix#FindProjectRoot(a:buffer)
 
     let l:temp_dir = ale#engine#CreateDirectory(a:buffer)
 
@@ -49,4 +57,5 @@ call ale#linter#Define('elixir', {
 \   'executable': 'mix',
 \   'command_callback': 'ale_linters#elixir#mix#GetCommand',
 \   'callback': 'ale_linters#elixir#mix#Handle',
+\   'lint_file': 1,
 \})
--- a/ale_linters/elixir/mix.vim
+++ b/ale_linters/elixir/mix.vim
@@ -44,8 +44,8 @@ function! ale_linters#elixir#mix#GetCommand(buffer) abort
     let l:temp_dir = ale#engine#CreateDirectory(a:buffer)
 
     let l:mix_build_path = has('win32')
-    \   ? 'set MIX_BUILD_PATH=' . ale#Escape(l:temp_dir) . ' && '
-    \   : 'MIX_BUILD_PATH=' . ale#Escape(l:temp_dir) . ' '
+    \   ? 'set MIX_BUILD_PATH=' . ale#Escape(l:temp_dir) . ' &&'
+    \   : 'MIX_BUILD_PATH=' . ale#Escape(l:temp_dir)
 
     return ale#path#CdString(l:project_root)
           \ . l:mix_build_path
--- a/ale_linters/elixir/mix.vim
+++ b/ale_linters/elixir/mix.vim
@@ -31,7 +31,7 @@ endfunction
 
 function! ale_linters#elixir#mix#Command(buffer) abort
     let l:project_dir = fnamemodify(ale#path#FindNearestFile(a:buffer, 'mix.exs'), ':h')
-    return  'cd ' . l:project_dir . ' && mix compile %s'
+    return  'cd ' . l:project_dir . ' && MIX_BUILD_PATH=/tmp/mix mix compile %s'
 endfunction
 
 call ale#linter#Define('elixir', {
--- a/ale_linters/elixir/mix.vim
+++ b/ale_linters/elixir/mix.vim
@@ -1,0 +1,42 @@
+" Author: evnu - https://github.com/evnu
+" Author: colbydehart - https://github.com/colbydehart
+
+function! ale_linters#elixir#mix#Handle(buffer, lines) abort
+    " Matches patterns like the following:
+    "
+    " Error format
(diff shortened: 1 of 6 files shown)


evnu:
Does the mix command place the resulting .beam file into `_build`? If so, that might surprise users when running an interactive shell in parallel. Maybe `:build_path` can be overriden.

(thanks for picking up that linter! :))

evnu:
> is there a way to configure that?

I spent some time a while ago to find such an option, but did not find any. It seems that the current Elixir master contains an environment variable to configure the output directory (`MIX_BUILD_PATH`, [see](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/project.ex#L433)), but this environment variable is not part of the current Elixir version v1.6.5 ([see](https://github.com/elixir-lang/elixir/blob/v1.6/lib/mix/lib/mix/project.ex#L420)).  I guess that it will be possible in the future to override the build path with this convenient environment variable, but for now this does not work. I tried to override the build path with the following script `override.exs` within a Mix project, but it I could not make it work either. It seems very brittle to even attempt this. For future reference, this is what I tried:

```elixir
#!/usr/bin/env elixir

:ok = Mix.start()

# Ensure that everything we might need is loaded
Mix.CLI.main()

# Ensure that we are within a mix project
Mix.Project.get!()
config = Mix.Project.config()

updated_config =
  Keyword.merge(
    config,
    build_path: "/tmp/foo/_bar"
  )

# FIXME The args should refer to the file being compiled.
args = []

# FIXME This path might need to be adapted
path = "."

# We need to pop the previously loaded project to allow reloading it
Mix.Project.pop()
Mix.Project.in_project(updated_config[:app], path, updated_config, fn module ->
  Mix.Project.compile(args)
end)
```

Note that the `:build_path` option is intended to be used for child apps of an umbrella and possibly does not work correctly in other cases.

evnu:
Maybe. It would be good to check against Elixir's master Branche (e.g. build it woth kex).

ale_linters/elixir/mix.vim
```diff
@@ -0,0 +1,42 @@
...
+        \   'text': l:text,
+        \})
+    endfor
+
+    return l:output
+endfunction
+
+function! ale_linters#elixir#mix#Command(buffer) abort
+    let l:project_dir = fnamemodify(ale#path#FindNearestFile(a:buffer, 'mix.exs'), ':h')
+    return  'cd ' . l:project_dir . ' && MIX_BUILD_PATH=/tmp/mix mix compile %s'
```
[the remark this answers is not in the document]
colbydehart: Cool  made those changes, also found out about the `ale#path#CdString` for properly changing directories while i was at it. Sorry about the errors this is my first vimscript PR, i'm a bit of a noob, but Vader is really neat!

colbydehart:
Cool  made those changes, also found out about the `ale#path#CdString` for properly changing directories while i was at it. Sorry about the errors this is my first vimscript PR, i'm a bit of a noob, but Vader is really neat!

ale_linters/elixir/mix.vim
```diff
@@ -0,0 +1,52 @@
...
+    return  ale#path#CdString(l:project_root)
+          \ . l:mix_build_path
+          \ . ' mix compile %s'
+endfunction
+
+call ale#linter#Define('elixir', {
+\   'name': 'mix',
+\   'executable': 'mix',
+\   'command_callback': 'ale_linters#elixir#mix#GetCommand',
+\   'callback': 'ale_linters#elixir#mix#Handle',
```
w0rp: Add `'lint_file': 1` here so this linter will only run when a file is opened, saved, etc.

ale_linters/elixir/mix.vim
```diff
@@ -0,0 +1,52 @@
...
+    let l:temp_dir = ale#engine#CreateDirectory(a:buffer)
+
+    let l:mix_build_path = has('win32')
+    \   ? 'set MIX_BUILD_PATH=' . ale#Escape(l:temp_dir) . ' && '
+    \   : 'MIX_BUILD_PATH=' . ale#Escape(l:temp_dir) . ' '
+
+    return  ale#path#CdString(l:project_root)
+          \ . l:mix_build_path
+          \ . ' mix compile %s'
+endfunction
```
w0rp: Add some tests for the output of this function here. See similar tests in `test/command_callback`.

colbydehart: sorry, been traveling. I couldn't find an example of how to test a command function where you make a temporary directory. Could you maybe point me in the right direction of how to do that?

colbydehart: I found one for javac and did what they did for testing here https://github.com/w0rp/ale/pull/1618/files#diff-9a91b3bcd6181a1601673b8b6c8999a6

ale_linters/elixir/mix.vim
colbydehart: I added this, it's just the line below.

colbydehart:
So i got testing the temp file working but now i can't figure out how to test the `CdString` portion of this in conjunction with `FindNearestFile` I made a dummy mix project but for some reason the command is not picking it up. This is what i'm doing in my before section
```
  " ./mix_paths/wrapped_project is the dummy project.
  call ale#test#SetFilename(ale#path#Simplify(g:dir .  '/mix_paths/wrapped_project/lib/main.ex'))

  let g:project_root = ale#path#Simplify(g:dir . '/mix_paths/wrapped_project')
```

Then when I am testing my command i am doing this: 
```
  AssertEqual
  \ GetCommand(),
  \ ale#path#CdString(g:project_root)
  \ . g:env_prefix
  \ . 'mix compile %s'
```
but for some reason the linter won't pick up the project file using `FindNearestFile` when running the test. Might be doing something wrong.

ale_linters/elixir/mix.vim
```diff
@@ -0,0 +1,61 @@
...
+function! ale_linters#elixir#mix#Handle(buffer, lines) abort
+    " Matches patterns like the following:
+    "
+    " Error format
+    " ** (CompileError) apps/sim/lib/sim/server.ex:87: undefined function update_in/4
+    "
+    " TODO: Warning format
+    " warning: variable "foobar" does not exist and is being expanded to "foobar()", please use parentheses to remove the ambiguity or change the variable name
+
+    let l:pattern = '\v\(([^\)]+Error)\) ([^:]+):([^:]+): (.+)$'
```
w0rp: I wouldn't be surprised if it turns out that we need to parse the filename from the output and handle it for reporting problems in other files. Let's just get people to try this and see.

w0rp:
Cheers! :beers:

colbydehart:
Oh thanks so much!

bbtdev:
Sorry noob here, how can I test this?

w0rp:
Talk to @colbydehart and install `mix` on your system.

PR #1625, opened 2018-06-03
Allow flake8 executable to be set to `pipenv`.

elebow:
It appends ` run flake8`, analogously to the Ruby tools when the executable is set to `bundle`


w0rp:
Cheers! :beers:

PR #1629, opened 2018-06-05
Allow all Python linter executables to be set to `pipenv`.

elebow:
It appends ` run {linter_name}`, analogously to the Ruby linters when the executable is set to `bundle`. Includes documentation update.


w0rp:
Nicely done. Cheers! :+1:

alok:
This may be easier to use if it checked that pipenv was active in some way (perhaps by using the exit code of `pipenv --venv`). Then it could be used out of the box on python files that aren't in a project, like a lot of scripts.

w0rp:
@alok That sounds like a better idea. I'll create an issue for it.

PR #1638, opened 2018-06-07
Use --vanilla switch for Rscript

ssiegel:
This prevents possibly bad interference with the user's R environment,
e.g. by an auto-activating packrat.


w0rp:
Cheers! :beers:

PR #1641, opened 2018-06-07
sh: add bash-language-server linter

docwhat:
I don't see any examples for how to add an LSP based fixer; feel free to point me at one.

This doesn't work without PR #1640

### DIFF
--- a/ale_linters/sh/language_server.vim
+++ b/ale_linters/sh/language_server.vim
@@ -1,0 +1,33 @@
+" Author: Christian Höltje (https://docwhat.org/)
+" Description: BASH Language server integration for ALE
+scriptencoding utf-8
+
+call ale#Set('sh_language_server_executable', 'bash-language-server')
+call ale#Set('sh_language_server_use_global', get(g:, 'ale_use_global_executables', 0))
+
+function! ale_linters#sh#language_server#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'sh_language_server', [
+    \   'node_modules/.bin/bash-language-server',
+    \])
+endfunction
+
+function! ale_linters#sh#language_server#GetCommand(buffer) abort
+    let l:exe = ale#Escape(ale_linters#sh#language_server#GetExecutable(a:buffer))
+
+    return l:exe . ' start'
+endfunction
+
+function! ale_linters#sh#language_server#GetProjectRoot(buffer) abort
+    let l:git_path = ale#path#FindNearestDirectory(a:buffer, '.git')
+
+    return !empty(l:git_path) ? fnamemodify(l:git_path, ':h:h') : ''
+endfunction
+
+call ale#linter#Define('sh', {
+\   'name': 'language_server',
+\   'lsp': 'stdio',
+\   'executable_callback': 'ale_linters#sh#language_server#GetExecutable',
+\   'command_callback': 'ale_linters#sh#language_server#GetCommand',
+\   'language': 'sh',
+\   'project_root_callback': 'ale_linters#sh#language_server#GetProjectRoot',
+\})
--- a/doc/ale-sh.txt
+++ b/doc/ale-sh.txt
@@ -3,6 +3,25 @@ ALE Shell Integration                                          *ale-sh-options*
 
 
 ===============================================================================
+sh-language-server                                     *ale-sh-language-server*
+
+g:ale_sh_language_server_executable        *g:ale_sh_language_server_executable*
+                                           *b:ale_sh_language_server_executable*
+  Type: |String|
+  Default: `'bash-language-server'`
+
+  See |ale-integrations-local-executables|
+
+
+g:ale_sh_language_server_use_global        *g:ale_sh_language_server_use_global*
+                                           *b:ale_sh_language_server_use_global*
+  Type: |Number|
+  Default: `get(g:, 'ale_use_global_executables', 0)`
+
+  See |ale-integrations-local-executables|
+
+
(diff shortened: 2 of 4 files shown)


doc/ale.txt
```diff
@@ -317,7 +318,7 @@ Notes:
 * API Blueprint: `drafter`
 * AsciiDoc: `alex`!!, `proselint`, `redpen`, `write-good`
 * Awk: `gawk`
-* Bash: `shell` (-n flag), `shellcheck`, `shfmt`
+* Bash: `language-server`, `shell` (-n flag), `shellcheck`, `shfmt`
```
w0rp: Add this to `README.md` too. Have a look at the output from Travis, which complains about some things like this.

doc/ale-sh.txt
```diff
@@ -2,6 +2,25 @@
 ALE Shell Integration                                          *ale-sh-options*
 
 
+===============================================================================
+sh-language-server                                     *ale-sh-language-server*
+
+g:ale_sh_language_server_executable        *g:ale_sh_language_server_executable*
+                                          *b:ale_sh_language_server_executable*
```
w0rp: Make sure to align tags to the right margin.

ale_linters/sh/language_server.vim
```diff
@@ -0,0 +1,34 @@
...
+function! ale_linters#sh#language_server#GetCommand(buffer) abort
+    let l:exe = ale#Escape(ale_linters#sh#language_server#GetExecutable(a:buffer))
+
+    return l:exe . ' start'
+endfunction
+
+function! ale_linters#sh#language_server#GetProjectRoot(buffer) abort
+    let l:git_path = ale#path#FindNearestDirectory(a:buffer, '.git')
+
+    return !empty(l:git_path) ? fnamemodify(l:git_path, ':h:h') : '.'
```
w0rp: Be aware that `'.'` will start one server per file on the system, and it will keep running until Vim closes or you use `:ALEStopAllLSPs`.

ale_linters/sh/language_server.vim
```diff
@@ -0,0 +1,34 @@
...
+" Description: BASH Language server integration for ALE
+scriptencoding utf-8
+
+call ale#Set('sh_language_server_executable', 'bash-language-server')
+call ale#Set('sh_language_server_use_global', get(g:, 'ale_use_global_executables', 0))
+
+function! ale_linters#sh#language_server#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'sh_language_server', [
+    \   'node_modules/.bin/bash-language-server',
+    \   'bash-language-server',
```
w0rp: You can remove this line here. These strings are for paths to search upwards for. The global executable will be used if a local one can't be found.

ale_linters/sh/language_server.vim
docwhat: Is there something better I should do to handle files that aren't in a "project"?

I'm still new to this whole LSP thing.

ale_linters/sh/language_server.vim
docwhat: done

doc/ale-sh.txt
docwhat: done

doc/ale.txt
docwhat: done in a7ec36d

ale_linters/sh/language_server.vim
w0rp: I suppose it could use some arbitrary directory for the project, like `/`.

docwhat: I honestly don’t understand what the root is for.

docwhat: > Be aware that '.' will start one server per file on the system, and it will keep running until Vim closes or you use :ALEStopAllLSPs.

Is that one server per _root_ on the system?  e.g. If you edit different files in the same directory would it be one server?   Or is it really per-file?

What happens if the file is outside the _root_?  e.g. if we use `$HOME` and the file is in `/etc/`?

I could write some code that looks if the file is in `$HOME` and use that or else use `/`.

w0rp: The issue is that ALE runs one instance of an LSP server per project, and you typically have to establish where a project is. I suppose `'.'` might so happen to work where the path is kind of ignored and one server is started for all files not in git repositories. LSP is designed to work with projects which have some project root path, instead of individual files spread across the file system.

For files which aren't in any project in particular, the best thing I can think of is just to set the project root to be the root of the filesystem the file is in.

docwhat:
@w0rp After some experimenting it seems that the "project root" is more important for some language-servers than others.

I think it is used to scan for classes, methods, etc. that can be used in the file that is being edited.

Using either `.` or `/` are _bad ideas_ since the language-server may be unresponsive as it scans too many files or because the returned list of suggestions may be **way** too big.

So I switched it back to just `''`.

The reason I thought it was more important was because at least one of the language-servers (I forget which) just flat-out didn't work if it the file-under-edit wasn't in a project root (i.e. it would exit immediately).

I think I'll file a bug to add something like a "language-server" status in the `:ALEInfo` information.  Is there a "ping" command you can send the language-server to ask it if it is up and running?

w0rp:
ALE keeps track of running processes, so we can see if the process is still running or not.

w0rp:
Cheers! :beers:

docwhat:
I don’t think that is an ALE problem...

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1651, opened 2018-06-12
Add support for qmlfmt fixer

pylipp:
Is it ok that `g:ale_qml_qmlfmt_executable` is also defined in
`ale_linters/qml/qmlfmt.vim`?


w0rp:
Cheers! :beers:

PR #1665, opened 2018-06-20
 Make CloudFormation linter work again with latest version

ashemedai:
As mentioned in the commit, I opted to not be smarter than ALE itself. Since ALE explicitly converts `lnum`, `col`, `end_lnum`, and `end_col` using `str2nr()`, let that do the heavy lifting. I opted to simply adjust the test to account for this by having the variables be strings, as they get out of the regexp matcher groups.

### DIFF
--- a/ale_linters/cloudformation/cfn_python_lint.vim
+++ b/ale_linters/cloudformation/cfn_python_lint.vim
@@ -4,8 +4,8 @@
 function! ale_linters#cloudformation#cfn_python_lint#Handle(buffer, lines) abort
     " Matches patterns line the following:
     "
-    " sample.template.yaml:96:7:96:15: [E3012] Property Resources/Sample/Properties/FromPort should be of type Integer
-    let l:pattern = '\v^(.*):(\d+):(\d+):(\d+):(\d+): \[([[:alnum:]]+)\] (.*)$'
+    " sample.template.yaml:96:7:96:15:E3012:Property Resources/Sample/Properties/FromPort should be of type Integer
+    let l:pattern = '\v^(.*):(\d+):(\d+):(\d+):(\d+):([[:alnum:]]+):(.*)$'
     let l:output = []
 
     for l:match in ale#util#GetMatches(a:lines, l:pattern)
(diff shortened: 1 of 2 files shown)


ale_linters/cloudformation/cfn_python_lint.vim
```diff
@@ -4,22 +4,22 @@
 function! ale_linters#cloudformation#cfn_python_lint#Handle(buffer, lines) abort
     " Matches patterns line the following:
     "
-    " sample.template.yaml:96:7:96:15: [E3012] Property Resources/Sample/Properties/FromPort should be of type Integer
-    let l:pattern = '\v^(.*):(\d+):(\d+):(\d+):(\d+): \[([[:alnum:]]+)\] (.*)$'
+    " sample.template.yaml:96:7:96:15:E3012:Property Resources/Sample/Properties/FromPort should be of type Integer
+    let l:pattern = '\v^(.*):(\d+):(\d+):(\d+):(\d+):([[:alnum:]]+):(.*)$'
```
w0rp: Has the output changed in a newer version of the linters? If so, put both the old and new formats in the tests, and ALE should support both, so both older and newer versions work.

ashemedai: Yes it has changed to a more parsable output from 0.2.2 onwards and should be considered stable now.

Given the relative age of the project (~2 months) I am not sure it is worthwhile to support the old format, but I can try to make it work if you want. Do you know from the top of your head another linter in ALE that does this?

w0rp: There are many linters where both the older and newer formats are supported, but I didn't realise the project was so young. I think it's fine to just switch to the new format at this stage, then.

w0rp:
Cheers! :beers:

ashemedai:
Thanks for merging.

And yes, I happen to be on the frontline here since my colleague actively helped the AWS folks kick it off and maintain (including the new linter format). :)

kevinold:
How do I configure this in my .vimrc?

ashemedai:
My current way might not be the best way to configure it (might need to add some detection if `AWSTemplateFormatVersion` is present), but in `.vim/after/ftplugin/yaml.vim` I have:

```vim
let b:ale_linter_aliases = {'yaml': ['cloudformation', 'yaml']}
```

You might even want to drop the second yaml if you only care about the CloudFormation linting.

PR #1672, opened 2018-06-23
Use the same options for all Clang tools

lukipuki:
The Clang compiler (and GCC) both use C++14 and -Wall, so clang-tidy and
clang-check should have the same options.

### DIFF
--- a/ale_linters/cpp/clangcheck.vim
+++ b/ale_linters/cpp/clangcheck.vim
@@ -2,7 +2,7 @@
 " Description: clang-check linter for cpp files
 
 call ale#Set('cpp_clangcheck_executable', 'clang-check')
-call ale#Set('cpp_clangcheck_options', '')
+call ale#Set('cpp_clangcheck_options', '-- -std=c++14 -Wall')
 call ale#Set('c_build_dir', '')
 
 function! ale_linters#cpp#clangcheck#GetExecutable(buffer) abort
--- a/ale_linters/cpp/clangtidy.vim
+++ b/ale_linters/cpp/clangtidy.vim
@@ -7,7 +7,7 @@ call ale#Set('cpp_clangtidy_executable', 'clang-tidy')
 call ale#Set('cpp_clangtidy_checks', ['*'])
 " Set this option to manually set some options for clang-tidy.
 " This will disable compile_commands.json detection.
-call ale#Set('cpp_clangtidy_options', '')
+call ale#Set('cpp_clangtidy_options', '-std=c++14 -Wall')
 call ale#Set('c_build_dir', '')
 
 function! ale_linters#cpp#clangtidy#GetExecutable(buffer) abort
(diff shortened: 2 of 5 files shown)


w0rp:
This isn't the right thing to do. Both `clangcheck` and `clang-tidy` read `compile_commands.json` files, and use the arguments found in those files.

lukipuki:
Please reopen.

`compile_commands.json` doesn't always exist, e.g. when editing a single file.

lukipuki:
Why does `clang` have default options then? See https://github.com/lukipuki/ale/blob/master/doc/ale-cpp.txt#L29

w0rp:
`clang` doesn't parse `compile_commands.json`. II'm not merging this.

lukipuki:
I still don't follow this logic.

This is what I did today:
* Stumbled upon ALE. Let's try it out!
* Wrote a simple program using C++14: http://paste2.org/cHdhX62M
* Got a bunch of warnings. Fixed the options for `clang-tidy` and `clang-check`.
* Wondered why `clang` didn't need its options fixed. Discovered it already has `-std=C++14` supplied.
* I've fixed the inconsistencies and created a pull request.

Not fixing this has a few drawbacks:
* Users writing simple one-file programs will get the wrong linter messages. I often do this at work outside of large projects to do simple mock-ups.
* Users checking that their ALE installation works out-of-the-box (using a simple one-file C++ program) will see inconsistent behavior between `clang`/`g++` and `clang-tidy`/ `clang-check`. It gives a bad impression about the project.

There are simple cases where there's no `compile_commands.json` available but they are still useful to users. Even your `README.md` says
> Some tools and build configurations can generate compile_commands.json files.

Why support only **some** tools and build configurations? Why not a few more?

w0rp:
Those tools support `compile_commands.json`, and `clang` is just a compiler. There are other users using these tools today who wouldn't be happy if the options were suddenly changed. At the very least, we'd need to give people time to update their vimrc files.

Setting the C++ standard and the warning level for `clangcheck` and `clang-tidy` isn't the right thing to do. You should configure projects using `compile_commands.json` files. The options in ALE are there if you want to override them. You can't impose the options you think are best on everyone else.

w0rp:
Another way of looking at this is, suppose someone has a C++ project with a `compile_commands.json` file which sets `--std=C++17`. Suppose I merge this code. Their experience will be that syntax which is valid in their project is no longer considered valid. `clang` and `gcc` require configuration in Vim, but `clang-tidy` and `clangcheck` ought not to require any configuration in Vim, but in project files instead. This is how other linters like `eslint` work.

lukipuki:
Aha, now I see that the underlying issue is that settings in `.vimrc` override project-specific settings. IMHO that's poor design leading to bad user experience (such as those I've described above). Unfortunately also YouCompleteMe does the same, where `~/.ycm_extra_conf.py` overrides `~/code/project/compile_commands.json`.

I've worked around the issues I had by updating to Clang 6, which has C++14 enabled by default.

PR #1675, opened 2018-06-25
Improve pyrex cython linter.

nicopauss:
Like many other linters, use variables for the executable and options
used by the linter.
By default, the linter now report every warnings as errors with
`--warning-errors`.
Also add include directory and set working directory to file directory.

### DIFF
--- a/ale_linters/pyrex/cython.vim
+++ b/ale_linters/pyrex/cython.vim
@@ -1,10 +1,27 @@
-" Author: w0rp <devw0rp@gmail.com>
+" Author: w0rp <devw0rp@gmail.com>,
+" Nicolas Pauss <https://github.com/nicopauss>
 " Description: cython syntax checking for cython files.
+
+call ale#Set('pyrex_cython_executable', 'cython')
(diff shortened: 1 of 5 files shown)


ale_linters/pyrex/cython.vim
```diff
@@ -1,10 +1,27 @@
-" Author: w0rp <devw0rp@gmail.com>
+" Author: w0rp <devw0rp@gmail.com>,
+" Nicolas Pauss <https://github.com/nicopauss>
 " Description: cython syntax checking for cython files.
 
+call ale#Set('pyrex_cython_executable', 'cython')
+call ale#Set('pyrex_cython_options', '--warning-extra --warning-errors')
```
w0rp: Could you remove `--warning-errors` here, so we can preserve the current behavior? If users want warnings to be errors, they can change the option themselves. I don't want to change the behavior without a very compelling reason, and whether warnings should be errors is subjective.

nicopauss: Done.

nicopauss:
Done

w0rp:
Good stuff. :+1: Cheers! :beers:

Issue #1680, opened 2018-06-26
Python prospector: E474 Attempt to decode blank string


fennerm:
## Information
Prospector linter is raising error when no warnings are present in file. Easy fix, I'll submit a PR in a few minutes.

**VIM version**
NVIM v0.3.0
Build type: Release

Operating System: WHAT OS WERE YOU USING?
Arch Linux

w0rp:
This was merged ages ago, but the issue was never closed.

Issue #779, opened 2017-07-19
tslint: E474: Attempt to decode a blank string


justanr:
Not sure if I have everything setup correctly, typescript, tslint and eslint were all installed with `npm i -g` on both 7.9 and 8.1 (this fails in both).

The full message:

```
Error detected while processing function <SNR>172_NeoVimCallback[29]..<SNR>167_HandleExit[40]..ale_linters#typescript#tslint#Handle:
line    3:
E474: Attempt to decode a blank string
```

And output of `:ALEInfo`

```
Current Filetype: typescript
Available Linters: ['eslint', 'tslint', 'tsserver', 'typecheck']
  Enabled Linters: ['eslint', 'tslint', 'tsserver', 'typecheck']
 Linter Variables:
let g:ale_typescript_tslint_config_path = ''
let g:ale_typescript_tslint_executable = 'tslint'
let g:ale_typescript_tslint_use_global = 0
let g:ale_typescript_tsserver_config_path = ''
let g:ale_typescript_tsserver_executable = 'tsserver'
let g:ale_typescript_tsserver_use_global = 0
 Global Variables:
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'E'
let g:ale_echo_msg_format = '[%linter%] %s [%severity%]'
let g:ale_echo_msg_warning_str = 'W'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'always'
let g:ale_linter_aliases = {}
let g:ale_linters = {}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_column_always = 1
let g:ale_sign_error = '>>'
let g:ale_sign_offset = 1000000
let g:ale_sign_warning = '--'
let g:ale_statusline_format = ['%d error(s)', '%d warning(s)', 'OK']
let g:ale_warn_about_trailing_whitespace = 1
  Command History:
[...]

PR #1682, opened 2018-06-27
Fix prospector empty string error

fennerm:
Prospector linter is raising error when no warnings are present in file
(#1680). Copied fix from #779.


w0rp:
Cheers! :beers:

PR #1691, opened 2018-07-01
add pyre lsp linter to python linters

dsifford:
**Note:** This is my first time contributing here, so please don't hesitate to scrutinize this and call out anything I may have done improperly.

This PR simply adds support for Facebook's new [`pyre-check`](https://github.com/facebook/pyre-check/graphs/contributors) LSP linter for python 3. ([background story here](https://www.facebook.com/notes/protect-the-graph/pyre-fast-type-checking-for-python/2048520695388071/))

The additions are pretty much a clean copy of the work done by @aurieh in his `pyls` contribution, with small modifications to instead invoke `pyre persistent` instead.

I've tested this locally on a project I'm using and it seems to work fine. However, I'm a brand new `pyre` user, so I haven't yet tested any crazy configuration modifications etc so there may be something missing still (Happy to submit a PR to fix these issues if/when I run into them).

Anyway, let me know if I missed anything or if there's anything else I need to do and I'd be happy to make the changes.

Thanks for the great plugin!

### DIFF
--- a/ale_linters/python/pyre.vim
+++ b/ale_linters/python/pyre.vim
@@ -1,0 +1,29 @@
+" Author: dsifford <dereksifford@gmail.com>
+" Description: A performant type-checker supporting LSP for Python 3 created by Facebook
+
+call ale#Set('python_pyre_executable', 'pyre')
+call ale#Set('python_pyre_use_global', get(g:, 'ale_use_global_executables', 0))
+
+function! ale_linters#python#pyre#GetExecutable(buffer) abort
+    return ale#python#FindExecutable(a:buffer, 'python_pyre', ['pyre'])
+endfunction
+
+function! ale_linters#python#pyre#GetCommand(buffer) abort
+    let l:executable = ale_linters#python#pyre#GetExecutable(a:buffer)
+
+    let l:exec_args = l:executable =~? 'pipenv$'
+    \   ? ' run pyre persistent'
+    \   : ' persistent'
+
+    return ale#Escape(l:executable) . l:exec_args
+endfunction
+
+call ale#linter#Define('python', {
+\   'name': 'pyre',
+\   'lsp': 'stdio',
+\   'executable_callback': 'ale_linters#python#pyre#GetExecutable',
+\   'command_callback': 'ale_linters#python#pyre#GetCommand',
+\   'language': 'python',
+\   'project_root_callback': 'ale#python#FindProjectRoot',
+\   'completion_filter': 'ale#completion#python#CompletionItemFilter',
+\})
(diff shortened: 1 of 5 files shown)


dsifford:
**Note:** I think the travis build cache is failing the build because the error is referring to a doc tag that doesn't appear to exist anymore in my branch. (specifically `ale-python-aardvark`)

w0rp:
Don't worry about that failing test. If you rebase your commits, that will go away.

This is pretty exciting. I've been trying to use `mypy` at work off and on, and one of the main issues with it is that it is very, very slow. This might end up being useful for me at work.

w0rp:
Because this is the first case where there are multiple LSP connections that could be made for one language, I'll have to think about how I can fix some instances where LSP features select the first LSP linter available, like completion, go to definition, etc.

ale_linters/python/pyre.vim
```diff
@@ -0,0 +1,29 @@
...
+
+function! ale_linters#python#pyre#GetCommand(buffer) abort
+    let l:executable = ale_linters#python#pyre#GetExecutable(a:buffer)
+
+    let l:exec_args = l:executable =~? 'pipenv$'
+    \   ? ' run pyre persistent'
+    \   : ' persistent'
+
+    return ale#Escape(l:executable) . l:exec_args
+endfunction
```
w0rp: Could you add some tests for these functions? Have a look at similar tests in the `test/command_callback` directory.

dsifford: Sure thing. I'll take a look more into this tomorrow afternoon(ish). I'll probably have to chime back here in this thread though at some point with questions since I've never actually used vader before (even though I've had it installed for like almost two years now :stuck_out_tongue_closed_eyes:)

dsifford:
@w0rp I gave it a shot (again, by mostly just copying the pyls test, since this is essentially ran in the exact same way). 

The tests are now all passing, but for whatever reason the test command is exiting with a 1 status, so that's why travis is still reporting to be failed.

Let me know what else you need.

w0rp:
Don't worry about Travis CI failing there. I recently switched from running tests in NeoVim 0.1.7 to NeoVim 0.2.0 and 0.3.0, and for whatever reason, the tests fail to run at all in 0.3.0 some times. I'm still trying to figure out how to fix that. For now, I just restart some builds that fail now and then.

w0rp:
Cheers! :beers:

I'll think about how to handle sending messages to multiple LSP servers later.

dsifford:
FYI: Not sure what happened, but I'm fairly certain that a recent pyre update made this linter now only work for one trigger and then crash (on my machine, maybe others too?).

So for whoever might also be running into this and coming here to see what's up: I have no idea.

PR #1697, opened 2018-07-04
puppet: Add puppet-languageserver linter

ananace:
Since ALE seems to have some other LSPs, I figure adding support for Puppet as well won't hurt.

And it means I don't have to run multiple linters, which is always nice.

### DIFF
--- a/ale_linters/puppet/languageserver.vim
+++ b/ale_linters/puppet/languageserver.vim
@@ -14,15 +14,25 @@ function! ale_linters#puppet#languageserver#GetCommand(buffer) abort
 endfunction
 
 function! ale_linters#puppet#languageserver#GetProjectRoot(buffer) abort
-    " Note: while manifest.json is a strong recommendation, the only
-    " *required* path for a Puppet module is the manifests folder.
+    " Note: The metadata.json file is recommended for Puppet 4+ modules, but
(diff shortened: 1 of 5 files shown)


ale_linters/puppet/languageserver.vim
```diff
@@ -0,0 +1,35 @@
...
+
+    return !empty(l:root_path) ? fnamemodify(l:root_path, ':h') : ''
+endfunction
+
+call ale#linter#Define('puppet', {
+\   'name': 'languageserver',
+\   'lsp': 'stdio',
+\   'executable_callback': 'ale_linters#puppet#languageserver#GetExecutable',
+\   'command_callback': 'ale_linters#puppet#languageserver#GetCommand',
+\   'language': 'puppet',
```
w0rp: The language is optional now, and defaults to the filetype given above, but having it here won't hurt.

ananace:
I just realized that the language server itself probably has the least use for the path detection - it contains root detection code after all, but at the same time all of the Puppet linters should really share path detection code.

Both the Puppet parser and puppet-lint have support for doing name-based autoloading, which requires them to be run from the correct module path - the folder above the project root. Though I'm not going to do that de-duplication as part of the scope of this PR.

w0rp:
Cheers! :beers:

PR #1698, opened 2018-07-04
add options variable to isort

dsifford:
This PR simply adds the ability to pass options to isort.

This is particularly handy if also using the black linter (see https://github.com/ambv/black#the-black-code-style)

In short, black recommends the following:

```
[settings]
multi_line_output=3
include_trailing_comma=True
force_grid_wrap=0
combine_as_imports=True
line_length=88
```

which equates to the following from the command line:

```
$ isort --multi-line=3 --trailing-comma --force-grid-wrap=0 --combine-as --line-width=88 [ file.py ]
```

I updated the docs as well. Let me know if there's something I missed.

### DIFF
--- a/doc/ale-python.txt
+++ b/doc/ale-python.txt
@@ -145,6 +145,14 @@ g:ale_python_isort_executable                   *g:ale_python_isort_executable*
   See |ale-integrations-local-executables|
 
 
+g:ale_python_isort_options                      *g:ale_python_isort_options*
+                                                *b:ale_python_isort_options*
+  Type: |String|
+  Default: `''`
+
+  This variable can be set to pass extra options to isort.
+
+
 g:ale_python_isort_use_global                   *g:ale_python_isort_use_global*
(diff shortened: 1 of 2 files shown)


dsifford:
You might be wondering "why not just use a global `isort.cfg`"?

Because I am vehemently against any config files that do not comply with the XDG base directory specification. I refuse to put those sorts of files in my home directory.

dsifford:
Should be good to go. Let me know if I missed anything

doc/ale-python.txt
```diff
@@ -145,6 +145,14 @@ g:ale_python_isort_executable                   *g:ale_python_isort_executable*
   See |ale-integrations-local-executables|
 
 
+g:ale_python_isort_options                      *g:ale_python_isort_options*
```
w0rp: Travis complained about the alignment of the doctag here. I'll fix it.

dsifford: sorry about that!

Issue #1296, opened 2018-01-17
Add support for checking Vue files with vls


Dcryfm:
## Information

**neovim version**: 0.2.2

Operating System: elementary

I saw this in the help document:

```txt
LSP linters can be used in combination with any other linter, and
will automatically connect to LSP servers when needed.
```

then I searched for the information in the #517

and then I try to do this;

```vim
let g:ale_linters = {
  \'vue': ['vls'],
\}
```

but it didn't work, I have installed vue-language-server.
What should I do? Thanks.

w0rp:
Someone needs to define a linter for it. That one doesn't exist.

Dcryfm:
@w0rp So, like css-language-server, html-language-server, json-language-server and so on can't be used yet?

Dcryfm:
I want to report the diagnostic information of the language server with the other linter tool's diagnostic information by ale.

w0rp:
That can be used as soon as someone defines a linter for them, like everything else. Look at the `ale_linters` directory.

w0rp:
Implemented in #1699

PR #1699, opened 2018-07-05
vue: Add the Vue 'vls' Language Server (#1296)

ananace:
Sadly there's currently an [issue with the vim-vue plugin](https://github.com/posva/vim-vue/blob/master/ftplugin/vue.vim#L24-L25), where it will alias `'vue'` to `'javascript'` for ALE, which means vue-specific linters aren't checked.

As vls is not intended to be used for anything but vue code, adding it as a javascript linter doesn't really make sense, so most likely people will have to explicitly alias `'vue'` to `'vue'` in case they want to use vls while also using `vim-vue` or `vim-polyglot` (Which bundles `vim-vue`).

Perhaps ALE should read available linters from both the alias, and the real language name, if there are linters available for both?

### DIFF
--- a/ale_linters/vue/vls.vim
+++ b/ale_linters/vue/vls.vim
@@ -1,0 +1,32 @@
+" Author: Alexander Olofsson <alexander.olofsson@liu.se>
+" Description: Vue vls Language Server integration for ALE
+
+call ale#Set('vue_vls_executable', 'vls')
+call ale#Set('vue_vls_use_global', get(g:, 'ale_use_global_executables', 0))
+
+function! ale_linters#vue#vls#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'vue_vls', [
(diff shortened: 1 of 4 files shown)


ale_linters/vue/vls.vim
```diff
@@ -0,0 +1,32 @@
...
+
+    return !empty(l:package_path) ? fnamemodify(l:package_path, ':h') : ''
+endfunction
+
+call ale#linter#Define('vue', {
+\   'name': 'vls',
+\   'lsp': 'stdio',
+\   'executable_callback': 'ale_linters#vue#vls#GetExecutable',
+\   'command_callback': 'ale_linters#vue#vls#GetCommand',
+\   'language': 'vue',
```
w0rp: I made the `'language'` key optional now. It won't do any harm, though.

w0rp:
Cheers! :beers:

PR #1719, opened 2018-07-13
Add python_[linter]_auto_pipenv options for python linters (fixes #1656)

### LINKED ISSUE #1656: Automatically detect if `pipenv` should be used for linters
First mentioned in a comment for #1629

`pipenv --venv` will print information that can be used to check if a file is instead of a `pipenv` managed project or not. If a file is in such a project, ALE could automatically load executables from `pipenv` projects instead. This `pipenv` detection ought to be configurable, and probably off by default, as not everyone will be using `pipenv`.

elebow:
When set to true, and the buffer is currently inside a pipenv, GetExecutable will return "pipenv", which will trigger the existing functionality to append the correct pipenv arguments to run each linter.

Defaults to false.

I was going to implement ale#python#PipenvPresent by invoking `pipenv --venv` or `pipenv --where`, but it seemed to be abominably slow, even to the point where the test suite wasn't even finishing ("Tried to run tests 3 times"). The diff is:

```diff
diff --git a/autoload/ale/python.vim b/autoload/ale/python.vim
index 7baae079..8c100d41 100644
--- a/autoload/ale/python.vim
+++ b/autoload/ale/python.vim
@@ -106,5 +106,9 @@ endfunction

" Detects whether a pipenv environment is present.
function! ale#python#PipenvPresent(buffer) abort
-    return findfile('Pipfile.lock', expand('#' . a:buffer . ':p:h') . ';') isnot# ''
+    let l:cd_string = ale#path#BufferCdString(a:buffer)
+    let l:output = systemlist(l:cd_string . 'pipenv --where')[0]
+    " `pipenv --where` returns the path to the dir containing the Pipfile
+    " if in a pipenv, or some error text otherwise.
+    return strpart(l:output, 0, 18) !=# "No Pipfile present"
 endfunction
```

Using vim's `findfile` is much faster, behaves correctly in the majority of situations, and also works reliably when the `pipenv` command doesn't exist.

### DIFF
--- a/ale_linters/python/flake8.vim
+++ b/ale_linters/python/flake8.vim
@@ -5,12 +5,17 @@ call ale#Set('python_flake8_executable', 'flake8')
 call ale#Set('python_flake8_options', '')
 call ale#Set('python_flake8_use_global', get(g:, 'ale_use_global_executables', 0))
 call ale#Set('python_flake8_change_directory', 1)
+call ale#Set('python_flake8_auto_pipenv', 0)
 
 function! s:UsingModule(buffer) abort
     return ale#Var(a:buffer, 'python_flake8_options') =~# ' *-m flake8'
 endfunction
 
 function! ale_linters#python#flake8#GetExecutable(buffer) abort
+    if ale#Var(a:buffer, 'python_flake8_auto_pipenv') && ale#python#PipenvPresent(a:buffer)
+        return 'pipenv'
+    endif
+
     if !s:UsingModule(a:buffer)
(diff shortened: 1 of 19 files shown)


w0rp:
I just finished making the tests a lot easier to write. Now you can write something like `AssertLinter 'flake8', ale#Escape('flake8') . ' --whatever'`. for checking commands. You can update the tests to use that.

You could try chaining commands if you want for running `pipenv` commands. See `:help ale-command-chain`. Testing chains of commands is now a lot easier.

w0rp:
Nice, looks good to me. :+1: 

Cheers! :beers:

PR #1724, opened 2018-07-14
Pipenv virtualenv and project path detection

randrej:
Added support for Pipenv's local virtualenv paths (.venv/).

The project directory can also be detected by the presence of Pipenv or Pipenv.lock files.


w0rp:
Cheers! :beers:

PR #1711, opened 2018-07-10
Support remark-lint installed locally

watagashi:
Support remark-lint installed locally and enables lint without saving it to disk.
It is like textlint.

PR #1730, opened 2018-07-16
Support remark-lint installed locally with new test

watagashi:
Support remark-lint installed locally and enables lint without saving it to disk.

I rebased #1711 onto master, and applied feedback.


w0rp:
There are some missing lines in the documentation, but I'll fix those. Otherwise this looks just fine to me. Cheers! :beers:

PR #1734, opened 2018-07-19
Add language server source in ALEDetail

sridhars:
pyls reports one of several sources per the language server protocol - sources such as pyflakes and pycodestyle. This line allows ALEDetail to show them without changing the default %linter% value.


autoload/ale/lsp/response.vim
```diff
@@ -50,6 +50,10 @@ function! ale#lsp#response#ReadDiagnostics(response) abort
             let l:loclist_item.nr = l:diagnostic.code
         endif
 
+        if has_key(l:diagnostic, 'source')
+           let l:loclist_item.detail = printf("[%s] %s", l:diagnostic.source, l:diagnostic.message)
```
RyanSquared: use single quotes here instead of double quotes to fix the linter errors

sridhars:
Thanks @RyanSquared  and @neersighted !! Had trouble running the tests on my setup but it works now. Please lmk know if I need to do anything.

RyanSquared:
Cheers! :beers:

Issue #1578, opened 2018-05-13
Add support for Haskell IDE Engine (aka hie)


unode:
[Haskell IDE Engine](https://github.com/haskell/haskell-ide-engine) is a server implementing Language Server Protocol.

It encapsulates [most of the linters currently supported by ALE](https://github.com/haskell/haskell-ide-engine/blob/master/docs/Tools.md) in a unified interface simplifying use and providing somewhat better management of system resources.

Saul-Mirone:
+1, since ghc-mod is not maintained anymore, HIE is really important.

alex404:
I'm also eagerly waiting for this.

w0rp:
Go ahead and create a pull request for it if you know how to start the server. Have a look at existing linters in the `ale_linters` directory and `:help ale#linter#Define()` for a description of how that works.

alex404:
Thanks for the tips. I'll look into it and give it a shot.

Luxed:
@alex404 
I'm working on that right now, here is the branch:
https://github.com/Luxed/ale/tree/hie
I don't think it's ready to be a PR yet, but it works pretty much as intended.
Simply enable the `hie` linter for haskell and you're good to go !
( I basically ripped off the Rust rls code to use for the HIE one)
I'll make a PR later today.

w0rp:
Sounds good to me. :+1:

Luxed:
@alex404 @Saul-Mirone , PR has been merged ! Please go ahead and test it to see if it is working correctly.
It worked for me, but I'm a begginer so I don't work on any real codebase, just a simple stack project.
Everything should work as intended as long as you have `hie` in your path. Then it searches for the stack file, the cabal file and it will use the current directory as a default to work on "plain GHC projects".

alex404:
Seems to work well so far! Thanks for your hard work! I'll report if any problems come up.

PR #1735, opened 2018-07-19
Add Haskell IDE Engine (hie) support

Luxed:
This PR should address the issue #1578.

I don't think it is ready to merge yet.
For me, hie is working in a stack project or on a single file. I don't know about cabal projects (it shouldn't because it only searches for "stack.yaml", maybe I should search for the nearest "*.cabal" file too ?).

Help is appreciated :beer:.

TODO list:
- [x] Make a function to find the nearest file corresponding to `*.cabal`
- [x] Do more testing on plain GHC project (no file to identify them)
- [x] Add tests for hie


ale_linters/haskell/hie.vim
```diff
@@ -0,0 +1,29 @@
...
+    let l:executable = ale_linters#haskell#hie#GetExecutable(a:buffer)
+    let l:arguments = '--lsp'
+
+    return ale#Escape(l:executable) . ' ' . l:arguments
+endfunction
+
+function! ale_linters#haskell#hie#GetProjectRoot(buffer) abort
+    let l:stack_file = ale#path#FindNearestFile(a:buffer, 'stack.yaml')
+
+    return !empty(l:stack_file) ? fnamemodify(l:stack_file, ':h') : getcwd()
```
w0rp: If you can't find a project file, return `''` and the language server won't be started. You can only use a language server connection for projects. As you say, it's worth looking for Cabal files too.

ale_linters/haskell/hie.vim
```diff
@@ -0,0 +1,29 @@
...
+function! ale_linters#haskell#hie#GetExecutable(buffer) abort
+    return ale#Var(a:buffer, 'haskell_hie_executable')
+endfunction
+
+function! ale_linters#haskell#hie#GetCommand(buffer) abort
+    let l:executable = ale_linters#haskell#hie#GetExecutable(a:buffer)
+    let l:arguments = '--lsp'
+
+    return ale#Escape(l:executable) . ' ' . l:arguments
+endfunction
```
w0rp: Try writing just `'command': '%e --lsp',` below. I added support for a special `%e` format marker for adding in the escaped executable recently.

ale_linters/haskell/hie.vim
w0rp: Make sure not to use `getcwd()`. See `:help ale-development` for lots of information, including why `getcwd()` is banned.

Luxed:
Everything should be good now !
This is how a project is found: it searches for a `stack.yaml`, if it doesn't find it, it will search for a `*.cabal` file and if it fails again, it will use the buffers directory.

w0rp:
Cheers! :beers:

Luxed:
Cheers! :beers:

PR #1736, opened 2018-07-19
Add support for Fortran language server by @hansec

unpairedbracket:
Adds hansec/fortran-language-server support to ALE. Pretty simple LSP linter so I'm following the example of others like it (such as #1063), but if Vader tests and/or individual documentation are necessary I'll get to adding those.


unpairedbracket:
I've added some documentation and testing for the `fortls` LSP integration as best as I could work out how to - most of the linter testing advice in `:help ale-development` seemed to be intended for compiler-/linter-based integrations rather than LSP, so I don't *think* they apply to this one.

I wasn't sure about how the indentation in the doc file should work as the gfortran docs already in `ale-fortran.txt` were a bit inconsistent and the strings themselves are quite long, so if you'd like me to change them let me know how and I'll do it

w0rp:
Chees! :beers:

PR #1737, opened 2018-07-19
purescript linter

dakom:
This is a starting point for the purescript linter, which _should_ work since it passes all the checkboxes on
https://langserver.org/ and relies on the same server used in the VSCode and Atom plugins.

It requires that [purescript-language-server](https://github.com/nwolverson/purescript-language-server) be installed, as well as the purescript runtime.

However - I am not getting any actual reporting from Ale, despite the server being correctly started/stopped when I open/close vim with a valid purescript file.

Help is appreciated!

### DIFF
--- a/ale_linters/purescript/pure_ls.vim
+++ b/ale_linters/purescript/pure_ls.vim
@@ -1,0 +1,39 @@
+" Author: David Komer <david.komer@gmail.com> 
+" Description: Integrate ALE with purescript-language-server.
+
+call ale#Set('pure_ls_executable', 'purescript-language-server')
+call ale#Set('pure_ls_use_global',
+\    get(g:, 'ale_use_global_executables', 0)
+\)
+
+function! ale_linters#purescript#pure_ls#GetExecutable(buffer) abort
+    return ale#node#FindExecutable(a:buffer, 'pure_ls', [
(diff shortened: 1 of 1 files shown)


dakom:
Latest commit now fixes it, note that the project must be built first before linting (this is the same behavior as VSCode and Atom plugins)

dakom:
I tried to set it up but `docker` didn't install well into WSL... will try to give it a shot when I get to the OSX box :)

Btw - one thing I did notice is that I'm not getting constant linting as a type, though I _am_ getting that with typescript. any idea what might cause that?

xla:
Very interested in this addition. How can this be moved forward?

ale_linters/purescript/pure_ls.vim
```diff
@@ -0,0 +1,38 @@
+" Author: David Komer <david.komer@gmail.com> 
+" Description: Integrate ALE with purescript-language-server.
+
+call ale#Set('pure_ls_executable', 'purescript-language-server')
```
w0rp: The options should be named `ale_<language>_<linter>_<option>`, so you can write `purescript_pure_ls_...` here. `purels` might be better.

w0rp:
Closed due to inactivity. Update the PR and let us know to review it again when ready.

PR #1742, opened 2018-07-20
Add cabal-ghc linter

typetetris:
cabal-ghc calls ghc via cabal exec and so ghc has access to packages
in cabal sandboxes for example


typetetris:
The situation with cabal and stack is as follows: You can manage your haskell projects with one of them or even both. So you could have a stack configuration file, a cabal configuration file or both.

But I can't imagine, that you would want to use stack-ghc or stack-build and cabal-ghc at the same time. So the user should select one.

I can't use stack though, as I don't use stack and need cabal-ghc for ale to be useful to me.

Maybe cabal-ghc could be disabled as default, but I don't know, how to do that, I am not a vim script hacker, did mostly C&P here.

w0rp:
Cheers! :beers;

typetetris:
Cheers! :beers:

PR #1745, opened 2018-07-21
Work around hot-reloading issue again

sharils:
See phoenixframework/phoenix#1165 for more detail

---
This change is [Reviewable](https://reviewable.io/reviews/w0rp/ale/1745)

### DIFF
--- a/ale_linters/elixir/dialyxir.vim
+++ b/ale_linters/elixir/dialyxir.vim
@@ -28,7 +28,7 @@ endfunction
 call ale#linter#Define('elixir', {
 \   'name': 'dialyxir',
 \   'executable': 'mix',
-\   'command': 'mix dialyzer',
+\   'command': 'mix help dialyzer && mix dialyzer',
 \   'callback': 'ale_linters#elixir#dialyxir#Handle',
 \})
 


sharils:
@w0rp 

Sure, I can explain.

When dializer isn't a dependency, `mix dialyzer` recompiles the whole project because it's not possible to know if this command `dialyzer` exist or not until recompilation is done. Then the timestamps of the project is then messed up which results in broken hot-loading. In this case, `mix help dialyzer` would return zero which prevents compilation of the whole project since dialyzer isn't installed, it's help manual doesn't exist.

When dialyzer is a dependency, `mix dialyzer` would just run the command. In this case, `mix help dialyzer` would return 1 which allows `mix dialyzer` to run.

Let me know if you have further question. It's my pleasure to answer any further question.

BTW, this kind of work around has been proposed before, https://github.com/w0rp/ale/pull/1266 . It just that at the time when 1266 was still being review, the dialyzer linter is also being reviewed https://github.com/w0rp/ale/pull/1257 , thus in that PR, dialyzer wasn't included.

Best

sharils:
I've also amended the commit message to include the reason this work around works.

w0rp:
Cheers! :beers:

PR #1750, opened 2018-07-22
Make drafter linter use stdin instead of writing to tmp file

kylef:
Writing to a tmp file is unnecessary as drafter will use stdin if a path is not provided.

This can be demonstrated via:

```shell
$ cat test.apib | drafter -ul

OK.
warning: (6)  no identifier specified; line 8, column 11 - line 8, column 18
```


w0rp:
Cheers! :beers:

PR #1752, opened 2018-07-23
Only run stack if a stack.yaml config is found

destiny-index:
It is necessary to check for a stack.yaml file to distinguish between
cabal-only projects or stack projects (which are also cabal projects
since stack is built on top of cabal).


w0rp:
I was saying in a previous pull request that this would probably come up.

autoload/ale/handlers/haskell.vim
```diff
@@ -1,5 +1,15 @@
...
 " Description: Error handling for the format GHC outputs.
+"
+function! ale#handlers#haskell#GetStackExecutable(bufnr) abort
+    if ale#path#FindNearestFile(a:bufnr, 'stack.yaml') isnot# ''
+        return 'stack'
+    else
+        " if there is no stack.yaml file, we don't use stack even if it exists,
+        " so we return '', because executable('') apparently always fails
+        return ''
+    endif
```
w0rp: Move `return ''` to the end of the function, instead of putting it inside `else`.

destiny-index: Fixed

RyanSquared:
I'm not confident in the tests. I think that the tests should create `stack.yml`, and then in the negative tests, remove it.

PR #1756, opened 2018-07-24
feat(linters: python): add vulture; fixes #1729

### LINKED ISSUE #1729: Add support for vulture
**Name:** vulture
**URL:** https://github.com/jendrikseipp/vulture

Vulture finds unused/dead python code using static code analysis. Uses the same output format as pyflakes.

actionless:


actionless:
@w0rp what exactly do you mean, test for Exception message from vulture or something more?

test/handler/test_vulture_handler.vader
```diff
@@ -0,0 +1,28 @@
...
+  AssertEqual
+  \ [
+  \   {
+  \     'lnum': 34,
+  \     'text': "unused variable 'foo' (60% confidence)",
+  \     'type': 'W',
+  \   },
+  \ ],
+  \ ale_linters#python#vulture#Handle(bufnr(''), [
+  \   "/testplugin/test/handler/something_else.py:34: unused variable 'foo' (60% confidence)",
```
actionless: @w0rp is there any helper function to get that path dynamically in tests instead of hardcoding it?

it seems to be the reason of appveyor testsuite to fail

w0rp: Look at existing tests for uses of `SetDirectory`, `RestoreDirectory`, and `SetFilename`. They will set up paths so an filenames so they work inside and outside of the Docker app. Use the `g:dir` variable and `ale#path#Simplify` so the paths work on both Linux and Windows.

test/handler/test_vulture_handler.vader
```diff
@@ -0,0 +1,28 @@
...
+
+  silent file something_else.py
+
+Execute(Basic vulture check should be handled):
+  call ale#test#SetFilename('something_else.py')
+  AssertEqual
+  \ [
+  \   {
+  \     'lnum': 34,
+  \     'text': "unused variable 'foo' (60% confidence)",
```
w0rp: For strings containing single quotes, just escape the single quotes by doubling them. `'foo '' bar'` -> `foo ' bar`.

ale_linters/python/vulture.vim
```diff
@@ -0,0 +1,67 @@
...
+    endfor
+
+    " Matches patterns line the following:
+    let l:pattern = '\v^([a-zA-Z]?:?[^:]+):(\d+): (.*)$'
+    let l:bufpath = bufname(a:buffer)
+    let l:output = []
+
+    for l:match in ale#util#GetMatches(a:lines, l:pattern)
+        let l:path = l:match[1]
+        if l:path isnot# l:bufpath && ale#Var(a:buffer, 'python_vulture_change_directory')
```
w0rp: I think instead of filtering based on the buffer name here, you can include the filename below in the `'filename'` key. Then you'll be able to jump to errors in other files. Does vulture always return absolute paths, or relative paths? If the paths are relative, use `ale#path#GetAbsPath` to get an absolute path to the files.

actionless:
thanks for the reviewing! hope now it gonna work well

actionless:
i've just noticed it's not added to the list of "file linters" in the docs

w0rp:
Cool, create a pull request for that.

PR #1765, opened 2018-07-27
doc(readme: faq): add note about manually calling completion instantly

actionless:


w0rp:
These functions aren't part of the public API. Instead, we should add a command for this, and tests for the command.

actionless:
command for `function! MyAleCompletion()` or for both un-exposed functions which i am calling there?

w0rp:
The `ale#completion` functions there shouldn't be called outside of the codebase. They could be changed at any time. We should add a command for triggering completion manually,  with a Plug keybind, and a setting for disabling automatic completion.

actionless:
>  and a setting for disabling automatic completion.

mb just to make it use `g:ale_completion_enabled`?

w0rp:
Hmm, yeah I guess you could just left the option off and it would work if you use the new command.

RyanSquared:
@actionless do you want to redo this PR or should we add an issue with a feature request?

actionless:
i wanna redo it (since i am still using `ale`) but not soon (vacation) so i'm closing it

actionless:
but of course i would be more happy if something similar will be implemented as a feature :-)

w0rp:
There's an issue open for it.

actionless:
this one? https://github.com/w0rp/ale/issues/1830

PR #1769, opened 2018-07-31
Javascript prettier filetype detect fix

paulreimer:
Noticed that javascript `prettier` fixer had some filetype->parser detection support, but I believe there was a bug inverting the logic in filetype detection, preventing it from being used (301e85f).

As well, it is simply good luck that the name of the parsers available in `prettier` match the vim filetype string, so that was converted to a dictionary to handle the case where they are different (9b11c84).

Finally, `prettier` does support additional filetypes/parsers not yet added, so I added yaml in (41da5e4), and the experimental html parser support which is coming soon (!) in (9838e45). The name of that parser is `parse5`, thus it is the first case where the filetype will not match the parsing string, motivating the dictionary approach above.


paulreimer:
Fair enough! But now I have some design/req questions (bear with me as I am not a vimscript dev):
• Overall, it seems like there are tests for this (based on `set filetype`), however the actual code to select a parser is enabled only for files with NO extension (that's why `testfile` works in the tests, but not `myfile.css` in the wild). I think this guard should be removed and any `l:filetype` that matches `has_key` should select the correct parser.
• It seems the intent was to supply a default `prettier` parser value of 'babylon', which _is_ currently the default `prettier` parser. But due to above, it wasn't getting detected and by default there was no `--parser XXX` flag supplied (the tests expected this). This allows `prettier` to change their mind later, and so maybe there should be no `--parser babylon` by default?

(that is to say, both of these are behaviour changes to make the existing tests pass, but I think that might have been the intended behaviour in the first place?)

paulreimer:
1. Fixed in 192112c
2. Fixed in 9e44e2d

Tests added for additional filetypes: YAML, HTML
FYI HTML support is not officially landed in Prettier, but much review on the current PR has occurred.
(https://github.com/prettier/prettier/pull/4753)

paulreimer:
Finally, noticed there were already some empty files with extensions (e.g. .css, .ts, .json) and updated the tests to use them. So should I create some more empty files -- one per file extension with an alternate parser? Currently approx 50% of them have file extensions, the rest are relying on filetypes with `testfile` (no extension).

w0rp:
Put the code for only setting `--parser` if the file extension is set back in. The `--parser` option is there as a fallback option. Prettier will automatically detect which parser to use, but it can't do that when the file doesn't have an extension. When Vim knows what the filetype is, it can figure out which parser to use.

w0rp:
Read these pull requests.

https://github.com/w0rp/ale/pull/1094
https://github.com/w0rp/ale/pull/1620

Setting `--parser` causes problems when `prettier` is updated to support different parsers, but needs to be set when there's no file extension.

paulreimer:
OK, reverted those changes in 9459a6a.

The extent of the PR is now adding support for YAML parser (when no file extension is present), and changing the list of parsers to a dictionary.

paulreimer:
Also, still unsure whether a default `--parser babylon` should be supplied (when no file extension and no matching filetype). At this point it would not be provided e.g. for a JS file with no extension, it would rely on `prettier`'s default parser.

PR #1771, opened 2018-07-31
Show LSP diagnostic codes for `%code%` in message formats

### LINKED ISSUE #1767: `%code%` in message format doesn't work for LSP
## Information

**VIM version**

```
VIM - Vi IMproved 8.1 (2018 May 18, compiled Jul 22 2018 05:24:30)
macOS version
```

Operating System: MacOS High Seirra

### :ALEInfo

```

 Current Filetype: 
Available Linters: []
  Enabled Linters: []
 Suggested Fixers: 
 Linter Variables:

 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = v:null
let g:ale_command_wrapper = v:null
let g:ale_completion_delay = 100
let g:ale_completion_max_suggestions = 50
let g:ale_echo_cursor = 1
let g:ale_echo_msg_format = '[%linter%] %code% %s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = v:null
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_filetype_changed = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_insert_leave = 0
let g:ale_linter_aliases = {}
let g:ale_linters = {}
let g:ale_linters_explicit = 0
let g:ale_list_window_size = v:null
let g:ale_list_vertical = v:null
let g:ale_loclist_msg_format = v:null
let g:ale_max_buffer_history_size = v:null
let g:ale_max_signs = v:null
let g:ale_maximum_file_size = v:null
let g:ale_open_list = v:null
let g:ale_pattern_options = v:null
let g:ale_pattern_options_enabled = v:null
let g:ale_set_balloons = 1
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_column_always = v:null
let g:ale_sign_error = v:null
let g:ale_sign_info = v:null
let g:ale_sign_offset = v:null
let g:ale_sign_style_error = v:null
let g:ale_sign_style_warning = v:null
let g:ale_sign_warning = v:null
let g:ale_statusline_format = v:null
let g:ale_type_map = v:null
let g:ale_use_global_executables = v:null
  Command History:
```
## What went wrong

`%code%` is always the empty string when using lsp-based linters.

## Reproducing the bug

1. Put `%code%` in your `g:ale_echo_msg_format`
1. Open a file using an LSP server
1. introduce an error
1. Expected: `%code%` to be replaced by the code from publishDiagnostics. Actual: `%code%` replaced with empty string

## Cause

A `.nr` field is populated here:
- https://github.com/w0rp/ale/blob/a7b8cb4fe32b43dbeeadc8cb94f378b4e2112723/autoload/ale/lsp/response.vim#L50
- https://github.com/w0rp/ale/blob/a7b8cb4fe32b43dbeeadc8cb94f378b4e2112723/autoload/ale/lsp/response.vim#L73

Problems:
- https://github.com/w0rp/ale/blob/a7b8cb4fe32b43dbeeadc8cb94f378b4e2112723/autoload/ale.vim#L255 does not check for `.nr` - it only checks for the `code` field, which is not documented in `ale-loclist-format`.
- I'm uncertain if it's safe to do this: it looks like `.nr` is expected to be an int, however the LSP allows string error codes

fredemmott:
fixes #1767


fredemmott:
Will fix the test failures. Missed them due to some pre-existing failures when I ran without modifications :/

fredemmott:
More to hunt down, but out of time for today.

autoload/ale.vim
```diff
@@ -252,10 +252,10 @@ endfunction
 function! ale#GetLocItemMessage(item, format_string) abort
     let l:msg = a:format_string
     let l:severity = g:ale_echo_msg_warning_str
-    let l:code = get(a:item, 'code', '')
+    let l:code = get(a:item, 'code', -1)
     let l:type = get(a:item, 'type', 'E')
     let l:linter_name = get(a:item, 'linter_name', '')
-    let l:code_repl = !empty(l:code) ? '\=submatch(1) . l:code . submatch(2)' : ''
+    let l:code_repl = l:code != -1 ? '\=submatch(1) . l:code . submatch(2)' : ''
```
w0rp: Undo the changes to this file. The LSP code should be fixed so it doesn't set the key with `-1` instead.

autoload/ale/engine.vim
```diff
@@ -348,13 +348,10 @@ function! ale#engine#FixLocList(buffer, linter_name, loclist) abort
         \   'vcol': 0,
         \   'type': get(l:old_item, 'type', 'E'),
         \   'nr': get(l:old_item, 'nr', -1),
+        \   'code': get(l:old_item, 'code', -1),
         \   'linter_name': a:linter_name,
         \}
 
-        if has_key(l:old_item, 'code')
-            let l:item.code = l:old_item.code
-        endif
```
w0rp: Undo the changes here. The code shouldn't be set if it's invalid.

autoload/ale/lsp/response.vim
```diff
@@ -47,7 +47,10 @@ function! ale#lsp#response#ReadDiagnostics(response) abort
         endif
 
         if has_key(l:diagnostic, 'code')
-            let l:loclist_item.nr = l:diagnostic.code
+            let l:loclist_item.code = l:diagnostic.code
+            if type(l:diagnostic.code) == v:t_number
+                let l:loclist_item.nr = l:diagnostic.code
+            endif
```
w0rp: Check the value of the code before you set it. Don't set it if it's `-1`. Convert numbers to strings.

Put blank lines above if statements that don't start a new level of indentation.

autoload/ale/lsp/response.vim
```diff
@@ -70,7 +73,10 @@ function! ale#lsp#response#ReadTSServerDiagnostics(response) abort
         \}
 
         if has_key(l:diagnostic, 'code')
-            let l:loclist_item.nr = l:diagnostic.code
+            let l:loclist_item.code = l:diagnostic.code
+            if type(l:diagnostic.code) == v:t_number
+                let l:loclist_item.nr = l:diagnostic.code
+            endif
```
w0rp: Same as the above.

doc/ale.txt
```diff
@@ -2318,6 +2318,12 @@ ale#linter#Define(filetype, linter)                       *ale#linter#Define()*
                          `type` - Defaults to `'E'`.
                          `nr` - Defaults to `-1`.
 
+                           Numeric error code. If `nr` is not `-1`, `code` should have
+                           the same value.
+                         `code` - Defaults to `-1`.
```
w0rp: There should be no default.

autoload/ale/lsp/response.vim
fredemmott: > Check the value of the code before you set it. Don't set it if it's -1

Will do for `nr`; not planning on doing that for `code` given that `-1` has no special meaning either here (after removing the default), or in the LSP

w0rp: The `code` key should be a String, and shouldn't be set if the code isn't valid.

fredemmott: > shouldn't be set if the code isn't valid.

`-1` is a valid code for an LSP server to send; the only requirement in the specification is that it is a number or a string.

> The code key should be a String

Sure, planning on setting it as `'-1'` in this case

w0rp: I wouldn't bother setting the code if it's `-1`. It doesn't tell anyone anything useful. They see a message like `-1: Some error`. It's better if they just see `Some error` instead.

w0rp: Do you know if any servers use `-1` for "no error code," or if any servers use `-1` for a specific error code?

fredemmott: Fair enough. That said, both this behavior and casting to strings will likely lead to problems if you decide to implement `textDocument/codeAction` in the future, as most servers seem to expect the code provided by the client in context/diagnostics to exactly match what was sent.

w0rp: We could add another key for that. The existing key named `code` has a specific meaning in ALE, which is an error code to print to the user. This is typically a descriptive code like `'no-const-assign'` or something.

fredemmott:
mostly fixed. todo:
- add tests for -1 behavior
- figure out why `code` is not present in some of the tsserver tests when `nr` is

doc/ale.txt
```diff
@@ -2318,6 +2318,12 @@ ale#linter#Define(filetype, linter)                       *ale#linter#Define()*
                          `type` - Defaults to `'E'`.
                          `nr` - Defaults to `-1`.
 
+                           Numeric error code. If `nr` is not `-1`, `code` should have
```
fredemmott: oh, and I should change this to say something about being cast to a string.

w0rp: Say that the code is required to be a String. This is documentation for people looking up which keys they can export, and what the values should look like.

fredemmott:
I think this is mostly done now; the only thing remaining is:

```
    (1/3) [EXECUTE] tsserver syntax error responses should be handled correctly
    (1/3) [EXECUTE] (X) [{'lnum': 1, 'bufnr': 3, 'col': 14, 'pattern': '', 'valid': 1, 'vcol': 0, 'nr': 1005, 'type': 'E', 'text': '1005: '','' expected.'}] should be equal to [{'lnum': 1, 'bufnr': 3, 'col': 14, 'pattern': '', 'valid': 1, 'vcol': 0, 'code': '1005', 'nr': 1005, 'type': 'E', 'text': ''','' expected.'}]
    (2/3) [EXECUTE] tsserver semantic error responses should be handled correctly
    (2/3) [EXECUTE] (X) [{'lnum': 1, 'bufnr': 3, 'col': 14, 'pattern': '', 'valid': 1, 'vcol': 0, 'nr': 1005, 'type': 'E', 'text': '1005: Some semantic error'}] should be equal to [{'lnum': 1, 'bufnr': 3, 'col': 14, 'pattern': '', 'valid': 1, 'vcol': 0, 'code': '1005', 'nr': 1005, 'type': 'E', 'text': 'Some semantic error'}]
```

I see how to fix the tests, but I don't know:
- why `code` is missing
- if adding `code:` to the text is desired here

fredemmott:
Thanks - still have those tsserver test errors though; what's the expected behavior for these?

w0rp:
I think I'll fix those tsserver tests later this evening and merge this. They probably aren't hard to fix.

fredemmott:
Here's /a/ fix:

```
diff --git a/test/test_engine_lsp_response_handling.vader b/test/test_engine_lsp_response_handling.vader
index ddcd7b2e..818659db 100644
--- a/test/test_engine_lsp_response_handling.vader
+++ b/test/test_engine_lsp_response_handling.vader
@@ -62,9 +62,8 @@ Execute(tsserver syntax error responses should be handled correctly):
   \     'col': 14,
   \     'vcol': 0,
   \     'nr': 1005,
-  \     'code': '1005',
   \     'type': 'E',
-  \     'text': ''','' expected.',
+  \     'text': '1005: '','' expected.',
   \     'valid': 1,
   \     'pattern': '',
   \   },
@@ -136,9 +135,8 @@ Execute(tsserver semantic error responses should be handled correctly):
   \     'col': 14,
   \     'vcol': 0,
   \     'nr': 1005,
-  \     'code': '1005',
   \     'type': 'E',
-  \     'text': 'Some semantic error',
+  \     'text': '1005: Some semantic error',
   \     'valid': 1,
   \     'pattern': '',
   \   },
```

w0rp:
Cheers! :beers:

PR #1778, opened 2018-08-02
Tell HHAST to only look at open files

fredemmott:
Given ALE only cares about open files, this has no observable change,
except for significantly reduced resource usage.

Didn't include this in the original PR as I thought there was some subtle incompatibly. Turns out that actually, HHAST was ignoring `rootUri`/`rootPath`, and always used $CWD - vscode and atom spawn it from `cwd`, but that's not required. Have hopefully fixed that issue in the next release of hhast.

### DIFF
--- a/ale_linters/hack/hhast.vim
+++ b/ale_linters/hack/hhast.vim
@@ -26,10 +26,15 @@ function! ale_linters#hack#hhast#GetExecutable(buffer) abort
     return !empty(l:absolute) ? l:absolute : ''
 endfunction
 
+function! ale_linters#hack#hhast#GetInitializationOptions(buffer) abort
+    return {'lintMode': 'open-files'}
+endfunction
+
 call ale#linter#Define('hack', {
 \   'name': 'hhast',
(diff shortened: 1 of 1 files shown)


fredemmott:
(the default behavior is to prioritize open files, but to start linting the whole tree the moment the initialize command is received)

ale_linters/hack/hhast.vim
```diff
@@ -26,10 +26,15 @@ function! ale_linters#hack#hhast#GetExecutable(buffer) abort
     return !empty(l:absolute) ? l:absolute : ''
 endfunction
 
+function! ale_linters#hack#hhast#GetInitializationOptions(buffer) abort
+    return {'lintMode': 'open-files'}
```
w0rp: I'll add this in now. Later I plan to set up a default `initialization_options_callback` which reads `ale_FILETYPE_LINTER_options` by default, and then I'll move this into an option and document it.

w0rp:
Cheers! :beers:

PR #1783, opened 2018-08-03
Add support for yang-lsp

dimbleby:
[yang-lsp](https://github.com/theia-ide/yang-lsp)

I hope I've done everything right!  

One thing I did hesitate over: I've gone with just `ale-yang-lsp` etc where perhaps the consistent thing would be to spell out the filetype and the linter both in full - giving `ale-yang-yang-lsp` etc.  But that stutter seemed pretty ugly to me, so that's why I made the choice that I did.  Happy to change if that's preferred.

(This particular language server will be more useful if and when theia-ide/yang-lsp#140 is fixed, allowing us to benefit from going to definitions, finding references, etc.  Meanwhile we still get linting via `publishDiagnostics` so it's not nothing).


w0rp:
Cheers! :beers:

PR #1787, opened 2018-08-05
doc: LSP can provide completions

polyzen:


w0rp:
Cheers! :beers: I missed that.

PR #1789, opened 2018-08-05
Add prettier as a YAML option

petermbenjamin:
This is a minor README update to add `prettier` as an option for yaml.


w0rp:
Cheers! :beers:

severo:
OK, thanks for the explanation.

PR #1793, opened 2018-08-06
Use .cquery file to detect the project root

kodemeister:
Hi @w0rp,

In addition to `compile_commands.json` cquery also supports `.cquery` file placed into the project root. This is a simple text file containing a list of clang compilation options.

https://github.com/cquery-project/cquery/wiki/.cquery

I have several projects using `.cquery` files so it would be good to support them in ALE as well.

Thanks!

### DIFF
--- a/ale_linters/c/cquery.vim
+++ b/ale_linters/c/cquery.vim
@@ -6,6 +6,9 @@ call ale#Set('c_cquery_cache_directory', expand('~/.cache/cquery'))
 
 function! ale_linters#c#cquery#GetProjectRoot(buffer) abort
     let l:project_root = ale#path#FindNearestFile(a:buffer, 'compile_commands.json')
+    if empty(l:project_root)
+        let l:project_root = ale#path#FindNearestFile(a:buffer, '.cquery')
+    endif
     return !empty(l:project_root) ? fnamemodify(l:project_root, ':h') : ''
 endfunction
 
--- a/ale_linters/cpp/cquery.vim
+++ b/ale_linters/cpp/cquery.vim
@@ -6,7 +6,9 @@ call ale#Set('cpp_cquery_cache_directory', expand('~/.cache/cquery'))
 
 function! ale_linters#cpp#cquery#GetProjectRoot(buffer) abort
     let l:project_root = ale#path#FindNearestFile(a:buffer, 'compile_commands.json')
-
+    if empty(l:project_root)
+        let l:project_root = ale#path#FindNearestFile(a:buffer, '.cquery')
+    endif
     return !empty(l:project_root) ? fnamemodify(l:project_root, ':h') : ''
 endfunction
 


ale_linters/c/cquery.vim
```diff
@@ -6,6 +6,9 @@ call ale#Set('c_cquery_cache_directory', expand('~/.cache/cquery'))
 
 function! ale_linters#c#cquery#GetProjectRoot(buffer) abort
     let l:project_root = ale#path#FindNearestFile(a:buffer, 'compile_commands.json')
+    if empty(l:project_root)
```
w0rp: Put a blank line before every`if` that doesn't start a new level of indentation.

ale_linters/c/cquery.vim
```diff
@@ -6,6 +6,9 @@ call ale#Set('c_cquery_cache_directory', expand('~/.cache/cquery'))
...
+        let l:project_root = ale#path#FindNearestFile(a:buffer, '.cquery')
+    endif
     return !empty(l:project_root) ? fnamemodify(l:project_root, ':h') : ''
```
w0rp: Put a blank line before every`return` that doesn't start a new level of indentation.

kodemeister:
Thank you for reviewing this! As usual, excellent points and all are valid! I've added tests not only for `.cquery` file but also for `compile_commands.json`. Please let me know if you need anything else.

kodemeister:
Cheers! :beers:

MaskRay:
Actually, I think neither `compile_commands.json` nor `.cquery` is a good choice.

Here is a common setup where users create a build directory in the project root and create a compile_commands.json symlink pointing to the real one in the build directory.
```
% tree
.
├── build
│   ├── a.cc
│   └── compile_commands.json
├── compile_commands.json -> build/compile_commands.json
└── src
    └── a.cc
```

`build/a.cc` will have an in correct project root.

In emacs-ccls, the default logic I use (MaskRay/emacs-ccls@b02a5d8#diff-2093f0a348c7bfdc7e0a9006283e081fR100):

* Locate `.ccls-root` if it exists
* `(projectile-project-root)` otherwise. projectile is a widely-used project extension so the setting is fine. But this is tricky in Vim land I know as there is no single dominant project plugin.

When subprojects are involved:
```
% tree -a        
.
└── llvm
    ├── .git
    └── tools
        └── clang    # users expect files to be considered part of the llvm umbrella project
            └── .git
```

I solve this problem by locating `.ccls-root`.

A `.cquery` file can reside in subdirectories to control compile command lines only in that directory. ccls has a similar `.ccls` file.


And BTW a feature request for adding ccls https://github.com/MaskRay/ccls/wiki/Getting-started as a supported linter :) The latency of its completion and diagnostics should be on par with clangd now, but it offers references as cquery does (and better, but this is subjective).

w0rp:
Add in an ad for ShamWow while you're at it.

Issue #1482, opened 2018-04-08
ale resets the location list opened by gv.vim


ahmedelgabri:
## Information

**VIM version**
```
NVIM v0.2.2
Build type: Release
```
and

```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Mar 31 2018 09:12:38)
macOS version
Included patches: 1-1650
```

Operating System: macOS High Sierra 10.13.4

### :ALEInfo

```
 Current Filetype: vim
Available Linters: ['vint']
  Enabled Linters: ['vint']
 Linter Variables:

let g:ale_vim_vint_show_style_issues = 1
 Global Variables:

let g:ale_cache_executable_check_failures = 0
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = 100
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = 50
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '[%linter%] %code% %s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_format = '[%linter%] %code% %s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 1
let g:ale_fixers = {'reason': ['refmt'], 'markdown': ['prettier'], 'scss': ['prettier'], 'json': ['prettier'], 'javascript': ['prettier'], 'css': ['prettier'
]}
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_filetype_changed = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'always'
let g:ale_lint_on_insert_leave = 0
let g:ale_linter_aliases = {'html': ['html', 'css'], 'mail': 'markdown'}
let g:ale_linters = {'javascript': ['eslint', 'flow']}
let g:ale_linters_explicit = 0
let g:ale_list_window_size = 10
let g:ale_list_vertical = 0
let g:ale_loclist_msg_format = '%code: %%s'
let g:ale_max_buffer_history_size = 20
let g:ale_max_signs = -1
let g:ale_maximum_file_size = 0
[...]

PR #1796, opened 2018-08-07
Never write to location list when disabled

idbrii:
Improves #1482.

Fixes ale clobbering populated location-list value when ale's disabled
immediately before populating.

Consider a command that pipes disable with populating (`GV?` is
junegunn/gv.vim's command to populate the location-list):

    command! Ghistory ALEDisableBuffer| echo "ALEDisableBuffer"| GV?

On execution, the location-list is opened, but empty. Running the
command again works (because ALE is still disabled).

Also reproducible with a simple loclist:

    :ALEDisableBuffer | call setloclist(0, [{'lnum': 1, 'text': 'important text'}])

### DIFF
--- a/autoload/ale/list.vim
+++ b/autoload/ale/list.vim
@@ -75,6 +75,13 @@ function! s:BufWinId(buffer) abort
 endfunction
 
 function! s:SetListsImpl(timer_id, buffer, loclist) abort
+    if ale#ShouldDoNothing(a:buffer)
+        " It's possible someone tried to turn off ale in a buffer and then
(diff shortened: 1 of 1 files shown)


w0rp:
`:ALEDisableBuffer` needs to clear the loclist when it is used, and this patch will break that. If you want to use the loclist to display something from `git`, you shouldn't also use it to show problems in a file. I recommend setting `g:ale_set_loclist` to `0`. You can jump through problems in files without using the loclist with `:ALENext` and so on.

w0rp:
No, but I'll open an issue for it. Here you go: #1810

Issue #1758, opened 2018-07-25
ALEGoTo commands should add to jumplist


dimbleby:
After doing (say) `:ALEGoToDefinition` I expect to be able to jump back to where I just came from.  Actually, that's probably what I want to do most times!

(cf #1537.  Though I'm not sure what the problem is there - for me, navigating via the location list already seems to populate the jumplist just fine).

Presumably the fix would be to add something like 

        if !empty(l:result)
            normal! m`
        endif

to `ale#definition#HandleLSPResponse`?

dimbleby:
No it doesn't.  That's exactly what I want to be able to do.

w0rp:
I'll mark it as a bug then. If you could share all of your vim settings and your Vim version, that may be helpful.

dimbleby:
I'm using nightly neovim appimage, currently `v0.3.2-30-g72ddbb675`, and this reproduces with as minimal a `.vimrc` as I can manage:

```
set nocompatible
call plug#begin('~/.vim/plugged')
Plug 'w0rp/ale'
call plug#end()

let g:ale_completion_enabled = 1
let g:ale_linters = { 'rust': ['rls'] }
let g:ale_rust_rls_toolchain = 'stable'
```

dimbleby:
I think the problem comes when the jump is to a definition in the current file.  Then all that happens is `cursor(a:line, a:column)` - which does not change the jumplist.

References: [ale code](https://github.com/w0rp/ale/blob/06132954b12ba87c7bba8fb7d5a40e4ff8b774e8/autoload/ale/util.vim#L99), [vim help](https://github.com/vim/vim/blob/b6c8cd8dc3fbbe9da1180ce1b75f2252df2c1318/runtime/doc/eval.txt#L3468).

w0rp:
The marks should now be set. There are no tests for this because Vader can't test it.

dimbleby:
super, many thanks!

PR #1797, opened 2018-08-07
Additionally adjust the jumplist with line numbers when opening a new file

bradlarsen:
This further addresses Issue #1758.

In particular, without these changes, when you use `ALEGoToDefinition` and you are taken to a different buffer, the particular line you end up on in that buffer isn't included in the position in the jump list.  So, while you can jump back to where you were with C-o, once you do that, you can't return to where you jumped with C-i — instead, C-i takes you to the top of the buffer containing the definition in question.

These changes use the optional `+LINE_NUM` argument to the `:edit` and `:tabedit` commands, which adjusts the jumplist with the line number as well as the buffer, making C-o and C-i both work as expected.  (Well, not strictly true, as the particular column number isn't preserved in the jump list in these cases.  But at least we get line numbers now.)

### DIFF
--- a/autoload/ale/util.vim
+++ b/autoload/ale/util.vim
@@ -88,10 +88,10 @@ endfunction
 
 function! ale#util#Open(filename, line, column, options) abort
     if get(a:options, 'open_in_tab', 0)
-        call ale#util#Execute('tabedit ' . fnameescape(a:filename))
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

bradlarsen:
Cheers! :beers:  :-D

PR #1799, opened 2018-08-08
Add ale_php_phpmd_suffixes option for phpmd linter

ppppisme:
Hey!

This PR adds functionality to specify extensions of the files to check with phpmd (by default it only checks `.php` files)

Cheers! :beers:

### DIFF
--- a/ale_linters/php/phpmd.vim
+++ b/ale_linters/php/phpmd.vim
@@ -6,9 +6,18 @@ let g:ale_php_phpmd_executable = get(g:, 'ale_php_phpmd_executable', 'phpmd')
 " Set to change the ruleset
 let g:ale_php_phpmd_ruleset = get(g:, 'ale_php_phpmd_ruleset', 'cleancode,codesize,controversial,design,naming,unusedcode')
 
+" Set to tell phpmd files with which suffixes to check
+let g:ale_php_phpmd_suffixes = get(g:, 'ale_php_phpmd_suffixes', '')
+
 function! ale_linters#php#phpmd#GetCommand(buffer) abort
+    let l:suffixes = ale#Var(a:buffer, 'php_phpmd_suffixes')
+    let l:suffixes_option = !empty(l:suffixes)
+    \   ? ' --suffixes ' . l:suffixes
+    \   : ''
+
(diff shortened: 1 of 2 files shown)


ale_linters/php/phpmd.vim
```diff
@@ -6,9 +6,18 @@ let g:ale_php_phpmd_executable = get(g:, 'ale_php_phpmd_executable', 'phpmd')
...
 let g:ale_php_phpmd_ruleset = get(g:, 'ale_php_phpmd_ruleset', 'cleancode,codesize,controversial,design,naming,unusedcode')
 
+" Set to tell phpmd files with which suffixes to check
+let g:ale_php_phpmd_suffixes = get(g:, 'ale_php_phpmd_suffixes', '')
+
 function! ale_linters#php#phpmd#GetCommand(buffer) abort
+    let l:suffixes = ale#Var(a:buffer, 'php_phpmd_suffixes')
+    let l:suffixes_option = !empty(l:suffixes)
+    \   ? ' --suffixes ' . l:suffixes
+    \   : ''
```
w0rp: I wonder, could we set this automatically? We could grab the file extension from the filename and pass that to the command, and then add tests for a few examples. Then you wouldn't need to configure anything for it to work. `expand('#' . a:buffer . ':e')` will get you the file extension without a dot, or an empty string if the file has no file extension.

vladgor: Good point. But what if user wants to specify an exact set of file extensions that phpmd should work with? Can it be done with ALE built-in functionality?

What do you think of following approach: we could use your solution when suffixes are not specified in `php_phpmd_suffixes` variable. Otherwise, only lint files with given extensions. It may be a bit overcomplicated, but then we will keep ability to prevent linting of files with certain extensions and user will not have to set up anything initially to make phpmd work

w0rp: Yeah, I think that could be a good option. The the variable is an empty string, default to using the current file extension, if there is one. Otherwise use the extensions from the variable.

w0rp:
Closed due to inactivity. Update the PR and let us know to review it again when ready.

Issue #1178, opened 2017-11-29
eslint fixer with eslint_d replaces file content on save


miuirussia:
1. Make index.js with any content
2. init node project without eslint (package.json)
3. Saving index.js results replaced file content with:
```
Error: No ESLint configuration found.
```
:ALEInfoToClipboard
```
 
 Current Filetype: javascript.jsx
Available Linters: ['eslint', 'flow', 'jscs', 'jshint', 'standard', 'xo']
  Enabled Linters: ['eslint']
 Linter Variables:

let g:ale_javascript_eslint_executable = 'eslint_d'
let g:ale_javascript_eslint_options = ''
let g:ale_javascript_eslint_suppress_eslintignore = 0
let g:ale_javascript_eslint_suppress_missing_config = 0
let g:ale_javascript_eslint_use_global = 1
 Global Variables:

let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 1
let g:ale_fixers = {'javascript': ['eslint'], 'css': ['prettier']}
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'always'
let g:ale_linter_aliases = {}
let g:ale_linters = {'haskell': ['stack-ghc'], 'javascript': ['eslint']}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_column_always = 0
let g:ale_sign_error = '•'
let g:ale_sign_offset = 1000000
let g:ale_sign_warning = '•'
let g:ale_statusline_format = ['%d error(s)', '%d warning(s)', 'OK']
let g:ale_warn_about_trailing_whitespace = 1
  Command History:

(executable check - success) eslint_d
(started) ['/bin/sh', '-c', '''eslint_d'' -f unix --stdin --stdin-filename ''/Users/kirill/Projects/experiments/index.js'' < ''/var/folders/h9/3syz1kss55j3srkcmk6kfx_h0000gn/T/nvimFJBM9J/2/index.js''']
[...]

PR #1802, opened 2018-08-09
Don't use error output from prettier_d for fixing files

flybayer:
`prettier_d` has the same problem as `eslint_d` where errors are output via stdout instead of stderr.

This PR ports the eslint fix for #1178 to fix the problem for prettier_d.


w0rp:
Cheers! :beers:

PR #1806, opened 2018-08-10
Fix #1684: Use the correct flag name in rustc linter

### LINKED ISSUE #1684: Consider using `rustc --emit metadata` by default
See #1031

`-Z no-trans` only works with `nightly` Rust versions. `--emit metadata` doesn't generate a binary, and might work in more versions. Someone should find out which versions of Rust support that option and consider using that option by default. The default shouldn't be changed until we know that it's safe to do so.

jimenezrick:
The rust compiler renamed the option '-Z no-trans' to '-Z no-codegen'.

https://github.com/rust-lang/rust-enhanced/issues/281

### DIFF
--- a/ale_linters/rust/rustc.vim
+++ b/ale_linters/rust/rustc.vim
@@ -1,7 +1,7 @@
 " Author: Daniel Schemala <istjanichtzufassen@gmail.com>
 " Description: rustc for rust files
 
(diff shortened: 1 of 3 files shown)


w0rp:
Cheers! :beers:

PR #1809, opened 2018-08-12
Update README with the `*` ft option for the fixers object

nrhodes91:
It wasn't immediately obvious that the `g:ale_fixers` cannot be a list,
and would allow the use of `*` to match all filetypes. I was hoping to
add a bit more detail to the README to make this clearer.


nrhodes91:
Let me know what you think of the latest changes - I've merged it in with the previous paragraph to keep things concise.

w0rp:
Cheers! :beers:

PR #1815, opened 2018-08-15
If present, use new-ish LSP 'relatedInformation' field for :ALEDetail

fredemmott:
screen shot 2018-08-15 at 2 30 05 pm


w0rp:
Cheers! :beers:

PR #1820, opened 2018-08-16
Added phpcs options support as described in #257

SuRaMoN:
Added phpcs options support as described in #257

added unit tests
tested it and works
i ran everything with vint and it have an ok
i also updated the doc files


w0rp:
Cheers! :beers:

PR #1836, opened 2018-08-22
Add hlint refactoring as a fixer

eborden:
`hlint` exposes a command for automatically refactoring its lint
suggestions. This fits perfectly into the form of a fixer. This commit
provides `hlint` as a fixer with a simple invocation.

This commit presents two problems:

1. It introduces `haskell_hlint_executable` as a global var for setting
   the desired `hlint` executable. This global is not currently respected
   by the `hlint` linter.
2. `ale#Escape` does not play nicely with a common pattern in haskell
   executable usage. Many `stack` users invoke compiler tools via
   `stack exec [command] -- [options]`. `ale#Escape` wraps this command in
   quotes, because it includes spaces. This results in silent failure of
   the fixer.

This commit does not present resolutions for either of these problems.

### DIFF
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -170,6 +170,11 @@ let s:default_registry = {
 \       'suggested_filetypes': ['haskell'],
 \       'description': 'Fix Haskell files with brittany.',
 \   },
+\   'hlint': {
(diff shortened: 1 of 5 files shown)


eborden:
So the user would set `haskell_hlint_executable` to `stack` and we'd append `hlint --`?

PR #1837, opened 2018-08-22
Add stylish-haskell as a fixer

eborden:
`stylish-haskell` is a common formatting tool for the haskell toolchain.
It is not as advanced as `brittany` or `hindent`, but it is commonly
used for formatting of imports and data declarations. This adds it as a
fixer in ALE.


w0rp:
Cheers! :beers:

Issue #1834, opened 2018-08-22
Unexpected "go build" errors not handled gracefully


arp242:
```
$ export GOTMPDIR=/tmp/gotmpdir
$ go build a.go
stat /tmp/gotmpdir: no such file or directory
$ echo $?
1
```

Fair enough; but the error reporting is not so good:

```
:let g:ale_linters = {'go': ['go build']}
:ALELint
Error detected while processing function <SNR>88_VimCloseCallback[14]..<SNR>99_HandleExit[33]..<SNR>99_InvokeChain[1]..ale#engine#ProcessChain[19]..ale_linters#go#gobuild#GetCommand:
line    4:
E684: list index out of range: 0
E15: Invalid expression: {   'GOPATH': a:goenv_output[0],   'GOROOT': a:goenv_output[1],}
```

From `:ALEInfo` the failed call seems to be:

    (finished - exit code 1) ['/bin/zsh', '-c', 'go env GOPATH GOROOT']

Which errors out in the same way as `go build`:

    $ go env GOPATH GOROOT
    stat /tmp/gotmpdir: no such file or directory

---

When using `:let g:ale_linters = {'go': ['gometalinter']}` it doesn't error out, but simply doesn't do anything:

```
(finished - exit code 1) ['/bin/zsh', '-c', 'cd ''/home/martin/work/src/github.com/teamwork/hubc'' && ''gometalinter'' --include=''hubc\.go'' --disable-all --tests --enable=vet --enable=errch
eck --enable=ineffassign --enable=goimports --enable=misspell --enable=lll --line-length=120 --enable=golint .']
<<<NO OUTPUT RETURNED>>>
```

---

I'll see if I can fix this later, just wanted to report/document the issue so I don't forget :-)

<details>
<summary>Version information</summary>

## Information

**VIM version**

    VIM - Vi IMproved 8.1 (2018 May 18, compiled Aug 13 2018 14:59:01)
    Included patches: 1-279

Operating System: Arch Linux

### :ALEInfo

```
 Current Filetype: go
Available Linters: ['gobuild', 'gofmt', 'golint', 'gometalinter', 'gosimple', 'gotype', 'govet', 'staticcheck']
   Linter Aliases:
'gobuild' -> ['go build']
'govet' -> ['go vet']
  Enabled Linters: ['gometalinter']
 Suggested Fixers: 
  'gofmt' - Fix Go files with go fmt.
  'goimports' - Fix Go files imports with goimports.
[...]

PR #1840, opened 2018-08-23
Remove "go env" from gobuild linter

Carpetsmoker:
I see no reason to do this? It is just setting the environment to what
it already is?

It was originally added in #297, but that entire PR is not a great idea
in the first place; that PR (together with #270) tried to make the Go do
non-standard and non-supported stuff like compiling packages outside of
GOPATH.

That's not something that works well (I tried), so was eventually
removed in #465, but these "go env" calls remained, for no reason in
particular, as far as I can think of.

This will improve on #1834; you will now no longer get a confusing error
(but still won't get a meaningful error; need to think how to do that).

### DIFF
--- a/ale_linters/go/gobuild.vim
+++ b/ale_linters/go/gobuild.vim
@@ -5,35 +5,11 @@
 
 call ale#Set('go_gobuild_options', '')
 
-function! ale_linters#go#gobuild#ResetEnv() abort
-    unlet! s:go_env
-endfunction
-
-function! ale_linters#go#gobuild#GoEnv(buffer) abort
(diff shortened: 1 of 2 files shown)


w0rp:
@joshuarubin Can you recall why `go env` was needed here? I don't know much about Go myself.

joshuarubin:
Looking at the code, I can't say I remember. Perhaps we had to temporarily modify the GOPATH for some operation. Doesn't look necessary now though.

w0rp:
Let's remove this and see what happens. Cheers! :beers:

PR #1841, opened 2018-08-23
importjs fixer not reading correct executable variable name

dabbeg:
importjs fixer is not reading the correct executable variable name.
Like referenced in the help section the variable name should be "javascript_..." but not "js_..."

### DIFF
--- a/autoload/ale/fixers/importjs.vim
+++ b/autoload/ale/fixers/importjs.vim
@@ -1,7 +1,7 @@
 " Author: Jeff Willette <jrwillette88@gmail.com>
 " Description: Integration of importjs with ALE.
 
-call ale#Set('js_importjs_executable', 'importjs')
+call ale#Set('javascript_importjs_executable', 'importjs')
 
 function! ale#fixers#importjs#ProcessOutput(buffer, output) abort
     let l:result = ale#util#FuzzyJSONDecode(a:output, [])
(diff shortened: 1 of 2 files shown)


autoload/ale/fixers/importjs.vim
```diff
@@ -9,7 +9,7 @@ function! ale#fixers#importjs#ProcessOutput(buffer, output) abort
 endfunction
 
 function! ale#fixers#importjs#Fix(buffer) abort
-    let l:executable = ale#Var(a:buffer, 'js_importjs_executable')
+    let l:executable = ale#Var(a:buffer, 'javascript_importjs_executable')
```
w0rp: Update the variable above too, and in the documentation, if it's documented there. If the option is documented, write `get(g:, 'ale_js_importjs_executable', 'importjs')` above so we won't break anything for people who are using the old variable name.

dabbeg: It is documented as:
```
importjs                                              *ale-javascript-importjs*

g:ale_javascript_importjs_executable     *g:ale_javascript_importjs_executable*
                                         *b:ale_javascript_importjs_executable*
  Type: |String|
  Default: `'importjs'`


```

So the code is not in sync with the documentation or vice versa.
Do you still want the 'get(...)' for backwards compatibility?

w0rp: Nah, if it's not documented, then don't worry about it.

w0rp:
Cheers! :beers:

PR #1845, opened 2018-08-24
Fix 1844 - Fix broken kotlin lsp linter.

hsanson:
This fixes kotlin lsb linter that broke three weeks ago after some heavy refactoring in ALE ( 2172843 ).


w0rp:
I'll fix that by using `'%e' for the command, and I'll add a test for it.

Issue #1403, opened 2018-03-06
ALE not working with a locally installed Rubocop


ryanma:
## Information

**VIM version**
```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Feb 12 2018 14:37:54)
Included patches: 1-2
```

Operating System: CentOS 6

### :ALEInfo
```
Current Filetype: ruby
Available Linters: ['brakeman', 'rails_best_practices', 'reek', 'rubocop', 'ruby']
Enabled Linters: ['rubocop', 'ruby']
Linter Variables:
let g:ale_ruby_rubocop_executable = '/export/home/ryanma/.gem/ruby/2.4.0/bin/rubocop'
let g:ale_ruby_rubocop_options = ''
let g:ale_ruby_ruby_executable = 'ruby'
Global Variables:
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {'ruby': ['rubocop']}
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'always'
let g:ale_linter_aliases = {}
let g:ale_linters = {'ruby': ['ruby', 'rubocop']}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_column_always = 1
let g:ale_sign_error = '>>'
let g:ale_sign_offset = 1000000
let g:ale_sign_warning = '--'
let g:ale_statusline_format = ['%d error(s)', '%d warning(s)', 'OK']
let g:ale_warn_about_trailing_whitespace = 1
Command History:
(executable check - success) /export/home/ryanma/.gem/ruby/2.4.0/bin/rubocop
(finished - exit code 1) ['/bin/bash', '-c', '''/export/home/ryanma/.gem/ruby/2.4.0/bin/rubocop'' --format json --force-exclusion --stdin ''<path_to_file>/config/deploy.rb'' < ''/tmp/vEUIODm/5/deploy.rb''']
<<<NO OUTPUT RETURNED>>>
[...]

PR #1850, opened 2018-08-24
Update all Ruby linters to work consistently with bundler

rrosenblum:
This is in response to #1403

### DIFF
--- a/ale_linters/ruby/brakeman.vim
+++ b/ale_linters/ruby/brakeman.vim
@@ -1,8 +1,17 @@
 " Author: Eddie Lebow https://github.com/elebow
 " Description: Brakeman, a static analyzer for Rails security
 
+call ale#Set('ruby_brakeman_options', '')
+call ale#Set('ruby_brakeman_executable', 'brakeman')
+
 let g:ale_ruby_brakeman_options =
 \   get(g:, 'ale_ruby_brakeman_options', '')
+
+function! ale_linters#ruby#brakeman#GetExecutable(buffer) abort
+    let l:executable = ale#Var(a:buffer, 'ruby_brakeman_executable')
+
+    return ale#handlers#ruby#EscapeExecutable(l:executable, 'brakeman')
+endfunction
 
 function! ale_linters#ruby#brakeman#Handle(buffer, lines) abort
     let l:output = []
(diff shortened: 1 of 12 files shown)


rrosenblum:
@w0rp to clarify, are you wanting me to implement `ale#handlers#haskell_stack#EscapeExecutable` for ruby?

```vim
function! ale#handlers#haskell_stack#EscapeExecutable(executable, stack_exec) abort
    let l:exec_args = a:executable =~? '^stack$'
    \   ? ' exec ' . ale#Escape(a:stack_exec) . ' --'
    \   : ''
    return ale#Escape(a:executable) . l:exec_args
endfunction
```

rrosenblum:
@w0rp I am having some trouble implementing the `EscapeExecutable` function the same way it is written in #1851.

The problem is that by using `EscapeExecutable` inside of `GetExecutable` causes `ale#engine#IsExecutable` to not think the command is executable. This appears to be caused by the command now being made up of multiple words `bundle exec rubocop` rather than a single command `bundle`. I have tried using different escaping, but I haven't found anything that causes `IsExecutable` to pass. What is the best way to move forward with this change?

ale_linters/ruby/brakeman.vim
```diff
@@ -1,9 +1,17 @@
...
 
+call ale#Set('ruby_brakeman_options', '')
+call ale#Set('ruby_brakeman_executable', 'brakeman')
+
 let g:ale_ruby_brakeman_options =
 \   get(g:, 'ale_ruby_brakeman_options', '')
 
+function! ale_linters#ruby#brakeman#GetExecutable(buffer) abort
+    let l:executable = ale#Var(a:buffer, 'ruby_brakeman_executable')
+    return ale#handlers#ruby#EscapeExecutable(l:executable, 'brakeman')
```
w0rp: The answer here is that you should just return the executable here wihout escaping it, and you should use `ale#handlers#ruby#EscapeExecutable` in `GetCommand` below. The reason being that `executable_callback` should return the path to the program that will be run, and `command_callback` should return a command to run, where the part of the command with an executable name has to be escaped.

ale_linters/ruby/brakeman.vim
```diff
@@ -27,20 +35,22 @@ function! ale_linters#ruby#brakeman#Handle(buffer, lines) abort
...
+    return ale#Escape(l:executable)
+    \    . ' -f json -q '
     \    . ale#Var(a:buffer, 'ruby_brakeman_options')
     \    . ' -p ' . ale#Escape(l:rails_root)
 endfunction
 
 call ale#linter#Define('ruby', {
 \    'name': 'brakeman',
-\    'executable': 'brakeman',
+\    'executable_callback': 'ale_linters#ruby#brakeman#GetExecutable',
```
w0rp: When the callbacks for `executable_callback` just `return ale#Var(a:buffer, variable_name)`, you can write `executable_calback: ale#VarFunc(variable_name)` instead to get the same result.

ale_linters/ruby/brakeman.vim
rrosenblum: Ok, cool. That is what I was thinking, but I wanted to check with you first.

rrosenblum:
@w0rp this is ready for review again.

w0rp:
The callbacks only need to be adjusted slightly. I'll sort that out and then merge this.

w0rp:
Cheers! :beers:

stefansedich:
This is great, but one thing I want to be able to do is run `bundle` when there is a project Gemfile and fall back to my locally installed rubocop if I am editing a random ruby file without a Gemfile. Is there a nice way that I could achieve this currently?

rrosenblum:
Executable fallback would be useful. I don't think that there is a way to get it working with the current implementation. There is some code that checks to make sure the executable command is executable so it might be possible to add this functionality. What are your thoughts @w0rp?

PR #1851, opened 2018-08-24
Add better support for Haskell stack compiler tools

eborden:
`ale#Escape` does not play nicely with a common pattern in haskell
executable usage. Many `stack` users invoke compiler tools via
`stack exec [command] -- [options]`. `ale#Escape` wraps this command in
quotes, because it includes spaces. This results in silent failure of
the fixer.

This commit adds support for `stack` as the executable of a tool. This
follows a pattern that has been implemented for `bundler`'s tool chain.


eborden:
This solution was mentioned here https://github.com/w0rp/ale/pull/1836

eborden:
Still need to add some tests for `hlint` and convert some other tools to `stack` support.

eborden:
@w0rp I've run into an issue where linters that use `executable_callback` fail with `'stack' exec 'foo' --`. Do you know of way to fix this?

ale_linters/haskell/ghc_mod.vim
```diff
@@ -0,0 +1,18 @@
...
+function! ale_linters#haskell#ghc_mod#GetCommand (buffer) abort
+    let l:executable = ale#Var(a:buffer, 'haskell_ghc_mod_executable')
+    return ale#handlers#haskell_stack#EscapeExecutable(l:executable, 'ghc-mod')
+\       . ' --map-file %s=%t check %s'
+endfunction
+
+call ale#linter#Define('haskell', {
+\   'name': 'ghc_mod',
+\   'aliases': ['ghc-mod'],
+\   'executable': 'ghc-mod',
```
w0rp: You should add `executable_callback` here and you can use `ale#VarFunc(''haskell_ghc_mod_executable')` for that. Then ALE will check if the stack executable is executable before running the linter, instead of just `ghc-mod`.

ale_linters/haskell/hlint.vim
```diff
@@ -26,9 +26,16 @@ function! ale_linters#haskell#hlint#Handle(buffer, lines) abort
...
+function! ale_linters#haskell#hlint#GetCommand(buffer) abort
+    return ale#handlers#hlint#GetExecutable(a:buffer)
+\        . ' --color=never'
+\        . ' --json'
+\        . ' -'
+endfunction
+
 call ale#linter#Define('haskell', {
 \   'name': 'hlint',
 \   'executable': 'hlint',
```
w0rp: Use `executable_callback` here too.

ale_linters/haskell/ghc_mod.vim
eborden: So with this method the `%e` is not viable in the command?

w0rp: The command you have is fine, you just need to tell ALE what the executable is too, so it can check if that's executable.

eborden:
@w0rp This is good to go. All relevant haskell executables have been updated to support `stack exec`.

eborden:
@w0rp this has been rebased to account for conflicts in #1908 and is good to go.

autoload/ale/handlers/haskell_stack.vim
```diff
@@ -0,0 +1,7 @@
+function! ale#handlers#haskell_stack#EscapeExecutable(executable, stack_exec) abort
+    let l:exec_args = a:executable =~? '^stack$'
```
w0rp: This checks for an executable which exactly matches `'stack'`, so this is the same as `is? 'stack'`. Use `=~? 'stack$'` instead so any path ending in `'stack'` will be treated like a stack executable.

ale_linters/haskell/ghc_mod.vim
```diff
@@ -0,0 +1,23 @@
...
+function! ale_linters#haskell#ghc_mod#GetCommand (buffer) abort
+    let l:executable = ale#Var(a:buffer, 'haskell_ghc_mod_executable')
+
+    return ale#handlers#haskell_stack#EscapeExecutable(l:executable, 'ghc-mod')
+\       . ' --map-file %s=%t check %s'
+endfunction
+
+function! ale_linters#haskell#ghc_mod#GetExecutable(buffer) abort
+    return ale#Var(a:buffer, 'haskell_ghc_mod_executable')
+endfunction
```
w0rp: This function isn't used any more, so you can remove it.

ale_linters/haskell/ghc_mod.vim
```diff
@@ -0,0 +1,23 @@
+" Author: wizzup <wizzup@gmail.com>
+" Description: ghc-mod for Haskell files
+
+call ale#Set('haskell_ghc_mod_executable', 'ghc-mod')
+
+function! ale_linters#haskell#ghc_mod#GetCommand (buffer) abort
+    let l:executable = ale#Var(a:buffer, 'haskell_ghc_mod_executable')
+
+    return ale#handlers#haskell_stack#EscapeExecutable(l:executable, 'ghc-mod')
+\       . ' --map-file %s=%t check %s'
```
w0rp: Put the line continuation character on the same level of indentation as the previous line.

w0rp:
Cheers! :beers:

zkourouma:
could @eborden give a quick explanation of what this does and what changes need to be made to one's vimrc? from https://github.com/w0rp/ale/pull/1836 it seems like we can just set `let g:ale_haskell_hlint_executable = 'stack'` and ale will use whichever hlint stack has access to?

eborden:
@zkourouma Yup, you just need to set the executable to `stack` and then the package name will be used to execute it from `stack`. You'll have to do `stack build hlint --copy-compiler-tool` to get the package available though. 

Alexis King covers reason for `--copy-compiler-tool` in this great post https://lexi-lambda.github.io/blog/2018/02/10/an-opinionated-guide-to-haskell-in-2018/

Issue #1443, opened 2018-03-22
Support settings options for pyls


w0rp:
As with many other options, add an `ale_python_pyls_options` variable for defining options for `pyls`, with documentation and tests.

andrewminer:
According to the folks over at Palantir (palantir/python-language-server#312), this needs to be done by sending a `workspace/didChangeConfiguration` message to `pyls` once it has been started.  This will allow you to configure which linters are used.

etaoins:
This is useful for RLS as well: https://github.com/rust-lang-nursery/rls#configuration

eugene-bright:
As I understand the work is not done yet and the value is not passed to `ale#linter#Define`.
Direct  call to `DidChangeConfiguration` doesn't make an effect either.
```
:let a = ale#lsp#message#DidChangeConfiguration(bufnr('%'), g:ale_python_pyls_config)
```
Maybe `0` value is not equivalent to JSON's `false`.
Also I struggle to debug pyls without running it with options `-vv --log-file ~/pyls.log`. Could you hint me what to do?

eugene-bright:
As I now see `ale#lsp#message#DidChangeConfiguration` function is only do value wrapping.
The yielding value is consumed by `ale#lsp#Send` while `ale#lsp_linter#StartLSP` is executing.
To make all this work the dictionary must be passed as argument `lsp_config` to function `ale#linter#Define`.

I've passed a value manually to check my hypothesis.
```
diff --git a/ale_linters/python/pyls.vim b/ale_linters/python/pyls.vim
index eceb63f6..7b00cc1f 100644
--- a/ale_linters/python/pyls.vim
+++ b/ale_linters/python/pyls.vim
@@ -44,4 +44,5 @@ call ale#linter#Define('python', {
 \   'command_callback': 'ale_linters#python#pyls#GetCommand',
 \   'project_root_callback': 'ale#python#FindProjectRoot',
 \   'completion_filter': 'ale#completion#python#CompletionItemFilter',
+\   'lsp_config': {'pyls': {'plugins': {'pycodestyle': {'enabled': 0}}}},
 \})
```

I've prepared a patch #2088 to pass more args to pyls. Now debug logging can be forced by a few strings in init.vim:
```
let g:ale_python_pyls_extra_args = [
\   '-vv',
\   '--log-file', expand('~/pyls.log')
\]
```

Now I can see that settings are passed correctly
```
[...]

PR #1852, opened 2018-08-24
LSP configuration via didChangeConfiguration

danielwelch:
Addressing #1443

>Disclaimer: This is my first time writing vimscript. Feedback is extremely welcome. I'm sure I missed the boat on something here but I'm happy to help any way I can with some guidance.

Adding the ability to configure LSPs via the [`workspace/didChangeConfiguration`](https://microsoft.github.io/language-server-protocol/specification#workspace_didChangeWorkspaceFolders) method. I've added the `lsp_config` key to linter definition, added a new `ale#lsp#message#DidChangeConfiguration`, and added a call to `ale#lsp#Send` this message during `ale#lsp#StartLSP` if `lsp_config` is set.

This leaves it up to each LSP linter to define its own configuration variable and pass that to the linter definition. I've added an example implementation for `pyls`.

### DIFF
--- a/autoload/ale/linter.vim
+++ b/autoload/ale/linter.vim
@@ -255,6 +255,13 @@ function! ale#linter#PreProcess(filetype, linter) abort
         elseif has_key(a:linter, 'initialization_options')
             let l:obj.initialization_options = a:linter.initialization_options
         endif
+
+        if has_key(a:linter, 'lsp_config')
+            if type(a:linter.lsp_config) isnot v:t_dict
+                throw '`lsp_config` must be a Dictionary'
+            endif
+        endif
+
     endif
 
     let l:obj.output_stream = get(a:linter, 'output_stream', 'stdout')
--- a/autoload/ale/linter.vim
+++ b/autoload/ale/linter.vim
@@ -260,6 +260,7 @@ function! ale#linter#PreProcess(filetype, linter) abort
             if type(a:linter.lsp_config) isnot v:t_dict
                 throw '`lsp_config` must be a Dictionary'
             endif
+            let l:obj.lsp_config = a:linter.lsp_config
         endif
 
     endif
--- a/autoload/ale/linter.vim
+++ b/autoload/ale/linter.vim
@@ -260,9 +260,9 @@ function! ale#linter#PreProcess(filetype, linter) abort
             if type(a:linter.lsp_config) isnot v:t_dict
                 throw '`lsp_config` must be a Dictionary'
             endif
+
             let l:obj.lsp_config = a:linter.lsp_config
         endif
-
     endif
 
     let l:obj.output_stream = get(a:linter, 'output_stream', 'stdout')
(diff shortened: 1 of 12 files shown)


danielwelch:
Are we sure these are the same thing? I'm most familiar with the implementation of RLS, so I'll use that as an example... Here's the [`initialization_options`](https://github.com/rust-lang-nursery/rls/blob/f92b4e8db8c9e4cfa032972053ce6491fb25712f/src/lsp_data.rs#L263) available to RLS and [here are the configuration options](https://github.com/rust-lang-nursery/rls/blob/master/src/config.rs#L118) available via `workspace/didChangeConfiguration`.

w0rp:
I guess they are different. For `pyls`, the options people will really care about can all be given in `initialization_options`, like disabling or enabling particular plugins.

w0rp:
Add support for `didChangeConfiguration`, but take away your changes for `pyls` for now.

danielwelch:
I removed the pyls changes. I'll probably add the RLS changes in a separate pull request once this is merged.

autoload/ale/lsp_linter.vim
```diff
@@ -180,6 +180,14 @@ function! ale#lsp_linter#StartLSP(buffer, linter) abort
 
     let l:language_id = ale#util#GetFunction(a:linter.language_callback)(a:buffer)
 
+    if has_key(a:linter, 'lsp_config')
```
w0rp: You can use `if !empty(get(a:linter, 'lsp_config'))` here.

You'll have to update `ale#linter#PreProcess` to accept `lsp_config` as an option, and check the type, and update the tests to check that it's allowed and that it appears in the output okay.

tmc:
would love to see this land.

autoload/ale/lsp_linter.vim
w0rp: Look at what the function does. The `ale#Define` options are validated and new Dictionary values are created from the options. That function hasn't been modified, so this option doesn't work yet.

autoload/ale/linter.vim
```diff
@@ -255,6 +255,13 @@ function! ale#linter#PreProcess(filetype, linter) abort
         elseif has_key(a:linter, 'initialization_options')
             let l:obj.initialization_options = a:linter.initialization_options
         endif
+
+        if has_key(a:linter, 'lsp_config')
+            if type(a:linter.lsp_config) isnot v:t_dict
+                throw '`lsp_config` must be a Dictionary'
+            endif
+        endif
```
w0rp: You need to copy the dictionary to `l:obj`, and add tests for the new option.

danielwelch: Thanks for the feedback. I'll work on tests when I get the chance.

tmc:
Great to see progress here!

tmc:
What is missing here, tests?

RyanSquared:
Some tests that make sure that bad input is thrown would be nice, I think. You should always test every possible branch, *especially* the ones you want to fail.

danielwelch:
Yep, tests are still missing. Specifically the PreProcess tests still need to be written.

danielwelch:
@w0rp added tests to check PreProcess (to check that `lsp_config` is accepted and processed appropriately, and that it throws if `lsp_config` is not a dictionary)

w0rp:
Cheers! :beers:

jparise:
Would this also benefit from an `lsp_config_callback` variant so that `lsp_config` doesn't need to be a static dictionary and to support `ale#VarFunc`?

PR #1865, opened 2018-08-27
Support options for puppet parser validate

TheLocehiliosan:
It can be necessary to pass options to the puppet parser validation. The
most glaring example of this is when using Puppet 3, with the
`parser = future` option enabled. This update allows adding
`--parser=future` to the options passed to Puppet.

### DIFF
--- a/ale_linters/puppet/puppet.vim
+++ b/ale_linters/puppet/puppet.vim
@@ -1,5 +1,8 @@
 " Author: Alexander Olofsson <alexander.olofsson@liu.se>
 
+call ale#Set('puppet_puppet_executable', 'puppet')
+call ale#Set('puppet_puppet_options', '--color=false')
+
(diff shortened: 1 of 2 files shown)


ale_linters/puppet/puppet.vim
```diff
@@ -1,5 +1,8 @@
 " Author: Alexander Olofsson <alexander.olofsson@liu.se>
 
+call ale#Set('puppet_puppet_executable', 'puppet')
+call ale#Set('puppet_puppet_options', '--color=false')
```
w0rp: We should probably take `--color=false` out of the configurable options here, and write that below. Otherwise people will be able to take the option away and make it so ALE can't parse the output.

TheLocehiliosan: Yes. I initially coded it that way, but then somehow talked myself out of it. I’ll push another commit later tonight when I’m at a computer again.

PR #1870, opened 2018-08-30
Fix #1822 - support go-langserver lsp.

### LINKED ISSUE #1822: Golang Language Server support
At the moment, there doesn't seem to be support for the go-langserver.  Only for type script, can we please add in support for Go?

hsanson:
Add support for go-langserver LSP.

Closes #1822


hsanson:
I need some help getting the tests passing on Windows. They work perfectly on my Ubuntu/KdeNeon machine.  The test depends on the GOPATH env variable being set correctly and that is what I think is not working on windows.

autoload/ale/path.vim
```diff
@@ -47,6 +47,24 @@ function! ale#path#FindNearestDirectory(buffer, directory_name) abort
     return ''
 endfunction
 
+" Given a buffer and a directory name, find the parent directory by searching upwards
+" through the paths relative to the given buffer.
+function! ale#path#FindParentDirectory(buffer, directory_name) abort
+    let l:buffer_filename = fnamemodify(bufname(a:buffer), ':p')
+    let l:buffer_filename = fnameescape(l:buffer_filename)
+
+    let l:relative_path = finddir(a:directory_name, l:buffer_filename . ';')
```
w0rp: I don't think we need to search the filesystem here. See below.

autoload/ale/go.vim
```diff
@@ -0,0 +1,23 @@
...
+" Description: Functions for integrating with Go tools
+
+" Find the nearest dir listed in GOPATH and assume it the root of the go
+" project.
+function! ale#go#FindProjectRoot(buffer) abort
+
+    let l:sep = has('win32') ? ';' : ':'
+
+    for l:name in split($GOPATH, l:sep)
+        let l:dir = fnamemodify(
```
w0rp: Paths in GOPATH must be absolute paths, so you can just do this. 

```vim
let l:filename = ale#path#Simplify(expand('#' . a:buffer . ':p'))

for l:name in split($GOPATH, l:sep)
    let l:path_dir = ale#Path#Simplify(l:name)

    " Use the directory from GOPATH if the current filename starts with it.
    if l:filename[: len(l:path_dir) - 1] is? l:path_dir
        return l:path_dir
    endif
endfor

return ''
```

test/command_callback/test_golangserver_command_callback.vader
```diff
@@ -0,0 +1,25 @@
+
+Before:
+  call ale#assert#SetUpLinterTest('go', 'golangserver')
+  let sep = has('win32') ? ';' : ':'
+  let $GOPATH=ale#path#Simplify(g:dir . '/go_paths/go1')
```
w0rp: Write `Save $GOPATH` above, so the previous value of `GOPATH` will be restored after running these tests.

ale_linters/go/golangserver.vim
```diff
@@ -0,0 +1,17 @@
...
+function! ale_linters#go#golangserver#Executable(buffer) abort
+    return ale#Var(a:buffer, 'go_langserver_bin')
+endfunction
+
+call ale#linter#Define('go', {
+\   'name': 'golangserver',
+\   'lsp': 'stdio',
+\   'executable_callback': 'ale_linters#go#golangserver#Executable',
+\   'command': '%e',
+\   'language': 'go',
```
w0rp: You can omit the language here if you want. It will be taken from the first argument to `Define`.

ale_linters/go/golangserver.vim
```diff
@@ -0,0 +1,17 @@
...
+call ale#Set('go_langserver_bin', 'go-langserver')
+
+function! ale_linters#go#golangserver#Executable(buffer) abort
+    return ale#Var(a:buffer, 'go_langserver_bin')
+endfunction
+
+call ale#linter#Define('go', {
+\   'name': 'golangserver',
+\   'lsp': 'stdio',
+\   'executable_callback': 'ale_linters#go#golangserver#Executable',
```
w0rp: You have the option of using `ale#VarFunc` here to create a callback which returns the value of the option.

ale_linters/go/golangserver.vim
```diff
@@ -0,0 +1,17 @@
+" Author: Horacio Sanson <https://github.com/hsanson>
+" Description: Support for go-langserver https://github.com/sourcegraph/go-langserver
+
+call ale#Set('go_langserver_bin', 'go-langserver')
```
w0rp: Name the variable `_executable` to be consistent with other linters.

hsanson:
@w0rp took care of all your recommendations.

doc/ale-go.txt
```diff
@@ -114,5 +114,16 @@ g:ale_go_staticcheck_lint_package           *g:ale_go_staticcheck_lint_package*
   current file.
 
 
+===============================================================================
+golangserver                                              *ale-go-golangserver*
+
+g:ale_go_langserver_executable                 *g:ale_go_langserver_executable*
+                                               *b:ale_go_langserver_executable*
+  Type: |String|
+  Default: `'go-langserver'`
```
diegs: Recent `go-langserver` supports completion with the `-gocodecompletion` flag.

Would it be possible to add this flag here and use ALE for Go completion as well? I've been successful in using it with `vim-lsp` and `asyncomplete-vim` but it would be great to be able to drop those plugins and rely on ALE 100% for linting, LSP, and completion.

I can try tackling it myself in a follow-up PR too if you want to submit this one first.

diegs: Ok, actually took a crack at this and it works like a treat. Here is my augmented PR:

https://github.com/diegs/ale/pull/1

Feel free to add this in here, or I can rebase once you submit and send a follow-up PR.

hsanson:
@diegs very nice addition... did not know that go-langserver supported autocompletion. I added this improvement to this PR and works perfectly. Thanks.

w0rp:
Check out the new errors on Travis CI. I recently made it so Travis CI checks for more style gripes. You should see the same messages as you edit ALE project files magically if Python is installed.

autoload/ale/go.vim
```diff
@@ -0,0 +1,23 @@
...
+    for l:name in split($GOPATH, l:sep)
+      let l:path_dir = ale#path#Simplify(l:name)
+
+      " Use the directory from GOPATH if the current filename starts with it.
+      if l:filename[: len(l:path_dir) - 1] is? l:path_dir
+          return l:path_dir
+      endif
+    endfor
+
+    return ''
```
Carpetsmoker: Note this function won't work with go modules. It should return any directory with a `go.mod` file as a root directory too.

autoload/ale/go.vim
```diff
@@ -0,0 +1,23 @@
...
+
+" Find the nearest dir listed in GOPATH and assume it the root of the go
+" project.
+function! ale#go#FindProjectRoot(buffer) abort
+
+    let l:sep = has('win32') ? ';' : ':'
+
+    let l:filename = ale#path#Simplify(expand('#' . a:buffer . ':p'))
+
+    for l:name in split($GOPATH, l:sep)
```
Carpetsmoker: `$GOPATH` may be empty; from `go help gopath`:

> If the environment variable is unset, GOPATH defaults
to a subdirectory named "go" in the user's home directory
($HOME/go on Unix, %USERPROFILE%\go on Windows),
unless that directory holds a Go distribution.
Run "go env GOPATH" to see the current GOPATH.

ale_linters/go/golangserver.vim
```diff
@@ -0,0 +1,19 @@
+" Author: Horacio Sanson <https://github.com/hsanson>
+" Description: Support for go-langserver https://github.com/sourcegraph/go-langserver
+
+call ale#Set('go_langserver_executable', 'go-langserver')
+
+function! ale_linters#go#golangserver#GetCommand(buffer) abort
+    let l:executable = ale#Var(a:buffer, 'go_langserver_executable')
+    let l:completion_enabled = ale#Var(a:buffer, 'completion_enabled')
+    let l:exec_args = l:completion_enabled == 1 ? ' -gocodecompletion' : ''
```
Carpetsmoker: Could you add a general `go_langserver_flags` variable? There are several options that may be useful to set by the user.

autoload/ale/go.vim
Carpetsmoker: Actually, what is this intended to return? If the code is in `/home/martin/go/src/github.com/me/project/foo/code.go`, then should it return `/home/martin/go/src/github.com/me/project` or `/home/martin/go`? The function name suggests the former, but it actually seems to do the latter.

hsanson: @Carpetsmoker I have tested with both root paths and go-langserver works perfectly. Even setting the root path to /tmp or $USER works ok. I can get auto-completion for files in any project installed in my machine. 

So I opted to set as root the first path that matches an entry in $GOPATH.

autoload/ale/go.vim
hsanson: Added code to return $HOME/go on Unix or  %USERPROFILE%\go on Windows if  they exists and GOPATH is not defined.

doc/ale-go.txt
hsanson: Added it but slightly different as now we have ale_go_langserver_options variable.

autoload/ale/go.vim
Carpetsmoker: I don't know what it does exactly, but it seems to me that if you're going to set it, you might as well set it to a correct value, and not a surprising incorrect one?

hsanson: It is setting the value based exactly on what `go help gopath` says it should be. So this should be the correct path for any environment:

 - 1. first try to find a root defined in GOPATH
 - 2. If GOPATH is not defined then set it to $HOME/go on Unix or %USERPROFILE%\go on Windows.
 - 3 If none of the above are defined or not exist then returns empty string, that disables the linter.

Carpetsmoker: But it's not the project's directory, it's the GOPATH?

hsanson: As I mentioned on a separate conversation the root path is irrelevant for go-langserver. The cases were we want the actual root path is for language servers that support special configuration files (e.g. .solargraph.yml for ruby language server). In this case we can search for these special configuration files and assume their directory as root project path.

The go-langserver does not support any configuration files so I avoid the trouble of finding the actual root path. Getting the actual root path of a go project is difficult (except for go modules that have a go.mod file) and has no actual use for go-langserver.

w0rp:
Cheers! :beers:

autoload/ale/go.vim
Carpetsmoker: Yes, I read that but "it just happens to work" is no reason to do the wrong thing. When it *does* matter in 6 months or 2 years everyone will be confused. Or people looking for this "get project root" utility function for something else will be confused.

It would be much better to just not do anything (e.g. set it to the current dir) than to a different value than you're telling people to set it.

Issue #1283, opened 2018-01-13
Scala support with SBT


Ophirr33:
It'd be nice if Scala projects could be linted by SBT, as currently the scalac checker basically just checks syntax. SBT recently [announced support](https://developer.lightbend.com/blog/2017-11-30-sbt-1-1-0-RC1-sbt-server/) for the language server protocol, so maybe that could be leveraged to help integrating with it?

I'd be willing to do some leg work on this if I can get some general pointers on how to about this.

Ophirr33:
I have a working prototype of this, but it needs some polishing before it's ready.

TK009:
@Ophirr33 Are you still working on this?

I tried this but didn't get it to work. Here is some documentation about sbt language server protocol: https://www.scala-sbt.org/1.x/docs/sbt-server.html

There seems to be two different modes
* Unix domain socket (on windows: named pipe)
* TCP socket

The first mode is the default in sbt 1.1.x and a bit simpler. It's the default mode and doesn't require token authentication, but [`ch_open`](https://github.com/vim/vim/blob/master/src/channel.c#L655) doesn't seem to have implementation for Unix domain sockets. Is there a way to connect to unix domain sockets in ALE/vim?

The TCP mode can be activated by setting `serverConnectionType` to `ConnectionType.Tcp` and seems to require an authentication token in `initializationOptions` which can be read from a file. 

I would be happy to help with this, but I'm quite beginner with vimscript and vim plugins, so I appreciate if someone can help or give me pointers.

Ophirr33:
Hey @TK009, I had dropped this but was actually thinking about picking it back up since I'm back to full time scala dev. You can see some of the old stuff I did over in https://github.com/ophirr33/ale. It currently works if you set the sbt server to tcp mode (and I have some debug printing stuff you'll want to delete), but unix sockets ofc don't work.
The three solutions I see going forward are:
1. create a program that connects to the sbt server and lets ale communicate to it over stdin and stdout (might need to implement file watching on active.json)
2. submit a patch to implement this stdin/stdout functionality into sbt itself
3. submit a patch to allow vim to open channels from unix sockets
Option 1 is probably the easiest solution, but it does make distribution a bit of a pain.
If I do jump back into this I'll comment back here

w0rp:
You can write a linter which connects to a TCP socket. See this recently added linter: https://github.com/w0rp/ale/blob/master/ale_linters/ruby/solargraph.vim

[...]

PR #1871, opened 2018-08-31
added sbtserver linter

Ophirr33:
Per w0rp's request and to make it easier for ongoing vim maintainability/portability, this pull request only supports sbtserver in tcp mode


Ophirr33:
addresses #1283

ale_linters/scala/sbtserver.vim
```diff
@@ -0,0 +1,37 @@
...
+    return l:project_root
+endfunction
+
+function! ale_linters#scala#sbtserver#GetAddress(buffer) abort
+    let l:address = ale#Var(a:buffer, 'scala_sbtserver_address')
+    if l:address is? ''
+        let l:project_root = ale_linters#scala#sbtserver#GetProjectRoot(a:buffer)
+        let l:active_file = l:project_root . '/project/target/active.json'
+        if !empty(glob(l:active_file))
+            let l:active = json_decode(join(readfile(l:project_root . '/project/target/active.json')))
```
w0rp: The `readfile` call here will block editing in Vim. Can we just use a setting for the address instead?

autoload/ale/linter.vim
```diff
@@ -99,7 +99,7 @@ function! ale#linter#PreProcess(filetype, linter) abort
     endif
 
     if index(['', 'socket', 'stdio', 'tsserver'], l:obj.lsp) < 0
-        throw '`lsp` must be either `''lsp''` or `''tsserver''` if defined'
+        throw '`lsp` must be either `''lsp''`, `''stdio''` or `''tsserver''` if defined'
```
w0rp: Change he other suggested string to `'socket'` while you're at it.

test/scala_fixtures/no_active_sbt_project/Main.scala
```diff
@@ -0,0 +1,5 @@
+object HelloAle {
+  def main(args: Array[String]): Int = {
+    println("Hello, ale!")
+  }
+}
```
w0rp: You can delete this file and the other files. The tests won't read them, and we shouldn't make the `readfile()` call.

Ophirr33:
active.json can potentially be different every time the sbtserver starts up, especially if there are multiple running servers (i.e., working on two separate projects at once necessitates different ports for the different server). Requiring the user to change the setting manually every time would be poor UX; is there a way to read files in vim asynchronously? If not, I can go back to using the wrapper at https://github.com/Ophirr33/sbtio

Ophirr33:
I guess there are two other options we could do:
1. recommend that the user override the serverPort value in their global.sbt to a default that matches the var's default value. This would still break on attempts to run more than one project at a time, however.
2. We could try to replicate the algorithm they use to generate the port, which is computing the sha1 hash of the directory name, transforming it into a hex string, then converting it into a number via hashcode computation and adding the result % 1000 to 5000: https://github.com/sbt/sbt/blob/4c3cd4e535a89d220a0d20f360f36212ceff9b09/main/src/main/scala/sbt/Defaults.scala#L288

Thoughts on what would be best for me to implement?

Ophirr33:
Well at least for now I'll just leave it as a variable and recommend that the user overrides their global sbt config to connect to the specified port. It still won't work out of the box for multiple concurrent projects, but I can set up a separate pull request for that later.

PR #1886, opened 2018-09-05
Fixed error when quitting vim after perl script edit

smhc:
Fixed the below error when quickly quiting vim/neovim after a perl script change with :wq

Error detected while processing function <SNR>101_NeoVimCallback[29]..<SNR>98_HandleExit[43]..ale_linters#perl#perl#Handle:
line    6:
E684: list index out of range: -1
E15: Invalid expression: a:lines[-1] =~# 'syntax OK'


w0rp:
I'll fix this and add a test for it. Note that `if` doesn't need parentheses.

PR #1889, opened 2018-09-06
Autocmd for automatically opening and closing the preview window based on the cursor

jpsouzasilva:
**Adding 'Open Preview Window on CursorMove' as a global option
Adding 'Close Preview Window on InsertEnter' as a global option**

1. I'm not sure how I'd write tests for this
2. Where is the most appropriate place to go about restricting this option per buffer or per filetype?

### DIFF
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -56,7 +56,7 @@ function! s:StopCursorTimer() abort
 endfunction
 
 function! ale#cursor#EchoCursorWarning(...) abort
-    if !g:ale_echo_cursor
+    if !ale#util#WarningsEnabled()
         return
     endif
 
@@ -72,21 +72,28 @@ function! ale#cursor#EchoCursorWarning(...) abort
     let l:buffer = bufnr('')
     let [l:info, l:loc] = s:FindItemAtCursor()
 
-    if !empty(l:loc)
-        let l:format = ale#Var(l:buffer, 'echo_msg_format')
-        let l:msg = ale#GetLocItemMessage(l:loc, l:format)
-        call ale#cursor#TruncatedEcho(l:msg)
-        let l:info.echoed = 1
-    elseif get(l:info, 'echoed')
-        " We'll only clear the echoed message when moving off errors once,
-        " so we don't continually clear the echo line.
-        execute 'echo'
-        let l:info.echoed = 0
+    if ale#Var(bufnr(''), 'echo_cursor')
+        if !empty(l:loc)
+            let l:format = ale#Var(l:buffer, 'echo_msg_format')
+            let l:msg = ale#GetLocItemMessage(l:loc, l:format)
+            call ale#cursor#TruncatedEcho(l:msg)
+            let l:info.echoed = 1
+        elseif get(l:info, 'echoed')
+            " We'll only clear the echoed message when moving off errors once,
+            " so we don't continually clear the echo line.
+            execute 'echo'
+            let l:info.echoed = 0
+        endif
+    endif
+
+    if ale#Var(bufnr(''), 'cursor_detail')
+        call ale#preview#CloseIfTypeMatches('ale-preview')
+        call ale#cursor#ShowCursorDetail({'stay_here': 1})
     endif
 endfunction
 
 function! ale#cursor#EchoCursorWarningWithDelay() abort
-    if !g:ale_echo_cursor
+    if !ale#util#WarningsEnabled()
         return
     endif
 
@@ -95,22 +102,29 @@ function! ale#cursor#EchoCursorWarningWithDelay() abort
         return
     endif
 
-    call s:StopCursorTimer()
+    if ale#Var(bufnr(''), 'echo_cursor')
+        call s:StopCursorTimer()
 
-    let l:pos = getcurpos()[0:2]
+        let l:pos = getcurpos()[0:2]
 
-    " Check the current buffer, line, and column number against the last
-    " recorded position. If the position has actually changed, *then*
-    " we should echo something. Otherwise we can end up doing processing
-    " the echo message far too frequently.
-    if l:pos != s:last_pos
-        let l:delay = ale#Var(bufnr(''), 'echo_delay')
+        " Check the current buffer, line, and column number against the last
+        " recorded position. If the position has actually changed, *then*
+        " we should echo something. Otherwise we can end up doing processing
+        " the echo message far too frequently.
+        if l:pos != s:last_pos
+            let l:delay = ale#Var(bufnr(''), 'echo_delay')
 
-        let s:last_pos = l:pos
-        let s:cursor_timer = timer_start(
-        \   l:delay,
-        \   function('ale#cursor#EchoCursorWarning')
-        \)
+            let s:last_pos = l:pos
+            let s:cursor_timer = timer_start(
+                        \   l:delay,
+                        \   function('ale#cursor#EchoCursorWarning')
+                        \)
+        endif
+    endif
+
+    if ale#Var(bufnr(''), 'cursor_detail')
+        call ale#preview#CloseIfTypeMatches('ale-preview')
+        call ale#cursor#ShowCursorDetail({'stay_here': 1})
     endif
 endfunction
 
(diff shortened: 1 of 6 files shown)


w0rp:
This is an interesting idea. I'll have a think about it.

jpsouzasilva:
I've added an option to only do this on certain filetypes, configurable by a global variable.

https://gfycat.com/AnnualNimbleDouglasfirbarkbeetle

autoload/ale/events.vim
```diff
@@ -138,6 +138,32 @@ function! ale#events#Init() abort
...
+            function! AleShowCursorDetailCmd() abort
+                if exists('*ale#engine#Cleanup')
+                    call ale#preview#CloseIfTypeMatches('ale-preview')
+                    " Don't jump to the preview window during cursor movement.
+                    call ale#cursor#ShowCursorDetail({'stay_here': 1})
+                endif
+            endfunction
+
+            if g:ale_show_cursor_detail
+                if len(g:ale_cursor_detail_fileextensions)
```
w0rp: Instead of adding an option to specify which file extensions to enable the option for, use `ale#Var` to get either a `b:` or `g:` variable for the option, and people can turn the option on in `ftplugin` files. That's how most of the other ALE options work.

Remove the additional `autocmd` command here, and check for if we should automatically open the preview window in the `ale#cursor#EchoCursorWarning` instead. The `InsertEnter` event will probably still be useful, but you should also check if we should close the preview when the cursor moves off the line in `ale#cursor#EchoCursorWarning`.

autoload/ale/events.vim
```diff
@@ -138,6 +138,32 @@ function! ale#events#Init() abort
...
+                    let ftStr=""
+
+                    for ft in g:ale_cursor_detail_fileextensions
```
w0rp: Install [Vint](https://github.com/Kuniwak/vint) and check your code with it. You should use `l:` for all local variables. Use `snake_case` for variable names, and `PascalCase` for function names.

autoload/ale/cursor.vim
```diff
@@ -130,8 +130,14 @@ function! ale#cursor#ShowCursorDetail() abort
 
     if !empty(l:loc)
         let l:message = get(l:loc, 'detail', l:loc.text)
-
-        call ale#preview#Show(split(l:message, "\n"))
+        " In case options have been received, pass them down to the called
+        " method.
+        let options = get(a:000, 0, {})
+        if len(options) 
+            call ale#preview#Show(split(l:message, "\n"), options)
```
w0rp: Instead of passing on the same options this function received, create a new Dictionary which shows exactly which options you want to pass on explicitly.

autoload/ale/events.vim
```diff
@@ -138,6 +138,32 @@ function! ale#events#Init() abort
...
+                    endfor
+
+                    execute "autocmd CursorMoved,CursorHold *.{" . strpart(ftStr, 0, len(ftStr)-1) . "} call AleShowCursorDetailCmd()"
+                else
+                    autocmd CursorMoved,CursorHold * call AleShowCursorDetailCmd()
+                endif
+            endif
+
+            if g:ale_close_preview_on_insert
+                autocmd InsertEnter * call ale#preview#CloseIfTypeMatches('ale-preview')
```
w0rp: Wrap the command here in `if exists('*ale#preview#CloseIfTypeMatches(') | ... | endif`. That way we won't load the `preview.vim` file when you enter Insert mode, without having yet shown a preview window.

autoload/ale/cursor.vim
jpsouzasilva: What would be the options I'm trying to pass down?

I thought about this and concerns were raised for me

1. Implicit coupling. Following your approach, when the called function changes, this would also need to change. And in case you were to forget about this detail, it would bug the code.
2. Assuming the first problem _isn't a problem_, how would I know which options need to be passed down, since it isn't explicit in `ale#preview#Show`'s signature?
3. Assuming this check could be be triggered for every CursorMove and there's no cost into passing the whole options dictionary as it comes, isn't it interesting to not overload this code with parsing and keep it slim, since it's tied to an AutoCmd? I'm thinking performance overhead there

w0rp: Don't pass on the same Dictionary arguments given to this function, as it's too hard to figure out which options are being given to the other one.

autoload/ale/events.vim
jpsouzasilva: What if the person doesn't have g:ale_echo_cursor set? It wouldn't enter the if there:

```
 if g:ale_echo_cursor                                                                                                                         
    autocmd CursorMoved,CursorHold * if exists('*ale#engine#Cleanup') | call ale#cursor#EchoCursorWarningWithDelay() | endif
```

Which is why I have it separated under a different block with an extra AutoCmd instead. Do you want to rename EchoCursorWarning to some other function that does both?

By the way, feel free to push commits to this branch and organize the code however you see fit. I'm sadly not able to follow what are your intentions with those tweaks. If you want to close this PR and do this by yourself later, that's fine too.

w0rp: If `g:ale_echo_cursor` isn't enabled, you don't get detailed messages either.

jpsouzasilva:
@w0rp I've tried to incorporate the changes you suggested.

The warning feedback behavior in general is now organized inside of `ale#cursor#EchoCursorWarning` (checks for both 'ale_cursor_detail' and 'ale_echo_cursor'). I haven't renamed the function because it's widely used currently and it might break somewhere else.

From my usage, `ale#Var` correctly works after enabling/disabling the variables in the buffer scope.

autoload/ale/events.vim
```diff
@@ -131,13 +131,17 @@ function! ale#events#Init() abort
                 autocmd InsertLeave * call ale#Queue(0)
             endif
 
-            if g:ale_echo_cursor
+            if ale#util#WarningsEnabled()
```
w0rp: Check `g:ale_echo_cursor || g:ale_cursor_detail` here. The variables used for enabling the `autocmd` events have to be global and set before ALE loads. That restriction makes it possible to avoid setting up the `autocmd` events if they aren't needed. You can remove the new function, because it's trivial.

autoload/ale/cursor.vim
```diff
@@ -130,8 +146,16 @@ function! ale#cursor#ShowCursorDetail() abort
 
     if !empty(l:loc)
```
w0rp: Take this part of the function here, and move it to a script local function which accepts the arguments `(loc, stay_here)` Then you can call that function instead of this one. You can avoid all of the logic above for showing detailed information automatically, including searching for the item again.

autoload/ale/cursor.vim
```diff
@@ -95,26 +102,35 @@ function! ale#cursor#EchoCursorWarningWithDelay() abort
...
 
-        let s:last_pos = l:pos
-        let s:cursor_timer = timer_start(
-        \   l:delay,
-        \   function('ale#cursor#EchoCursorWarning')
-        \)
+    if ale#Var(l:buffer, 'cursor_detail')
+        call ale#preview#CloseIfTypeMatches('ale-preview')
+        call ale#cursor#ShowCursorDetail({'stay_here': 1})
     endif
```
w0rp: You should remove the changes to this function, apart from the first line. What this function does is check if the cursor position has actually changed, and echo messages only after a short delay. If you don't wait for a short delay, the preview window will be closed and opened on every single keypress that moves the cursor, and that will slow down editing significantly.

w0rp:
I'll just do it how I want it done.

jpsouzasilva:
As said [at the end of the comment](https://github.com/w0rp/ale/pull/1889#discussion_r216522145), I supposed that would be better. I'll be closing this.

w0rp:
https://github.com/w0rp/ale/commit/43d7e8fde9aedb5dc3d76014a64de04fbdbaa276 Okay, there you go.

Thank you for all of your work on this. For something like this, it's difficult to introduce a new feature without breaking the existing functionality. Now it should be possible to do what you wanted to do here without stepping on anyone's toes.

I didn't add a buffer-local option yet. That could be done later. Someone might want to have an option fro controlling the height of the preview window. Additionally, it might be a good idea to only show the preview window for simple messages if the option for echoing messages is turned off. (Only show the window for strings in `detail` and not `text`.)

PR #1890, opened 2018-09-06
Add golangci-lint

saschagrunert:
I think this would be a nice addition. The only sad thing is that it is currently not possible to distinguish between warnings and errors.


w0rp:
Cheers! :beers:

ale_linters/go/golangci_lint.vim
```diff
@@ -0,0 +1,56 @@
...
+
+    if l:lint_package
+        return ale#path#BufferCdString(a:buffer)
+        \   . '%e run '
+        \   .  l:options
+    endif
+
+    return ale#path#BufferCdString(a:buffer)
+    \   . '%e run '
+    \   . ale#util#EscapePCRE(l:filename)
```
Carpetsmoker: Is `EscapePCRE` correct here? `l:filename` doesn't seem to be a regular expression, but just a regular filename. The documentation isn't clear, but this tool also doesn't seem to accept regular expressions here:

```
[~tw/utils](master)% golangci-lint run '.*'
ERRO Running error: context loading failed: can't eval symlinks for path .*: lstat .*: no such file or directory
 3 [~tw/utils](master)% golangci-lint run ./...
[~tw/utils](master)%
```
I think it should be `ale#util#EscapeVim` or `ale#Escape()`?

w0rp: Change it to `ale#Escape` if that works instead.

PR #1897, opened 2018-09-08
Add support for sqlfmt

LEI:
As a side note; I think [sqlint](https://github.com/purcell/sqlint) is missing from the list of integrations in `ale.txt`, and maybe the documentation should mention that:

> At this stage, SQLint checks SQL against the ANSI syntax, and uses the PostgreSQL SQL parser to achieve this. In time, we hope to add support for non-standard SQL variants (e.g. MySQL).

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ formatting.
 | SML | [smlnj](http://www.smlnj.org/) |
 | Solidity | [solhint](https://github.com/protofire/solhint), [solium](https://github.com/duaraghav8/Solium) |
 | Stylus | [stylelint](https://github.com/stylelint/stylelint) |
-| SQL | [sqlint](https://github.com/purcell/sqlint) |
+| SQL | [sqlint](https://github.com/purcell/sqlint), [sqlfmt](https://github.com/jackc/sqlfmt) |
 | Swift | [swiftlint](https://github.com/realm/SwiftLint), [swiftformat](https://github.com/nicklockwood/SwiftFormat) |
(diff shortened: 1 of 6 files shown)


w0rp:
Cheers! :beers:

Carpetsmoker:
It is worth pointing out that `sqlfmt` is not a very robust tool; [the author says](https://github.com/jackc/sqlfmt/issues/1):

> Unfortunately, sqlfmt is far from complete. It can handle pretty much anything that can happen in a select statement except CTEs, but not much else. 

Not much has changed since 2015, and [more recently](https://github.com/jackc/sqlfmt/issues/7):

> Sadly, my work on sqlfmt has been on an indefinite hiatus. Not sure when I'll get back to it. https://github.com/darold/pgFormatter could be a good alternative except that it is a perl script.

It also [supports only PostgreSQL](https://github.com/jackc/sqlfmt/issues/6):

> The parser is based on PostgreSQL syntax and PostgreSQL doesn't support backticks. I don't have any objection to supporting MySQL specific syntax as well, but I suspect it may be rather difficult.

Which makes me wonder how valuable it is to support this in ALE? I haven't tried it myself, but judging by the issues, it's not very usable in the vast majority of cases.

LEI:
Good point: it's the more robust formatter I could find out there, and it handles basic MySQL nicely (once the backticks removed).

I suggest to keep it but to clarify the limitations in the docs, idem for `sqlint`.

PR #1905, opened 2018-09-10
Fix tsserver `className` responses 

ngavalas:
While I was reading this file to look into another issue I had, I noticed that this was misspelled. I compared with the `tsserver` spec and verified that it really should be `className`. Minor bug, but it does mean that classname references were being miscategorized in the omnifunc dropdown.


ngavalas:
Updated the test.

w0rp:
Cheers! :beers:

PR #1906, opened 2018-09-10
fix non-. prefix triggers overfiltering results

### LINKED ISSUE #1904: Autocomplete doesn't suggest method names in languages that don't use . for members
When using the LSP with Hack (or any language that doesn't use `.` to access member functions, like Rust with `::` for modules), ALE doesn't autocomplete the available functions/members. You need to guess the first character of the method you want to call, type that, and then ALE properly shows all of the available autocomplete entries starting with that character. LSP implementations usually support this type of autocomplete.

For a concrete example, in Hack (running the `hh_server` LSP implementation), if I type `FooClass::`, I'd expect to see all of the available static class variables/methods/constants available on `FooClass`. If I manually send an LSP query at that position, the response correctly has that information.

ngavalas:
Currently, ALE overfilters results that come from `trigger_character_map` queries when the trigger character isn't `.`.

As an example, in Hack (if you add the proper trigger characters in, which I will do in a separate diff), typing `FooClass::` won't actually show any completion suggestions. This makes it hard to seamlessly chain the autocompletions, since you need to guess the first character of the method you're looking for in order for any suggestions to show up.

Fixes #1904.

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -98,14 +98,15 @@ function! ale#completion#GetTriggerCharacter(filetype, prefix) abort
     return ''
 endfunction
 
-function! ale#completion#Filter(buffer, suggestions, prefix) abort
+function! ale#completion#Filter(buffer, filetype, suggestions, prefix) abort
     let l:excluded_words = ale#Var(a:buffer, 'completion_excluded_words')
+    let l:triggers = s:GetFiletypeValue(s:trigger_character_map, a:filetype)
 
     " For completing...
     "   foo.
     "       ^
     " We need to include all of the given suggestions.
-    if a:prefix is# '.'
+    if index(l:triggers, a:prefix) >= 0
         let l:filtered_suggestions = a:suggestions
     else
         let l:filtered_suggestions = []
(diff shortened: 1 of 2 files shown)


ngavalas:
I amended the commit to fix a small oversight and tried to retrigger the CI, let's see if that worked...

ngavalas:
Updated the tests.

w0rp:
Cheers! :beers: 

At some point, ALE can be set up so it automatically pulls the trigger characters from the language servers. It's reading them now, but I just haven't finished the job and set it all up yet.

ngavalas:
If you want to point me at the work you've already done, I might be able to just finish it all up. I believe I saw that we were reading the triggers from the LSP response, but didn't dig further. Is that the extent of what's been done so far?

w0rp:
If you look at `autoload/ale/lsp.vim`, you'll see that ALE is saving the characters to `.capabilities.completion_trigger_characters` for each connection. A function could be added for pulling the characters out. `completion.vim` will have to be changed to use the characters from the language server instead of from the script's default values.

PR #1907, opened 2018-09-11
Fix bug where last C compile flag was ignored

guill:
There is currently a check that tries to prevent c-flags that contain
'-' in them from being unintentionally split and included in the list of
commands. For example, we wouldn't want "-fno-exceptions " to appear as
"-fno" and "-exceptions ". The way this check was done was by making sure
the last character of the split string was a space.

This meant that the very last option to appear in the compile command
was ignored (as it doesn't end with a space). This fix explicitly skips
the ends-with-space check on the last option in the command-line.

This isn't the best fix. Really we should be using the same
argument-processing rules as a shell would rather than just splitting on
'-'. That's a much larger and more complicated change though.


guill:
Test has been added.

PR #1919, opened 2018-09-14
Change PCRE escape to simple escape

saschagrunert:
As mentioned in https://github.com/w0rp/ale/pull/1890#discussion_r216167903, it would be more useful to use a simple escape here instead of PCRE.


w0rp:
Cheers! :beers:

PR #1925, opened 2018-09-16
Fix markdown error in README solargraph entry

elebow:
Fixes a minor markdown error in the Solargraph entry of the README.


w0rp:
Cheers! :beers:

PR #1926, opened 2018-09-16
c#ParseCFlags: don't index empty list

gnustomp:


gnustomp:
I've modified an existing cflags test to use `--sysroot=subdir`, which triggers the bug. The flag should be eventually ignored, but it needs to not crash first.

Before:
```
    ( 7/14) [EXECUTE] (X) Vim(if):E684: list index out of range: -1
      > function ale#c#ParseCFlags, line 11
```

After:
```
    ( 7/14) [EXECUTE] ParseCFlags should ignore -T and other arguments
```

PR #1928, opened 2018-09-16
To avoid blocking build tools, suspend ALE when suspending vim

### LINKED ISSUE #1898: Support VimSuspend and VimResume
When programming in Rust, I am used to `CTRL-Z`ing to background vim, run `cargo test` and then foreground vim again to keep working.

What interrupts this flow is that about half the time `cargo` says that it is blocked waiting on a lock file. I think that is `ale` running cargo to show compile errors in the buffer.

Given that https://github.com/neovim/neovim/issues/3648 has been merged, would it be possible to add a hook to abort any running jobs when `VimSuspend` is triggered?

I was looking around the code and I spotted `StopCurrentJobs` but it takes a buffer as an argument... maybe there is value in creating a hook to stop all jobs irrespective of the current buffer?

If you can point me in the right direction I'll create a PR 👋

felipesere:
Adds support for having ALE stop any linting jobs when vim is dropped into the backgorund (CTRL-Z).

Closes #1898

### DIFF
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -18,6 +18,22 @@ if !has_key(s:, 'executable_cache_map')
     let s:executable_cache_map = {}
 endif
 
+
+function! ale#engine#CleanupEveryBuffer() abort
+    for l:key in keys(g:ale_buffer_info)
+        " The key could be a filename or a buffer number, so try and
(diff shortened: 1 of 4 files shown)


felipesere:
This should be better now, as CI is happy 😄

w0rp:
Cheers! :beers:

PR #1932, opened 2018-09-18
shfmt: Use Vim's indent config as default indent width

rhysd:
[`shfmt`](https://github.com/mvdan/sh) uses a hard tab for indentation by default. To avoid this, `g:ale_sh_shfmt_options` variable can be set in vimrc. It's not bad. However, Vim already has a config for indentation. I thought it's better to set an indent value to `shfmt` command by looking Vim's indent option values.

This PR makes `shfmt` to format a `sh` buffer with the same indentation as its buffer's configuration as default behavior.


w0rp:
Cheers! :beers:

PR #1944, opened 2018-09-23
Bugfix: python add blank lines

bertjwregeer:
This fixes one issue I've had for a long time:

```

def some():
    continue_some_var = True
    forward_something = False

    if (
        continue_some_var and
        forward_something
    ):
        return True

    return False
```

would become:

```

def some():
    continue_some_var = True

    forward_something = False

    if (
        continue_some_var and

        forward_something
    ):
        return True

    return False
```

This commit also fixes the issue of adding extra newlines to docstrings accidentally by detecting when we are inside a docstring and not adding newlines even if we do match on control statements.


w0rp:
Cheers! :beers:

bertjwregeer:
Awesome, thanks very much!

PR #1948, opened 2018-09-25
Call QuickFixCmdPre and QuickFixCmdPost autocmds

Carpetsmoker:
The concrete problem that this solves is that I want to do "stuff" when
there are quickfix/location list errors. I initially tried hooking in to
ALELintPost, but that doesn't work as ALE sets errors asynchronously.

By hooking in to QuickFixCmd* this can be done in a generic (non-ALE
specific) way, that will work for e.g. :make as well; for example to
switch the colour of the statusline depending on the quickfix errors:

    fun! s:errcolor()
        if len(getloclist(bufnr('%'))) > 0 || len(getqflist()) > 0
            hi StatusLine term=bold,reverse cterm=bold,reverse gui=bold,reverse
                        \ ctermfg=1 guifg=#cd0000
        else
            hi StatusLine term=bold,reverse cterm=bold,reverse gui=bold,reverse
                        \ ctermfg=NONE guifg=NONE
        endif
    endfun

    autocmd QuickFixCmdPost ?expr,*make call s:errcolor() | redrawstatus

IMHO this is a more elegant/Vim-native solution than using ALELintPost
anyway (I saw there is an undocumented setting for excecuting it in
sync).

I'll add appropriate documentation and tests if the approach in this PR
is approved (no need to needlessly write docs/struggle with vader).

### DIFF
--- a/autoload/ale/list.vim
+++ b/autoload/ale/list.vim
@@ -81,17 +81,23 @@ function! s:SetListsImpl(timer_id, buffer, loclist) abort
     if g:ale_set_quickfix
         let l:quickfix_list = ale#list#GetCombinedList()
 
+        silent doautocmd <nomodeline> QuickFixCmdPre cexpr
+
         if has('nvim')
             call setqflist(s:FixList(a:buffer, l:quickfix_list), ' ', l:title)
         else
             call setqflist(s:FixList(a:buffer, l:quickfix_list))
             call setqflist([], 'r', {'title': l:title})
         endif
+
+        silent doautocmd <nomodeline> QuickFixCmdPost cexpr
     elseif g:ale_set_loclist
         " If windows support is off, bufwinid() may not exist.
         " We'll set result in the current window, which might not be correct,
         " but it's better than nothing.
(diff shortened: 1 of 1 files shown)


w0rp:
That undocumented setting is intended for use with the test suite only.

w0rp:
What does this let you do that `ALELintPost` doesn't? Is it purely just because the loclist and quickfix list have to be set after a 0 second timer, to work around Vim bugs? If so, that could be fixed by putting `ALELintPost` on a 0 second timer too.

Carpetsmoker:
> Is it purely just because the loclist and quickfix list have to be set after a 0 second timer

Pretty much, yeah. `getloclist()` would always return `[]` in `ALELintPost`.

As mentioned, I think that calling `QuickFixCmd*` is a good idea anyway; it's kind of the expected behaviour I think?
But I could look at fixing `ALELintPost` too (or we could do both). I don't have very strong opinions.

Carpetsmoker:
This seems to work correctly with the `ALELintPost` autocmd now; I don't know if something changed in the meanwhile, or if I did something wrong last time.

PR #1950, opened 2018-09-26
Add ccls support for C/C++/ObjC 

yejingchen:
ccls is a C/C++/ObjC language server supporting cross references, hierarchies, completion and semantic highlighting. As ccls originates from cquery, most of the code and tests are derived from the corresponding cquery part.

One change worths noting is that I exposed the initialization options to `ale_c_ccls_init_options` (also `cpp`, `objc` ones) as a dictionary to let users do more complicated customizations, as ccls support various options. By default it is an empty dictionary `{}`.

All tests passed.


yejingchen:
Strange, these tests passed on my local run, trying to identify the cause.

ale_linters/c/ccls.vim
```diff
@@ -0,0 +1,32 @@
...
+
+call ale#Set('c_ccls_executable', 'ccls')
+call ale#Set('c_ccls_init_options', {})
+
+function! ale_linters#c#ccls#GetProjectRoot(buffer) abort
+    let l:project_root = ale#path#FindNearestFile(a:buffer, '.ccls-root')
+
+    if empty(l:project_root)
+        let l:project_root = ale#path#FindNearestFile(a:buffer, 'compile_commands.json')
+    endif
```
w0rp: This will do for now. Later on, I'll add a function for finding the project root based on `compile_commands.json` being inside of a directory like `build`.

ale_linters/cpp/ccls.vim
```diff
@@ -0,0 +1,32 @@
...
+    if empty(l:project_root)
+        let l:project_root = ale#path#FindNearestFile(a:buffer, 'compile_commands.json')
+    endif
+
+    if empty(l:project_root)
+        let l:project_root = ale#path#FindNearestFile(a:buffer, '.ccls')
+    endif
+
+    return !empty(l:project_root) ? fnamemodify(l:project_root, ':h') : ''
+endfunction
```
w0rp: Could you move this function to `autoload/ale/handlers/ccls.vim` and use the the same function in all of the files?

yejingchen: Moved those functions to `autoload/ale/handlers/ccls.vim`. Should tests also be adjusted?

w0rp:
Cheers! :beers:

yejingchen:
Thanks!

PR #1958, opened 2018-09-28
Add Perl6 support via 'perl6 -c'

Garland-g:
Adds support for perl6 files using perl6 -c.

Because perl6 -c has the same security issues with BEGIN and CHECK blocks as perl -c, perl6 -c is disabled by default.

Implements https://github.com/w0rp/ale/issues/1104


w0rp:
Is a separate filetype and linter needed for Perl 6, or can the existing Perl linter be used instead by changing the executable option?

Garland-g:
I found some error messages that cause errors in the Handle function.

These changes fix this. There are also tests to ensure that handling these error messages is done properly.

The ErrorExtract function had concatenation instead of assignment in some places. That has been fixed as well.

The return value of ErrorExtract is now handled consistently in the Handle function.

w0rp:
Cheers! :beers:

Sorry for the delay. I've been slowly catching up with a backlog of pull requests and issues.

PR #1959, opened 2018-09-30
[WIP] JS-Beautify fixers for template portions and HTML files (+ some changes needed in the engine)

jpsouzasilva:
I haven't included the docs yet because it could still be subject to change.

Those fixers are coming from [JS-Beautify](https://github.com/beautify-web/js-beautify).

### The main intent here is to add two new fixers

- **Beautify-HTML** which formats HTML files.
- **Beautify-Template** which formats Javascript (JSX), Vue and HTML files.

### The way it works for HTML
Just pass the file's contents to html-beautify and let it format.
### The way it works for templates:
1. Extract the template portion of the file; the extraction is made by matching the opening and closing tags (e.g. `<html>...</html>`) or by regex (matching each of the lines with a regex until the whole regex chain is satisfied). 
2.

For React files it's extracting the template like this:
```
export default class App extends Component {
	render() { <--- match by regex (opening)
		return (  <--- match by regex (opening)
                       JSX content below gets formatted and then glued back
			<Provider restaurantStore={restaurantStore}>
				<RootStack />
			</Provider>
		);  <--- match by regex (closing)
	}
}
```

For Vue files and HTML files it's extracting the content like this:
```
<script lang="ts">
...
</script>

<template> <-- match with equals == "<template>"
   Vue template here gets formatted and then glued back
</template> <-- match with equals == "</template>"

<style>
...
</style>
```

I had to change the [fix.vim](https://github.com/w0rp/ale/blob/master/autoload/ale/fix.vim) engine file because the template extraction needs to be processed before the fixer runs and I didn't find a way to do it with the current implementation.

Thus the file has been refactored in order for allowing fixers to have a "pre init" phase where it can modify the fixer job options before it actually runs.


jpsouzasilva:
Also the reason the job is failing is because I can't use single quote ("string literals") for the tab character.

https://travis-ci.org/w0rp/ale/jobs/435193194#L475

The tab character has to be put within double quotes like [this](https://github.com/w0rp/ale/pull/1959/files#diff-a5341ffa70b1ce301e9d45f9a758bf61R133) for it to be considered a tab.

PR #1963, opened 2018-10-01
Fix Credo message types

Sgiath:
Credo uses more message types then currently present. This causes to fallback to error messages. So fo example "Software Design" message "TODO comment" is reported as error.

I would like to report "Software Design" and "Code Readability" as "Info message" I can see that there is config for `info_sign` but I was able to find only examples of warning and error messages. If it is possible to mark some message as "Info message" I would happily change that.

For comparison relevant lines form neomake credo checker https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/elixir.vim#L10


Sgiath:
@w0rp added tests.
So it is not possible to set message type as "Info"? Why is there `info_sign` then?

w0rp:
@Sgiath Use `'I'` for "info" messages. Search for signs in the documentation for how to set different signs for info. The warning signs are used for info by default.

Sgiath:
Hmm I guess I just didn't search enough :) thanks

Sgiath:
@w0rp fixed

Sgiath:
@w0rp what do you think? Can you approve it or should I make some other changes?

w0rp:
Cheers! :beers:

Sgiath:
Hope you are well now :) thanks for your effort!

PR #1964, opened 2018-10-02
Fix PMD not working with classes without package

dlresende:
PMD is currently not working properly for Java classes that use [unnamed packages](https://docs.oracle.com/javase/specs/jls/se11/html/jls-7.html#jls-7.4.2).

Consider the following Java class that does not contain a `package` declaration:

```java
public class App {
  String getGreeting() {
    return "Hello world.";
  }

  static void main(String... args) {
    System.out.println(new App().getGreeting());
  }
}
```

Running PMD in the command line agaist the Java class above produces an output with empty string `""` in the `"Package"` column:

```sh
$ pmd -R category/java/bestpractices.xml -f csv -d './src/main/java/App.java'
Oct 02, 2018 9:10:39 PM net.sourceforge.pmd.PMD processFiles
WARNING: This analysis could be faster, please consider using Incremental Analysis: https://pmd.github.io/pmd-6.7.0/pmd_userdocs_incremental_analysis.html
"Problem","Package","File","Priority","Line","Description","Rule set","Rule"
"1","","/Users/diego/Projects/github.com/dlresende/kata-fizz-buzz/src/main/java/App.java","2","7","System.out.println is used","Best Practices","SystemPrintln"
```

But the pmd.vim handler's current pattern refuses everything coming from a Java class that does not have a package name (2nd column):

```vim
let l:pattern = '"\(\d\+\)",".\+","\(.\+\)","\(\d\+\)","\(\d\+\)","\(.\+\)","\(.\+\)","\(.\+\)"$'
```

The solution I am proposing is to also accept empty strings as package names.

### DIFF
--- a/ale_linters/java/pmd.vim
+++ b/ale_linters/java/pmd.vim
@@ -2,7 +2,7 @@
 " Description: PMD for Java files
 
 function! ale_linters#java#pmd#Handle(buffer, lines) abort
-    let l:pattern = '"\(\d\+\)",".\+","\(.\+\)","\(\d\+\)","\(\d\+\)","\(.\+\)","\(.\+\)","\(.\+\)"$'
+    let l:pattern = '"\(\d\+\)",".*","\(.\+\)","\(\d\+\)","\(\d\+\)","\(.\+\)","\(.\+\)","\(.\+\)"$'
     let l:output = []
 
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

PR #1971, opened 2018-10-03
Add the dockerfile_lint linter for Dockerfiles

ananace:
Since we've started using dockerfile_lint and some custom rules to lint the images we're using for CI purposes, I figure I might as well add an ALE linter for local use.


ale_linters/dockerfile/dockerfile_lint.vim
```diff
@@ -0,0 +1,66 @@
...
+
+            if has_key(l:object, 'description') && l:object['description'] != 'None'
+                let l:message = l:message . '. ' . l:object['description']
+            endif
+
+            call add(l:messages, {
+                        \ 'lnum': l:line,
+                        \ 'text': l:message,
+                        \ 'type': ale_linters#dockerfile#dockerfile_lint#GetType(l:type),
+                        \})
```
w0rp: Align the line continuation characters so they are on the same level of indentation as the previous lines. See other files in the codebase for examples.

ale_linters/dockerfile/dockerfile_lint.vim
```diff
@@ -0,0 +1,66 @@
...
+
+    if empty(l:data)
+        " Should never happen, but it's better to be on the safe side
+        return []
+    endif
+
+    let l:messages = []
+
+    for l:type in ['error', 'warn', 'info']
+        for l:object in l:data[l:type]['data']
```
w0rp: Can any of these keys be missing? It might be best to use the `get()` function just in case.

ale_linters/dockerfile/dockerfile_lint.vim
```diff
@@ -0,0 +1,66 @@
...
+    \   . ' -p -j -f'
+    \   . ' %t'
+endfunction
+
+call ale#linter#Define('dockerfile', {
+\   'name': 'dockerfile_lint',
+\   'executable_callback': ale#VarFunc('dockerfile_dockerfile_lint_executable'),
+\   'command_callback': 'ale_linters#dockerfile#dockerfile_lint#GetCommand',
+\   'callback': 'ale_linters#dockerfile#dockerfile_lint#Handle',
+\})
```
w0rp: Add tests for the new executable and command. See `:help ale-development`.

ale_linters/dockerfile/dockerfile_lint.vim
ananace: Well, if any of them are missing, then that's because dockerfile_lint has crashed completely when trying to lint the Dockerfile, which means the object won't be valid JSON either way.

RyanSquared: So, there will always be an `error`, `warn`, and `info` output, even if they're empty?

ananace: If the the linter runs - and exits with a non-255 code, then it will spit out a JSON object that always looks like; (Though with data in place)
```json
{
  "error": {
    "count": 0,
    "data": []
  },
  "warn": {
    "count": 0,
    "data": []
  },
  "info": {
    "count": 0,
    "data": []
  },
  "summary": []
}
```

If - on the other hand - it exits with 255 then you'll get a JavaScript error backtrace as an output, which is not valid JSON and therefore won't reach the extraction part.

RyanSquared: Alright. In that case, we should be fine.

ale_linters/dockerfile/dockerfile_lint.vim
```diff
@@ -0,0 +1,66 @@
...
+        " Should never happen, but it's better to be on the safe side
+        return []
+    endif
+
+    let l:messages = []
+
+    for l:type in ['error', 'warn', 'info']
+        for l:object in l:data[l:type]['data']
+            try
+                let l:line = l:object['line']
```
RyanSquared: This section can be replaced with: `let l:line = get(l:object, "line", -1)`

ale_linters/dockerfile/dockerfile_lint.vim
```diff
@@ -0,0 +1,66 @@
...
+            catch
+                let l:line = -1
+            endtry
+
+            let l:message = l:object['message']
+
+            if has_key(l:object, 'description') && l:object['description'] != 'None'
```
RyanSquared: Similar to above: `if get(l:object, 'description', 'None') != 'None'` compacts it to one call.

w0rp:
Cheers! :beers:

PR #1972, opened 2018-10-04
Support multiline text in loclist and quickfix

tokorom:
I want to pass multiple lines of text to loclist and quickfix. Some linters sometimes returns useful text of multiple lines.

### loclist sample

```vim
let loclist = [
\ {'filename': 'Foo.swift', 'lnum': 100, 'col': 20, 'text': ['line 1', 'line 2', 'line 3']},
\]
```

### DIFF
--- a/autoload/ale/list.vim
+++ b/autoload/ale/list.vim
@@ -60,7 +60,7 @@ function! s:FixList(buffer, list) abort
 
         let l:fixed_item.text = ale#GetLocItemMessage(l:item, l:format)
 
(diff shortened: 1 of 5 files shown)


w0rp:
The loclist and quickfix don't really support displaying errors split over many lines. Which you should do for your linter instead is write your multi-line message into the `detail` key that ALE supports, and then you can view the full error message with `:ALEDetail`. The changes here aren't needed, and don't fit the Vim APIs.

PR #1977, opened 2018-10-06
Allow custom executable for ansible linters

ix5:
Set `ale_ansible_ansible_lint_executable` to use ansible-lint from a virtualenv.

Tests are rudimentary, I don't know a thing about vader.

What do you think about renaming the linter to `ansible_lint`? It's already called that way internally, just the user-facing `name` attribute in
```
call ale#linter#Define('ansible', {
\   'name': 'ansible',
```
is confusing.

### DIFF
--- a/ale_linters/ansible/ansible_lint.vim
+++ b/ale_linters/ansible/ansible_lint.vim
@@ -7,13 +7,6 @@ function! ale_linters#ansible#ansible_lint#GetExecutable(buffer) abort
     return ale#Var(a:buffer, 'ansible_ansible_lint_executable')
 endfunction
 
-function! ale_linters#ansible#ansible_lint#GetCommand(buffer) abort
-    let l:executable = ale_linters#ansible#ansible_lint#GetExecutable(a:buffer)
-
-    return ale#Escape(l:executable)
-    \  . ' -p %t'
-endfunction
-
 function! ale_linters#ansible#ansible_lint#Handle(buffer, lines) abort
     for l:line in a:lines[:10]
         if match(l:line, '^Traceback') >= 0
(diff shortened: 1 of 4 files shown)


ale_linters/ansible/ansible_lint.vim
```diff
@@ -43,7 +56,7 @@ endfunction
 
 call ale#linter#Define('ansible', {
 \   'name': 'ansible',
-\   'executable': 'ansible',
-\   'command': 'ansible-lint -p %t',
+\   'executable_callback': 'ale_linters#ansible#ansible_lint#GetExecutable',
+\   'command_callback': 'ale_linters#ansible#ansible_lint#GetCommand',
```
w0rp: Write `'command': '%e -p %t',` instead. which is equivalent.

ale_linters/ansible/ansible_lint.vim
```diff
@@ -43,7 +56,7 @@ endfunction
 
 call ale#linter#Define('ansible', {
 \   'name': 'ansible',
```
w0rp: You can rename this to `ansible_lint`, if and only if you add an alias for the old name. See `aliases` under `:help ale#linter#Define()`.

ix5: Done

ale_linters/ansible/ansible_lint.vim
ix5: Done

ix5:
Documentation added as well.

Thank you w0rp for all your work!

w0rp:
Cheers! :beers:

ix5:
Thanks for merging!

PR #1979, opened 2018-10-06
Add prolog swipl linter

tyru:
Added SWI-Prolog linter, one of implementations of Prolog language.
It tries to load current file within sandbox.
It checks (syntactic / semantic) (warnings / errors).

ale-prolog-linter

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ formatting.
 | PO | [alex](https://github.com/wooorm/alex) !!, [msgfmt](https://www.gnu.org/software/gettext/manual/html_node/msgfmt-Invocation.html), [proselint](http://proselint.com/), [write-good](https://github.com/btford/write-good) |
 | Pod | [alex](https://github.com/wooorm/alex) !!, [proselint](http://proselint.com/), [write-good](https://github.com/btford/write-good) |
 | Pony | [ponyc](https://github.com/ponylang/ponyc) |
+| Prolog | [swipl](https://github.com/SWI-Prolog/swipl-devel) |
 | proto | [protoc-gen-lint](https://github.com/ckaznocha/protoc-gen-lint) |
 | Pug | [pug-lint](https://github.com/pugjs/pug-lint) |
 | Puppet | [languageserver](https://github.com/lingua-pupuli/puppet-editor-services), [puppet](https://puppet.com), [puppet-lint](https://puppet-lint.com) |
--- a/ale_linters/prolog/swipl.vim
+++ b/ale_linters/prolog/swipl.vim
@@ -88,7 +88,7 @@ endfunction
 " because what we want is syntactic or semantic check.
 function! s:Ignore(item) abort
     return a:item.type is# 'E' &&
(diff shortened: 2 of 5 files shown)


tyru:
I found infinite loop (`:- repeat, writeln('loop!'), fail.`) causes ALE not to work, but I could not find the workaround for this.

tyru:
> infinite loop causes ALE not to work

Okay, finally it can be detected by setting SIGALRM.
and if current file tries to overwrite SIGALRM (`:- alarm(1, writeln('hacked!'), _, []).`) in any line, it is blocked by sandbox.

ale-prolog-linter2

Also added help file. please review.

tyru:
For easier review, here is the test script.

```prolog
:- initialization main.
:- writeln(hello).

main :-
  fib(10, N), writeln(N),
  halt.

fib(0, 0).
fib(1, 1).
fib(N, M) :-
  N1 is N - 1, N2 is N - 2,
  fib(N1, M1), fib(N2, M2),
  M is M1 + M2.

:- repeat, writeln('loop!'), fail.
```

SWI-Prolog can be installed by `apt install swi-prolog-nox` on Debian.

w0rp:
If an infinite loop fails, does that mean that this command will execute the code you have written? If so, is it possible to write a command which stops execution immediately, and which finds syntax errors, so that no malicious code can be executed by viewing a file?

tyru:
> If an infinite loop fails, does that mean that this command will execute the code you have written?

No. Always the commands are executed until signal was received (3 seconds (`prolog_swipl_timeout`) after by default).
But it is executed in sandbox.
Please let me explain.

---

Prolog programs have two types of clauses, **rule** and **query**.

The format of rule is `Head :- Body.`. It is an definition of a function (it's called "predicate" in Prolog).

```prolog
hello :- writeln('hello!').
```

The format of query is defined like `:- Query.` or `?- Query.`.
The following code calls above `hello` predicate.

```prolog
:- hello.
```

So if you want to disallow running arbitrary code, you might think it's OK to disallow queries (like parsing viewing file, extract only definitions, save it to a temporary file, and run interpreter with the file).

But I want to execute queries because it is necessary to detect syntax errors.
In Prolog, a user can define arbitrary binary or unary operators.

```prolog
:- op(501, xfy, 'by').

L by R :- write(L), write(' by '), write(R).

:- hello by john.
```

`:- op(501, xfy, 'by').` defines binary operator `by`.
Without it, `L by R :- write(L), write(' by '), write(R).` causes syntax error.
interpreter also tries to execute `:- hello by john.` but it is blocked by sandbox.

```
ERROR: /home/tyru/config/home/volt/repos/github.com/w0rp/ale/test.pl:5:
        No permission to call sandboxed `write(_7730)'
        Reachable from:
                  user:A by B
```

Defining operators are commonly used in Prolog.
Without it, almost Prolog programs can not be parsed.

w0rp:
Okay, it runs in a sandbox. That's all I needed to know.

w0rp:
Cheers! :beers:

tyru:
Thanks! :smile:

Issue #1976, opened 2018-10-05
ALE cannot be enabled locally when disabled globally


plexigras:
I want to enable linting only for the current buffer.

https://github.com/w0rp/ale/blob/master/autoload/ale/toggle.vim#L63

plexigras:
why not have `ALEEnableBuffer` do this for me?

w0rp:
It can't work that way. The global flag disables the `autocmd` commands entirely, which allows people to completely disable autocmd commands. The commands would need to be left on for ALE to check buffer-local options. You can do what I suggested above, which is how you ought to only enable ALE for some filetypes/buffers.

plexigras:
what i mean is that `ALEEnableBuffer` could enable the autocommands just like `ALEEnable` does.

w0rp:
It can't do that.

w0rp:
You should do what I suggested above. Enable ALE globally with no linters enabled anywhere, and enable some linters for just one type of file.

plexigras:
the only thing i see that makes it so it can't do that is that `ale#events#Init` is only checking `g:ale_enabled` and not also if there is a buffer that has `b:buffer_enabled` before disabeling/enabeling the autocommands.

or is there something else that needs to be done  i mean other then adding `call ale#events#Init()` to  `ale#toggle#ToggleBuffer`?

PR #1983, opened 2018-10-08
[WIP] added support for linting locally without having ale enabled globally

plexigras:
solution for #1976 

I don't understand the tests, maybe someone can help me?


mareq:
@w0rp I found this issue by googling a solution for my use-case, where I would like to have ALE disabled for all buffers and enable it only on-demand for that one single buffer I need. I.e. not for all buffers having the same `filetype`, but literally just for that one buffer and one buffer only. Is it possible to somehow achieve this?

PR #1988, opened 2018-10-10
Add support for black usage with pipenv

mdtusz:
For many projects and just out of habit, I have black installed using pipenv - the standard black executable will work assuming `pipenv shell` has been run before opening vim, but I often forget or haven't run it. 

This adds support for `black` usage with `pipenv` the same way that it works with `mypy` and other python tools.

### DIFF
--- a/autoload/ale/fixers/black.vim
+++ b/autoload/ale/fixers/black.vim
@@ -25,7 +25,7 @@ function! ale#fixers#black#Fix(buffer) abort
     let l:options = ale#Var(a:buffer, 'python_black_options')
 
     return {
-    \   'command': ale#Escape(l:executable. l:exec_args)
+    \   'command': ale#Escape(l:executable) . l:exec_args
     \       . (!empty(l:options) ? ' ' . l:options : '')
     \       . ' -',
     \}
--- a/autoload/ale/fixers/black.vim
+++ b/autoload/ale/fixers/black.vim
@@ -18,10 +18,6 @@ endfunction
 function! ale#fixers#black#Fix(buffer) abort
     let l:executable = ale#fixers#black#GetExecutable(a:buffer)
 
-    if !executable(l:executable)
-        return 0
-    endif
-
     let l:exec_args = l:executable =~? 'pipenv$'
     \   ? ' run black'
     \   : ''
@@ -29,7 +25,7 @@ function! ale#fixers#black#Fix(buffer) abort
     let l:options = ale#Var(a:buffer, 'python_black_options')
 
     return {
-    \   'command': ale#Escape(l:executable) . l:exec_args
+    \   'command': ale#Escape(l:executable. l:exec_args)
     \       . (!empty(l:options) ? ' ' . l:options : '')
     \       . ' -',
     \}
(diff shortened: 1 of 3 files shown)


autoload/ale/fixers/black.vim
```diff
@@ -4,22 +4,32 @@
...
 
 function! ale#fixers#black#Fix(buffer) abort
-    let l:executable = ale#python#FindExecutable(
-    \   a:buffer,
-    \   'python_black',
-    \   ['black'],
-    \)
+    let l:executable = ale#fixers#black#GetExecutable(a:buffer)
 
     if !executable(l:executable)
```
w0rp: The test fails because of this check here. I think we can just remove it.

RyanSquared:
```
[00:01:01]   Starting Vader: C:\testplugin\test\fixers\test_black_fixer_callback.vader
[00:01:01]     (1/3) [EXECUTE] The black callback should return the correct default values
[00:01:01]     (1/3) [EXECUTE] (X) Vim(if):E735: Can only compare Dictionary with Dictionary
[00:01:01]     (2/3) [EXECUTE] The black callback should include options
[00:01:01]     (3/3) [EXECUTE] Pipenv is detected when python_black_auto_pipenv is set
[00:01:01]     (3/3) [EXECUTE] (X) {'command': 'xxxinvalid -'} should be equal to {'command': 'pipenv run mypy'}
[00:01:01]   Success/Total: 1/3
```

Latest failed segment.

autoload/ale/fixers/black.vim
RyanSquared: Tests are now failing due to this as 0 was an expected case.

autoload/ale/fixers/black.vim
```diff
@@ -4,22 +4,28 @@
...
-    endif
+    let l:exec_args = l:executable =~? 'pipenv$'
+    \   ? ' run black'
+    \   : ''
 
     let l:options = ale#Var(a:buffer, 'python_black_options')
 
     return {
-    \   'command': ale#Escape(l:executable)
+    \   'command': ale#Escape(l:executable. l:exec_args)
```
w0rp: This isn't right. This will produce a command string like `"pipenv run black" -`. You want something more like `"pipenv" run black -`. You should move `. l:exec_args` outside of the `ale#Escape` call.

test/fixers/test_black_fixer_callback.vader
```diff
@@ -37,3 +34,11 @@ Execute(The black callback should include options):
   AssertEqual
   \ {'command': ale#Escape(ale#path#Simplify(g:dir . '/python_paths/with_virtualenv/env/' . b:bin_dir . '/black')) . ' --some-option -' },
   \ ale#fixers#black#Fix(bufnr(''))
+
+Execute(Pipenv is detected when python_black_auto_pipenv is set):
+  let g:ale_python_black_auto_pipenv = 1
+  call ale#test#SetFilename('/testplugin/test/python_fixtures/pipenv/whatever.py')
+
+  AssertEqual 
+  \ {'command': ale#Escape('pipenv run black') . ' -'},
```
w0rp: Fix the test here, given the suggestion above.

RyanSquared:
@mdtusz do you mind fixing some things relating to w0rp's review? After that, it should be good to merge.

mdtusz:
Sorry for the radio silence! Will try to get around to it this week to close it up!

w0rp:
Cheers! :beers:

PR #1991, opened 2018-10-12
Expose ruby-solargraph's initialization options

alskdj21:
Expose ruby-solargraph's initialization option as `ale_ruby_solargraph_init_options` as a dictionary with a default value of `{}` to help users on customizing solargraph's default configurations.

### DIFF
--- a/ale_linters/ruby/solargraph.vim
+++ b/ale_linters/ruby/solargraph.vim
@@ -5,7 +5,7 @@
 " Description: updated to use stdio
 
 call ale#Set('ruby_solargraph_executable', 'solargraph')
-call ale#Set('ruby_solargraph_init_options', {})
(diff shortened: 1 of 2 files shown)


ale_linters/ruby/solargraph.vim
```diff
@@ -5,6 +5,7 @@
 " Description: updated to use stdio
 
 call ale#Set('ruby_solargraph_executable', 'solargraph')
+call ale#Set('ruby_solargraph_init_options', {})
```
w0rp: Rename that to just `_options` instead of `_init_options`, and I think this should be good.

w0rp:
Cheers! :beers:

alskdj21:
Thanks!

Issue #1922, opened 2018-09-15
Add support for clippy


loyd:
Clippy plugin interface [was deprecated](https://github.com/rust-lang-nursery/rust-clippy/pull/2712) =(.

Now you need to use `cargo clippy`.

Related #111.

taylorlee:
As a workaround, this seems to be functional for me so far:
```diff
diff --git a/ale_linters/rust/cargo.vim b/ale_linters/rust/cargo.vim
index 5aefe72..841c4b8 100644
--- a/ale_linters/rust/cargo.vim
+++ b/ale_linters/rust/cargo.vim
@@ -71,7 +71,7 @@ function! ale_linters#rust#cargo#GetCommand(buffer, version_output) abort
     endif
 
     return l:nearest_cargo_prefix . 'cargo '
-    \   . (l:use_check ? 'check' : 'build')
+    \   . (l:use_check ? 'clippy' : 'build')
     \   . (l:use_all_targets ? ' --all-targets' : '')
     \   . (l:use_examples ? ' --examples' : '')
     \   . (l:use_tests ? ' --tests' : '')
```

RyanSquared:
> 12:38 [LordRyan] For the purpose of linting support in an editor, I can use `cargo-clippy` and rely on the error output being the same as with `cargo build` and `cargo check`?
12:57 [oli_obk] well kind of. as long as `cargo-clippy` exists at all, yes
12:57 [oli_obk] but it's not clear whether we're going to support it forever
12:58 [oli_obk] but yes, the entire output mirrors that of `cargo check`, just with more errors

Would it be reasonable to have `l:use_clippy` instead? It'll have _more_ errors than traditional `cargo`.

w0rp:
A `use_clippy` option sounds like the way to go, complete with another version check to see if `cargo` supports `clippy`.

Rust tools are forever unstable and constantly changing.

RyanSquared:
> complete with another version check to see if cargo supports clippy.

It's actually an external tool. Clippy as of right now is not a part of rustc, so you have to install `cargo-clippy` - which can be checked by just seeing if `cargo-clippy` exists (but you should still use the `cargo` interface for now). Eventually, a lot of the checks should be integrated into rustc itself, AIUI: https://github.com/rust-lang/rust/issues/53224.

w0rp:
Thanks, I figured that out after reading the pull request, and suggested just adding in an option for using `clippy` which can be off by default.

[...]

PR #1995, opened 2018-10-15
Add `rust_cargo_use_clippy` option

devonhollowood:
I took a shot at #1922. Right now, this adds an option `ale_rust_cargo_use_clippy`, which defaults to `0`. If instead set to `1`, `ALE` uses `cargo clippy` instead of `cargo build` or `cargo check`.

Optimally, I'd like for this to check whether the `cargo-clippy` executable exists and if so default to using clippy, but there's a wrinkle: if `rustup` has `cargo-clippy` on one toolchain, but not all toolchains, then it's possible for `cargo-clippy` to exist but to exit with an error. So maybe a smarter way to handle the check is to see if `cargo-clippy -V` exits with status `0`, but I'm not sure the best way to do that (I'm new to vimscript, and it looks like `ALE` has it's own functions to handle this sort of thing anyways). Any advice there would be appreciated.

Edit: I figured out how to do the above with `systemlist()`. Let me know if that's acceptable, or if you would rather I use an `ALE`-specific function. Also, the tests no longer run successfully. I think this is because the docker image for ALE doesn't have `clippy` installed.

### DIFF
--- a/ale_linters/rust/cargo.vim
+++ b/ale_linters/rust/cargo.vim
@@ -4,7 +4,7 @@
 " Description: rustc invoked by cargo for rust files
 
 call ale#Set('rust_cargo_use_check', 1)
-call ale#Set('rust_cargo_use_clippy', 0)
+call ale#Set('rust_cargo_use_clippy', 1)
 call ale#Set('rust_cargo_check_all_targets', 0)
 call ale#Set('rust_cargo_check_examples', 0)
 call ale#Set('rust_cargo_check_tests', 0)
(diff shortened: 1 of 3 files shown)


devonhollowood:
I can see the logic in making `rust_cargo_use_clippy = 0` by default, but it might make sense to leave the version check so that clippy "fails" gracefully.

A pretty common set-up is to use `rustup` to manage multiple Rust installations, allowing one to switch versions of the Rust "toolchain" easily (and in some cases automatically). Some Rust toolchains don't have `clippy` available for them through `rustup`, so the availability of `clippy` might effectively change per-buffer. For example, for my current setup `clippy` is available on the `nightly` toolchain but not on the `stable` toolchain. So on the `nightly` toolchain `cargo-clippy -V` outputs `clippy 0.0.212 (32b1d1fc 2018-10-05)`, but on `stable`, `cargo-clippy -V` outputs ``error: toolchain 'stable-x86_64-apple-darwin' does not have the binary `cargo-clippy` ``. Keeping the version check allows me to set `rust_cargo_use_clippy = 1` in my `.vimrc` and have ALE automatically fallback to using `cargo check` when `clippy` isn't available.

Let me know your thoughts on this.

w0rp:
Any setting that changes per-buffer can be handled by using `b:` options.

Superseded by #2001.

PR #1999, opened 2018-10-16
Reek: --force-exclusion flag

stevenharman:
Expand the current buffer's full path name for the `--stdin-filename` argument. This is basically identical to how Rubocop does it. Before the `%s` wasn't being interpolated, but the pathname given to reek was '%s'. Now it gets the full pathname of the current buffer.

I also added the `--force-exclusion` argument to ensure any configured `excluded_paths` are respected. However, there's currently an issue open with Reed re: piped in source not respecting the configured `excluded_paths`: https://github.com/troessner/reek/issues/1429

### DIFF
--- a/ale_linters/ruby/reek.vim
+++ b/ale_linters/ruby/reek.vim
@@ -30,7 +30,7 @@ function! ale_linters#ruby#reek#GetCommand(buffer, version_output) abort
     \   : ''
 
     return ale#handlers#ruby#EscapeExecutable(l:executable, 'reek')
-    \   . ' -f json --no-progress --no-color'
+    \   . ' -f json --no-progress --no-color --force-exclusion'
     \   . l:display_name_args
 endfunction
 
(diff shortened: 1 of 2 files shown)


w0rp:
Did you check the output of `:ALEInfo`? `%s` will be expanded into the absolute path for the file, but not in the tests there. `--force-exclusion` should be good to add.

stevenharman:
> Did you check the output of `:ALEInfo`?

@w0rp I'm not sure what you're asking.

Part of my confusion/the issue here might be to do with the mentioned Reek issue (troessner/reek#1429), which causes this all to fail anyhow. But locally, once I hacked Reek to "work", the changes in the PR also "worked." Though, it may be that I didn't need the file expansion changes. Perhaps I should pull those out of this PR, leaving the `--force-exclusion` flag, and see if that works once Reek is fixed?

w0rp:
@stevenharman Try this again with just the `--force-exclusion` flag. `:ALEInfo` logs the commands that are run, so you'll see the absolute path to the file in the output there, but the `%s` format string isn't expanded in the tests for the command.

stevenharman:
@w0rp, done. Please have a look. Thanks!

w0rp:
Cheers! :beers:

PR #2001, opened 2018-10-17
Add support for `cargo clippy`

rhysd:
This PR adds support for `cargo clippy`. `cargo clippy` is official 'more aggressive' linter than `cargo check` for Rust projects. I usually use it instead of `cargo check` and want to use it with ALE.

https://github.com/rust-lang-nursery/rust-clippy

Since `cargo clippy` is a superset of `cargo check`, I added support of `cargo clippy` as an option of `cargo` linter.

This PR is actually work in progress since tests are not added yet since I couldn't. `cargo` linter checks `cargo-clippy` executable is available or not. If not, it falls back into `cargo check`. I could not understand how to test the behavior. To test it, I need to switch `cargo-clippy` is executable or not executable.

### DIFF
--- a/doc/ale-rust.txt
+++ b/doc/ale-rust.txt
@@ -132,6 +132,11 @@ g:ale_rust_cargo_use_clippy
 
   https://github.com/rust-lang-nursery/rust-clippy
 
+  Since `cargo clippy` is optional toolchain, it's safer to check whether
+  `cargo-clippy` is executable as follows:
+>
+    let g:ale_rust_cargo_use_clippy = executable('cargo-clippy')
+<
 
 g:ale_rust_cargo_clippy_options
                                               *g:ale_rust_cargo_clippy_options*


rhysd:
@w0rp New options is effective only when `cargo-clippy` exists. Should I create a faked `cargo-clippy` and add it to $PATH? It would enable to test options.

ale_linters/rust/cargo.vim
```diff
@@ -70,14 +72,23 @@ function! ale_linters#rust#cargo#GetCommand(buffer, version_output) abort
         let l:default_feature = ''
     endif
 
+    let l:subcommand = l:use_check ? 'check' : 'build'
+    let l:clippy_options = ''
+
+    if ale#Var(a:buffer, 'rust_cargo_use_clippy') && executable('cargo-clippy')
```
w0rp:
```suggestion
    if ale#Var(a:buffer, 'rust_cargo_use_clippy')
```

Could you just check if the variable is set here instead? People will have to manually enable `clippy` anyway.

rhysd:
Thank you for the suggestion. I'll write up tests tomorrow.

rhysd:
I've updated this branch. Could you review again?

rhysd:
I've updated this branch. Could you review again?

w0rp:
Cheers! :beers:

stevensonmt:
Is this an opt-in or does using `cargo` as the rust linter automatically use `clippy` if available and `cargo-check` if not?

w0rp:
You have to opt in by setting the option to `1`.

PR #2004, opened 2018-10-18
Use default setting of clang-tidy instead of enabling all checks by d…

### LINKED ISSUE #1962: clang-tidy shouldn't enable all checks by default
The default config for the clang-tidy linter for c and cpp passes in --checks=* to the clang-tidy command line. This enables all checks (even those that are disabled by default), and furthermore makes clang-tidy ignore any .clang-tidy config files that might be in the project tree. Seems like the default behavior of the linter should match the default behavior the tool, which is to enable some checks, and honor a .clang-tidy config file if found. I am able to get the desired behavior by setting `let g:ale_c_clangtidy_checks = []` in my .vimrc, in which case the linter honors the .clang-tidy file in my project. It seems like I shouldn't have to 'unconfigure' the linter clang-tidy just so I can use the defaults of clang-tidy itself.

I am using homebrew vim and llvm 7.0 on macos.

theimpostor:
…efault

Fixes #1962

### DIFF
--- a/ale_linters/c/clangtidy.vim
+++ b/ale_linters/c/clangtidy.vim
@@ -9,8 +9,9 @@ call ale#Set('c_clangtidy_executable', 'clang-tidy')
 "
 " Consult the check list in clang-tidy's documentation:
 " http://clang.llvm.org/extra/clang-tidy/checks/list.html
-
-call ale#Set('c_clangtidy_checks', ['*'])
+"
+" Use ['*'] to enable all checks.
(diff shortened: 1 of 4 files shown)


RyanSquared:
Can you fix the tests as well? Load up the log and search for `(X)` to see which tests are failing.

neersighted:
This pull request appears to have been made obsolete by #2031. Please feel free to re-open it if that is incorrect.

PR #2005, opened 2018-10-19
php: change phpstan's --errorFormat to --error-format

dreamlover:
The latest version of phpstan has marked the `--errorFormat` parameter as *deprecated* and recommend `--error-format` instead.

### DIFF
--- a/ale_linters/php/phpstan.vim
+++ b/ale_linters/php/phpstan.vim
@@ -12,7 +12,7 @@ endfunction
 
 function! ale_linters#php#phpstan#VersionCheck(buffer) abort
     let l:executable = ale_linters#php#phpstan#GetExecutable(a:buffer)
-    
+
     " If we have previously stored the version number in a cache, then
     " don't look it up again.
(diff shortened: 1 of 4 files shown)


RyanSquared:
@dreamlover this would require a forced update of phpstan if there's any possibility of someone not having phpstan updated since August, which I wouldn't say isn't a possibility. Can you add in a version check to see if it's >= 0.10.3 ?

dreamlover:
@RyanSquared `--error-format` was supported starting from phpstan 0.10.3, as specified [here](https://github.com/phpstan/phpstan/releases/tag/0.10.3).

w0rp:
Probably a good idea to add a version check. Have a look at the `flake8` linter for one example with `command_chain`.

dreamlover:
@w0rp OK, i'll trying to add this!

dreamlover:
@w0rp  Version check added.

w0rp:
Cheers! :beers:

PR #2008, opened 2018-10-19
Update Psalm plugin to use LSP mode

muglug:
Psalm was [just updated](https://github.com/vimeo/psalm/releases/tag/2.0.15) to support the Language Server Protocol.

This PR updates the Psalm plugin to use LSP, making everything faster (and allowing us to add more functionality down the line).

cc @twitchard


w0rp:
I second that suggestion. Add a `psalm-lsp` linter.

muglug:
OK, closing!

RyanSquared:
No. It breaks backwards compatibility with current setups. Either you could ghost all support for the previous version (which I wouldn't recommend; it would make users more likely to just not use your linter, as it shows a lack of history of support) and expect for the next ALE major release to have a fixed linter, or you could add in a new one.

PR #2016, opened 2018-10-24
fix: change google_java_format_* to java_google_java_format_* 

terryding77:
According to doc/ale-java.txt, we should use java_* for the variable in google_java_format


w0rp:
Cheers! :beers:

PR #2022, opened 2018-10-26
Add vim-plug installation instructions

bengadbois:
vim-plug installation is generally pretty straightforward, but it might be good to explicitly add


w0rp:
Cheers! :beers:

PR #2025, opened 2018-10-27
feat(ale_linters): add xo support for typescript files

zanona:
This PR extends [XO](https://github.com/xojs/xo/) linter support for typescript files besides javascript.
`ale_linters#javascript#xo#` methods were moved to `ale#handlers#xo#` to avoid duplication

### DIFF
--- a/ale_linters/typescript/xo.vim
+++ b/ale_linters/typescript/xo.vim
@@ -1,7 +1,7 @@
 " Author: Daniel Lupu <lupu.daniel.f@gmail.com>
 " Description: xo for JavaScript files
 
-call ale#linter#Define('javascript', {
+call ale#linter#Define('typescript', {
 \   'name': 'xo',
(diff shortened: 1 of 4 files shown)


zanona:
The only thing I am not quite sure is if we should leave option names scoped to `javascript` as in `javascript_xo_executable`? Maybe so since 'xo' is originally a js linter with extensions/plugins for typescript?

zanona:
Thanks, actually after a bit more type trying to develop using Typescript, I wasn't satisfied with performance at all, and the way xo would work at the moment was no different than TSLint either. So I believe it's best to not cultivate additions of linters that run elsewhere but through a language service plugin through tsserver.
For this reason, I will close this and perhaps after some work to integrate xo as a language service plugin we can submit another PR so performance is always favoured.

I also updated some info at #2052 just to point users in the right direction whenever choosing for a linter.

PR #2026, opened 2018-10-28
Do not enable all clang-tidy checks by default

### LINKED ISSUE #1962: clang-tidy shouldn't enable all checks by default

capjo:
Fixes #1962

Pull request #2004 for the same issue already exists, but seems to be abandoned???

Trying to fix the tests.

### DIFF
--- a/autoload/ale/fixers/rubocop.vim
+++ b/autoload/ale/fixers/rubocop.vim
@@ -9,7 +9,7 @@ function! ale#fixers#rubocop#GetCommand(buffer) abort
     return ale#handlers#ruby#EscapeExecutable(l:executable, 'rubocop')
     \   . (!empty(l:config) ? ' --config ' . ale#Escape(l:config) : '')
     \   . (!empty(l:options) ? ' ' . l:options : '')
-    \   . ' --auto-correct --force-exclusion %t'
+    \   . ' --auto-correct --force-exclusions %t'
 endfunction
 
 function! ale#fixers#rubocop#Fix(buffer) abort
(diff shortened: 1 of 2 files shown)


RyanSquared:
Couple of tests are still failing. Please open up the CI and look for `(X)` to see failed tests.

capjo:
Clazy check is a copy of the clang-tidy check??? Seems to be wrong.

`Starting Vader: /testplugin/test/command_callback/test_cpp_clazy_command_callback.vader
    (1/8) [EXECUTE] The clangtidy command default should be correct
    (1/8) [EXECUTE] (X) ['clang-tidy', '''clang-tidy'' %s'] should be equal to ['clang-tidy', '''clang-tidy'' -checks=''*'' %s']
    (2/8) [EXECUTE] You should be able to remove the -checks option for clang-tidy
    (3/8) [EXECUTE] You should be able to set other checks for clang-tidy
    (4/8) [EXECUTE] You should be able to manually set compiler flags for clang-tidy
    (4/8) [EXECUTE] (X) ['clang-tidy', '''clang-tidy'' %s -- -Wall'] should be equal to ['clang-tidy', '''clang-tidy'' -checks=''*'' %s -- -Wall']`

It seems I missed one test:

> test/command_callback/test_c_import_paths.vader

PR #2033, opened 2018-10-30
add nargs=* to ALEHover to allow it to be ran as keywordprg

dsifford:
This PR simply adds `-nargs=*` to `:ALEHover` to allow it to be set as the `keywordprg` for any given filetype in ftplugins.

**Example:**

```vim
" ftplugin/typescript.vim

setlocal keywordprg=:ALEHover
```

This is useful particularly for LSP linters to allow the normal `K` mapping to remain `keywordprg` so that it still remains intelligent on filetypes where ALE isn't around (specifically helpfiles).

Let me know if you have any questions.

### DIFF
--- a/plugin/ale.vim
+++ b/plugin/ale.vim
@@ -191,7 +191,7 @@ command! -bar ALEGoToDefinitionInTab :call ale#definition#GoTo({'open_in_tab': 1
 command! -bar ALEFindReferences :call ale#references#Find()
 
 " Get information for the cursor.
-command! -bar ALEHover :call ale#hover#Show(bufnr(''), getcurpos()[1],
+command! -bar -nargs=* ALEHover :call ale#hover#Show(bufnr(''), getcurpos()[1],
                                             \ getcurpos()[2], {})
 
 " <Plug> mappings for commands


dsifford:
Can't figure out where (or why?) appveyor is failing due to this PR... Seems hard to believe it would...

capjo:
This is an issue on the master branch.

w0rp:
The Windows tests should pass again on `master`. I'm unsure of this. I would like passing arguments to `:ALEHover` to continue to be an error.

dsifford:
I'm happy to brainstorm an alternative to this if that makes you uneasy... But I would imagine there are more users like me that would like the keep the keywordprg mapping intact. Seems like low-hanging fruit to just add this in.

dsifford:
Circling back to check up on this. Any ideas @w0rp? And if I may ask, what is it specifically that is concerning about this as it stands?

w0rp:
The problem is that `:ALEHover ` should throw an error if given any arguments.

dsifford:
Right, I got that... But my question really boils down to **why** should this error be thrown? What purpose does this error solve?

w0rp:
So people will know it's not vallid to use the command like that.

I think if you want to use this command for `keywordprg`, you can write a wrapper command for that.

dsifford:
No worries. That's what I'm doing now. Thanks for considering.

PR #2035, opened 2018-10-30
Add a `lsp_config_callback` linter option

jparise:
This is the callback-based variant of the existing `lsp_config` linter
option. It serves the same purpose but can be used when more complicated
processing is needed.

`lsp_config` and `lsp_config_callback` are mutually exclusive options;
if both an given, a linter preprocessing error will be raised.

The runtime logic has been wrapped in `ale#lsp_linter#GetConfig` for
convenience, similar to `ale#lsp_linter#GetOptions`.

This also adds documentation and an `AssertLSPConfig` test function for
completeness.

See #1852

### DIFF
--- a/autoload/ale/assert.vim
+++ b/autoload/ale/assert.vim
@@ -85,6 +85,14 @@ function! ale#assert#LSPOptions(expected_options) abort
     AssertEqual a:expected_options, l:initialization_options
 endfunction
 
+function! ale#assert#LSPConfig(expected_config) abort
+    let l:buffer = bufnr('')
+    let l:linter = s:GetLinter()
+    let l:config = ale#lsp_linter#GetConfig(l:buffer, l:linter)
+
+    AssertEqual a:expected_config, l:config
+endfunction
+
 function! ale#assert#LSPLanguage(expected_language) abort
     let l:buffer = bufnr('')
     let l:linter = s:GetLinter()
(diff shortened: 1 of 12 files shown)


jparise:
> Looks good. Could we do one more thing? Let's record the config in the lsp.vim file in the Dictionary for the connection, and only send a didChangeConfiguration message if the config Dictionary is different (!=)from the previous one. Otherwise, we'll send adidChangeConfiguration` message every time we check a buffer, ask for completion information, etc.

Yes, I'd be glad to make that change. I started to notice that the other day when I was debugging the ElixirLS connection, and I was surprised it was being sent (redundantly) so frequently. Thanks for hint!

w0rp:
You can put a `" no-custom-checks` comment at the end of the line to tell the script to ignore that.

jparise:
These AppVeyor failures appear to be unrelated to this change:

```
    (1/2) [EXECUTE] The default command should be correct
    (1/2) [EXECUTE] (X) ['jshint', 'jshint --reporter unix --extract auto --filename %s -'] should be equal to ['jshint', '''jshint'' --reporter unix --extract auto --filename %s -']
    (2/2) [EXECUTE] Setting a config location should add the config parameter
    (2/2) [EXECUTE] (X) ['jshint', 'jshint --reporter unix --extract auto --config /some/file --filename %s -'] should be equal to ['jshint', '''jshint'' --reporter unix --extract auto --config ''/some/file'' --filename %s -']
```

w0rp:
Cheers! :beers:

PR #2038, opened 2018-10-31
Adds -B flag to make when parsing a makefile.

bakpakin:
Its a pretty simple change.

Before, if `g:ale_c_parse_makefile` was enabled but the target was already built, many makefiles would output nothing. This should make all targets dry-run unconditionally.

### DIFF
--- a/autoload/ale/c.vim
+++ b/autoload/ale/c.vim
@@ -223,7 +223,7 @@ function! ale#c#GetMakeCommand(buffer) abort
         let l:makefile_path = ale#path#FindNearestFile(a:buffer, 'Makefile')
 
         if !empty(l:makefile_path)
-            return 'cd '. fnamemodify(l:makefile_path, ':p:h') . ' && make -n'
+            return 'cd '. fnamemodify(l:makefile_path, ':p:h') . ' && make -Bn'
         endif
     endif
 


bakpakin:
That's a bummer. There seems to be no BSD equivalent of the GNU -B flag. My other initial thought was to `touch` the current source file before running `make -n`, which should allow well structured makefiles to work correctly. This is a bit gross though and I don't like it, but I can't think of a better way for BSD.

RyanSquared:
Given that this is system-dependent, should we add an option, enabled by default (I suspect there'll be more GNU make users rather than BSD make), that determines whether or not to include the `-B` flag?

w0rp:
I'd check if there's some way to safely add the flag for GNU make, without breaking people who use BSD make.

RyanSquared:
I'm not sure about breaking BSD make, but you could use the long version of the flag: `--always-make`. That way, either you get success or you get an obvious failure.

w0rp:
My suggestion is to try using `-B` on bsd, which has a completely different meaning, but might happen to work.

RyanSquared:
It shouldn't; the only thing different is that it now uses a different jobs system. It still considers things "built" from previous iterations.

w0rp:

w0rp:
@bakpakin ALE now uses `--always-make` by default on both Linux and BSD, even though it will break BSD `make`. BSD users will have to use `let g:ale_c_always_make = 0` to get `make -n` parsing to work. I decided it was worth changing, even if it breaks `make` on BSD by default. There are barely any people on Earth who will care, and those who do can just set another option.

bakpakin:
Thanks, I appreciate the change. Makes ale work a bit better by default on some of my Make-based projects.

PR #2040, opened 2018-10-31
Pass the filename of the current file into jshint

leamingrad:
This MR updates the `jshint` linter so that the filename of the file being linted is passed in on the command line. Doing this allows file-specific override config to be used, which is helpful on large projects.

(Note that jshint is currently bugged in this area, so there will probably be little visible change to the end user until https://github.com/jshint/jshint/pull/3335 goes in).

I haven't added tests for this function, as there did not seem to be any existing ones in this area to jump off from - I'd be happy to add some with guidance if needed.


leamingrad:
Added some basic command callback tests, but happy to add a couple more if there is something obvious

w0rp:
Cheers! :beers:

PR #2044, opened 2018-11-01
PHP linter: make the path to the executable configurable

### LINKED ISSUE #1388: Specify php executable.
Most other linters let you specify the the path/location of the linter, but the php linter just uses the first php executable in `$PATH`. We should be able to specify the executable the same as we do for other linters.

For what it's worth, the use case here is developing locally, but having tools (such as PHP) installed on a vagrant virtual box. Also for having multiple instances of PHP installed (system php and brew PHP, for example)

Steap:
Fixes #1388


w0rp:
Cheers! :beers:

PR #2045, opened 2018-11-01
elixir-ls now recognizes umbrella projects

jparise:
Previously, elixir-ls would treat each sub-project within an umbrella as
standalone, which isn't desirable from a language server perspective.

Added `ale#handlers#elixir#FindMixUmbrellaRoot`, which locates the current
project's root and then continues searching upwards for a potential
umbrella project root. This literally looks just two levels up to keep
things simple while keeping in line with Elixir project conventions.

Use this new function to determine elixir-ls's LSP project root.


w0rp:
There are some conflicts here to fix now.

jparise:
> There are some conflicts here to fix now.

Resolved.

PR #2048, opened 2018-11-02
Allow configuration of hamllint executable

justinhoward:
The hamllint executable was hard-coded, preventing it from being
overridden. Fix the executable to be dynamic to allow custom executable
paths.


w0rp:
Cheers! :beers:

PR #2050, opened 2018-11-03
fix phpcs linter error: 'option --stdin-path' not known'

dementor5:
Good afternoon! the syntax of phpcss has changed:
ERROR: option "--stdin-path=/home/me/project/update/load.php" not known.

I simple remove key "--stdin-path" and all work again.

### DIFF
--- a/ale_linters/php/phpcs.vim
+++ b/ale_linters/php/phpcs.vim
@@ -14,7 +14,7 @@ function! ale_linters#php#phpcs#GetCommand(buffer) abort
     \   : ''
     let l:options = ale#Var(a:buffer, 'php_phpcs_options')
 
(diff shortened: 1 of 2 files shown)


capjo:
Tests in file "test_phpcs_command_callback.vader" are failing. Remove the --stdin-path also there.

capjo:
Which version of phpcs are you using?

I couldn't find any hints that this option was removed.

w0rp:
That option exists. You should update `phpcs` to a newer version.

dementor5:
> That option exists. You should update `phpcs` to a newer version.
Thank you! My version was 2.2.0.
All ok after update on 3.3.2.
Excuse me for troubling.

PR #2052, opened 2018-11-04
docs(typescript): update instructions for TSLint replacement

zanona:
After Typescript 2.3 language service plugins were natively implemented and a TSLint plugin released by Microsoft.
This PR only points users to the most up-to-date plugin repository as well as providing insights on what the language service plugin is with link to the wiki page.


w0rp:
The formatting isn't right here. The Vim help files don't use Markdown formating for hyperlinks. I'll update the link in the documentation and the text above it.

w0rp:
Thanks anyway!

PR #2053, opened 2018-11-04
Make ocamlformat work without temporary files

andreypopp:
Problem: ocamlformat is configured to format files in-place and thus go via creating a temporary file for that. Because the temporary file resides in a different directory ocamlformat can't find `.ocamlformat` configuration files in an original location of source files.

Solution: ocamlformat since version 0.8 can read sources of stdin and spur result on stdout. We reconfigure ocamlformat fixer to use this simpler interface.

cc @sbl


sbl:
Thanks for the fix. Tested the branch locally and this works fine with local `.ocamlformat` files.

w0rp:
Cheers! :beers:

Issue #2055, opened 2018-11-06
Add support for NeoVim virtual text (annotations).


hsanson:
NeoVim supports virtual text (annotations) within a buffer (see :h nvim_buf_set_virtual_text) that would allow ALE to put the error messages next to the line of code causing the error.

See demo video below:

   https://asciinema.org/a/qjRVAwEOFh6KzhIN9TDHDDjWr?t=20&speed=1.6

This combined with nvim_buf_add_highlight would make ALE even better than with it is now.

neersighted:
I believe this was implemented in #2056. Feel free to re-open if it was not.

RyanSquared:
There's a couple of bugs with it, from what I've seen, but yes the minimal implementation exists.

PR #2056, opened 2018-11-06
Add support for nvim's virtualtext on cursor

luan:
Addresses #2055. I haven't added tests for this yet and I may have inadvertently violated parts of the development guide. But I was just very excited to play around with this. The API being used here (`nvim_buf_set_virtual_text`) is not yet released on neovim so I'm not sure we'd want to merge this before that happens. I'll look into adding tests for this tomorrow! But I'd love some early feedback on it.

Summary of the changes:
- Add g:ale_virtualtext_cursor boolean to enable/disable it
- Add g:ale_virtualtext_prefix to configure what prefix to use (default:
'> ')
- Requires neovim 0.3.2's unreleased API `nvim_buf_set_virtual_text`

Demo:
[asciicast](https://asciinema.org/a/DUvYWKMpsEM6eLPeQHPKiGkjb)


w0rp:
It might be a good idea to put support for `virtualtext` messages entirely in their own file, and use separate `autocmd` events for it, which can be set only if the option is enabled. That way, you'll be able to turn echoing messages off, and turn `virtualtext` messages on.

luan:
@w0rp I moved everything to `autoload/ale/virtualtext.vim` as suggested. Decided to extract a shared function into `ale#util` for `ale#util#FindItemAtCursor`. I also split out `g:ale_virutaltext_delay` to have both echo and virtualtext be fully independent of each other.

Wasn't super sure about [this block](https://github.com/w0rp/ale/blob/ba9b056a57a605e1253b9f50e94074499bf808b4/autoload/ale/events.vim#L134-L148) seems like a lot of duplicaiton for the autocmds there. I also didn't quite understand why we needed to check `if exists('*ale#engine#Cleanup')` for those, but I just followed it for the virtualtext bit.

I fixed the tests that the original commit had broken with regards to `ALEInfo`. And I tried to write tests for the specific virutaltext functionality. I ran into two problems with that:
- There is no way (that I could find) to inspect what is the current virutaltext being shown on any given line.
- Even if there was, the tests would depend on an unreleased version of neovim, which the current infrastructure doesn't run against.

luan:
@w0rp, thoughts?

w0rp:
Cheers! :beers:

ahmedelgabri:
Not sure what is wrong here but the messages don't get updated when I move the cursor, you can see the normal echo messages in the command line but the virtualtext is not being updated the same way

error

babariviere:
@ahmedelgabri you have to enable option `ale_virtualtext_cursor`

ahmedelgabri:
@babariviere I already enabled it. Otherwise inline error messages wouldn’t have appeared at all.

babariviere:
oh yeah sorry, i haven't see the inline message

luan:
> Nice! I've noticed that for some linters (especially Rust), the error messages are across multiple lines. Does virtualtext only support being displayed on a single line? (As it is now, the error message just has the `\n` characters displayed as `^@` control sequences.)

@alok Yeah, it's on a single line. I actually thought about showing multiline ones, but I got stuck with thinking about when the next line is longer than the previous it would make it hard to present.

@ahmedelgabri I haven't seen anything like that, can you help me reproduce that issue?

ahmedelgabri:
@luan sure, I'm trying to disable plugins & settings one by one to see if I can find the issue. This is my [ale config](https://github.com/ahmedelgabri/dotfiles/blob/master/files/.vim/after/plugin/ale.vim) in case you see something odd or something.

megalithic:
> > Nice! I've noticed that for some linters (especially Rust), the error messages are across multiple lines. Does virtualtext only support being displayed on a single line? (As it is now, the error message just has the `\n` characters displayed as `^@` control sequences.)
> 
> @alok Yeah, it's on a single line. I actually thought about showing multiline ones, but I got stuck with thinking about when the next line is longer than the previous it would make it hard to present.
> 
> @ahmedelgabri I haven't seen anything like that, can you help me reproduce that issue?

@luan is it possible to a.) set textwidth for the virtualtext buffer; or b.) have `\n` properly render as newlines? /cc @alok

pwntester:
That is a cool addition, thanks @luan!

Would it be possible to add an option to justify the text to the rigth based on `winwidth(0)` and text length?

kristijanhusak:
Can we use separate highlighting group for this? I would like to use color that doesn't create so much contrast, but if i change any of the ALE* groups, it also changes the coloring on the exact spot where the error is.

kristijanhusak:
Thanks!

uri:
@ahmedelgabri did you ever find the problem? I'm having the same issue.

ahmedelgabri:
@uri Didn't have enough time to debug it yet, I disabled it for the time being.

uri:
@ahmedelgabri if you are looking for virtual text https://github.com/neomake/neomake does it pretty well.

maxim-usikov:
@luan this option just for nvim?

`g:ale_virtualtext_cursor`

meain:
@maxim-usikov yeah, vim does not have support for the feature `virtualtext`.

maxim-usikov:
@meain thanks. 

May be add some notes about that to docs?

w0rp:
I added a sentence to the documentation to clarify that now.

PR #2057, opened 2018-11-06
Add g:ale_echo_command variable

tyru:
I added g:ale_echo_command variable which is used for echoing message string to command-line.
I don't want ale to record echoed messages in history (:h message-history), so I want to use `:echo` command to echo.

```vim
let g:ale_echo_command = 'echo'
```

What do you think about this?


w0rp:
It's not a good idea. The command is written specifically so it truncates messages and doesn't show "Press Enter to continue" prompts. Anything else causes problems. `exec "norm! :echomsg l:message\n"` is the best option.

tyru:
Okay, thanks.

PR #2059, opened 2018-11-06
Remove test/command_callback/mix_paths/

jparise:
All of the other tests were already using equivalent fixtures under
`test/command_callback/elixir_paths/`, so let's use that path everywhere.


w0rp:
Cheers! :beers:

PR #2060, opened 2018-11-07
Add prettier to html formatters

meain:
Prettier now has support for formatting html documents.

https://prettier.io/blog/2018/11/07/1.15.0.html


w0rp:
Cheers! :beers:

PR #2062, opened 2018-11-07
Change the PHP project root detection method

lkebin:
Change method of find out project root directory.

For a PHP project, find out `composer.json` file to indicate the project root should be better than VCS folder.
Sometimes project use `git` but another using `hg` and `svn`, but if you are using php-language-server, you may also using composer.

### DIFF
--- a/autoload/ale/path.vim
+++ b/autoload/ale/path.vim
@@ -30,7 +30,6 @@ function! ale#path#FindNearestMarker(buffer, markers) abort
             if filereadable(l:fn) || isdirectory(l:fn)
                 return l:fn
             endif
-
         endfor
     endfor
 
--- a/autoload/ale/path.vim
+++ b/autoload/ale/path.vim
@@ -22,23 +22,16 @@ endfunction
 function! ale#path#FindNearestMarker(buffer, markers) abort
     let l:buffer_filename = fnamemodify(bufname(a:buffer), ':p')
     let l:buffer_filename = fnameescape(l:buffer_filename)
-
-    for l:marker in a:markers
-        let l:pivot = l:buffer_filename
-
-        while 1
-            let l:prev = l:pivot
-            let l:pivot = fnamemodify(l:pivot, ':h')
-            let l:fn = l:pivot.(l:pivot is# '/' ? '' : '/').l:marker
+    
+    for l:path in ale#path#Upwards(l:buffer_filename)
+        for l:marker in a:markers
+            let l:fn = l:path.(l:path is# '/' ? '' : '/').l:marker
 
             if filereadable(l:fn) || isdirectory(l:fn)
                 return l:fn
             endif
 
-            if l:pivot is# l:prev
-                break
-            endif
-        endwhile
+        endfor
     endfor
 
     return ''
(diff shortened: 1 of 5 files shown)


RyanSquared:
Is there a case where you're using Git but not using Composer? Is there any possibility that an ALE user might have their linter stop working because of this?

lkebin:
@RyanSquared 
Yes, some PHP projects using Composer with [Mercurial (hg)](https://www.mercurial-scm.org/) or [SVN](https://subversion.apache.org/).
Yes. Some projects linter function will be stop working with this changes.

If composer.json is not a good idea for project root detection, may be replace the static file/directory to dynamic variable should be better.

w0rp:
If some projects don't use composer, then support both. Check for composer.json first, and fall back on `.git`.

lkebin:
Yes, we can support both of them. 

Maybe define a global variable will be better.  Such as 
```vim
g:ale_php_project_root_marker = ['.git', '.hg', 'composer.json', '.svn', '...']
```
Check each element until the first found, if nothing found `getcwd()` should be fine.
How do you think ?

w0rp:
Yes, that might be a good option.

autoload/ale/path.vim
```diff
@@ -17,6 +17,33 @@ function! ale#path#Simplify(path) abort
     return substitute(simplify(l:win_path), '^\\\+', '\', 'g') " no-custom-checks
 endfunction
 
+" Given a buffer and a file or directory name list, find the nearest file or directory by searching upwards
+" through the paths relative to the given buffer.
+function! ale#path#FindNearestMarker(buffer, markers) abort
+    let l:buffer_filename = fnamemodify(bufname(a:buffer), ':p')
+    let l:buffer_filename = fnameescape(l:buffer_filename)
+
+    for l:marker in a:markers
```
w0rp: Use a `for` loop with `ale#path#Upwards` instead. Have a look at the functions in the `python.vim` file for examples.

autoload/ale/path.vim
```diff
@@ -129,8 +156,8 @@ endfunction
 function! ale#path#IsBufferPath(buffer, complex_filename) abort
     " If the path is one of many different names for stdin, we have a match.
     if a:complex_filename is# '-'
-    \|| a:complex_filename is# 'stdin'
-    \|| a:complex_filename[:0] is# '<'
+                \|| a:complex_filename is# 'stdin'
+                \|| a:complex_filename[:0] is# '<'
```
w0rp: Undo these changes.

autoload/ale/path.vim
```diff
@@ -153,7 +180,7 @@ function! ale#path#IsBufferPath(buffer, complex_filename) abort
     let l:buffer_filename = expand('#' . a:buffer . ':p')
 
     return l:buffer_filename is# l:test_filename
-    \   || l:buffer_filename[-len(l:test_filename):] is# l:test_filename
+                \   || l:buffer_filename[-len(l:test_filename):] is# l:test_filename
```
w0rp: Undo these changes.

autoload/ale/path.vim
```diff
@@ -187,13 +214,13 @@ function! ale#path#ToURI(path) abort
...
-    \   '/',
-    \   'g',
-    \)
+                \   ((l:has_drive_letter || a:path[:0] is# '/') ? 'file://' : '')
+                \       . (l:has_drive_letter ? '/' . a:path[:2] : '')
+                \       . ale#uri#Encode(l:has_drive_letter ? a:path[3:] : a:path),
+                \   '\\',
+                \   '/',
+                \   'g',
+                \)
```
w0rp: Undo these changes.

lkebin:
@w0rp I did some changes according to your suggestions.

lkebin:
Let me submit PR again when I have a free time to consider a better solution.

Issue #1305, opened 2018-01-22
Change checkstyle linter to allow for setting classpath


nemrod:
The current checkstyle implementation uses the checkstyle command, which doesn't allow for setting classpath. Without it we can't change checkstyle version or, as is the trouble in my case, specify an additional jar file that we have built ourselves that extends the functionality.

I suggest checkstyle is called in the way checkstyle themselves specify on [this page](http://checkstyle.sourceforge.net/cmdline.html), e.g. "java -classpath MyCustom.jar;checkstyle-8.7-all.jar com.puppycrawl.tools.checkstyle.Main -c config.xml Check.java". An option needs to be added, "ale_java_checkstyle_classpath". A bonus is that we don't need a checkstyle binary installed at all.

Syntastic does checkstyle in this way for this reason, you can check out their code for how to do it.

w0rp:
Feel free to create a pull request for this with tests.

languitar:
Along these lines it would also be good to make the use of a temporary file an option because this break the check for the existence of a package-info.java file.

languitar:
Just as a note, what I did to launch a custom checkstyle version is to set the following in an `.exrc` file:
```vim
let $PATH=getcwd() . ':' . $PATH
```
The project root then contains as `checkstyle` launcher script for the correct version.

hsanson:
Once #2542 is accepted and merged it would be possible to configure the binary path used to invoke checkstyle. With this we can create launch scripts with all the configuration options we want and tell ALE to use that launch script.

hsanson:
Creating an executable script (e.g. mycheckstyle.sh) that contains the invokation command and then setting is as checkstyle executable should do what is needed:

```bash
#!/usr/bin/env bash
# mycheckstyle.sh script to invoke checkstyle via JAR file.
java -classpath MyCustom.jar;checkstyle-8.7-all.jar com.puppycrawl.tools.checkstyle.Main "$@"
```

Then set in your vimrc:

```
let g:ale_java_checkstyle_executable = '/absolute/path/to/mycheckstyle.sh'
```

PR #2063, opened 2018-11-07
Checkstyle improvements

languitar:
This PR adds support for an older output format of checkstyle and removes the support for linting temporary files, which causes problems for some checks as discussed in #1305.


w0rp:
Cheers! :beers:

PR #2068, opened 2018-11-11
Add linter for ispc

m-pilia:
Hi w0rp! I added a linter for [ispc](https://github.com/ispc/ispc).

Unfortunately the output of `ispc` is quite inconsistent and messages can be split over multiple lines quite randomly, but this should be able to recover the entire error/warning message text.

I have basically zero experience with VimL, so if I did something stupid in the code please let me know and I will try to fix it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ formatting.
 | HCL | [terraform-fmt](https://github.com/hashicorp/terraform) |
 | HTML | [alex](https://github.com/wooorm/alex) !!, [HTMLHint](http://htmlhint.com/), [proselint](http://proselint.com/), [tidy](http://www.html-tidy.org/), [prettier](https://github.com/prettier/prettier), [write-good](https://github.com/btford/write-good) |
 | Idris | [idris](http://www.idris-lang.org/) |
-| ISPC | [ispc](https://ispc.github.io/) |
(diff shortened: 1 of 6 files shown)


m-pilia:
I was having a closer look at the source code of ispc, the annoyance comes from the fact that it [detects](https://github.com/ispc/ispc/blob/71b24470efc505880405e1fc5d34d234d16569c3/src/util.cpp#L80-L98) the terminal width and wraps output lines accordingly. So messages are minced in different ways according to the width of the Vim terminal. Great...

The current implementation of the handler copes with almost all error messages, searching through the source I found a few uncommon messages that are not terminated by punctuation. I am considering to open a PR to ispc to fix the format inconsistencies upstream, rather than bloat the handler here.

In any case, the while loop in the handler was designed such that, even if the termination of a message is missed, some garbage may be added to that message, but then handling recovers without missing any error/warning.

w0rp:
It looks like you can pass `--nowrap` to the compiler to tell it not to wrap long lines. https://github.com/ispc/ispc/blob/71b24470efc505880405e1fc5d34d234d16569c3/src/main.cpp#L639 Try that.

m-pilia:
Great, thanks! It works now. That option is totally undocumented.

w0rp:
Cheers! :beers:

PR #2071, opened 2018-11-12
Add specific highlight group for virtual text

jamestthompson3:
This changes the highlight group used in the `ale_virtualtext` function. It allows for the styling of virtual text to be different than the default `ALEError` or `ALEWarning` allowing vitualtext to stand out better.

Before:
old_highlight

After:
new_highlight

### DIFF
--- a/autoload/ale/virtualtext.vim
+++ b/autoload/ale/virtualtext.vim
@@ -63,11 +63,18 @@ function! ale#virtualtext#ShowCursorWarning(...) abort
         let l:type = get(l:loc, 'type', 'E')
 
         if l:type is# 'E'
-            let l:hl_group = 'ALEVirtualTextError'
+            if exists('g:ale_virtualtext_error_hi')
+              let l:hl_group = g:ale_virtualtext_error_hi
+            else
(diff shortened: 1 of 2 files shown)


autoload/ale/virtualtext.vim
```diff
@@ -63,9 +63,9 @@ function! ale#virtualtext#ShowCursorWarning(...) abort
         let l:type = get(l:loc, 'type', 'E')
 
         if l:type is# 'E'
-            let l:hl_group = 'ALEError'
+            let l:hl_group = 'ALEVirtualTextError'
         elseif l:type is# 'W'
-            let l:hl_group = 'ALEWarning'
+            let l:hl_group = 'ALEVirtualTextWarning'
```
w0rp: You should define a default for this group above, aliasing it to the old group names, and document the highlights in the section for highlighting in the main help file.

w0rp:
You should use the new highlight group names like you did in your first commit, and set them to default to the other highlight groups. Have a look at the signs or highlight files for examples. Document the new highlights in the section for highlighting.

PR #2077, opened 2018-11-13
refactor: sasslint linter

sQVe:
Previous implementation required one to have `sass-lint` globally. This
allows you to have it locally, override the executable and add options.


test/command_callback/test_scss_sasslint_command_callback.vader
```diff
@@ -0,0 +1,43 @@
...
+
+Execute(should fallback to bin, .bin/sass-lint):
+  call ale#test#SetFilename('sasslint-test-files/with-bin/test.scss')
+
+  let b:executable = ale#path#Simplify(
+  \   g:dir
+  \   . '/sasslint-test-files/with-bin/node_modules/.bin/sass-lint'
+  \)
+
+  AssertLinter b:executable, ale#Escape(b:executable) . ' -v -q -f compact %t'
```
w0rp: The test failed on Travis CI here because you need to commit hidden directories or files with `git add -f`. Hidden files are in `.gitignore`.

sQVe:
@w0rp Making progress. The implementation tests are now passing but failing for the added docs. I'm a bit unsure on why.

sQVe:
@neersighted :+1: I've had the exact same thoughts regarding duplication. What makes it a bit tricky is that the big difference between scss and sass makes it quite likely that you would want different options for them. I know cases where you want to specifically set a config file for legacy sass files and another for scss.

If you guys can think of a better solution I'm happy to implement it. I'd need some guidance though as I'm not super comfortable with Vimscript.

RyanSquared:
I don't see a good way to merge the two and like you said it is definitely possible people could have different options. I think this is fine to merge as-is unless anyone else sees an obvious issue. I'll possibly get to merging this sometime over the weekend.

PR #2083, opened 2018-11-16
Continue scalac compilation until just before bytecode generation

zackhsi:
Previous behavior does not compile deep enough to surface errors.

See compiler phases:
https://docs.scala-lang.org/overviews/compiler-options/index.html#compilation-phases

### DIFF
--- a/ale_linters/scala/scalac.vim
+++ b/ale_linters/scala/scalac.vim
@@ -9,7 +9,7 @@ endfunction
 call ale#linter#Define('scala', {
 \   'name': 'scalac',
 \   'executable_callback': {buf -> s:IsSbt(buf) ? '' : 'scalac'},
-\   'command': '%e -Ystop-after:parser %t',
+\   'command': '%e -Ystop-before:jvm %t',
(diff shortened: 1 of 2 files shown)


w0rp:
Just to be sure, is there any risk of executing arbitrary code with this change, or is this just a step for generating bytecode which isn't executed? Update the tests to match the changes to the command.

zackhsi:
Nope, no executables/classfiles are produced until the last step, `jvm`, so there is nothing to execute. The specific flag `-Ystop-before:jvm` tells the compiler to stop _before_ the `jvm` step.

w0rp:

w0rp:
I had to revert this, as it broke using `scalac` at all for some people. We can consider adding an option for this instead.

RyanSquared:
I did some snooping and came across this page:

https://docs.scala-lang.org/overviews/compiler-options/index.html#compilation-phases

I don't see either `jvm` or `parser`, but I do see `parse`. Is the documentation for this out of date or something? Do we need some kind of version check system?

PR #2101, opened 2018-11-24
fixing reading of ale_python_flake8_args

lo002:
I believe I fixed the reading of the ale_python_flake8_args from the vimrc.

Currently it is set to an empty string not taking into account the value set in the vimrc for the corresponding variable.


w0rp:
This option was removed a while ago now. Use `_options` instead.

PR #2103, opened 2018-11-25
Don't hardcode the namespace id for virtual text

bfredl:
This makes collisions with other plugins less likely.

Note: this function was merged just yesterday to Nvim master. A variant that works since earlier is to initialize `s:ns_id` to zero, and use `let s:ns_id = nvim_buf_set_virtual_text(l:buffer, s:ns_id, ...)`


w0rp:
I'm fine with just using the function, as this version of NeoVim isn't stable yet, so we don't have to expect the use of the API to be stable either yet. This is nice, I didn't know about this function.

w0rp:
Cheers! :beers:

I wish there was an API as easy to use as this for the `loclist`. I've seen something in recent version of Vim, but it is incredibly awkward to use, versus just "give me a unique ID" and "here is the unique ID for this list."

PR #2110, opened 2018-11-29
Only set the unnamed register at the end of :ALEInfoToClipboard

neersighted:
This fixes performance problems in Neovim, where every character results
in spawning a new clipboard-tool process.

Behaviour is not similarly pathological in Vim, but it still results in
an unnecessary amount of register churn.


RyanSquared:
My only concern with this is the potential for memory duplication versus what _should_ be an append-only mechanism, but I think with the current method it would be just as bad.

Cheers! :beers:

PR #2118, opened 2018-12-04
Move README images to gh-pages

idbrii:
Move pngs out of repo so users aren't pulling the pngs on every clone. (Saves 4.3 MB, which is more than half of the repo's size. I think worth removing.)

Also requires my [gh-pages](https://github.com/idbrii/ale/tree/gh-pages) branch commit 8a34ebc4aa4df92507c612fa691096c97c08dacf! (I don't know how to do a PR since you don't have a gh-pages branch.)

You can see that it works by browsing at my test commit (that used my repo's gh-pages): https://github.com/idbrii/ale/tree/no-img-in-repo-test

(I modeled this setup after what OmniSharp/vim-omnisharp does.)

Maybe we'd want to add the logo psd from 071c04da3d6cd5fa1a9f4d2d3d38448bf9ea0e1e to gh-pages?

Only three png links changed because README references to issues.png and no_issues.png were removed in ed8f79987d5a64c41f8b5c1c244243babde45d25. As far as I can tell (`git log -S issues.png`), they were not added anywhere else.


w0rp:
We've decided that we can't do this for now.  Removing images from the repository, such that the size of `git clone` is smaller requires rewriting the entire commit log, as if those files were never committed. That will cause a large number of problems that take more time to deal with than we have right now.

I do want the size of the repo to eventually be smaller.

idbrii:
This change doesn't rewrite the commit log and would provide a place outside of the repo to put images. Commits with new images would be more likely to follow the out-of-repo pattern.

While I understand why you wouldn't want to rewrite history, why not apply this change now to prevent further repo bloat?

Additionally, if someone wanted a tiny checkout, they could use `git clone --depth 20 ` and so long as there were 20 commits after this one, the images should not be included in their checkout.

Also, by this comment: 

> Saves 4.3 MB, which is more than half of the repo's size. I think worth removing.)

I meant the size of the images themselves and not their storage in git history.

w0rp:
The PR would only make the download size smaller of you used `--depth`, and the PR can't be merged until the second repository for the images is created. I don't think it's worth the hassle at the moment.

RyanSquared:
It should be noted that ALE does have a separate repository already - the wiki itself is a separate repo. Further images or other documents that should not be involved in ALE itself can be put there.

PR #2121, opened 2018-12-05
WIP Fix 1996 - Add eclipse LSP support.

### LINKED ISSUE #1996: Add support for Support eclipse.jdt.ls
**Name:** eclipse.jdt.ls
**URL:** https://github.com/eclipse/eclipse.jdt.ls

LSP for Java

hsanson:
Adds support for the [eclipse language server](https://github.com/eclipse/eclipse.jdt.ls) that seems to work better than javalsp for gradle and maven projects.

Closes #1996

### DIFF
--- a/ale_linters/java/eclipselsp.vim
+++ b/ale_linters/java/eclipselsp.vim
@@ -1,0 +1,130 @@
+" Author: Horacio Sanson <https://github.com/hsanson>
+" Description: Support for the Eclipse language server https://github.com/eclipse/eclipse.jdt.ls
+
+let s:version_cache = {}
+
+call ale#Set('java_eclipselsp_path', 'eclipse.jdt.ls')
+call ale#Set('java_eclipselsp_executable', 'java')
+
+function! ale_linters#java#eclipselsp#Executable(buffer) abort
+    return ale#Var(a:buffer, 'java_eclipselsp_executable')
+endfunction
+
+function! ale_linters#java#eclipselsp#TargetPath(buffer) abort
+    return ale#Var(a:buffer, 'java_eclipselsp_path')
+endfunction
+
+function! ale_linters#java#eclipselsp#JarPath(buffer) abort
+    let l:path = ale_linters#java#eclipselsp#TargetPath(a:buffer)
+    let l:path = l:path . '/org.eclipse.jdt.ls.product/target/repository/plugins'
+
+    let l:files = globpath(l:path, 'org.eclipse.equinox.launcher_*.jar', 1, 1)
+
+    if empty(l:files)
+        return ''
+    endif
+
+    return l:files[0]
+endfunction
+
+function! ale_linters#java#eclipselsp#ConfigurationPath(buffer) abort
+    let l:path = ale_linters#java#eclipselsp#TargetPath(a:buffer)
+    let l:path = l:path . '/org.eclipse.jdt.ls.product/target/repository'
+
+    if has('win32')
+        let l:path = l:path . '/config_win'
+    elseif has('macunix')
+        let l:path = l:path . '/config_mac'
+    else
+        let l:path = l:path . '/config_linux'
+    endif
+
+    return ale#path#Simplify(l:path)
+endfunction
+
+function! ale_linters#java#eclipselsp#VersionCheck(version_lines) abort
+    return s:GetVersion('', a:version_lines)
+endfunction
+
+function! s:GetVersion(executable, version_lines) abort
+    let l:version = []
+
+    for l:line in a:version_lines
+        let l:match = matchlist(l:line, '\(\d\+\)\.\(\d\+\)\.\(\d\+\)')
+
+        if !empty(l:match)
+            let l:version = [l:match[1] + 0, l:match[2] + 0, l:match[3] + 0]
+            let s:version_cache[a:executable] = l:version
+            break
+        endif
+    endfor
+
+    return l:version
+endfunction
+
+function! ale_linters#java#eclipselsp#CommandWithVersion(buffer, version_lines, meta) abort
+    let l:executable = ale_linters#java#eclipselsp#Executable(a:buffer)
+    let l:version = s:GetVersion(l:executable, a:version_lines)
+
+    return ale_linters#java#eclipselsp#Command(a:buffer, l:version)
+endfunction
+
+function! ale_linters#java#eclipselsp#Command(buffer, version) abort
+    let l:path = ale#Var(a:buffer, 'java_eclipselsp_path')
+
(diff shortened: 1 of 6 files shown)


ale_linters/java/eclipselsp.vim
```diff
@@ -0,0 +1,92 @@
...
+endfunction
+
+function! ale_linters#java#eclipselsp#Command(buffer) abort
+    let l:executable = ale_linters#java#eclipselsp#Executable(a:buffer)
+
+    if empty(l:executable)
+        return ''
+    endif
+
+    let l:version_lines = split(system(ale#Escape(l:executable) . ' -version'), '\n')
```
w0rp: The system call here we can't merge in, as `system()` calls are blocking. I or someone else will have to first implement support for chaining commands for LSP, which will be pretty complicated, but worth it.

w0rp:
Should be good once this is ready to go.

w0rp:
I opened an issue for supporting `command_chain` for LSP linters. See #2128.

ale_linters/java/eclipselsp.vim
nerdrew: If you just need the version to determine if the extra arguments are required, can you just make a config value "using java > 1.8"?

hsanson: Adding a configuration option for this can work in the short term but would not be a proper solution. I myself keep jumping between Java 8, 9, and 11 so I would have to update that configuration when switching projects.

On the other hand this branch works and I do my best to keep it updated with origin/master so is updated. Feel free to use this branch as I do myself. The only drawback is that you may experience some lockup of vim when the system() method is called but if your machine is fast enough it should not be noticeable at all.

w0rp:
@hsanson I finished work on #2132 last night, so now you can replace the `system()` call here with `ale#command#Run`. Have a look at the documentation for the function and existing uses in the codebase. You can effectively run the command in the background and receive the results in a callback, which can return the command to run for starting the language server.

w0rp:
You might also want to try using the internal `ale#semver#RunWithVersionCheck` function, which uses `ale#command#Run` to parse semver versions from results and then cache the results so we only need to run the version check roughly once per executable.

hsanson:
@w0rp first let me say, great work. I did not think this refactoring could be done so fast.

Have modified the Eclipse LSP to use the new async callback methods following as example the flake8 linter.  Also with caching of versions similar to ale#semver#RunWithVersionCheck.

But I have one test failing it travis CI that I am unable to pinpoint. I am sure I have listed the new LSP support everywhere it should (ale-supported-languages-and-tools.txt, ale.txt, and supported-tools.md).

Also I don't see the LSP being invoked in ALEInfo as before. I can see version checking is being executed but no LSP. 

And since a while back I am unable to get auto-completion working with any LSP, including this one. When time allows it, please check this MR and let me know if you find what I am doing wrong.

w0rp:
The test is failing because something isn't sorted correctly in the documentation.

I don't know why the command isn't being run on your machine. I'll have to try this linter myself on my own machine, at some point.

hsanson:
Fixed the sorting issue, a note in the ale-developer documentation about this new requirement would be nice. 

Using the ch_logfile trick in vim it was easy to find the issue with the LSP server:

```
  9.856030 on 0: Freeing job
  9.856536 RECV on 1: 'Error: Unable to access jarfile /home/ryujin/Apps/eclipse.jdt.ls/org.eclipse.jdt.ls.product/target/repository/plugins/org.eclipse.equinox.launcher_1.5.200.v20180922-1751.jar
'
```

The problem is that every time the LSP server is updated the jar file needed to start it changes name. It has the version and build time as part of the name (e.g. _1.5.200.v20180922-1751.jar).

I am unable to find a way in vim to search for a file using patterns that does not require using external tools like find. Is there a best practice on how to do this in ALE?

Also in ALEInfo there are no details of the LSP errors. Without the ch_logfile in vim is not possible to know if there are any errors. In neovim this is what I see:

```
  Command History:
(executable check - success) java
(finished - exit code 0) ['/bin/bash', '-c', '''java'' -version']
<<<NO OUTPUT RETURNED>>>
```

Is there a way I can make any errors reported by LSP appear as part of ALEInfo output?

w0rp:
See #2137 for adding LSP debugging information.

test/command_callback/test_eclipselsp_command_callback.vader
```diff
@@ -0,0 +1,87 @@
...
+  call ale#assert#SetUpLinterTest('java', 'eclipselsp')
+  let b:ale_java_eclipselsp_path = '/home/user/eclipse.dst.ls'
+  call ale#test#SetFilename('dummy.java')
+
+  let b:cfg = '/home/user/eclipse.dst.ls/org.eclipse.jdt.ls.product/target/repository/config_linux'
+
+  if has('win32')
+    let b:cfg = '\home\user\eclipse.dst.ls\org.eclipse.jdt.ls.product\target\repository\config_win',
+  elseif has('macunix')
+    let b:cfg = '/home/user/eclipse.dst.ls/org.eclipse.jdt.ls.product/target/repository/config_mac',
```
w0rp: The commas here break the tests on Windows and Mac. If you remove those, AppVeyor should pass.

hsanson: Fixed.

w0rp:
Cheers! :beers: Thank you for working on this, waiting for a few months for me to get the uniform asynchronous API done, and then coming back and finishing this.

JulioJu:
@hsanson  maybe in the ALE documentation, it could be cool to explain shortly how to customize compiler warning preferences ?  We should add customization under `root_project/.settings/org.eclipse.jdt.core.prefs`  with values presented at https://help.eclipse.org/neon/topic/org.eclipse.jdt.doc.isv/reference/api/org/eclipse/jdt/core/JavaCore.html ?

What do you think about that ?

hsanson:
@JulioJu sounds interesting but I have not used this feature and cannot see it documented anywhere on the eclipe lsp repo.  Is this a known feature?  I do not want to document something that is internal and may change in the future.

JulioJu:
@hsanson  you could check https://github.com/eclipse/eclipse.jdt.ls/issues/581

hsanson:
From the issues linked above I see:

> We do not have a proper way of defining those yet. unofficially though defining them on an eclipse style .settings/org.eclipse.jdt.core.prefs should work for now.

Once eclipse.jdt.ls has official way to add preferences that is documented I will be more than happy to add a note with a link to the official documentation explaining how to set them.

Of course anyone is free to modify the documentation and submit a MR for review.

PR #2122, opened 2018-12-05
Add support for Pod based SwiftLint

gfontenot:
It's common to add SwiftLint as a CocoaPod dependency, instead of as a global
binary. In this case we should use that version of SwiftLint before looking
for any others. Note that I'm also adding support for SwiftLint in ReactNative
projects here as well, where the Pods directory would be nested inside an ios
directory.

### DIFF
--- a/ale_linters/swift/swiftlint.vim
+++ b/ale_linters/swift/swiftlint.vim
@@ -5,19 +5,19 @@ call ale#Set('swift_swiftlint_executable', 'swiftlint')
 call ale#Set('swift_swiftlint_use_global', get(g:, 'ale_use_global_executables', 0))
 
 function! ale_linters#swift#swiftlint#GetExecutable(buffer) abort
-  return ale#node#FindExecutable(a:buffer, 'swift_swiftlint', [
-  \   'Pods/SwiftLint/swiftlint',
-  \   'ios/Pods/SwiftLint/swiftlint',
-  \   'swiftlint',
-  \])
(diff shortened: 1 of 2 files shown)


ale_linters/swift/swiftlint.vim
```diff
@@ -1,6 +1,24 @@
-" Author: David Mohundro <david@mohundro.com>
+" Author: David Mohundro <david@mohundro.com>, Gordon Fontenot <gordon@fonten.io>
 " Description: swiftlint for swift files
 
+call ale#Set('swift_swiftlint_executable', 'swiftlint')
+call ale#Set('swift_swiftlint_use_global', get(g:, 'ale_use_global_executables', 0))
+
+function! ale_linters#swift#swiftlint#GetExecutable(buffer) abort
+  return ale#node#FindExecutable(a:buffer, 'swift_swiftlint', [
```
gfontenot: Is it worth pulling this out of `ale#node` and into `ale#utils`? It's being used in a bunch of different linters that have nothing to do with node, and there doesn't _seem_ to be anything about it that's specific to node itself. There's also `ale#python#FindExecutable`, which might be able to be removed and replaced with this common util function.

neersighted: I believe both of those functions used to be specific to virtual environments and node_modules before being generalized. I'd be happy to take a follow-up PR factoring them out into a more generic, single function.

PR #2123, opened 2018-12-06
`arguments` needs to be handled for `compile_commands.json` in addition to `command`

MoZhonghua:
By [standard](https://clang.llvm.org/docs/JSONCompilationDatabase.html):

> command: The compile command executed. After JSON unescaping, this must be a valid command to rerun the exact compilation step for the translation unit in the environment the build system uses. Parameters use shell quoting and shell escaping of quotes, with ‘"’ and ‘\’ being the only special characters. Shell expansion is not supported.
arguments: The compile command executed as list of strings. __Either arguments or command is required__.

This PR will check the existence of 'arguments' and 'command' before using them. If there is no 'command' but with 'arguments', call join(item.arguments, ' ') to generate command, then follow the old code path.


autoload/ale/c.vim
```diff
@@ -174,11 +174,21 @@ function! s:GetListFromCompileCommandsFile(compile_commands_file) abort
     return []
 endfunction
 
+function! ale#c#GetCompileCommand(json_item) abort
+    if has_key(a:json_item, "command")
+        return a:json_item.command
```
RyanSquared: I think that further down the call chain, there's likely a call to `ale#Escape()` which quotes the commands used. this should be used here, as well as for each argument containing spaces in the below section - meaning manually looping instead of using the `join()` function. Additionally, somewhere, `ale#Escape()` needs to be removed.

MoZhonghua: You're right, we need to handle special characters more carefully.

[bear](https://github.com/rizsotto/Bear) will generate compile_commands.json like this:

```json
   "arguments": [
            "/usr/bin/cc",
            "-c",
            "-I/home/mozhonghua/playground/zimg/build space",
```
that is, no double quotes surrounding paths with space.

I don't have more time to fix this, maybe someone else would put more efforts on this issue?

RyanSquared: i'll look into fixing it up when i have the time, thanks for raising awareness of this issue in the first place

w0rp:
Write tests please.

autoload/ale/c.vim
RyanSquared: It looks like in the case of relative paths for the project (see line 72-80 of this file), they are escaped, but absolute paths are kept as-is. I'm not sure whether or not this is because they would already be escaped otherwise. @w0rp you might know more about this than I do.

PR #2131, opened 2018-12-09
Parse more C/C++ compiler options

0mco:
According gcc manual, in addition to '-I/path/to/include', there are many other flags can be used to add header search directory. Qt with cmake use '-isystem' to include Qt headers, if not adding these flags to cflag_line, ale cannot find Qt headers. This PR adds these flags to cflag_line.

### DIFF
--- a/autoload/ale/c.vim
+++ b/autoload/ale/c.vim
@@ -79,51 +79,43 @@ function! ale#c#AreSpecialCharsBalanced(option) abort
 endfunction
 
 function! ale#c#ParseCFlags(path_prefix, cflag_line) abort
-    let l:cflags_list = []
-    let l:previous_options = ''
-
-    let l:split_lines = split(a:cflag_line, ' ')
+    let l:split_lines = split(a:cflag_line)
     let l:option_index = 0
 
     while l:option_index < len(l:split_lines)
-        let l:option = l:previous_options . l:split_lines[l:option_index]
-        let l:option_index = l:option_index + 1
+        let l:next_option_index = l:option_index + 1
 
-        " Check if cflag contained an unmatched special character and should not have been splitted
-        if ale#c#AreSpecialCharsBalanced(l:option) == 0 && l:option_index < len(l:split_lines)
-            let l:previous_options = l:option . ' '
-            continue
-        endif
+        " Join space-separated option
+        while l:next_option_index < len(l:split_lines) &&
+            \ stridx(l:split_lines[l:next_option_index], '-') != 0
+            let l:next_option_index += 1
+        endwhile
 
-        " Check if there was spaces after -D/-I and the flag should not have been splitted
-        if l:option is# '-D' || l:option is# '-I'
-            let l:previous_options = l:option
-            continue
-        endif
+        let l:option = join(l:split_lines[l:option_index : l:next_option_index-1], ' ')
+        call remove(l:split_lines, l:option_index, l:next_option_index-1)
+        call insert(l:split_lines, l:option, l:option_index)
 
-        let l:previous_options = ''
-
-
-        " Fix relative paths if needed
-        if stridx(l:option, '-I') >= 0 &&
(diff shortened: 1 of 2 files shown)


0mco:
Well, adding these flags to c linter solves the header issue, but it also exposes another problem:
After adding this, I got error:
```
In file included from /home/***/Applications/Qt/5.10.1/gcc_64/include/QtGui/qtguiglobal.h:43:
/home/***/Applications/Qt/5.10.1/gcc_64/include/QtCore/qglobal.h:1162:4: error: "You must build your code with position independent code if Qt was built with -reduce-relocations. "         "Compile your code with -fPIC (-fPIE is not enough)."
#  error "You must build your code with position independent code if Qt was built with -reduce-relocations. "\
   ^
```
The` -fPIC `flag is in compile_commands.json, but it's not passed into c linter. Passing `-fPIC` to c linter solves this issue, but 
Why ale only pass `-D` and `-I` flags to c linter? Is it for compatibility consideration? Since some flags may not be recognized by c linter?

0mco:
Passing all flags found to c linter seems necessary, with the potential issue of 'unrecognized option'.

0mco:
Can you explain why we shouldn't pass `-o` option?
I've tested passing all flags to c linter, it seems working pretty well, even with deliberately wrong option like `-asdfasdf`.

w0rp:
Yes. Say if a C or C++ complier passes an option like `-o` for generating build files, then running the compiler with that option will cause linting a file to leave binary files lying around in your project. If ALE did that, people would be understandably annoyed. `-o` itself might not cause that problem, but some other flags might cause that problem.

The problem is just trying to figure out which flags shouldn't be included, and how to exclude those flags. The way you implement that is by adding a large series of tests which capture all of the cases, such that there are no important cases which are not tested, and everything is covered.

0mco:
@w0rp I found a summary of `GCC` [options](https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html#Option-Summary). I looked at `Options Controlling the Kind of Output` section, found there is actually only one option for controlling output, namely `-o`.
I've also tested with `-o` option, which made me sure that `-o` option is ignored automatically by the linter. This is great, since we don't have to distinguish which options should be passed into linter and which shouldn't.
Now that generating build files is not a problem at all, passing all options seems doable?

w0rp:
Yeah, if it works, try it. All that matters is that relatively accurate output is produced, and we don't generate binary files or execute code. Being able to delete the messy code for parsing the flags would be nice bonus.

0mco:
@w0rp Passing all options to `clang` seems to work pretty well, but `gcc` is bad at handling conflict options. In fact, even `-c` and `-o` will cause conflicts:
```
<<<OUTPUT STARTS>>>
gcc: fatal error: cannot specify -o with -c, -S or -E with multiple files
compilation terminated.
<<<OUTPUT ENDS>>>
```
So I passed all but `-o` option to linter. If this is ok, I'll add tests for this later.

0mco:
There may be other conflicting options in `gcc`, which will only be addressed when we know that since I cant find a list of conflicting options for `gcc`.

w0rp:
Sure, go ahead and update the tests.

w0rp:
Okay, I merged a PR for problems with minuses from another guy. Update the code and the tests here, so everything passes now.

w0rp:
The tests failed, so have a look at those.

w0rp:
I just merged a PR for caching, so you'll need to rebase your commits here. It looks like you used `git merge` before. You might want to rewrite all of your changes into one commit.

Fix the tests here.

0mco:
I'm preparing my final exams these days. I'll fix it later.

w0rp:
Sure thing. Thank you for trying to make this better!

0mco:
@w0rp I've updated.

w0rp:
Cheers! :beers:

PR #2133, opened 2018-12-09
Adds standardrb to ale

searls:
See: https://github.com/testdouble/standard ( https://github.com/testdouble/standard/issues/43 ) 

StandardRB is to RuboCop what StandardJS is to ESLint. This commit 
naively copies the RuboCop linter and fixer to point at the standardrb
executable. Any other adjustments are very minor (the only I can think 
of is that standardrb takes a `--fix` option instead of 
`--auto-correct`).

This raises a confusing point to me as both developer and a user: since
ale enables all linters by default, won't this run both RuboCop and 
StandardRB (the results of which will almost always be in conflict with
one another)? How does ale already solve for this for the similar case
of StandardJS and ESLint?

In practice, it seems to "just work" and only run standard but I'm completely confused as to how, given that both rubocop and standardrb are present in the bundle. If someone could advise me on how this works I'd be happy to adjust things


neersighted:
The tests should check the handler (error parser) against real (but hardcoded) output, and test that the command line parameters are built correctly. Check the existing tests (say, eslint or rubocop) for examples.

As far as running both and not getting any lints from rubocop, I would need to reproduce it myself to know what is going on.

searls:
Sorry that I'm struggling to add the tests you asked for:

* I added a test for standardrb's error callback handler [here](https://github.com/searls/ale/blob/master/test/command_callback/test_standardrb_command_callback.vader)

* I added a test for standardrb's fixer callback [here](https://github.com/searls/ale/blob/master/test/fixers/test_standardrb_fixer_callback.vader)

* Based on your advice, I looked for handler tests and so started with a duplicate of [test_rubocop_handler.vader](https://github.com/w0rp/ale/blob/master/test/handler/test_rubocop_handler.vader), but it wound up being virtually identical because my standardrb linter [is configured to invoke the RuboCop handler](https://github.com/searls/ale/blob/master/ale_linters/ruby/standardrb.vim#L22) directly, so I don't think tests in which we invoke the handler will add very much.

As for having both rubocop and standard installed, it seems to work fine in practice for me, but either way I am going to follow the example of standardjs (which is so closely analogous that it may as well be precedent) and just recommend standardrb users to explicitly specify what tools they want to run, as [documented here](
https://standardjs.com/#vim)

searls:
There was a test failure caused I think by an overly aggressive gitignore rule added in April here 7cf3ddf6c

It is ignoring all files that start with a `.`, recursively. For this reason my fixture file `test/command_callback/ruby_paths/with_config/.standard.yml` didn't get picked up by git and the build failed. I worked around this with `git add --force`, but I'm doubtful this was intentional. Maybe the gitignore rule should have been `/.*`?

ale_linters/ruby/standardrb.vim
```diff
@@ -0,0 +1,23 @@
...
+    \   . ale#Var(a:buffer, 'ruby_standardrb_options')
+    \   . ' --stdin ' . ale#Escape(expand('#' . a:buffer . ':p'))
+endfunction
+
+" standardrb is based on RuboCop so the callback is the same
+call ale#linter#Define('ruby', {
+\   'name': 'standardrb',
+\   'executable_callback': ale#VarFunc('ruby_standardrb_executable'),
+\   'command_callback': 'ale_linters#ruby#standardrb#GetCommand',
+\   'callback': 'ale_linters#ruby#rubocop#Handle',
```
w0rp: Could you move the `Handle` function into a location like `autoload/ale/handlers/ruby.vim` with some other name like `HandleRubocopOutput` and use it both files? I try to avoid having one linter file use a function from another linter file.

w0rp:
The `.gitignore` file intentionally ignores all hidden files at every level, and you're right to use `git add -f` to add the file. I prefer doing that to people accidentally committing hidden files.

w0rp:
I added a comment to `.gitignore` just now, so people will understand that's the intention.

ale_linters/ruby/standardrb.vim
searls: You got it. Just pushed bd79fadd3aeb78267c0a4304caf2a05d8544e90b

PR #2134, opened 2018-12-09
javac linter: fix handling of error messages containing ':' character

oaue:
Javac error messages containing a ':' character were reported as error by the javac linter.

### DIFF
--- a/ale_linters/java/javac.vim
+++ b/ale_linters/java/javac.vim
@@ -92,7 +92,7 @@ function! ale_linters#java#javac#Handle(buffer, lines) abort
     " Main.java:13: warning: [deprecation] donaught() in Testclass has been deprecated
     " Main.java:16: error: ';' expected
     let l:directory = expand('#' . a:buffer . ':p:h')
-    let l:pattern = '\v^(.*):(\d+): (.+):(.+)$'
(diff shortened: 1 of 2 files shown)


oaue:
Sorry, I wanted to write: Javac error messages containing a ':' character were reported as **warnings** by the javac linter.

The problem fixed it the pattern is that '+' is greedy. So, with the regexp `\v^(.*):(\d+): (.+):(.+)$`, the third () will match everything until the last ':'.

And l:match[3] is matched against 'error' to test if it's an error:

> 'type': l:match[3] is# 'error' ? 'E' : 'W',

It won't work for error messages containing ':' characters like "error: incompatible types: Bar cannot be converted to Foo", where l:match[3] will contain "error: incompatible types", instead of "error".

But it will work as expected if we replace '+' with it's non-greedy equivalent '{-1,}'  (:help non-greedy).

oaue:
I'm not sure if there is still something that block the merge of this request?

w0rp:
No sorry, I forgot about this and I thought it was already merged. I'll merge it now.

w0rp:
Cheers! :beers:

oaue:
Thanks!

Issue #368, opened 2017-02-28
chktex reports issues from the wrong files


w0rp:
When you run chktex on a file which imports other files, where those other files produce warnings/errors, those warnings or errors are reported in the original file incorrectly.

lgalke:
Same goes for lacheck.

w0rp:
ALE doesn't use `errorformat` and so on at all.

I don't use `chktex` myself. If anyone wants to have a go at fixing this, go for it. If not, I'll get to it eventually.

lgalke:
Alright I will give it a try for both `lacheck` and `chktex`.

Unfortunately I did not yet find the code which further processes the returned (by linter handler) list of dictonaries. 
```vim 
call add(l:output, {                                                                                                                                                   
      \   'bufnr': a:buffer,                                                                                                                                                 
      \   'lnum': l:match[1] + 0,                                                                                                                                            
      \   'col': 0,                                                                                                                                                          
      \   'text': l:match[2],                                                                                                                                                
      \   'type': 'W',                                                                                                                                                       
      \}) 
``` 
in `ale_linters/tex/lacheck.vim`

Is there support for a key:value pair for the filename already?

I would propose to not just ignore out-of-file errors, since tex's `\input{otherfile.tex}` calls act exactly like if the input was written in the current file.

w0rp:
>Is there support for a key:value pair for the filename already?

I'm not sure what you mean by this. All items returned from the `callback` function should be errors to display for the file being checked.

lgalke:
So the returned list is a list of dictionaries, and each dictionary contains keys like:
`bufnr`,`lnum`,`col`,`text`,`type`. What I meant is how this dictionary (for a single error) gets further processed for display in the loc list. And especially whether there exists some dedicated key in the dictionary for a filename?

w0rp:
Only the buffer number is used for setting errors. Each buffer in Vim has a number, and the buffer number is recorded when ALE starts checking a file. This buffer number is passed around everywhere, and then the buffer number is used to set the errors for a given buffer.

For fixing the bug for this GitHub issue, the way to go about it is to first check if the lines of output from `chktex` return the filenames the errors and warnings are for. If the command can either return the filename being checked, or something like `'<stdin>'`, etc., then either can be used to filter the matches down to only those for the file being checked.

I don't know too much about `chktex` myself.

lgalke:
[...]

PR #2136, opened 2018-12-10
Fix 368 - Lacheck reports errors from input{} files.

hsanson:
This PR adds additional check to lacheck linter to exclude any warnings
related to sourced files via latex \input{} command.

Closes: #368


hsanson:
Added some tests to lacheck linter. It actually had no tests at all.

I tried to get the absolute paths of the files reported by lacheck but was more difficult that expected. Once we change current paths lacheck will report all inputted files as "Cannot open".

w0rp:
Cheers! :beers:

PR #2143, opened 2018-12-13
Decouple completion engine internals

reaysawa:
#### I will fix documentation and tests after the API changes are discussed and reviewed

# Notions
- The completion engine and the Autocmd events should be decoupled.
- Users can override the behavior of what happens to the results.
- Allow for interaction with other plugins, like deoplete (mentioned on https://github.com/w0rp/ale/issues/1753).

# Changes
- The process for polling/queuing activation has been separated from the Omnifunc itself.
- The old `ale#completion#Show` method called at the end of the completion parsing chain has been repurposed as `ale#completion#QueryDone`, which calls
     - Overridable **ALEProcessCompletionResults** for processing the parsed candidates for completion.
     - Overridable **ALEAfterCompletionResults** for adding flexibility to what happens after the processing is finished.

# How to test
Follow the instructions inside of [minimal_autocompletion_vimrc](test/completion/vimrc/minimal_autocompletion_vimrc).

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -19,6 +19,7 @@ let g:ale_completion_max_suggestions = get(g:, 'ale_completion_max_suggestions',
 let s:timer_id = -1
 let s:timer_pos = []
 let s:last_done_pos = []
+let s:currently_queued_pos = []
 
 " CompletionItemKind values from the LSP protocol.
 let s:LSP_COMPLETION_TEXT_KIND = 1
@@ -184,40 +185,50 @@ function! ale#completion#RestoreCompletionOptions() abort
     endif
 endfunction
 
-function! ale#completion#OmniFunc(findstart, base) abort
-    if a:findstart
-        let [l:line, l:column] = getcurpos()[1:2]
-        let l:regex = s:GetFiletypeValue(s:omni_start_map, &filetype)
-        let l:up_to_column = getline(l:line)[: l:column - 2]
-        let l:match = matchstr(l:up_to_column, l:regex)
+function! ale#completion#FindCompletionStart() abort
+    let [l:line, l:column] = getcurpos()[1:2]
+    let l:regex = s:GetFiletypeValue(s:omni_start_map, &filetype)
+    let l:up_to_column = getline(l:line)[: l:column - 2]
+    let l:match = matchstr(l:up_to_column, l:regex)
 
-        let b:ale_completion_start = l:column - len(l:match) - 1
+    let b:ale_completion_start = l:column - len(l:match) - 1
 
-        return b:ale_completion_start
+    return b:ale_completion_start
+endfunction
+
+function! ale#completion#PollCompletionResults() abort
+    if ale#completion#Queue()
+        return []
     else
-        if ale#completion#Queue()
-            return []
-        else
-            return get(b:, 'ale_completion_result', [])
-        endif
+        return get(b:, 'ale_completion_result', [])
     endif
 endfunction
 
-function! ale#completion#Results() abort
+function! ale#completion#OmniFunc(findstart, base) abort
+    if a:findstart
+        return ale#completion#FindCompletionStart()
+    else
+        return ale#completion#PollCompletionResults()
+    endif
+endfunction
+
+function! ale#completion#QueryDone() abort
     let s:last_done_pos = getcurpos()[1:2]
 
-    call ale#completion#ProcessResults(get(b:, 'ale_completion_result', []))
+    let b:ale_completion_result = ALEProcessCompletionResults(get(b:, 'ale_completion_result', []))
 
-    call ale#completion#AfterResults()
+    call ALEAfterCompletionResults()
 endfunction
 
-function! ale#completion#ProcessResults(completion_result) abort
-    " Users may override this function to process the results however they
-    " see fit.
-    return a:completion_result
+function! ALEProcessCompletionResults(completion_results) abort
+    " Users may override this function to act on the results however they
+    " see fit. Needs to be not namespaced.
+    return a:completion_results
 endfunction
 
-function! ale#completion#AfterResults() abort
+function! ALEAfterCompletionResults() abort
+    " Users may override this function to act on the results however they
+    " see fit. Needs to be not namespaced.
     if !ale#util#Mode() is# 'i'
         return
     endif
@@ -430,7 +441,7 @@ function! ale#completion#HandleTSServerResponse(conn_id, response) abort
     elseif l:command is# 'completionEntryDetails'
         let b:ale_completion_result = ale#completion#ParseTSServerCompletionEntryDetails(a:response)
 
-        call ale#completion#Results()
+        call ale#completion#QueryDone()
     endif
 endfunction
 
@@ -442,7 +453,7 @@ function! ale#completion#HandleLSPResponse(conn_id, response) abort
 
     let b:ale_completion_result = ale#completion#ParseLSPCompletions(a:response)
 
-    call ale#completion#Results()
+    call ale#completion#QueryDone()
 endfunction
 
 function! s:OnReady(linter, lsp_details, ...) abort
@@ -563,11 +574,14 @@ endfunction
 function! ale#completion#Queue() abort
     let s:timer_pos = getcurpos()[1:2]
 
-    if s:timer_pos == s:last_done_pos
+    if s:timer_pos == s:last_done_pos || s:timer_pos == s:currently_queued_pos
         " Do not ask for completions if the cursor rests on the position we
-        " last completed on.
+        " last completed on, or if the position is the one currently being
+        " queried by the engine.
         return 0
     endif
+
+    let s:currently_queued_pos = s:timer_pos
 
     " If we changed the text again while we're still waiting for a response,
     " then invalidate the requests before the timer ticks again.
@@ -582,12 +596,17 @@ function! ale#completion#Queue() abort
     return 1
 endfunction
 
-function! ale#completion#Done() abort
+function! ale#completion#Finished() abort
     silent! pclose
 
     call ale#completion#RestoreCompletionOptions()
 
     let s:last_done_pos = getcurpos()[1:2]
+endfunction
+
+function! ale#completion#ClearBufferResults() abort
+    let b:ale_completion_result = []
+    let b:ale_completion_info = {}
 endfunction
 
 function! s:Setup(enabled) abort
@@ -596,7 +615,7 @@ function! s:Setup(enabled) abort
 
         if a:enabled
             autocmd TextChangedI * call ale#completion#Queue()
-            autocmd CompleteDone * call ale#completion#Done()
+            autocmd CompleteDone * call ale#completion#Finished()
         endif
     augroup END
 
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -1,6 +1,8 @@
 " Author: w0rp <devw0rp@gmail.com>
 " Description: Completion support for LSP linters
 
+let g:ale_autocompletion_enabled = 0
+
 " The omnicompletion menu is shown through a special Plug mapping which is
 " only valid in Insert mode. This way, feedkeys() won't send these keys if you
 " quit Insert mode quickly enough.
@@ -193,33 +195,19 @@ function! ale#completion#OmniFunc(findstart, base) abort
 
         return l:column - len(l:match) - 1
     else
-        " Parse a new response if there is one.
-        if exists('b:ale_completion_response')
-        \&& exists('b:ale_completion_parser')
-            let l:response = b:ale_completion_response
-            let l:parser = b:ale_completion_parser
-
-            unlet b:ale_completion_response
-            unlet b:ale_completion_parser
-
-            let b:ale_completion_result = function(l:parser)(l:response)
+        if g:ale_autocompletion_enabled
+            return get(b:, 'ale_completion_result', [])
+        else
+            return ale#completion#ProcessCompletions()
         endif
-
-        call s:ReplaceCompletionOptions()
-
-        return get(b:, 'ale_completion_result', [])
     endif
 endfunction
 
-function! ale#completion#Show(response, completion_parser) abort
-    if ale#util#Mode() isnot# 'i'
+function! ale#completion#Show() abort
+    if !g:ale_autocompletion_enabled || !g:ale#util#Mode() isnot# 'i'
         return
     endif
 
-    " Set the list in the buffer, temporarily replace omnifunc with our
-    " function, and then start omni-completion.
-    let b:ale_completion_response = a:response
-    let b:ale_completion_parser = a:completion_parser
     " Replace completion options shortly before opening the menu.
     call s:ReplaceCompletionOptions()
 
@@ -228,6 +216,11 @@ endfunction
 
 function! s:CompletionStillValid(request_id) abort
     let [l:line, l:column] = getcurpos()[1:2]
+
+    return has_key(b:, 'ale_completion_info')
+    \&& b:ale_completion_info.request_id == a:request_id
+    \&& b:ale_completion_info.line == l:line
+    \&& b:ale_completion_info.column == l:column
 
     return ale#util#Mode() is# 'i'
     \&& has_key(b:, 'ale_completion_info')
@@ -418,10 +411,9 @@ function! ale#completion#HandleTSServerResponse(conn_id, response) abort
             \)
         endif
     elseif l:command is# 'completionEntryDetails'
-        call ale#completion#Show(
-        \   a:response,
-        \   'ale#completion#ParseTSServerCompletionEntryDetails',
-        \)
+        let b:ale_completion_result = ale#completion#ParseTSServerCompletionEntryDetails(a:response)
+
+        call ale#completion#Show()
     endif
 endfunction
 
@@ -431,10 +423,9 @@ function! ale#completion#HandleLSPResponse(conn_id, response) abort
         return
     endif
 
-    call ale#completion#Show(
-    \   a:response,
-    \   'ale#completion#ParseLSPCompletions',
-    \)
+    let b:ale_completion_result = ale#completion#ParseLSPCompletions(a:response)
+
+    call ale#completion#Show()
 endfunction
 
 function! s:OnReady(linter, lsp_details, ...) abort
@@ -489,6 +480,12 @@ function! s:OnReady(linter, lsp_details, ...) abort
     endif
 endfunction
 
+function! ale#completion#ProcessCompletions() abort
+    call ale#completion#GetCompletions()
+
+    return get(b:, 'ale_completion_result', [])
+endfunction
+
 function! s:GetLSPCompletions(linter) abort
     let l:buffer = bufnr('')
     let l:lsp_details = ale#lsp_linter#StartLSP(l:buffer, a:linter)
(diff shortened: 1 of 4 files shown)


autoload/ale/completion.vim
```diff
@@ -183,52 +185,71 @@ function! ale#completion#RestoreCompletionOptions() abort
...
+    endif
 endfunction
 
 function! s:CompletionStillValid(request_id) abort
     let [l:line, l:column] = getcurpos()[1:2]
 
+    return has_key(b:, 'ale_completion_info')
+    \&& b:ale_completion_info.request_id == a:request_id
+    \&& b:ale_completion_info.line == l:line
+    \&& b:ale_completion_info.column == l:column
```
w0rp: You need to remove the `return` statement below. Something will have to be done about checking that the mode is `'i'` still, for insert mode. Make sure we still check that the mode is `'i'` (Using the Mode wrapper function so we can test it.) before using `feedkeys()` with `<C-x><C-o>`.

autoload/ale/completion.vim
```diff
@@ -183,52 +185,71 @@ function! ale#completion#RestoreCompletionOptions() abort
...
-    endif
+function! ALEProcessCompletionResults(completion_results) abort
+    " Users may override this function to act on the results however they
+    " see fit. Needs to be not namespaced.
+    return a:completion_results
 endfunction
 
-function! ale#completion#Show(response, completion_parser) abort
-    if ale#util#Mode() isnot# 'i'
+function! ALEAfterCompletionResults() abort
```
w0rp: Doesn't deoplete accept autoload functions? Why not?

w0rp: Does deoplete need to use this function at all? This function opens up the completion menu itself, so I'd imagine it's only useful when you aren't using deoplete.

autoload/ale/completion.vim
```diff
@@ -183,52 +185,71 @@ function! ale#completion#RestoreCompletionOptions() abort
...
+endfunction
+
+function! ale#completion#PollCompletionResults() abort
+    if ale#completion#Queue()
+        return []
+    else
+        return get(b:, 'ale_completion_result', [])
+    endif
+endfunction
+
```
w0rp: I think you can pretty much undo the changes here. The code is basically the same, and the changes don't really improve on what was there. The previous behavior of looking at the line and column as requested instead of the current cursor position may be more appropriate.

autoload/ale/completion.vim
```diff
@@ -17,7 +17,9 @@ let g:ale_completion_excluded_words = get(g:, 'ale_completion_excluded_words', [
 let g:ale_completion_max_suggestions = get(g:, 'ale_completion_max_suggestions', 50)
 
 let s:timer_id = -1
+let s:timer_pos = []
 let s:last_done_pos = []
+let s:currently_queued_pos = []
```
w0rp: As mentioned before, you can remove this.

autoload/ale/completion.vim
reaysawa: When a result is parsed into menu candidates, after that, the caller needs a way to know that the results are ready to be used on their end. For deoplete I've overrode this function with:

```
function! ALEAfterCompletionResults() abort
    call deoplete#send_event('CompleteDone', 'ale')
endfunction
```

For other completion plugins, people can override this "call be when it's done callback" with something else, hence the name _**After**_CompletionResults.

autoload/ale/completion.vim
```diff
@@ -557,18 +572,17 @@ function! ale#completion#StopTimer() abort
 endfunction
 
 function! ale#completion#Queue() abort
-    if !g:ale_completion_enabled
-        return
-    endif
```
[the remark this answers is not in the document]
reaysawa: Shouldn't the person know that `ale#completion#Disable()` is the right way to do this? TextChangedI is a pretty costly autocmd to trigger if it's not being used.

autoload/ale/completion.vim
```diff
@@ -557,18 +572,17 @@ function! ale#completion#StopTimer() abort
...
+    if s:timer_pos == s:last_done_pos || s:timer_pos == s:currently_queued_pos
         " Do not ask for completions if the cursor rests on the position we
-        " last completed on.
-        return
+        " last completed on, or if the position is the one currently being
+        " queried by the engine.
+        return 0
     endif
 
+    let s:currently_queued_pos = s:timer_pos
```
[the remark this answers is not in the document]
reaysawa: 1. Outside source asks for completion and it's already there (**s:last_done_pos**)
2. Outside source asks for completion and it's not done yet, but it has been dispatched for this position so there's no need to call this function again (**s:currently_queued_pos**)

How do I signal situation two without this?

autoload/ale/completion.vim
reaysawa: It is a mistake that there are two return statements,

but the name `CompletionStillValid` speaks to me that it checks if completion is still valid, not if Vim is still in insert mode or if ALE is supposed to show this in the screen. The outside source shouldn't be bound by this behavior.

autoload/ale/completion.vim
reaysawa: Is this referring to the whole block of changes above? `ale#completion#PollCompletionResults `and `ale#completion#FindCompletionStart`

Those were exactly what was before inside of ale#completion#OmniFunc, only turned into functions because external sources don't use the flow of omnifunctions to command their behavior.

reaysawa:
@w0rp as with #1889 it seems like we have a lot of different opinions on the internal design of this file. You are able to push commits to this branch too, so can you redesign those changes when you have the time? It will be clear to me if I can adjust your changes to my deoplete plugin, or if it needs some tweaking.

I've pushed the plugin code (it's only this 70 line file) to [deoplete_ale.py](https://github.com/reaysawa/deoplete-ale/blob/master/rplugin/python3/deoplete/sources/deoplete_ale.py) for you to take a look.

It stops the ALE source when 'completion_done' is set (line 32-33) from a manual event dispatched with

```
function! ALEAfterCompletionResults() abort
    call deoplete#send_event('CompleteDone', 'ale')
endfunction
```

autoload/ale/completion.vim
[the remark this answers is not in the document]
w0rp: They should, but I'd leave the check in here anyway. It doesn't cost much, and it will be useful later if we add `b:ale_completion_enabled`.

autoload/ale/completion.vim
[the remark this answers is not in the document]
w0rp: You should remove the `s:currently_queued_pos` disabled check, and let the code run through to stopping and restarting the timer. The timer should be reset from when the last request was made, say if the delay is 200ms, and you make two requests at 0ms and 100ms, then the timer function should be called at 300ms.

autoload/ale/completion.vim
w0rp: Yes. Remove the `return` statement below this one, as I said.

autoload/ale/completion.vim
w0rp: Replacing an existing function like that is bad design, and it doesn't make sense. Instead of doing that, let's add the deoplete source to ALE, and we don't need to call the function for opening the omnicomplete menu when ALE completion integration with LSP is being used as a completion source.

w0rp:
See if you can implement deoplete integration without defining any global functions at all.

You can push the changes, and I can review them later.

autoload/ale/completion.vim
w0rp: Cool, thank you for clarifying. That makes sense.

Remember to add tests for these function names.

Let's add the deoplete source file to ALE's repo itself.

reaysawa:
Finally got some room back to act on this. I'll try to follow up on your inquiries but let me explain why it's set up this way and also comment on your comments.

---

[Python code for reference](https://github.com/reaysawa/deoplete-ale/blob/master/rplugin/python3/deoplete/sources/deoplete_ale.py)

Line 57 and forward of the Python code reads like this

        context['initial_position'] = self.get_completion_position()

        if context['initial_position'] < 0:
            return []
        else:
            context['is_async'] = True


When the position is invalid for the filetype we return right away without setting the `['is_async'] = True` flag in the context object. What happens is that  **at the end of gather_candidates**, if this flag is turned off, it won't try to get results from the source again. The idea is to let deoplete keep calling this function until we trigger the manual event to stop it at line 33.

By default it polls every source with _is_async_ every 1 second until it sets the flag to false, regardless of it having results or not. This is useful as compared to stopping/restarting timers because you could use those "heartbeats" to process LSP results sparsely and keep adding to deoplete, similar to what you said about

>I had set it up so parsing of results was deferred until just before the menu is shown, to spread out the processing between frames a bit, but I think it should be fine to just parse the results right away.

---

On line 53 a check is done for

    if 'initial_position' not in context:

Checking if this source has just been initialized.

I thought that instead of calling repeatedly _ale#completion#PollCompletionResults_ and returning its result until the manual trigger is received (relies on _s:currently_queued_pos_), I could also use this block of code to call Queue only once, and then not call ALE again until we get the manual trigger back. The problem with the latter is the lack of coordination; with the current approach, if the user types again halfway through the function, we incidentally call Queue again which will invalidate the results and keep the asynchronicity running smoothly. Otherwise we'd have to go through, repeatedly

- Deoplete activation delay
- Deoplete async setup (heartbeats every 1 second)
- ALE restarting timers
- ALE completion delay

Plus other potential "unknown problems" of inconsistency between the menu candidates.

Furthermore I don't agree with the notion of "polluting the global namespace" because those functions give more flexibility to use the engine without having to be bound by the behavior of what ALE considers as autocompletion. For instance, you can't call the omnifunction manually as of now and this could more broadly remedy that to some extent.

---

> Replacing an existing function like that is bad design, and it doesn't make sense. Instead of doing that, let's add the deoplete source to ALE, and we don't need to call the function for opening the omnicomplete menu when ALE completion integration with LSP is being used as a completion source.

ALEAfterCompletionResults is opening as an avenue for users (in general) to customize what happens after completion results, it could be anything related to any other plugin like MUComplete [which does not support asynchronous calls](https://github.com/lifepillar/vim-mucomplete/issues/70#issuecomment-375415566).

---

> Let's add the deoplete source file to ALE's repo itself.

Doable as shown by [OmniSharp-vim](https://github.com/OmniSharp/omnisharp-vim/tree/master/rplugin/python3/deoplete/sources), but the path will have to be created at the root of the repo. I agree that it is way better, so that the ALE community can contribute directly to it.

reaysawa:
Like I said I will try to follow on your proposals and see if it clicks with this setup, but it already syncs up decently well after the tweaks I have made. This is how it is currently with

    let g:ale_completion_enabled = 0
    let g:ale_completion_delay = 200
    call deoplete#custom#source('ale', 'min_pattern_length', 0)

[Demo](https://my.mixtape.moe/wfnaaw.mp4)

It can be very much improved and fine grained as far as controls over results go using some deoplete filters or a with a good strategy for clearing lingering results from previous quick completions, if any... But it is smooth as of now. I can control how "spammy" it is directly through

    let g:ale_completion_delay = 200

Changing the behavior could make it more clunky but I don't know that as of now, so I'll use it some more in the next few days and see if I can come up with something more conforming to your requests.

w0rp:
Defining a global function like that to be replaced is bad design. ALE's function could be loaded after some user-supplied function, so it's difficult to replace. We can't support users replacing particular functions, that will make maintenance very difficult. It's much better to offer a documented setting, like a variable naming a function to call, or an autocmd event.

Could you add the Python part of this to the pull request? I'll look at getting all of this to work with tests for everything, including the Python code.

reaysawa:
@w0rp here you go, I have added the source in respect to the filepath needed by deoplete to load and recognize the plugin. I have also not touched the code in order to not disrupt your previous review.

If you have more questions about deoplete, though I don't know much, I can venture deeper into the source. I appreciate the time you invested in the discussion and the work done in the whole of ALE. : D

w0rp:
Thanks! I'll have a look when I can. Christmas holiday family time is essentially over now.

andreypopp:
Would it be possible to provide synchronous completion interface via omnifunction? That would greatly benefit those not using large completion frameworks but just plain omnicompletion.

reaysawa:
@andreypopp that was a smaller goal of this PR. If you look at [completion.vim#L149](https://github.com/w0rp/ale/blob/master/autoload/ale/completion.vim#L149), you'll see that ALEdoes some trickery to keep the popup window opened without interfering with your native completion options, which can be a good thing IMO.

Aside from that, multiple `if !g:ale_completion_enabled` checks scattered across the file made it so you couldn't have **completion** without **autocompletion** by default.

Due to the replacement trick I mentioned above, which is embedded directly into the omnifunction, it might not work. Try to manually disable the autocmd events (ALECompletionGroup) after `ale#completion#Enable` and see if it clicks for you.

w0rp:
@andreypopp It would still be asynchronous, but yes, that is possible, and that's a separate problem. This PR is just about making it so ALE can be used as a source for deoplete or something else.

reaysawa:
Kinda related: does TSServer have any issues on providing completions around the dot? I've introduced some variables to debug how the prefix was being sent after some changes and it seemed to be identified correctly (e.g. Array. , Object. , etc...) and the word seemed to be captured correctly, but it still didn't work 100% of the time. Below recording is from the latest ALE master without any adulteration.

https://asciinema.org/a/sV3tAmgg07MrcX7OWeHeIiO3Q

I can reproduce it kinda easily there by going back-and-forth on the dot after showing the suggestions for the position, but I've felt like this also happened "at random" sometimes, where I had to type the last letter from the preceding word and the dot again for the suggestions to show up.

Obviously I don't suggest that people become too submissive on completing from LSP (they can use tune their needs and manual completion when it comes around), but I wish it to always be reliable when needed. I wonder if it also happens for other sources.

Perhaps related to https://github.com/w0rp/ale/issues/2041?

w0rp:
I would just like to note that working on deoplete integration with your help is near the top of the list of my priorities for ALE, alongside the unrelated task of working on more asynchronous programming support for ALE fixers and linters that require complicated logic. The next version of ALE will be tagged as 3.0.0 due to a breaking change I have already added for fixers, which doesn't seem to have caused any problems for anyone, as maybe only I used the feature in particular.

For 3.0.0, I would like to accomplish the following for completion.

1. Get support for getting completions via a custom command in.
2. Get support for using ALE as a deoplete source in, complete with unit tests for the Python code.
3. Maybe support a documented omnicomplete function. (I'm not entirely certain this is possible, or a good idea yet.)
4. Ensure the asynchronous completion support that exists already, which I use pretty much every day at work, continues to function properly.

reaysawa:
Cool. I'll recap the aftermath from the previous discussion from my understanding.

## ALEProcessCompletionResults

The function that will get all of the results and filter them according to some criteria (e.g. the [deoplete filter API](https://github.com/Shougo/deoplete.nvim/tree/master/rplugin/python3/deoplete/filter), any user defined function to his choosing). It's misplaced in the code because it needs access to the results before ALE caps the count with `g:ale_completion_max_suggestions`.

## ALEAfterCompletionResults

What gets done after the LSP server has handed the candidates over. I could not be interested in what ALE plans to do after that so I want to replace the "after" behavior with my own. In the case of deoplete, it could be:

```
function! ALEAfterCompletionResults() abort
    call deoplete#send_event('CompleteDone', 'ale')
endfunction
```


MUComplete, on the other hand, doesn't work asynchronously, so even if the omnifunction could be triggered manually,  (which is what the plugins does, trigger each completion method manually until one of them has something to show), due to ALE it returning an empty array in the first call, it would simply skip it. A way to get around that could be something like:

```
function! ALEAfterCompletionResults() abort
    let b:mucomplete_ale_ready = 1
endfunction
```

From the user's point of view, I think this is needed to differentiate between empty list (not ready) and empty list (I've tried but got nothing). If you don't want to "pollute the global scope" then it could be two user supplied lambdas like:

```
let g/b:ale_completion_filter_func = function('something')
let g/b:ale_completion_after_func = function('something')
```

Then put in if checks in the relevant parts to see if the behavior needs to be overriden.

## s:currently_queued_pos
The way I chose to open up the completion querying was through queueing/polling. The consumer will queue the current position and keep asking for results until ALE tells him to stop (ALEAfterCompletionResults). He will do so involuntarily (which is why resetting the timers and sending new requests won't work) using timer_start or something else until he chooses to stop, which, in the case of deoplete, is setting `context['is_async'] = False`, as I already mentioned above.

The [deoplete source will receive the event sent in ALEAfterCompletionResults](https://github.com/w0rp/ale/pull/2143/files#diff-62c6f532e8e2b6f9c1a574b6b697a71cR32)  and set the  custom property

`context['completion_done'] = True`

If we're halfway through the cycle or just starting `gather_candidates` when this happens, it does not matter since the check happens near the end. If the user types something else just before this happens, since we called

`candidates = self.vim.call(self.poll_func)`

The poll function would have invalidated the previous results just before the check, because candidates got overridden with the empty array returned on [line 201](https://github.com/w0rp/ale/pull/2143/files#diff-5d92fd447d59e3c6fa7e6eccdc172f7aR201); queue knows it's invalid because the position is kept track of in s:currently_queued_pos, and that has changed with user input.

Nevertheless the custom flag ('completion_done') is still set, so it would enter the final if condition and set _is_async_ to false. I think this could be a bug in some situations and the decision to continue should be presented if the position is still valid. We'd need to either:

1. Notify deoplete that the position has changed with another event, set another custom flag, check for that before shutting off.
2. Call `get_completion_position()` again, as it's done at the start, and check if the position is still valid; that being the case, still invalidate the results for the previous position, but keep going.

---

Overall this is the way I chose to go about making the two coordinate together, but you can come up with your own scheme and see if it works well.

reaysawa:
> Get support for using ALE as a deoplete source in, complete with unit tests for the Python code.

It already works to some extent if you want to test it. The setup needed is

```
call deoplete#custom#option({
... other options
  \ 'sources': {
  \   'javascript': ['ale']
  \ }
\ }

call deoplete#custom#source('ale', 'min_pattern_length', 0)

function! ALEAfterCompletionResults() abort
    call deoplete#send_event('CompleteDone', 'ale')
endfunction

call deoplete#enable()
```

Granted I've only used it with tsserver. It has the problems I mentioned with the dot trigger but I don't know exactly who's at fault in that case, it could be some bug in there.

> Maybe support a documented omnicomplete function. (I'm not entirely certain this is possible, or a good idea yet.)

I think people looking to use the completion capabilities are not interested in the omnicomplete pattern, which is something particular to Vim. If I'm an outside caller I need to know:

- How to trigger the queue
- How to know when they're done
- How to fetch the results when they're done

These need to be as direct as possible.

> Ensure the asynchronous completion support that exists already, which I use pretty much every day at work, continues to function properly.

I've included the minimal vimrc with instructions for testing default ALE completion, with as little variable usage as possible, and it's not broken or anything. You're free to see for yourself with test/completion/vimrc/minimal_autocompletion_vimrc.

w0rp:
Do whatever, as long as the following criteria are met.

1. Make sure everything currently supported continues to work.
2. Don't over-complicate the code too much, and don't design any completion functions to be something you can replace.
3. Add unit tests to cover all of the changes in VimL, and unit tests for Python code.

The thread for this PR is _way, way_ too long. Just write some code and don't overthink it.

reaysawa:
@w0rp Oh, I'm sorry, but since you said

> Could you add the Python part of this to the pull request? I'll look at getting all of this to work with tests for everything, including the Python code.

I had understood that you would take care of handling this yourself and were already working out some code. After the seemingly unending back-and-forth and the period of inactivity I've given up on the idea of driving this forward myself.

Since then I'm merely providing information that might be useful for the next person (I thought, you) to use it. The only reason why I commented after this week of a downtime is because I thought you might be still trying to figure out the best way to do this, so I tried to chime in more, as there's no hurt to it.

Let me stress that **I don't have the will or the availability to go forward with this anymore**, as I've already spent time and thought adjusting to not having it. Best of luck and I hope this is understandable.

w0rp:
Okay, thank you for all of the information. I will work on this eventually, but that could be weeks away from now. I'm slowly working on adding more async support in general, and I'll either finish that first or get part of the way into that and then work on this.

w0rp:
This will be implemented later.

PR #2146, opened 2018-12-14
Racket linting using raco

aqui18:
Simple 'linter' that uses the same approach as [neomake](https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/racket.vim) using `raco expand` and regular expressions to match output. 

Will probably look into [racket language server](https://github.com/jeapostrophe/racket-langserver/tree/master) in the future.


test/command_callback/test_racket_raco_command_callback.vader
```diff
@@ -0,0 +1,10 @@
+Before:
+  call ale#assert#SetUpLinterTest('racket', 'raco')
+
+After:
+  call ale#assert#TearDownLinterTest()
+
+Execute(The default command and executable should be correct):
+  AssertLinter 'raco', 'raco expand %s'
```
aqui18: I initially just plugged my code from what was in the docs. However, the ale#Escape function wasn't working for me. 
I assumed that this was because I was calling `raco expand %s` directly in `ale#linter#Define` under "command".

```
    " AssertLinter checks the executable and command.                                                                                                                                                               
    " Pass expected_executable, expected_command                                                                                                                                                                    
    AssertLinter 'some-command', ale#Escape('some-command') . ' --foo'
```

RyanSquared: `ale#Escape` is used for a _single_ file, not multiple separate arguments. In this case, it would be `ale#Escape("raco")`

aqui18: Yeah I tried that first. 
`AssertLinter 'raco', ale#Escape('raco') . ' expand %s'`

Unfortunately, that gives me an error. Is there something I'm doing wrong?

test/command_callback/test_racket_raco_command_callback.vader
```diff
@@ -0,0 +1,10 @@
+Before:
+  call ale#assert#SetUpLinterTest('racket', 'raco')
+
+After:
+  call ale#assert#TearDownLinterTest()
+
+Execute(The default command and executable should be correct):
+  AssertLinter 'raco', ale#Escape('raco') . ' expand %s'
```
w0rp:
```suggestion
  AssertLinter 'raco', 'raco expand %s'
```

You can remove `ale#Escape` on the right here. The executable for the linter isn't escaped because it isn't configurable, it's just written into the command string.

w0rp: Remove the extra blank lines at the end of the file.

aqui18:
Ok I've reverted the test back to `AssertLinter 'raco', 'raco expand %s'` so tests pass now.

w0rp:
Cheers! :beers:

PR #2150, opened 2018-12-14
Update the perl-linter's l:pattern to catch missing errors

mvgrimes:
In some situations, errors reported by `perl -c` can have multiple
listings of "at <file> line <number>". If the l:pattern is changed to
use non-greedy matching it will also match these.

For example:
```
use strict;
use DateTime;
$asdf=1;
```

Results in:
```
Global symbol "$asdf" requires explicit package name (did you forget to declare "my $asdf"?) at /Users/mgrimes/t.pl line 3, <DATA> line 1.
/Users/mgrimes/t.pl had compilation errors.
```

I am not 100% sure why `perl -c` generates errors with the extra "file
line <num>". It only happens in some versions of perl when certain
modules are used.

### DIFF
--- a/ale_linters/perl/perl.vim
+++ b/ale_linters/perl/perl.vim
@@ -18,7 +18,7 @@ function! ale_linters#perl#perl#Handle(buffer, lines) abort
         return []
     endif
 
-    let l:pattern = '\(.\+\) at \(.\+\) line \(\d\+\)'
+    let l:pattern = '\(.\+\) at \(.\+\) line \(\d\+\),'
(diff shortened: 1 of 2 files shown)


mvgrimes:
Sorry, that was the wrong patch. Not sure how that happened, but I've just pushed the correct patch. This passes `run-tests` on my machine.

w0rp:
Cool, that looks better. Add some more tests for the output we're able to capture now.

w0rp:
Cheers! :beers:

PR #2153, opened 2018-12-18
changed gqlint to lint the file on disk

deltaskelta:
- gqlint gathers a config file which searches the directory
  which contains the file being linted. This config file cannot
  be found without linting the file on disk

### DIFF
--- a/ale_linters/graphql/gqlint.vim
+++ b/ale_linters/graphql/gqlint.vim
@@ -1,9 +1,15 @@
 " Author: Michiel Westerbeek <happylinks@gmail.com>
 " Description: Linter for GraphQL Schemas
 
+function! ale_linters#graphql#gqlint#GetCommand(buffer) abort
+    return ale#path#BufferCdString(a:buffer)
+    \   . 'gqlint'
+    \   . ' --reporter=simple %t'
+endfunction
+
 call ale#linter#Define('graphql', {
 \   'name': 'gqlint',
 \   'executable': 'gqlint',
-\   'command': 'gqlint --reporter=simple %t',
+\   'command_callback': 'ale_linters#graphql#gqlint#GetCommand',
 \   'callback': 'ale#handlers#unix#HandleAsWarning',
 \})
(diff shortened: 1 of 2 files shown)


ale_linters/graphql/gqlint.vim
```diff
@@ -1,9 +1,18 @@
 " Author: Michiel Westerbeek <happylinks@gmail.com>
 " Description: Linter for GraphQL Schemas
 
+function! ale_linters#graphql#gqlint#GetCommand(buffer) abort
+    let l:filename = expand('#' . a:buffer . ':t')
+
+    return ale#path#BufferCdString(a:buffer)
+    \   . 'gqlint'
+    \   . ' --reporter=simple '
```
RyanSquared: Linter says there's a space on the end of this line. Please remove, and this should be good to merge.

RyanSquared:
ALE should be able to lint files in-place. Does this get rid of that functionality? What happens if the file is not saved but significant changes are made?

deltaskelta:
For this linter, yes it does forego linting the files in place because I dont foresee any other way to use a project specific config file.

This linter searches the directory that contains the file, the directory the command was run from, and the home directory for the config file

Is it possible to cd into the buffer dir and still run the lint on the buffer file? If that is possible then I think we can maintain the in-place linting

RyanSquared:
> Is it possible to cd into the buffer dir and still run the lint on the buffer file? If that is possible then I think we can maintain the in-place linting

There was another linter that had a similar issue. I do not recall which one. It is possible, but I don't remember how.

deltaskelta:
Ill play with it a little bit and see if I can get it to work in place

w0rp:
Try using `BufferCdString` and `'%t'` together.

deltaskelta:
I got it to work and added a test as well.

w0rp:
Cheers! :beers:

w0rp:
Commit message says the wrong thing, but _meh_.

PR #2161, opened 2018-12-23
Improve perf for compile dbs in large projects

guill:
When using a compilation database (compile_commands.json) in very large
projects, significant delays would occur when changing files --
particularly those that happened to be far down the db. Rather than
iterating over the whole list every time, we now build up a lookup table
based on the tail of the filename (and tail of the directory for
widening searches) and iterate over the much smaller list of compile
commands for files with the given name.

Test metrics (from compile_database_perf/test.sh) show a 90% performance
improvement -- from 25 seconds to 2.5 seconds per run.

Because behavior should be functionally identical to the old version, I didn't have a whole lot of tests to add. I added one test for an old behavior that wasn't tested -- falling back to files in the same directory if we can't find an exact match.

As far as I can tell, vader doesn't have any sort of support for performance profiling. Instead, I added a bash script to verify the performance improvement. Because this particular issue only becomes noticeable in large projects (and I didn't think people would be happy with me checking in a couple hundred tiny test files), the bash script generates all required files for testing.

### DIFF
--- a/autoload/ale/c.vim
+++ b/autoload/ale/c.vim
@@ -157,15 +157,17 @@ if !exists('s:compile_commands_cache')
     let s:compile_commands_cache = {}
 endif
 
-function! s:GetListFromCompileCommandsFile(compile_commands_file) abort
+function! s:GetLookupFromCompileCommandsFile(compile_commands_file) abort
+    let l:empty = [{}, {}]
+
     if empty(a:compile_commands_file)
-        return []
+        return l:empty
     endif
 
     let l:time = getftime(a:compile_commands_file)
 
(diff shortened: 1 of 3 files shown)


guill:
Huh, the AppVeyor build failure seems to be an assertion entirely unrelated to this change. Any idea what might be going on?

`(3/4) [EXECUTE] ale#Escape for cmd.exe should use caret escapes on special characters`

w0rp:
Looks pretty good. Have a look at the conflicts now I've merged something else.

w0rp:
Cheers! :beers:

PR #2163, opened 2018-12-24
Fall back to detecting LSP project roots based on Git repositories

bk2204:
Currently, we detect the linter root based on a variety of techniques. However, these techniques are not foolproof. For example, clangd works fine for many things without a compile_commands.json file, and Go projects may be build outside of the GOPATH to take advantage of Go 1.11's automatic module support.

If the language-specific technique does not produce a root, then the LSP linter is not enabled, which is suboptimal. As a fallback, detect whether we are in a Git repository, and if so, use that as the project root. Handle both regular repositories and submodules.

### DIFF
--- a/autoload/ale/assert.vim
+++ b/autoload/ale/assert.vim
@@ -109,6 +109,14 @@ function! ale#assert#LSPProject(expected_root) abort
     AssertEqual a:expected_root, l:root
 endfunction
 
+function! ale#assert#LSPProjectFull(expected_root) abort
+    let l:buffer = bufnr('')
+    let l:linter = s:GetLinter()
+    let l:root = ale#lsp_linter#FindProjectRoot(l:buffer, l:linter)
+
+    AssertEqual a:expected_root, l:root
+endfunction
+
 function! ale#assert#LSPAddress(expected_address) abort
     let l:buffer = bufnr('')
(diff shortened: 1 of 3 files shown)


w0rp:
There will be some servers where we can't assume .git is the project root, so we can't fall back on .git for all of them. You'll have to fix individual linters instead.

bk2204:
Would it be possible to instead add a hook so that people can provide their own fallback root? There's literally no other way to detect the root of any C project I work on, so LSP functionality is silently nonfunctional there.

w0rp:
Yeah, that sounds like a better idea. Then people can add additional methods for finding project roos.

Carpetsmoker:
>  Go projects may be build outside of the GOPATH to take advantage of Go 1.11's automatic module support.

The algorithm for Go is just wrong. I mentioned this on the PR that introduced it, but the author wasn't very receptive to this feedback: https://github.com/w0rp/ale/pull/1870#pullrequestreview-152689350 and https://github.com/w0rp/ale/pull/1870#discussion_r215426260

I should have a branch somewhere that checks for the existence of `go.mod`, `Gopkg.toml`, `vendor`, and some other stuff, which should be more correct, but I'm not sure what happened to that. It was probably on my old desktop that I ditched when I moved a few months ago >_<

Issue #2112, opened 2018-12-01
Can't get completion working


bradwood:
Hi,

I can't seem to get Python completion working...  I've go the kitchen sink in here right now so I know I need to prune some stuff down, but just can't figure this out!

Any help would be much appreciated.

Thanks!

vimrc:
```
" Enable completion where available.
" This setting must be set before ALE is loaded.
let g:ale_completion_enabled = 1

" Set up the gutter (left margin)
let g:ale_sign_column_always = 1
let g:ale_sign_error = '>>'
let g:ale_sign_warning = '--'

" Configure how ale reports messages/errors on the status line
let g:ale_echo_msg_error_str = 'E'
let g:ale_echo_msg_warning_str = 'W'
let g:ale_echo_msg_format = '[%linter%] %s [%severity%]'

" ALE respects PIPENV
" Favour all pipenv options if they are found
let g:ale_python_auto_pipenv=1
let g:ale_python_flake8_auto_pipenv=1
let g:ale_python_mypy_auto_pipenv=1
let g:ale_python_prospector_auto_pipenv=1
let g:ale_python_pycodestyle_auto_pipenv=1
let g:ale_python_pyflakes_auto_pipenv=1
let g:ale_python_pylint_auto_pipenv=1
let g:ale_python_pyls_auto_pipenv=1

```

ALEInfo:
```
 Current Filetype: python
Available Linters: ['flake8', 'mypy', 'prospector', 'pycodestyle', 'pyflakes', 'pylint', 'pyls', 'pyre', 'vulture']
  Enabled Linters: ['flake8', 'mypy', 'pylint']
 Suggested Fixers: 
  'add_blank_lines_for_python_control_statements' - Add blank lines before control statements.
  'autopep8' - Fix PEP8 issues with autopep8.
  'black' - Fix PEP8 issues with black.
  'isort' - Sort Python imports with isort.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'yapf' - Fix Python files with yapf.
 Linter Variables:

let g:ale_python_flake8_auto_pipenv = 1
let g:ale_python_flake8_change_directory = 1
let g:ale_python_flake8_executable = 'flake8'
let g:ale_python_flake8_options = ''
let g:ale_python_flake8_use_global = 0
let g:ale_python_mypy_auto_pipenv = 1
let g:ale_python_mypy_executable = 'mypy'
[...]

PR #2165, opened 2018-12-26
Add support for bingo

jeremija:
[bingo](https://github.com/saibing/bingo) is a fork of [go-langserver](https://github.com/sourcegraph/go-langserver), and the authors of `go-langserver` [recommend to use `bingo` instead](https://github.com/sourcegraph/go-langserver/blob/master/README.md):

> Note: We have deprioritized work on this language server for use in editors in favor of Google's upcoming Go language server. It is in the best interests of the community to only have a single language server. **If you want to use a Go language server with your editor in the meantime, try https://github.com/saibing/bingo, which is a partial fork of this repository with fixes for Go modules and other editor bugs.**

There doesn't seem to be any mention of `bingo` in the issues except in [#2112](https://github.com/w0rp/ale/issues/2112#issuecomment-443556449). I was not able to make `bingo` work by modifying `go_langserver_executable` and `go_langserver_options` because:

1. `ale_go_langserver_options` become sorted and when I try to specify `--mode stdio --logfile /tmp/lspserver.log --trace` it ends up being something like `--mode --logfile --trace /tmp/lspserver.log stdio`.
2. there was an issue with the way `ale#go#FindProjectRoot` uses `GOPATH` to detect project root. I would always get an error:

```
~/src/go is out of GOPATH workspace ~/src/go, but not a go module project
```

from [this line in bingo source](https://github.com/saibing/bingo/blob/86fe0f2a747bb60c1ba419a2da6a4020ef00cf29/langserver/internal/cache/module_cache.go#L103).

I ended up resolving the issue by detecting the `.git` folder similar to `ale_linters/php/langserver.vim`:

```vim
function! ale_linters#go#bingo#FindProjectRoot(buffer) abort
    return ale#path#FindNearestDirectory(a:buffer, '.git')
endfunction
```

I am not sure if this is the best solution since I haven't been actively using Go, but it seems to be working for me.

Would appreciate any feedback!


ale_linters/go/bingo.vim
```diff
@@ -0,0 +1,34 @@
...
+call ale#Set('go_bingo_executable', 'bingo')
+call ale#Set('go_bingo_options', '--mode stdio')
+
+function! ale_linters#go#bingo#GetCommand(buffer) abort
+    let l:executable = [ale#Escape(ale#Var(a:buffer, 'go_bingo_executable'))]
+    let l:options = ale#Var(a:buffer, 'go_bingo_options')
+    let l:options = filter(split(l:options, ' '), 'empty(v:val) != 1')
+    let l:command = join(extend(l:executable, l:options), ' ')
+
+    return l:command
```
w0rp: Just use `return '%e ' . ale#Pad(ale#Var(a:buffer, 'go_bingo_options'))` instead. The output will be the same, and it's simpler.

jeremija:
Thanks, @w0rp!

PR #2167, opened 2018-12-28
Add support for ale option to override default shell used by ale

stegmanh:
This PR adds the `g:ale_shell` and `g:ale_shell_argument` options. Allowing the user to customize which shell ale will use to execute commands and allowing the user to specify any additional arguments they wish to pass to the execute function.

The reasoning for this change it to address an issue that came up in https://github.com/w0rp/ale/issues/1968. The TL;DR is that there are some instances where Neovim attempts to run a command via `zsh` and it fails due to odd argument passing, which can be fixed by just using `bash` to run the commands.

By default the `g:ale_shell` option is set to the `&shell` vim variable. And the `g:ale_shell_argument` option is set to `-c` as was default in ale.

Files changed in this PR:
* `autoload/ale/job.vim` - Add support for `g_ale_shell` and `g:ale_shell_argument` commands and use those when running commands
* `doc/ale.txt` - Add documentation for new options
* `test/*` - Fix tests


autoload/ale/job.vim
```diff
@@ -188,11 +194,11 @@ function! ale#job#PrepareCommand(buffer, command) abort
         return 'cmd /s/c "' . l:command . '"'
     endif
 
-    if &shell =~? 'fish$\|pwsh$'
-        return ['/bin/sh', '-c', l:command]
+    if g:ale_shell =~? 'fish$\|pwsh$'
```
RyanSquared: i think this should be documented in the section for `g:ale_shell`

autoload/ale/job.vim
```diff
@@ -188,11 +194,11 @@ function! ale#job#PrepareCommand(buffer, command) abort
...
+        return ['/bin/sh', g:ale_shell_arguments, l:command]
```
RyanSquared: please revert this to be hardcoded `'-c'` to avoid conflicts with non-sh-compatible shells; because `sh` is standardized we know we can pass `-c` to it.

doc/ale.txt
```diff
@@ -1990,6 +1990,24 @@ g:ale_windows_node_executable_path         *g:ale_windows_node_executable_path*
...
 
+g:ale_shell                                                       *g:ale_shell*
+
+  Type: |String|
+  Default: `'&shell'`
+
+  This variable is used to determine which shell ale will use to execute 
+  commands. This variables defaults to the value of the vim option '&shell'
+  which corresponds to the $SHELL environment variable. For example 
+  if `$SHELL == '/bin/bash'`, but you want to use zsh, set `g:ale_shell = '/bin/zsh'.`
```
RyanSquared: add in a comment to suggest using `g:ale_shell_arguments` if overriding this, as `g:ale_shell_arguments` can pull incompatible flags from `&shellcmdflag` even if the user changes shell

stegmanh:
Thanks for reviewing that @RyanSquared! I just pushed a PR that addressed those comments. I think I might need to take a look at the appveyor logs though. That build appears to be failing

RyanSquared:
It breaks due to: https://github.com/w0rp/ale/blob/3ec20a730d1c6485ab28704f930a61c6ba426b27/autoload/ale/job.vim#L194

@w0rp both lists and strings are returned from ale#job#PrepareCommand - is that intentional, and if so, can that behavior be changed to be consistent?

w0rp:
> 
> 
> It breaks due to:
> 
> [ale/autoload/ale/job.vim](https://github.com/w0rp/ale/blob/3ec20a730d1c6485ab28704f930a61c6ba426b27/autoload/ale/job.vim#L194)
> 
> Line 194 in [3ec20a7](/w0rp/ale/commit/3ec20a730d1c6485ab28704f930a61c6ba426b27)
>  return 'cmd /s/c "' . l:command . '"' 
> 
> @w0rp both lists and strings are returned from ale#job#PrepareCommand - is that intentional, and if so, can that behavior be changed to be consistent?

Yes. Windows needs Strings, Unix needs Lists. Windows is weird.

w0rp:
The deal with Lists not working on Windows is that the Windows shell has no general support for quoting shell arguments. For `cmd`, argument quoting is handled by the programs, instead of by the shell itself. This means that quoting arguments can sometimes result in say literally `"foo"` instead of `foo` for `"foo"`. Windows is weird.

stegmanh:
Thanks for the review @w0rp. Ill make those changes

stegmanh:
Alright I think these changes are in the direction you were talking about @w0rp.

I was running into issues testing locally with the `readfile` command not being able to read a tmp file. Hopefully it doesn't come up in travis, but if it does Ill investigate further

w0rp:
Cheers! :beers:

stegmanh:
Cool! Thanks for merging this & helping with improving the PR. I'll keep an eye out for your PR so I know how to do `unset` and better formatting in the future. 

:beers:

RyanSquared:
Correct:

```
    (5/5) [EXECUTE] (X) ['/foo/bar', '/c', 'foobar'] should be equal to ['/foo/bar', '-c', 'foobar']
```

w0rp:
Yeah, I broke it.

w0rp:
I pushed a commit which should fix that now. I'll keep an eye on the test results.

PR #2169, opened 2018-12-29
Optionally use neovim's api-highlights

### LINKED ISSUE #1274: Implement support for NeoVim's more robust hightlighting API
I am not sure whether it is a limitation of VIM or ALE, so I hope anyone can clarify.

The problem is the following:

Steps:
1) Suppose I'm editing on any file type, and ALE has marked lines 1, 4, and 5 as warnings
2) I remove line 2. Now, ALE should move the highlights of 4 and 5 to line 3 and 4, immediately, and before even re-running the linter. However, it does not happen.

Depending on the file type and on other configs, I need to save or not save the file I'm editing for ALE to relint the file. Depending on the linter, it may take some time to for ALE to update the highlights (suppose a second or so).

The most troubling scenario is where I have, say - 10 lines consisting of 'imports' relevant to the programming language I am editing, and a few of them are marked 'unused imports'. If I go about and try to delete the unused imports starting from the top, I mess up deleting the wrong lines, because the highlights do not move only until I let ALE relint the file.

This issue also applies to intra-line highlights.

The bottom question is: Shouldn't ALE immediately update highlights according to removed added/removed text regarding-less of the linting process?

andreypopp:
Fixes #1274

Took a stub at it and it seems to work well. Highlights are indeed being
adapted to line instertions so we see less stalled highlights.

I'm going to use it for a while to see how it works.

### DIFF
--- a/autoload/ale/highlight.vim
+++ b/autoload/ale/highlight.vim
@@ -37,11 +37,13 @@ let s:nvim_api = exists('*nvim_buf_add_highlight') && exists('*nvim_buf_clear_na
 
 function! s:ale_nvim_highlight_id(bufnr) abort
     let l:id = getbufvar(a:bufnr, 'ale_nvim_highlight_id', -1)
+
     if l:id is -1
         " NOTE: This will highlight nothing but will allocate new id
         let l:id = nvim_buf_add_highlight(a:bufnr, 0, '', 0, 0, -1)
         call setbufvar(a:bufnr, 'ale_nvim_highlight_id', l:id)
     endif
+
     return l:id
 endfunction
 
--- a/autoload/ale/highlight.vim
+++ b/autoload/ale/highlight.vim
@@ -35,7 +35,7 @@ let s:MAX_COL_SIZE = 1073741824 " pow(2, 30)
 " rebases during edits so we see less stalled highlights.
 let s:nvim_api = exists('*nvim_buf_add_highlight') && exists('*nvim_buf_clear_namespace')
 
-function! ale#highlight#has_nvim_api() abort
+function! ale#highlight#HasNeovimApi() abort
     return s:nvim_api
 endfunction
 
@@ -86,7 +86,7 @@ endfunction
 " except these which have matching loclist item entries.
 
 function! ale#highlight#RemoveHighlights() abort
-    if ale#highlight#has_nvim_api()
+    if ale#highlight#HasNeovimApi()
         if get(b:, 'ale_nvim_highlight_id', 0)
             let l:bufnr = bufnr('%')
             " NOTE: 0, -1 means from 0 line till the end of buffer
@@ -106,7 +106,7 @@ function! ale#highlight#RemoveHighlights() abort
 endfunction
 
 function! s:highlight_line(bufnr, lnum, group) abort
-    if ale#highlight#has_nvim_api()
+    if ale#highlight#HasNeovimApi()
         let l:highlight_id = s:ale_nvim_highlight_id(a:bufnr)
         call ale#highlight#nvim_buf_add_highlight(
         \   a:bufnr, l:highlight_id, a:group,
@@ -118,7 +118,7 @@ function! s:highlight_line(bufnr, lnum, group) abort
 endfunction
 
 function! s:highlight_range(bufnr, range, group) abort
-    if ale#highlight#has_nvim_api()
+    if ale#highlight#HasNeovimApi()
         let l:highlight_id = s:ale_nvim_highlight_id(a:bufnr)
         " NOTE: lines and columns indicies are 0-based in nvim_buf_* API.
         let l:lnum = a:range.lnum - 1
--- a/autoload/ale/highlight.vim
+++ b/autoload/ale/highlight.vim
@@ -36,12 +36,13 @@ let s:MAX_COL_SIZE = 1073741824 " pow(2, 30)
 let s:nvim_api = exists('*nvim_buf_add_highlight') && exists('*nvim_buf_clear_namespace')
 
 function! s:ale_nvim_highlight_id(bufnr) abort
-    if get(b:, 'ale_nvim_highlight_id', -1) == -1
-        " NOTE: This will highlights nothing but will allocate new id
-        let b:ale_nvim_highlight_id = nvim_buf_add_highlight(a:bufnr, 0, '', 0, 0, -1)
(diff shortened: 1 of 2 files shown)


andreypopp:
This is how it looks:
vim-ale-nvim

RyanSquared:
Please check linter errors. Unfortunately, I'm not sure what to do about the errors with the outdated neovim. There's not much we can do about that unless the linter gets updated, I think?

autoload/ale/highlight.vim
```diff
@@ -49,6 +58,18 @@ endfunction
...
+    if get(b:, 'ale_highlight_id', 0)
+        let l:buf = bufnr('%')
+        " NOTE: 0, -1 means from 0 line till the end of buffer
+        call nvim_buf_clear_namespace(l:buf, b:ale_highlight_id, 0, -1)
+    endif
+endfunction
+
+else
+
 function! ale#highlight#RemoveHighlights() abort
```
w0rp: I think you should move the version check inside the function here, and define it once. That will be easier to test, and won't cost too much. It looks like the changes here break earlier versions of NeoVim. You might have to check if `nvim_buf_clear_namespace` exists too. it might be that `nvim_buf_add_highlight` exists in some versions where `nvim_buf_clear_namespace` does not exist.

autoload/ale/highlight.vim
```diff
@@ -57,6 +78,79 @@ function! ale#highlight#RemoveHighlights() abort
...
+            else
+                let l:group = 'ALEWarning'
+            endif
+        elseif l:item.type is# 'I'
+            let l:group = 'ALEInfo'
+        elseif get(l:item, 'sub_type', '') is# 'style'
+            let l:group = 'ALEStyleError'
+        else
+            let l:group = 'ALEError'
+        endif
```
w0rp: Large chunks of code here are identical. Instead of optionally defining one function or the other, separate the code logically into the common parts between the general Vim API and the NeoVim API, and use a check at the top to call one of the two different implementations. You could, for example, build a `List` containing `Dictionary` objects with all of the data to be set, like the `l:group` value and so on, and then use that data for both implementations.

andreypopp:
I've been using this functionality with no problems for a couple of weeks, proceeding with it.

Thanks for the feedback @w0rp, I believe I've addressed all points.

autoload/ale/highlight.vim
```diff
@@ -26,6 +26,24 @@ endif
...
+"
+" Below we define some functions' implementation conditionally if this API
+" exists or not.
+"
+" The API itself is more ergonomic and neovim performs highlights positions
+" rebases during edits so we see less stalled highlights.
+let s:nvim_api = exists('*nvim_buf_add_highlight') && exists('*nvim_buf_clear_namespace')
+
+function! s:ale_nvim_highlight_id(bufnr) abort
+    if get(b:, 'ale_nvim_highlight_id', -1) == -1
```
w0rp: Use `getbufvar` instead of `b:` variables. The variables for `b:` can differ from from buffer for the given buffer number.

autoload/ale/highlight.vim
```diff
@@ -26,6 +26,24 @@ endif
...
+        " NOTE: This will highlights nothing but will allocate new id
+        let b:ale_nvim_highlight_id = nvim_buf_add_highlight(a:bufnr, 0, '', 0, 0, -1)
```
w0rp: You can use `setbufvar` here.

autoload/ale/highlight.vim
```diff
@@ -49,12 +67,83 @@ endfunction
...
+                        \ a:bufnr, l:highlight_id, a:group,
+                        \ l:cur, 0, -1
+                        \ )
+                let l:cur += 1
+            endwhile
+
+            call nvim_buf_add_highlight(
+                \ a:bufnr, l:highlight_id, a:group,
+                \ l:end_lnum, 0, l:end_col
+                \ )
```
w0rp: Put line continuation characters on the same level of indentation as previous lines, and up-indent the content within. See most of the code in the codebase for examples.

andreypopp: Fixed. We'll need a lint rule for that!

andreypopp:
> Is there a way to get highlights via this NeoVim API, as well as set them? If so, add tests which check that the NeoVim-specific highlights are set.

Unfortunately no at this time, this should be possible soon though (asked on neovim gitter channel).

andreypopp:
Lint errors addressed.

w0rp:
I think we should wrap all of the NeoVim API calls in functions we can replace in tests, and use those calls to capture what input is being given. That way we'll be able to test that this code works, and more importantly prevent regressions later. See most of the LSP tests for examples of mocking function calls.

w0rp:
Have a look at some existing tests. You have to make sure the file is loaded first with `runtime`, replace it, and then load it again in an After block.

andreypopp:
@w0rp is there a way to execute tests "locally" w/o docker and only a subset of tests? I find it very hard to iterate (I'm vimscript beginner!) when tests take so much time to run. I really want to finish my PRs to ALE and bring more.

w0rp:
If you install Vader, you can run tests for a Vader file with the `:Vader` command. You can run subset of tests with the Docker image with a glob like `./run-tests test/directory/*.vader`. Use `--help` or check the documentation to see how you can limit the Vim versions to run some tests more quickly. Make sure to install Docker for running all of the tests locally.

andreypopp:
This is ready for review.

AppVeyor task should be cancelled as I froze it with leaving `:echo` in test code in past commit.
For some reason tests fail locally on my neovim 0.4.0 due to not being able to override functions (I suspect neovim 0.4.0 has problems with Vader) but it's ok on CI.

davidsierradz:
Hi, this commit somehow needs `let g:ale_set_signs = 1` to work correct, let me explain:

Without commit `114198e`:

With `let g:ale_set_signs = 0` (good):

2019-05-05-100823_1920x1080_scrot

With `let g:ale_set_signs = 1` (good):

2019-05-05-100835_1920x1080_scrot

With commit `114198e`:

With `let g:ale_set_signs = 0` (bad, send error highlight to next line):

2019-05-05-100717_1920x1080_scrot

With `let g:ale_set_signs = 1` (good):

2019-05-05-100637_1920x1080_scrot

andreypopp:
@davidsierradz indeed, thanks for catching that, it's off by one error. Should be fixed with #2475

PR #2174, opened 2019-01-01
Add initial ameba (crystal linter) support

HarrisonB:
Note that this depends on the saved files as `ameba` does not have STDIN
support.

Tests have been added for handling the JSON output of `ameba`. Let me know if you think additional tests are needed.

### DIFF
--- a/ale_linters/crystal/ameba.vim
+++ b/ale_linters/crystal/ameba.vim
@@ -20,11 +20,11 @@ call ale#linter#Define('crystal', {
 
 " Handle output from ameba
 function! ale_linters#crystal#ameba#HandleAmebaOutput(buffer, lines) abort
-    try
-        let l:errors = json_decode(a:lines[0])
-    catch
-        return []
-    endtry
+    if len(a:lines) == 0
+      return []
+    endif
+
+    let l:errors = ale#util#FuzzyJSONDecode(a:lines[0], {})
 
     if !has_key(l:errors, 'summary')
     \|| l:errors['summary']['issues_count'] == 0
--- a/ale_linters/crystal/ameba.vim
+++ b/ale_linters/crystal/ameba.vim
@@ -35,15 +35,15 @@ function! ale_linters#crystal#ameba#HandleAmebaOutput(buffer, lines) abort
     let l:output = []
 
     for l:error in l:errors['sources'][0]['issues']
-        let l:start_col = l:error['location']['column'] + 0
-        let l:end_col = l:error['end_location']['column'] + 0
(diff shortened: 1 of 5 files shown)


RyanSquared:
Tests for the customization of the executable path would also be preferred.

ale_linters/crystal/ameba.vim
```diff
@@ -0,0 +1,56 @@
...
+    if !has_key(l:errors, 'summary')
+    \|| l:errors['summary']['issues_count'] == 0
+    \|| empty(l:errors['sources'])
+        return []
+    endif
+
+    let l:output = []
+
+    for l:error in l:errors['sources'][0]['issues']
+        let l:start_col = l:error['location']['column'] + 0
```
RyanSquared: why + 0? type coersion? i'd prefer str2nr() just because it's more obvious what it's doing

ale_linters/crystal/ameba.vim
```diff
@@ -0,0 +1,56 @@
...
+        let l:end_col = l:error['end_location']['column'] + 0
+
+        if !l:end_col
+          let l:end_col = l:start_col + 1
+        endif
+
+        call add(l:output, {
+        \   'lnum': l:error['location']['line'] + 0,
```
RyanSquared: see above

ale_linters/crystal/ameba.vim
RyanSquared: from looking below at the tests it looks like it should be a number

ale_linters/crystal/ameba.vim
```diff
@@ -0,0 +1,56 @@
+" Author: Harrison Bachrach - https://github.com/HarrisonB
+" Description: Ameba, a linter for crystal files
+
+call ale#Set('crystal_ameba_executable', 'bin/ameba')
```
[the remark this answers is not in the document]
HarrisonB: I did this based on the fact that crystal doesn’t have a global install for its package manager (`shards`), so most devs wouldn’t necessarily have it in their path. There is a homebrew formula for it, however.

I can change it if you’d prefer.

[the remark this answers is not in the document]
w0rp: Okay, sounds good me. Anything is good if there's a reason for it.

veelenga:
Hey @w0rp, any update here? Thank you guys for doing this.

w0rp:
Cheers! :beers:

I'll additionally set `'lint_file': 1` so the linter doesn't run as you type, as it doesn't support that.

HarrisonB:
Thanks @w0rp, @RyanSquared, and @veelenga! 😄

Issue #1830, opened 2018-08-20
Add a command for manually triggering completion


nerdrew:
I can't find a way to manually trigger completion, i.e. via my own shortcut. I only see the timer based completions? If there isn't a way to manually trigger completion, can I request the feature?

I'm thinking of something like [this commit](https://github.com/nerdrew/ale/commit/a8562b42a600c036fa0501a717eadb94ee79c7a1) which would allow me to use "tab" to trigger completion.

w0rp:
We can't output a function to be used with `omnifunc`, as Vim only supports searching for completion results synchronously. Instead, we can do what the feature for searching for completion results as you type does now, which is make a request for completion results, get the response from the language server, temporarily replace `omnifunc`, show the results, and reset `omnifunc` after `CompleteDone`. The command will just call the same function that is called after a delay as you type.

w0rp:
I recently realised that we may also be able to offer an `omnifunc` function that just returns nothing and executes the rest of the logic as usual, before being replaced with the temporary function. It might work, it might not.

w0rp:
Another area where this is also useful is when you are using completion suggestions, but where the completions have to be triggered manually, and not by typing. For example:

```ts
const x: T = {
  // <- Cursor rests here
}
```

In a position like this, manually requesting completion will allow you to select a key to input in an object. As you input more keys in TypeScript, the completion results narrow to suggest only the remaining keys you have not provided values for.

dakom:
As it is right now - how do you trigger the autocompletion in the `foo.bar` case where we just type `foo.`

Currently I find myself needing to type `b` and only then does ALE kick in. Would be great if it gave suggestions immediately after `.`

dakom:
Hmmm I'm also trying in typescript. Here's my current ALE config: https://github.com/dakom/dotfiles/blob/5f23fd69220e5d230a5ebbc3f397249c074fec45/.config/nvim/init.vim#L209

dakom:
Hmmm - actually I can get it to work, but then sometimes it doesn't. Here's a reproducable situation where it doesnt:

1. `foo.` (let's assume it works)
2. cancel the auto complete
3. delete the `.` (now just `foo`)
4. `foo.` - no longer works

w0rp:
[...]

PR #2177, opened 2019-01-03
Add ability to manually trigger completion menu

jeremija:
Following the discussion in #1830, a part of `ale#completion#GetCompletions` was extracted into a new function: `ale#completion#AlwaysGetCompletions`. A new `:ALEComplete` command was added, along with a plug mapping: `<Plug>(ale_complete)`. Auto-completion can be triggered manually whether or not `g:ale_completion_enabled` is set.

Sample bindings:

```vim
inoremap <silent> <C-Space> <C-\><C-O>:GetCompletions<CR>
" or
imap <C-Space> <Plug>(ale_complete) 
```

### DIFF
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -820,6 +820,9 @@ with |g:ale_completion_max_suggestions|.
 If you don't like some of the suggestions you see, you can filter them out
 with |g:ale_completion_excluded_words| or |b:ale_completion_excluded_words|.
 
+The |ALEComplete| command can be used to show completion suggestions manually,
+even when |g:ale_completion_enabled| is set to `0`.
+
                                                 *ale-completion-completopt-bug*
 
 ALE implements completion as you type by temporarily adjusting |completeopt|
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -2235,6 +2235,17 @@ ALE will use to search for Python executables.
 ===============================================================================
 8. Commands/Keybinds                                             *ale-commands*
 
+ALEComplete                                                       *ALEComplete*
+
+  Manually trigger LSP autocomplete and show the menu. Works only when called
+  from insert mode. >
+
+    inoremap <silent> <C-Space> <C-\><C-O>:GetCompletions<CR>
+<
+  A plug mapping `<Plug>(ale_complete)` is defined for this command. >
+
+    imap <C-Space> <Plug>(ale_complete)
+<
 ALEDocumentation                                             *ALEDocumentation*
 
   Similar to the |ALEHover| command, retrieve documentation information for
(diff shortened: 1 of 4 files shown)


doc/ale.txt
```diff
@@ -2208,6 +2208,17 @@ ALE will use to search for Python executables.
 ===============================================================================
 8. Commands/Keybinds                                             *ale-commands*
 
+ALEComplete                                                       *ALEComplete*
+
+  Manually trigger LSP autocomplete and show the menu. Works only when called
+  from insert mode. >
+
+    inoremap <silent> <C-Space> <C-\><C-O>:GetCompletions<CR>
```
w0rp:
```suggestion
    inoremap <silent> <C-Space> <C-\><C-O>:ALEComplete<CR>
```

I think you need to update the command here.

jeremija: Thanks, that should be fixed now!

jeremija:
Hi @w0rp, I updated the docs and added the test. In the test I noticed that I cannot override the `ale#completion#AlwaysGetCompletions()` function without it being called first, because it is an autoloaded function. Do you know of a better way to do this, so that I can remove the initial call in the test?

w0rp:
If you rebase your code, that test on Windows will pass.

You can replace the function in Vader tests. Look through existing tests for completion. You can define the function in `Before` or in an `Execute` block, and then reload the original function again with `runtime autoload/ale/completion.vim` in `After`.

jeremija:
I guess I'm confused why I get this error in Vim 8.1:
```
Vim(function):E746: Function name does not match script file name: ale#completion#AlwaysGetCompletions
```

for this simple test case:

```
Before:
  let g:get_completions_called = 0
  function! ale#completion#AlwaysGetCompletions() abort
    let g:get_completions_called = 1
  endfunction

Execute(ale#completion#AlwaysGetCompletions should be called when ALEComplete is executed):
  AssertEqual 0, g:get_completions_called
  ALEComplete
  AssertEqual 1, g:get_completions_called
```

But the following works:

```
Before:
  function! MockAlwaysGetCompletions() abort
    let g:get_completions_called = 0

    function! ale#completion#AlwaysGetCompletions() abort
      let g:get_completions_called = 1
    endfunction
  endfunction

  call MockAlwaysGetCompletions()

Execute(ale#completion#AlwaysGetCompletions should be called when ALEComplete is executed):
  AssertEqual 0, g:get_completions_called
  ALEComplete
  AssertEqual 1, g:get_completions_called

```

w0rp:
You'll need to add a `runtime` line before you define the function, probably, or try moving it to the `Execute` block. Have a look at existing test files. They do similar things and the tests pass.

jeremija:
Thanks, I think I've already tried that and it didn't work. Did you get a chance to take a look at the latest commit? I did something similar to what I wrote above and the tests pass.

w0rp:
Cheers! :beers:

PR #2182, opened 2019-01-05
Ruby/Brakeman: only lint current file

geoffharcourt:
When the Brakeman linter was added, Brakeman didn't yet have the ability
to only check a single file. This results in ALE filling the location
list with violations from elsewhere in the project that aren't relevant
to the current buffer.

This change uses the `--only-files` argument to scope the Brakeman check
to the file of the current buffer. Note that this change will cause some
controller/view interactions to be missed, as Brakeman won't consider
violations that require scanning the current buffer's file in the
context of other files in the project (such as XSS with a model fetched
in a controller but output unescaped in a view), but it cuts down
significantly on unactionable noise.

Brakeman docs:
https://brakemanscanner.org/docs/options/

### DIFF
--- a/ale_linters/ruby/brakeman.vim
+++ b/ale_linters/ruby/brakeman.vim
@@ -3,7 +3,6 @@
 
 call ale#Set('ruby_brakeman_options', '')
 call ale#Set('ruby_brakeman_executable', 'brakeman')
-call ale#Set('ruby_brakeman_options', '')
 
 function! ale_linters#ruby#brakeman#Handle(buffer, lines) abort
     let l:output = []
@@ -40,6 +39,7 @@ function! ale_linters#ruby#brakeman#GetCommand(buffer) abort
     \    . ' -f json -q '
     \    . ale#Var(a:buffer, 'ruby_brakeman_options')
     \    . ' -p ' . ale#Escape(l:rails_root)
+    \    . ' --only-files ' . expand('#' . a:buffer . '...')
 endfunction
 
 call ale#linter#Define('ruby', {
--- a/test/command_callback/test_brakeman_command_callback.vader
+++ b/test/command_callback/test_brakeman_command_callback.vader
@@ -15,6 +15,8 @@ Execute(The brakeman command callback should find a valid Rails app root):
   AssertLinter 'brakeman', ale#Escape('brakeman')
   \ . ' -f json -q  -p '
   \ . ale#Escape(ale#path#Simplify(g:dir . '/../ruby_fixtures/valid_rails_app'))
+  \ . ' --only-files '
+  \ . ale#path#Simplify(g:dir . '/../ruby_fixtures/valid_rails_app/db/test.rb')
 
 Execute(The brakeman command callback should include configured options):
   call ale#test#SetFilename('../ruby_fixtures/valid_rails_app/db/test.rb')
@@ -24,6 +26,17 @@ Execute(The brakeman command callback should include configured options):
   AssertLinter 'brakeman', ale#Escape('brakeman')
   \ . ' -f json -q --combobulate -p '
   \ . ale#Escape(ale#path#Simplify(g:dir . '/../ruby_fixtures/valid_rails_app'))
+  \ . ' --only-files '
+  \ . ale#path#Simplify(g:dir . '/../ruby_fixtures/valid_rails_app/db/test.rb')
+
+Execute(The brakeman command callback should use the relative path of the buffer):
+  call ale#test#SetFilename('../ruby_fixtures/valid_rails_app/db/test.rb')
+
+  AssertLinter 'brakeman', ale#Escape('brakeman')
+  \ . ' -f json -q  -p '
+  \ . ale#Escape(ale#path#Simplify(g:dir . '/../ruby_fixtures/valid_rails_app'))
+  \ . ' --only-files '
+  \ . ale#path#Simplify(g:dir . '/../ruby_fixtures/valid_rails_app/db/test.rb')
 
 Execute(Setting bundle appends 'exec brakeman'):
   call ale#test#SetFilename('../ruby_fixtures/valid_rails_app/db/test.rb')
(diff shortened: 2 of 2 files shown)


ale_linters/ruby/brakeman.vim
```diff
@@ -3,7 +3,6 @@
 
 call ale#Set('ruby_brakeman_options', '')
 call ale#Set('ruby_brakeman_executable', 'brakeman')
-call ale#Set('ruby_brakeman_options', '')
```
geoffharcourt: This looked like it was accidentally duplicated.

ale_linters/ruby/brakeman.vim
```diff
@@ -40,6 +39,7 @@ function! ale_linters#ruby#brakeman#GetCommand(buffer) abort
     \    . ' -f json -q '
     \    . ale#Var(a:buffer, 'ruby_brakeman_options')
     \    . ' -p ' . ale#Escape(l:rails_root)
+    \    . ' --only-files ' . expand('#' . a:buffer . '...')
```
geoffharcourt: https://brakemanscanner.org/docs/options/

geoffharcourt:
An alternative approach would be to still run against all files but to filter out errors not tagged with the current buffer's file. That approach would be less likely to miss an issue that required two files to spot, but has the disadvantage of taking a much longer time to check.

geoffharcourt:
@w0rp on second look, it appears that the functionality was added in Brakeman 2.0.0, which is now six years old. Is it worth supporting? I'm happy to write logic for this differentiation, but wasn't sure if that's so far in the past that Brakeman 1.x support is desired.

w0rp:
Oh, if it has been there for 6 years, then no. That would be silly.

w0rp:
@elebow What do you think of the changes here? I'd like to know your thoughts. I don't write Ruby code myself.

elebow:
:thumbsup: I agree with this change. Brakeman can be very slow on large applications, so `--only-files` would significantly increase usability at the cost of incompleteness.

My only additional comment would be to consider allowing users to choose the old behavior, or perhaps the "filter out errors for other buffers" behavior discussed above (https://github.com/w0rp/ale/pull/2182#issuecomment-451747516). Some users may prefer the thoroughness of the full brakeman scan.

w0rp:
Okay, thanks for the input. @geoffharcourt Could you add an option for enabling the old behavior, which can be set to `0` by default?

presidentbeef:
Hi all,

May I suggest having `--only-files` be opt-in, instead? Results can be very, very wrong when Brakeman does not have access to the entire project. In particular, if Brakeman cannot figure out the version of Rails in use, it defaults to assuming the *oldest* supported version (2.3.x.), which is usually wrong.

Additionally, as @geoffharcourt's original comment points out, using `--only-files` will cause Brakeman to miss security issues. The main strength of Brakeman's analysis is the MVC dataflow, which is completely lost with single file analysis.

geoffharcourt:
@presidentbeef thanks for weighing in!

I think scanning everything and then trimming down the results to the relevant file is going to be the best approach in the end (which means the current solution should be tossed out), as it will show MVC problems in views with variables set in the controller, etc.

My pushback on opt-in is that the Brakeman behavior as it exists in ALE is very noisy and often not actionable in the current buffer. I might be editing a model and I see warnings about files that have no relationship to what I'm working on. It's also a massive difference in execution time for in-buffer linting, in our app scanning a single file takes 1.0s vs. 9-10s.

w0rp:
I agree. Using the option which gives you results much faster for only the file you are looking at should be the default. The option which takes longer to give you results and looks at more files should be opt-in.

w0rp:
Another way of looking it is that if you give someone a linting tool which runs quickly and gives them 90% accurate results, they are more likely to actually use that instead of a tool which takes 5X as long and gives 100% accurate results. The difference is actually then between 0% and 90% for most people, because people just disable slow things entirely, and then there's 100% for people who know enough to wait a bit longer for more accurate results.

presidentbeef:
So, I was wrong above. Brakeman will grab relevant files like `Gemfile` regardless of the `--only-files` option, so it should know the correct Rails version.

However, I still dislike this use of `--only-files` and wish I had never added it. Hopefully people won't rely solely on the use of ALE and will run full Brakeman scans elsewhere, like in CI.

w0rp:
Yeah, you shouldn't really on developers to catch linting issues on their machines. It's the same as relying on clients to santize HTML.

I have a similar workflow at my job. We run TSLint as a tsserver plugin, and also with a plugin for detecting import cycles. The latter doesn't work unless it scans the entire project. I can see most TSLint issues in Vim, but I can only see import cycles by running a script. Though I did also add an ESLint AST rule which just looks at the filename of specific imports for files in one specific directory, because there was one common cause of import cycles coming from one module in one directory, that new devs could add by accident.

geoffharcourt:
Getting all the results and filtering the lines turns out to be quite difficult for the controller-view interaction scenario, as Brakeman isn't a linter for ERB. It looks like there's three options:

1. Keep the above PR as-is, run for file-specific errors and rely on CI for a project-wide scan. (Lose interaction checks, but gain speed and reduce response time).
2. Add Brakeman as an ERB linter, run against the whole project, filter result sets down to the file (get the interaction checks, but expands the scope).
3. Abandon the change and stick with whole-project in every file (what's currently in ALE, gets everything but takes longer and shows violations from other files).

w0rp:
Add an option for this, and you can leave it off by default. Then people who want it can turn it on.

presidentbeef:
> shows violations from other files

How are violations in other files shown? Or is it just that they are reported but not displayed to the user? (I haven't seen these in my experience with ale, so I'm genuinely curious.)

geoffharcourt:
@presidentbeef for our team, we have some warnings that are either false positives or low-level issues on our backlog. When ALE parses the output, it shows with warnings in the location list (and indicates there are violations) despite the current file being unrelated to the issues.

presidentbeef:
@geoffharcourt and @w0rp  can you take a look at https://github.com/presidentbeef/brakeman/issues/1368 and provide any feedback you might have?

w0rp:
Sure, sorry about the delay. I'll have a look now. It sounds like it's worth leaving this pull request open for now, as what happens in brakeman will decide what we do here.

w0rp:

PR #2183, opened 2019-01-06
Allows ale to leverage npx on-demand / no-install cli task for bash-l…

devendor:
…anguage-server execution.

### DIFF
--- a/ale_linters/sh/language_server.vim
+++ b/ale_linters/sh/language_server.vim
@@ -8,13 +8,19 @@ call ale#Set('sh_language_server_use_global', get(g:, 'ale_use_global_executable
 function! ale_linters#sh#language_server#GetExecutable(buffer) abort
     return ale#node#FindExecutable(a:buffer, 'sh_language_server', [
     \   'node_modules/.bin/bash-language-server',
+    \   'node_modules/.bin/npx',
+    \   exepath('npx')
     \])
 endfunction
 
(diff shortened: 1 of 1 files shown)


devendor:
I agree with you and got half way through a failed attempt before bumping into my own lack of experience with vimscript and the codebase.

How about something like this:

```
g:ale_sh_language_server_alias = '%s -s bash-language-server ' 
g:ale_sh_language_server_executable = 'npx'
```

I don't have the time to make it so right now since I'd have to bone up on my vimscript, and I've to pay some bills, but it's a better design.

andreypopp:
Just as a data point I want a similar feature but for [esy](https://esy.sh)
(which is a sandboxed dev workflow) where all tools installed (even native) is
available locally:

```
% esy prettier
% esy ocamlmerlin
```

Thus I want ALE to know how to invoke them and also lookup commands within the
env with:

```
% esy which prettier
```

w0rp:
Okay, you can open another issue for that.

w0rp:
What I'll do for now is close this PR, and we can come up with a better design that covers everything elegantly. In the meantime, you can use `g:ale_command_wrapper` to run linter commands via a script, so you can use `npx` personally, or you can maintain your own fork with this in, or write your own custom linter. You can replace ALE linters with your own overrides, at your own risk.

PR #2189, opened 2019-01-08
Manually trigger autocomplete even when prefix is ""

jeremija:
This is also related to #1830. Adds a fix for the following scenario:

```typescript
const x: T = {
  // <- Cursor rests here
}
```

Previously, the manual completion would not be triggered when the prefix was empty.


autoload/ale/completion.vim
```diff
@@ -509,17 +509,18 @@ function! ale#completion#GetCompletions() abort
...
     endif
 
-    call ale#completion#AlwaysGetCompletions()
+    call ale#completion#AlwaysGetCompletions(1)
 endfunction
 
 " This function can be used to manually trigger autocomplete, even when
 " g:ale_completion_enabled is set to false
-function! ale#completion#AlwaysGetCompletions() abort
+function! ale#completion#AlwaysGetCompletions(...) abort
```
w0rp: This function isn't part of ALE's public API, so just add a new argument here and update the command. This is why I asked for the command to be part of the public API, but not the function. :wink:

autoload/ale/completion.vim
```diff
@@ -106,7 +106,7 @@ function! ale#completion#Filter(buffer, filetype, suggestions, prefix) abort
     "   foo.
     "       ^
     " We need to include all of the given suggestions.
-    if index(l:triggers, a:prefix) >= 0
+    if index(l:triggers, a:prefix) >= 0 || empty(a:prefix)
```
w0rp: Instead of writing this here, move `if empty(a:prefix)` to the top of the function, before `l:triggers` is loaded, and return `a:suggestions` right away. That will be more efficient when there's no prefix.

w0rp: In `GetTriggerCharacter` above, you should probably add `if empty(a:prefix) ... return ''` to the top of the function, to avoid processing we don't need again.

w0rp: Ah, I said _return_ `a:suggestions` right away. What I should say is set the variable right away without the other processing above. The suggestions still need to be filtered below, as some LSP configurations have filters in Vim for removing noise.

w0rp: Add a test which ensures that Python filtering on suggestions still works when there's no prefix.

jeremija: Roger! When you say:

> The suggestions still need to be filtered below, as some LSP configurations have filters in Vim for removing noise.

are you referring to the block starting with `if !empty(l:excluded_words)`?

jeremija: > Add a test which ensures that filtering on suggestions for Python still works when there's no prefix.

Do you think the test I added to `test_completion_filtering.vader` is enough?

```vim
  AssertEqual
  \ ['Deutsch'],
  \ ale#completion#Filter(bufnr(''), '', ['Deutsch'], '')
```

w0rp: Yes, we the function shouldn't return early, I realised, and should continue down to the part for processing filtering with functions. Look through the tests and make sure that we have a test for the Python language server filtering function which still filters the results when the prefix is empty.

PR #2190, opened 2019-01-08
Add textlint support for html

doronbehar:
I'm not sure if I should add tests, or where to add them if I do because I haven't seen tests that are exclusively needed to test a particular filetype that is handled by textlint.

### DIFF
--- a/test/handler/test_textlint_handler.vader
+++ b/test/handler/test_textlint_handler.vader
@@ -1,5 +1,7 @@
 Before:
   runtime! ale_linters/markdown/textlint.vim
+  runtime! ale_linters/text/textlint.vim
+  runtime! ale_linters/html/textlint.vim
 
 After:
   call ale#linter#Reset()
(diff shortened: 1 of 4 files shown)


test/handler/test_textlint_handler.vader
```diff
@@ -39,3 +41,75 @@ Execute(textlint handler should no error output):
   \ [],
   \ ale#handlers#textlint#HandleTextlintOutput(bufnr(''), [
   \ ])
+
```
w0rp: I think you got the wrong idea. Undo the changes to this file, and add a test in the `test/command_callback` directory for the HTML filetype. Read `:help ale-development-linter-tests`. You want to add an `AssertLinter` line to a new file.

doronbehar: Well TBH it's the 1st time in my life I'm writing tests. I was thinking of adding something to [`test_textlint_command_callbacks.vader`](https://github.com/w0rp/ale/blob/f23811770a8104346b7e9ccc6e586da828c8f41d/test/command_callback/test_textlint_command_callbacks.vader) but it doesn't seem like there are any `Execute` statements which relate to `.md` files, there's only a reference to `foo.txt` file. Since currently only `markdown` files are linted by `textlint` I'd expect to find there a reference to a `foo.md` path.

w0rp: That's okay, all you need to test is that the linter can be loaded and produces the correct executable and command. The filename doesn't matter.

w0rp: The existing tests are in this file: test/command_callback/test_textlint_command_callbacks.vader

You should add something like this.

```vim
Before:
  call ale#assert#SetUpLinterTest('html', 'textlint')

After:
  call ale#assert#TearDownLinterTest()

  AssertLinter 'textlint',
  \ ale#Escape('textlint') . ' -f json --stdin --stdin-filename %s'
```

doronbehar: I've added `call ale#assert#SetUpLinterTest('html', 'textlint')` but since the final command is the same for html as well I'm not sure if there's anything else to add. I assume the detection of the correct linter when a certain filetype is detected is tested once somewhere else?

w0rp: You just need to add a file with almost exactly the code I recommend above. That will test that the `textlint` linter can be loaded for the `html` filetype, and produces the correct executable string and command.

w0rp: That commit adds test for the function that parses the output. You need to also add tests that cover the command and executable that are produced. Use the code I shared above in a file in `test/command_callback`. Edit the code slightly if it doesn't work for some reason.

w0rp:

PR #2193, opened 2019-01-09
Add textlint for Asciidoc and add it to Fixers

ta2gch:
I made this patch to run textlint at asciidoc too, and also to work `textlint --fix` in `ALEFix`.
Thank you!


ta2gch:
I got it. However, I think you don't need tests for the linter because the tests are in this repository as for markdown. For example, A linter `write-good` have just one test sets for only markdown but `write-good` covers `asciidoc` too. 

Should I add just the tests for fixer function?

ta2gch:
I added them.

w0rp:
Add tests for the linter. You need to use the `AssertLinter` command. See similar comments in #2190.

ta2gch:
I'm appriciate you to tell me again, I got what you say. I added it, Please check it,

w0rp:
Cheers! :beers:

PR #2168, opened 2018-12-29
Fix CFLAGS & Makefile issues

fx-carton:
This PR fixes #2049
The arguments are split by `' '` instead of `'-'`, so the bugs where other filenames are appended to the flags are fixed. The cases with quoted arguments containing spaces are handled, as well as arguments separated with spaces (eg. `-D directory`).

PR #2194, opened 2019-01-09
Fix cflags parsing

fx-carton:
This fixes the bug mentionned in #2168. A relevant test case has been added.


davidvandebunte:
My manual test is probably covering the same code as the automatic test, but I can confirm this is working for me. Thanks for fixing this up!

w0rp:
Cheers! :beers:

PR #2206, opened 2019-01-14
Add hindent fixer support

gilligan:
This adds [hindent](https://github.com/chrisdone/hindent) as a fixed to ALE. 

**Note**: As you can tell by looking at the PR it is missing tests. I did add documentation and on my setup it works fine. I did have a look at the stylish-haskell test implementation but from what I can tell that test adds next to no value. It only checks some default options and doesn't invoke stylish-haskell at all.

If the lack of tests mean my PR can't be merged that that is quite OK. It only took me around 5min to add. I just don't have the time now to look into how to create a proper, useful test. I don't want to copy/paste the stylish-haskell one.

Thanks :tada:


asymmetric:
@gilligan as a `hindent` user I'd be really happy to see this get merged :)

w0rp:
Superseded by #2462.

PR #2207, opened 2019-01-15
Fixed parsing of pydocstyle errors

pmacosta:
ale#Escape function seems to prepend and append ' to the file name, which
are not present in the pydocstyle output. Having the parsing regexp match
the file name was overkill anyway, since there is an obvious 1:1
correspondence between the buffer number and the (potential) errors
reported by pydocstyle.


w0rp:
Cheers! :beers:

PR #2220, opened 2019-01-18
Add `babylon` parser for `javascript` filetypes

chaucerbao:
I'm not sure why the default `let l:parser = 'babylon'` [was removed](https://github.com/w0rp/ale/commit/cb8ad9fbd82df92993e526a8e48d4400ba08936b#diff-3c6cea148687a701cb21dbf2e3e4b5ddL40). `babylon` is the default parser for **prettier**.

But since it was removed, this fixer doesn't work for regular Javascript files. This PR will connect `javascript` file-types to the `babylon` parser, while keeping the current `let l:parser = ''`.


chaucerbao:
Actually, the first is a better solution. There's never a case, when there is no file extension, where you'd want **prettier** to auto-detect the file type, because it basically cannot. That's why there's a default parser.

I'll change this PR to reflect that.

w0rp:
I wonder if we might have to use the newer name for the parser for newer `prettier` versions.

PR #2223, opened 2019-01-19
End position in LSP range is exclusive

andreypopp:
From LSP spec:

> A range in a text document expressed as (zero-based) start and end
> positions. A range is comparable to a selection in an editor. Therefore
> the end position is exclusive.

https://microsoft.github.io/language-server-protocol/specification

### DIFF
--- a/autoload/ale/lsp/response.vim
+++ b/autoload/ale/lsp/response.vim
@@ -33,7 +33,7 @@ function! ale#lsp#response#ReadDiagnostics(response) abort
         \   'lnum': l:diagnostic.range.start.line + 1,
         \   'col': l:diagnostic.range.start.character + 1,
         \   'end_lnum': l:diagnostic.range.end.line + 1,
-        \   'end_col': l:diagnostic.range.end.character + 1,
+        \   'end_col': l:diagnostic.range.end.character,
         \}
 
         if l:severity == s:SEVERITY_WARNING
--- a/test/lsp/test_read_lsp_diagnostics.vader
+++ b/test/lsp/test_read_lsp_diagnostics.vader
@@ -17,7 +17,7 @@ Execute(ale#lsp#response#ReadDiagnostics() should handle errors):
   \     'lnum': 3,
   \     'col': 11,
   \     'end_lnum': 5,
-  \     'end_col': 16,
+  \     'end_col': 15,
   \     'code': 'some-error',
   \   }
   \ ],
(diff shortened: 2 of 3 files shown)


andreypopp:
Will update the tests if you agree with this change.

w0rp:
I think this might be right. It's hard to tell. Can you share some examples of highlights or something being corrected by this change? Some screenshots with example language servers would be helpful.

andreypopp:
This is what I see with `ocamlmerlin-lsp`:
screenshot 2019-01-21 at 17 58 33

This is from @TrySound with `flow lsp`:
unnamed

As you can see the `(` shouldn't be highlighted.

andreypopp:
Can't find what's wrong with appveyor by looking at logs... it seem to repo everything is ok...

andreypopp:
Found the failed tests but not sure why they are failing:
```
[00:01:38]   Starting Vader: C:\testplugin\test\smoke_test.vader
[00:01:38]     (1/3) [  GIVEN] Some imaginary filetype
[00:01:38]     (1/3) [EXECUTE] Linters should run with the default options
[00:01:38]     (2/3) [  GIVEN] Some imaginary filetype
[00:01:38]     (2/3) [EXECUTE] Linters should run in PowerShell too
[00:01:38]     (2/3) [EXECUTE] (X) [{'lnum': 4, 'bufnr': 3, 'col': 3, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'E', 'pattern': '', 'text': 'foo bar'}] should be equal to [{'lnum': 1, 'bufnr': 3, 'vcol': 0, 'pattern': '', 'valid': 1, 'nr': -1, 'type': 'E', 'col': 3, 'text': 'foo'}, {'lnum': 2, 'bufnr': 3, 'vcol': 0, 'pattern': '', 'valid': 1, 'nr': -1, 'type': 'E', 'col': 3, 'text': 'bar'}]
[00:01:38]     (3/3) [  GIVEN] Some imaginary filetype
[00:01:38]     (3/3) [EXECUTE] Previous errors should be removed when linters change
[00:01:38]     (3/3) [EXECUTE] (X) Vim(return):E684: list index out of range: 0
[00:01:38]       > function ale#engine#WaitForJobs[51]..<SNR>37_VimCloseCallback[14]..<SNR>34_HandleExit[44]..TestCallback, line 2
[00:01:38]   Success/Total: 1/3
```
Seem unrelated to this PR.

w0rp:
That's probably just a test that fails randomly some times.

w0rp:
Cheers! :beers:

PR #2224, opened 2019-01-19
Adjust column to be 0-based for LSP messages

andreypopp:
Per LSP spec column is zero based while vim's `getcurpos()` returns
1-based columns. This fixes it by making an LSP request with 0-based
column instead.

### DIFF
--- a/autoload/ale/lsp/message.vim
+++ b/autoload/ale/lsp/message.vim
@@ -3,6 +3,10 @@
 "
 " Messages in this movie will be returned in the format
 " [is_notification, method_name, params?]
+"
+" All functions which accept line and column arguments expect them to be 1-based
+" (the same format as being returned by getcurpos() and friends), those then
+" will be converted to 0-based as specified by LSP.
 let g:ale_lsp_next_version_id = 1
 
 " The LSP protocols demands that we send every change to a document, including
@@ -98,7 +102,7 @@ function! ale#lsp#message#Completion(buffer, line, column, trigger_character) ab
     \   'textDocument': {
     \       'uri': ale#path#ToURI(expand('#' . a:buffer . ':p')),
     \   },
-    \   'position': {'line': a:line - 1, 'character': a:column},
+    \   'position': {'line': a:line - 1, 'character': a:column - 1},
     \}]
 
     if !empty(a:trigger_character)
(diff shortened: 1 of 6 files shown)


w0rp:
Closely related to #2223

w0rp:
I'm not sure if this is correct. Can you share some example of a bug you found? ALE is already using 0-based columns by subtracting the 1-based columns from Vim by 1.

andreypopp:
@w0rp I only found it does so for lines here:
https://github.com/w0rp/ale/blob/c2c0c194723ea9f0a21e4acec4260e3938c6712a/autoload/ale/lsp/message.vim#L144
but not for columns, but this now suggests a better fix.

w0rp:
I don't understand. The LSP servers want to read a 0-based column. The column is 1-based in Vim. The code subtracts by `1` to send the right number. Maybe it might make sense to move the `- 1` part of the code to inside that function.

w0rp:
What helps with changes like these is first describing a bug you've found with a particular server, maybe with a screenshot, and explanation of how changing the code has fixed something concrete.

w0rp:
Ah sorry, I misread the code. Now it makes sense to me. Sorry for the confusion. I'm slightly red-green colorblind, so that can cause confusion sometimes.

andreypopp:
@w0rp no problem!

I check that with `flow lsp` too and it exhibits the same issue.

I'll update this PR to make an adjustment in `ale/lsp/message.vim` instead and include fixes for other messages too (hope it's ok to bundle those in the same PR).

w0rp:
In this case, I'll just merge this and get more people to try it. It's pretty hard to test this.

w0rp:
Okay, I'll wait for that.

andreypopp:
This is ready for review.

autoload/ale/lsp/message.vim
```diff
@@ -98,7 +102,7 @@ function! ale#lsp#message#Completion(buffer, line, column, trigger_character) ab
     \   'textDocument': {
     \       'uri': ale#path#ToURI(expand('#' . a:buffer . ':p')),
     \   },
-    \   'position': {'line': a:line - 1, 'character': a:column},
+    \   'position': {'line': a:line - 1, 'character': a:column - 1},
```
w0rp: I don't think this one works. I think for completion, you have to send the character after the character the cursor is on, so you tell the server you want to get a result like `adcdef` for `abc`, like so:

```
abc
   ^
```

We may what to decrement the column number here, but also increment it in `completion.vim`, which might be less confusing. I tried using completion with `rls` with and without this change, and it seems to work without this, and not with it.

andreypopp: I think you are right, I've been testing with `asyncomplete-ale` and `asyncomplete` seems to be doing the right thing (supplying cur column + 1) in its context so it worked for me. Fixing `completion.vim` then.

andreypopp: I went this the simplest/safest approach there, which fixes the position just for LSP, hope it's ok.

w0rp:
Cheers! :beers:

PR #2225, opened 2019-01-19
Pass --compiler flag to elm-test when linting 0.19 tests

ravicious:
This makes elm make linter work when elm is not installed globally.

### DIFF
--- a/ale_linters/elm/make.vim
+++ b/ale_linters/elm/make.vim
@@ -189,7 +189,10 @@ endfunction
 " Return the command to execute the linter in the projects directory.
 " If it doesn't, then this will fail when imports are needed.
 function! ale_linters#elm#make#GetCommand(buffer) abort
+    let l:executable = ale_linters#elm#make#GetExecutable(a:buffer)
     let l:root_dir = ale_linters#elm#make#GetRootDir(a:buffer)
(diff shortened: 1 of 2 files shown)


ale_linters/elm/make.vim
```diff
@@ -189,19 +189,31 @@ endfunction
 " Return the command to execute the linter in the projects directory.
 " If it doesn't, then this will fail when imports are needed.
 function! ale_linters#elm#make#GetCommand(buffer) abort
+    let l:executable = ale_linters#elm#make#GetExecutable(a:buffer)
     let l:root_dir = ale_linters#elm#make#GetRootDir(a:buffer)
+    let l:is_v19 = ale_linters#elm#make#IsVersionGte19(a:buffer)
+    let l:is_using_elm_test = l:executable =~ "elm-test$"
```
ravicious: In case where `ale_linters#elm#make#GetExecutable` is supposed to use `elm-test` but `elm-test` is unavailable, it falls back to `elm` (1f21eb0c422e8a). I don't know why exactly, presumably for the linter to still attempt to lint if somebody is editing files from within `/tests` folder?

`elm` doesn't support the `--compiler` flag and will crash if provided, so we have to check that first.

w0rp:
Cheers! :beers:

PR #2238, opened 2019-01-23
Use relative paths when previewing file locations

chaucerbao:
Absolute paths can be long and repetitive, and cause the more interesting details (in a file location preview) to be truncated. Instead, display file locations relative to the current working directory.

### DIFF
--- a/autoload/ale/references.vim
+++ b/autoload/ale/references.vim
@@ -122,7 +122,7 @@ function! ale#references#Find(...) abort
 
     if len(a:000) > 0
         for l:option in a:000
-            if l:option ==? '-relative'
+            if l:option is? '-relative'
                 let l:options.use_relative_paths = 1
             endif
         endfor
--- a/autoload/ale/references.vim
+++ b/autoload/ale/references.vim
@@ -1,5 +1,4 @@
 let s:references_map = {}
-let s:use_relative_paths = 0
 
 " Used to get the references map in tests.
 function! ale#references#GetMap() abort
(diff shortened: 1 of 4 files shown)


chaucerbao:
Linter is failing with:
```
autoload/ale/preview.vim:54 Do not use getcwd(), as it could run from the wrong buffer. Use expand('#' . a:buffer . ':p:h') instead.
```

But we (probably) don't want the results relative to specific buffer; we want it relative to the current working directory, which is precisely what `getcwd()` provides.

@w0rp is there a better way around this?

chaucerbao:
I added an optional **use_relative_paths** flag to the `ale#preview#ShowSelection()` function to enable the display of relative paths, defaulting to absolute paths.

```viml
" Absolute paths
ale#preview#ShowSelection([])
ale#preview#ShowSelection([], { 'use_relative_paths': 0 })

" Relative paths
ale#preview#ShowSelection([], { 'use_relative_paths': 1 })
```

But I'm not familiar with how to parse a `-relative` option from an ex-command, and then, the best way to pass that through multiple function calls to reach `ale#preview#ShowSelection()`

1. ale#references#Find()
2. s:FindReferences()
3. s:OnReady
    1. ale#references#HandleTSServerResponse
    2. ale#references#HandleLSPResponse
6. ale#preview#ShowSelection

autoload/ale/references.vim
```diff
@@ -115,7 +122,17 @@ function! s:FindReferences(linter) abort
...
+        for l:option in a:000
+            if l:option == '-relative'
+                let s:use_relative_paths = 1
+            endif
+        endfor
+    endif
+
     for l:linter in ale#linter#Get(&filetype)
         if !empty(l:linter.lsp)
             call s:FindReferences(l:linter)
```
w0rp: Instead of modifying a script variable, pass the options in here. Forward that on to the `OnReady` function. You can then save the options in `s:references_map`. You can write `let l:options = remove(s:references_map, a:response.id)` above to get the options out at the end.

chaucerbao: Thanks for pointing me in the right direction. I applied your solution and added tests.

Lemme know if there's anything else I can do to clean up (or correct) the code.

w0rp:
Cheers! :beers:

Issue #2132, opened 2018-12-09
Implement a uniform API for asynchronous processing for most ALE features


w0rp:
## Problem description

This issue outlines a design for how to add more asynchronous execution support
to ALE's codebase. Currently, ALE can run the following in the background.

* Jobs for getting linting results. (`callback`, `command_callback`)
* Jobs for figuring out how to build the commands for jobs. (`command_chain`)
* Jobs for fixing files. (`:ALEFix` functions)
* Jobs for chaining fixing files (`chain_with`)
* Sending messages to LSP servers or `tsserver`, etc.

ALE *does not* support asynchronous execution in the following common scenarios, at the time this issue was written.

* Figuring out how to build an executable path for a linter. (You can do this for a fixer at the moment, but not a linter.)
* Figuring out how to build a command to run for a language server.  (`command_chain` isn't currently implemented for LSP commands.)
* Running additional jobs when parsing linter results.

If ALE did add support for all the above, it would be possible to improve on the user experience in the following ways.

* ALE could discover valid execution paths for running linters in the background, so I/O never blocks editing in Vim. This isn't a problem for most people right now, but it **is** a problem for _some_ people, and that makes it a problem.
* ALE could support `--version` checks and more for LSP commands.
* ALE could support sending the results of one linter to another program for additional processing, where that program might be able to improve on the output in the background, without blocking Vim.
* There will be more uses I can't imagine, and that is a good thing.

## Proposal

All of the problems above can be solved with a uniform API for signaling when results are ready with a callback. A new function will be added which returns a special Dictionary representing a deferred object and accepts a callback. The callbacks for linters, fixers, and more will be able to return one of these deferred objects to indicate that something will be run in the background, and the result of the callback will be used as the eventual result.

Here is just one example of how this API could be used.

```vim
" This example is partly adapted from what the DMD linter does today.

function! s:ParseDubResults(buffer, output, metadata) abort
  " Do something with a:output here.

  return 'dmd ...'
endfunction

function! ale_linters#d#dmd#Command(buffer) abort
  " The result of this function is a deferred object that will be given to the 
  " ALE engine code. The callback below will be called later and will return
  " the eventual result.
  return ale#command#Run(
  \ a:buffer, 
  \ '... && dub describe --import-paths',
  \ function('s:ParseDubResults'),
  \)
endfunction
```

The `_callback` forms of options for `ale#linter#Define` will be deprecated and
later removed, and `command_chain` will be deprecated and removed now it can
[...]

PR #2241, opened 2019-01-24
Add a hook to detect LSP project root

bk2204:
Currently, we detect the linter root based on a variety of techniques. However, these techniques are not foolproof. For example, clangd works fine for many things without a compile_commands.json file, and Go projects may be built outside of the GOPATH to take advantage of Go 1.11's automatic module support.

Add a variable for the user to customize which points to a hook that detects the project root if the linter-specific configuration fails. Note that we allow only strings, since Vim requires that funcrefs are stored in variables that begin with an uppercase letter, and ALE's variables do not.

I fully expect discussion and revision of this PR, so comments are welcome.

### DIFF
--- a/autoload/ale/assert.vim
+++ b/autoload/ale/assert.vim
@@ -109,6 +109,14 @@ function! ale#assert#LSPProject(expected_root) abort
     AssertEqual a:expected_root, l:root
 endfunction
 
+function! ale#assert#LSPProjectFull(expected_root) abort
+    let l:buffer = bufnr('')
+    let l:linter = s:GetLinter()
+    let l:root = ale#lsp_linter#FindProjectRoot(l:buffer, l:linter)
+
+    AssertEqual a:expected_root, l:root
+endfunction
+
 function! ale#assert#LSPAddress(expected_address) abort
     let l:buffer = bufnr('')
     let l:linter = s:GetLinter()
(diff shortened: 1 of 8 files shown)


w0rp:
Another advantage is that later on when #2132 is done, we could support functions as values if we wanted to, which could themselves execute asynchronous code. I wouldn't worry about that right away.

w0rp:
I also recommend making it so the user's specified project path is _always_ preferred over whatever ALE is trying to figure out. That way you can override what ALE is trying to do by default, including skipping any potentially complex processing ALE might be doing, when you already know what the path should be.

bk2204:
Part of my goal here is to allow for global settings, such as always picking the top level of the Git repo as the project root. While there are merits to the design you've proposed, I think it lacks the ability to handle that case without specifically enumerating every possible LSP tool to use, which I would like to avoid.

Do you have any suggestions about a design that incorporates both the flexibility of your proposal with the ability to honor global settings of my original design? Would a `default` key be sufficient here in the dictionary, along with reading both `b:ale_lsp_root` and `g:ale_lsp_root`?

I'm fine with overriding the autodetection here, and I was considering that behavior myself.

bk2204:
Okay, I've updated the series to fit with that design. I additionally fixed a test that was missing a `name` entry in its linter dictionary, since we now rely on the fact that all linters have such an entry.

PR #2245, opened 2019-01-24
Add better ksh support

andrewimeson:
Ksh was not being detected properly for me by ALE

A file with this code would be run by ALE through `/bin/sh -n` and by shellcheck with `-s sh`, leading to it incorrectly complaining about process substitution being undefined in POSIX sh.

```ksh
#!/bin/ksh

diff -w <(find expected_logs/ -maxdepth 1 | sed 's/.expected//' | sort) <(find logs/ -maxdepth 1 | sed -E 's/\.[^.]+\.log$//' | sort)
```

I tested it locally and it fixes my issue.


w0rp:
Cheers! :beers:

PR #2248, opened 2019-01-25
Add tsserver as option for JavaScript files

ashleybaldock:
This pull request added support for this, but docs don't mention it - https://github.com/w0rp/ale/pull/1653


w0rp:
Needs to be in the other file too. I'll just add it to both.

PR #2266, opened 2019-02-01
Support pylama for python

kevinoid:
[pylama](https://github.com/klen/pylama) is a popular (>500 stars) plugin-based code audit tool for Python which includes plugins for several other popular tools.

The pylama ALE linter is based on existing python linters and attempts to adopt the same conventions and code as much as possible.

Thanks for considering,
Kevin

### DIFF
--- a/ale_linters/python/flake8.vim
+++ b/ale_linters/python/flake8.vim
@@ -74,15 +74,11 @@ let s:end_col_pattern_map = {
 \}
 
 function! ale_linters#python#flake8#Handle(buffer, lines) abort
-    for l:line in a:lines[:10]
-        if match(l:line, '^Traceback') >= 0
-            return [{
-            \   'lnum': 1,
-            \   'text': 'An exception was thrown. See :ALEDetail',
-            \   'detail': join(a:lines, "\n"),
-            \}]
-        endif
-    endfor
+    let l:output = ale#python#HandleTraceback(a:lines, 10)
+
+    if !empty(l:output)
+        return l:output
+    endif
 
     " Matches patterns line the following:
     "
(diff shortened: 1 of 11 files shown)


RyanSquared:
Just to clarify, a "Traceback" message is a possible flaw in the linter itself and not something like a syntax error?

kevinoid:
> Just to clarify, a "Traceback" message is a possible flaw in the linter itself and not something like a syntax error?

A traceback is printed whenever a linter plugin encounters an unhandled exception.  (See [pylama/core.py:85-87](https://github.com/klen/pylama/blob/79aaaba/pylama/core.py#L85-L87))  The unhandled exception could be due to an unexpected internal error or a syntax error severe enough that the linter plugin can't analyze the file.

For example, the pydocstyle plugin throws `ParseError` when `from __future__ import with_statement` occurs after a non-`__future__` import, which is what I used to generate the [traceback test case](https://github.com/w0rp/ale/blob/849389cc2b8a7f391da06976e8025c4c4730310c/test/handler/test_pylama_handler.vader#L105-L121).

Does that answer your question?

RyanSquared:
If this is a common case, I think there should be a better way to solve it than hide the message behind `:ALEDetail`, such as putting only the `repr`-like format of the error as the message, *then* mentioning `:ALEDetail`.

kevinoid:
I updated the pull request with an additional commit that consolidates the traceback parsing between the Python linters and uses the exception line (with the type and message) as the ALE text.  Let me know what you think.

w0rp:
Cheers! :beers:

kevinoid:
Thanks @w0rp!  :clinking_glasses:

PR #2270, opened 2019-02-03
linter/cypher: add cypher-lint

oblitum:


oblitum:
The `cypher-lint` tool supports input from `stdin`, I'm not sure whether that can be used for instant linter, initially I just provided linting on file save, passing a file as parameter. If it can be changed for instant linting by leveraging stdin, please tell me what I need to change.

PR #2277, opened 2019-02-06
Center the result after `ale#util#Open` jumps

chaucerbao:
In certain cases, when I use `:ALEGoToDefinition` (or other jumping call) and it leads to a location in the same file, the cursor moves, but the screen does not redraw.  Then, when I use a motion movement or wait a few seconds, the screen will finally update. This may be due to Vim's `lazyredraw` setting.

This PR forces a redraw after the cursor moves.

### DIFF
--- a/autoload/ale/util.vim
+++ b/autoload/ale/util.vim
@@ -111,6 +111,7 @@ function! ale#util#Open(filename, line, column, options) abort
     endif
 
     call cursor(a:line, a:column)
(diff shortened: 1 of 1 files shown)


chaucerbao:
Okay, well since the `zz` is a movement which also causes a redraw, I removed the `redraw!`

w0rp:
Cheers! :beers:

Issue #2198, opened 2019-01-11
Adding ALENextError which would ignore warnings


Lattay:
Would it be possible to append a command ALENextError which would do just as ALENext but ignoring warnings.
I think it can be really interesting when editing big old file from someone else with a lot of bad style practices.
For exemple I have some python files thousands lines longs, full of bad spaces and visual indentations that make flake8 crying. I have ~500 warnings for only 4 errors (which are just variable not use probably) and I do not want to fix every style problem but I would like to fix these errors but it is pretty hard to find them here.
By the way I think if you had to fix an old file you would always start fixing errors before warnings.

w0rp:
Good idea. It might make sense to support `:ALENext` with options instead of a new command. Either way, we'd need more plug mappings.

Lattay:
I decided to give it a try. It works fine though I have to write tests I guess.

PR #2279, opened 2019-02-06
Supporting filtered jump

Lattay:
This pull request address #2198.
I used extra argument to generalized `ale#loclist_jumping#Jump` and `ale#loclist_jumping#FindNearest` to the use of a filter on the type of error. You can now use `:ALENext 1 E` to jump to the next error ignoring warnings with wrapping. I added the corresponding <plug> binding to ease creating of custom shortcut.

PS: This is my first attempt at coding in a Vim plugin and at making a pull request, so I may have forgotten something but I would be glad to fix anything not orthodox.

### DIFF
--- a/autoload/ale/loclist_jumping.vim
+++ b/autoload/ale/loclist_jumping.vim
@@ -20,7 +20,13 @@ function! ale#loclist_jumping#FindNearest(direction, wrap, ...) abort
     if a:0 > 0
         let l:filter = a:1
     else
-        let l:filter = 'A'
+        let l:filter = 'any'
+    endif
+
+    if a:0 > 1
+        let l:subtype_filter = a:2
+    else
+        let l:subtype_filter = 'any'
     endif
 
     " When searching backwards, so we can find the next smallest match.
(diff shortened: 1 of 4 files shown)


Lattay:
Ok, I think that flag combinations are better than lots of specialized flags so I propose the following:
* default behavior is no wrapping, all types and all subtypes
* a flag -wrap enable wrapping
* three flags -error, -warning and -info (mutually exclusives) enable type filtering
* two flags -style and -nostyle (mutually exclusives) enable subtype filtering
The three kind of flags can be used in any order.
A wrapper prepare the arguments of ale#loclist_jumping#Jump from the list of flags and call the function.

w0rp:
Sounds good to me. Keep the existing commands around, and we can add deprecation warnings for them.

w0rp:
Check out 19cc724807de4cdf9f8147c70932b49a79a7f7ee and examples in `test/test_parse_command_args.vader`. I have now added a function for parsing command arguments we can use everywhere.

Lattay:
> Check out [19cc724](https://github.com/w0rp/ale/commit/19cc724807de4cdf9f8147c70932b49a79a7f7ee) and examples in `test/test_parse_command_args.vader`. I have now added a function for parsing command arguments we can use everywhere.

I can use ale#args#Parse but I will have the same content in the wrapper: a loop over the list of flags that modify the arguments to be passed.
So I think using it will only duplicate the logic.

By the way I have now updated \<plug\> bindings and tests as well as documentation and I think it may be ready.
What do you think ?

w0rp:
Use the new function to parse the arguments.

Lattay:
OK, done @w0rp. The behavior in case of multiple exclusive flags is a bit less intuitive but since it should not happen normally it is not a problem.

w0rp:
Cheers! :beers:

PR #2287, opened 2019-02-10
Document and test ale_python_vulture_options

kevinoid:
The Vulture linter already supports `ale_python_vulture_options`, but it is not documented or tested.  Since Vulture only supports configuration via options, it is an important use case.  Add docs and test.

Cheers,
Kevin

### DIFF
--- a/doc/ale-python.txt
+++ b/doc/ale-python.txt
@@ -671,6 +671,15 @@ g:ale_python_vulture_executable               *g:ale_python_vulture_executable*
   See |ale-integrations-local-executables|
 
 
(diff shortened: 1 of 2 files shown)


kevinoid:
Apparently the build failed on Travis for `--neovim-only`.  However, I don't see any failure messages in the output and the build [passed on my repo](https://travis-ci.org/kevinoid/ale/builds/491265558) before submission.  Can the failure be ignored (or the build restarted)?

w0rp:
It just fails randomly some times. Cheers! :beers:

kevinoid:
No worries.  Good to know.  Thanks again @w0rp!  :beers:

PR #2289, opened 2019-02-10
Support ale_sh_shellcheck_dialect to set shellcheck dialect

kevinoid:
As discussed in w0rp/ale#1051, there are cases (including my own) where it would be useful to be able to specify the dialect explicitly or allow `shellcheck` to determine the dialect using its own logic and considering [`# shellcheck shell=` directives](https://github.com/koalaman/shellcheck/wiki/Directive#shell) in the file.  This commit allows users to do so using the `ale_sh_shellcheck_dialect` variable, which can be set empty to disable auto-detection or set to a particular dialect to use it unconditionally.

Fixes: w0rp/ale#1051

Thanks for considering,
Kevin


w0rp:
Yep, you can sometimes figure out something better than a machine can.

w0rp:
Cheers! :beers:

kevinoid:
Yes indeed.  Thanks @w0rp! :beers:

PR #2291, opened 2019-02-12
pylama: Use %s instead of %t

kevinoid:
My apologies for yet another Pylama pull request.  After more testing, I think my decision to use `%t` (and suppressing C0103) was the wrong approach.  I am finding more errors which depend on the real path and surrounding files (e.g. `E0401 - Unable to import %s`, `E0402 - Attempted relative import beyond top-level package`).  I suspect there will be even more from Pylama plugins that I haven't used and that switching to `%s` is the better choice.  Thoughts?

Thanks for considering,
Kevin

### DIFF
--- a/ale_linters/python/pylama.vim
+++ b/ale_linters/python/pylama.vim
@@ -34,10 +34,13 @@ function! ale_linters#python#pylama#GetCommand(buffer) abort
     \   ? ' run pylama'
     \   : ''
 
+    " Note: Using %t to lint changes would be preferable, but many pylama
+    " checks use surrounding paths (e.g. C0103 module name, E0402 relative
+    " import beyond top, etc.).  Neither is ideal.
     return l:cd_string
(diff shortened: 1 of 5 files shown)


w0rp:
Add `'lint_file': 1` to the linter so it only runs when the file is saved to disk, and update the lists of supported tools to indicate that.

kevinoid:
Good catch!  Sorry I missed those (and for the noise repeatedly amending/rebasing the commit).  Should be fixed now.

w0rp:
Don't forget to add a floppy disk to the Markdown version too. :floppy_disk:

kevinoid:
Good catch!  Your CI checks are excellent!  Should be fixed now.

w0rp:
Cheers! :beers:

kevinoid:
Thanks again @w0rp! :clinking_glasses:

PR #2303, opened 2019-02-17
python/bandit: Use .bandit configuration file

kevinoid:
Bandit automatically [uses any `.bandit` file](https://github.com/PyCQA/bandit/blob/1.5.1/bandit/cli/main.py#L70-L73) within the directories on which it is invoked.  Since ALE invokes `bandit` on `-`, it does not load a `.bandit` file automatically.  Add support for automatically finding a `.bandit` file and passing it to bandit via the `--ini` option along with a variable to disable this behavior if desired.

Note: This is useful for the `skips` and `tests` configuration options, but not `exclude` which would require invoking `bandit` using a file name, which may or may not be a good trade-off.

Thanks for considering,
Kevin

### DIFF
--- a/ale_linters/python/bandit.vim
+++ b/ale_linters/python/bandit.vim
@@ -3,6 +3,7 @@
 
 call ale#Set('python_bandit_executable', 'bandit')
 call ale#Set('python_bandit_options', '')
+call ale#Set('python_bandit_use_config', 1)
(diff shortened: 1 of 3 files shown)


w0rp:
Cheers! :beers:

kevinoid:
Thanks again @w0rp!  :clinking_glasses:

PR #2305, opened 2019-02-18
javac: Don't assume src/main/java always exists

campbellr:
Some projects I have only have src/test/java (no /src/main/java), which
seemed to break some incorrect assumptions the linter had.


campbellr:
Sure, I'll try to add a test.

campbellr:
@w0rp I finally added a test that I believe covers this (sorry it took so long).

Let me know if there's anything else.

w0rp:
Cheers! :beers:

PR #2307, opened 2019-02-20
Clear the virtualtext on disable

lukas-reineke:
Currently if virtualtext is displayed when ale gets disabled with `ALEDisable` or `ALEDisableBuffer` it does not get cleared.

This PR clears the text.

I didn't write tests, because I don't know how I can test that `nvim_buf_clear_highlight` was called.


w0rp:
Cheers! :beers:

PR #2309, opened 2019-02-21
Fix syntax errors when using eruby with graphql-client gem

technicalpickles:
[graphql-client](https://github.com/github/graphql-client) has this interesting feature where you can do inline
[graphql in ERB](https://github.com/github/graphql-client#rails-erb-integration):

> These <%graphql sections are simply ignored at runtime but make their definitions available through constants. The module namespacing is derived from the .erb's path plus the definition name.

The problem is this isn't actually valid ERB, similar to how Rails has
it's own idiosyncrasies. https://github.com/w0rp/ale/issues/580 covers when this came up for Rails

This adds a gsub similar to Rails, but for the graphql helper, like it
does internally: https://github.com/github/graphql-client/blob/116bb3ea2c2abac13ce9ed95a99d1cf98ccc201e/lib/graphql/client/erubis_enhancer.rb#L17

### DIFF
--- a/ale_linters/eruby/erb.vim
+++ b/ale_linters/eruby/erb.vim
@@ -11,7 +11,10 @@ function! ale_linters#eruby#erb#GetCommand(buffer) abort
     " Rails-flavored eRuby does not comply with the standard as understood by
     " ERB, so we'll have to do some substitution. This does not reduce the
     " effectiveness of the linter—the translated code is still evaluated.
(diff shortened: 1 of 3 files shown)


technicalpickles:
I did a little more manual testing, and it looks like this needs some more work (plus automated tests). This PR works when there aren't syntax errors, but when there are, the line numbers are off by the number of lines of in the graphql query.

w0rp:

PR #2310, opened 2019-02-21
Detect notes in mypy's messages

delphinus:
`mypy` can show some useful advices with the `note` level. But with the default regexp, `note`'s are all ignored. This patch add an option to show them in addition `errors`'s and `warning`'s.

Here's an example. If you write code below,

```python
from typing import cast

def double(value: int) -> int:
    doubled = value * 2
    return cast(int, doubled)
```

`mypy` can warn the redundant `cast`.

```sh
$ mypy --warn-redundant-casts test.py
test.py:6: note: Redundant cast to "int"
```

You can catch this with `let g:ale_python_mypy_detect_notes = 1`.

### DIFF
--- a/test/handler/test_mypy_handler.vader
+++ b/test/handler/test_mypy_handler.vader
@@ -1,7 +1,9 @@
 Before:
   Save g:ale_python_mypy_ignore_invalid_syntax
+  Save g:ale_python_mypy_detect_notes
 
   unlet! g:ale_python_mypy_ignore_invalid_syntax
+  unlet! g:ale_python_mypy_detect_notes
 
   runtime ale_linters/python/mypy.vim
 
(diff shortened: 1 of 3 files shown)


RyanSquared:
Looks nice. I'll check it out tomorrow - haven't used mypy but I have some Python projects filled with linter testing fuel.

test/handler/test_mypy_handler.vader
```diff
@@ -69,6 +71,67 @@ Execute(The mypy handler should parse lines correctly):
...
+  \     'type': 'E',
+  \     'text': 'Argument 1 to "somefunc" has incompatible type "int"; expected "str"'
+  \   },
+  \   {
+  \     'lnum': 14,
+  \     'col': 0,
+  \     'filename': ale#path#Simplify(g:dir . '/another_module/bar.py'),
+  \     'type': 'I',
+  \     'text': 'In module imported here,',
+  \   },
```
w0rp: The notes provide additional information about previous problems, so we should instead include the lines in the `detail` key of the error above instead. Then using `:ALEDetail` or moving your mouse over an error will show you some more information about the error.

w0rp:

PR #2319, opened 2019-02-26
WIP: Add support for textDocument/codeLens

andreypopp:
This adds implementation of `textDocument/codeLens` request.

The result is rendered through neovim's virtualtext API and thus it's only usable with neovim.

codelens

TODO:
- [ ] Tests (I know!)
- [x] Commands to enable/disable/toggle (also per buffer variants)

### DIFF
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -109,7 +109,10 @@ function! s:Lint(buffer, should_lint_file, timer_id) abort
     \   || (a:should_lint_file && filereadable(expand('#' . a:buffer . ':p')))
 
     call ale#engine#RunLinters(a:buffer, l:linters, l:lint_file)
-    call ale#code_lens#Request(a:buffer)
+
+    if get(g:, 'ale_code_lens_enabled', 0)
+        call ale#code_lens#Request(a:buffer)
+    endif
 endfunction
 
 " (delay, [linting_flag, buffer_number])
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -109,6 +109,7 @@ function! s:Lint(buffer, should_lint_file, timer_id) abort
     \   || (a:should_lint_file && filereadable(expand('#' . a:buffer . ':p')))
 
     call ale#engine#RunLinters(a:buffer, l:linters, l:lint_file)
+    call ale#code_lens#Request(a:buffer)
 endfunction
 
 " (delay, [linting_flag, buffer_number])
--- a/autoload/ale/code_lens.vim
+++ b/autoload/ale/code_lens.vim
@@ -17,10 +17,10 @@ endfunction
 " TODO: maybe check with exists instead?
 let s:supported = exists('*nvim_buf_set_virtual_text')
 
-let s:enabled = 0
+let g:ale_code_lens_enabled = 0
 
 function! ale#code_lens#Toggle() abort
-    if s:enabled
+    if g:ale_code_lens_enabled
         call ale#code_lens#Disable()
     else
         call ale#code_lens#Enable()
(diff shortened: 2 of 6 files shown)


autoload/ale.vim
```diff
@@ -109,6 +109,7 @@ function! s:Lint(buffer, should_lint_file, timer_id) abort
     \   || (a:should_lint_file && filereadable(expand('#' . a:buffer . ':p')))
 
     call ale#engine#RunLinters(a:buffer, l:linters, l:lint_file)
+    call ale#code_lens#Request(a:buffer)
```
andreypopp: I'm not sure this is the right way — what I need is to make requests in the same cases we do requests diagnostics (also after we synced changed to LSP).

autoload/ale/code_lens.vim
```diff
@@ -0,0 +1,185 @@
...
+function! ale#code_lens#SetMap(map) abort
+    let s:code_lens_map = a:map
+endfunction
+
+function! ale#code_lens#ClearLSPData() abort
+    let s:code_lens_map = {}
+endfunction
+
+" TODO: maybe check with exists instead?
+let s:supported = exists('*nvim_buf_set_virtual_text')
```
andreypopp: You have `has('nvim-0.3.2')` in `virtualtext.vim`, why?

autoload/ale/code_lens.vim
```diff
@@ -0,0 +1,185 @@
...
+    if !ale#code_lens#IsEnabled(a:buffer)
+        return
+    endif
+
+    " Set a flag so we only make one request.
+    call setbufvar(a:buffer, 'ale_code_lens_request_made', 0)
+
+    for l:linter in ale#linter#Get(getbufvar(a:buffer, '&filetype'))
+        if !empty(l:linter.lsp) && l:linter.lsp isnot# 'tsserver'
+            call s:Request(l:linter, a:buffer)
```
andreypopp: This is incorrect now as requests to different LSPs will override each others codelens, I'm not sure though if we need to support the case with multiple LSPs + codelens, thoughts?

autoload/ale.vim
w0rp: Check if the lens feature is enabled before calling functions from `code_lens.vim`, and then the code won't be loaded for users who don't turn the feature on.

andreypopp:
Thanks! Addressed first round of feedback.

autoload/ale/code_lens.vim
w0rp: I can't remember, but I think that was the minimum version requirement due to changes for the functions.

autoload/ale/code_lens.vim
w0rp: You can either combine the results of multiple servers, or ignore responses from other servers. Note that if you choose to ignore responses, you should make a request to all available servers, but keep responses from only one.

autoload/ale.vim
```diff
@@ -109,7 +109,10 @@ function! s:Lint(buffer, should_lint_file, timer_id) abort
     \   || (a:should_lint_file && filereadable(expand('#' . a:buffer . ':p')))
 
     call ale#engine#RunLinters(a:buffer, l:linters, l:lint_file)
-    call ale#code_lens#Request(a:buffer)
+
+    if get(g:, 'ale_code_lens_enabled', 0)
```
w0rp: You might want to use `ale#Var(a:buffer, 'code_lens_enabled)` here, so the code lens feature can be enabled for a single buffer, while it's disabled globally.

andreypopp: That's useful, thank you!

Do you know how we can guarantee codelens request is sent after didChange notification is sent to server?

w0rp: The easiest way would be to always attempt to send a message about the buffer being changed before every codeLens request, and then the check for `b:changedtick` will prevent the message from being sent more than it needs to be.

w0rp:

PR #2323, opened 2019-03-01
Use new cabal exec command

danieltdt:
Cabal [build has changed](https://cabal.readthedocs.io/en/latest/nix-local-build-overview.html).
Using cabal v2.4.0.0, the current command `cabal exec` prints a deprecation warning like this:

```
Warning: The exec command is a part of the legacy v1 style of cabal usage.

Please switch to using either the new project style and the new-exec command
or the legacy v1-exec alias as new-style projects will become the default in
the next version of cabal-install. Please file a bug if you cannot replicate a
working v1- use case with the new-style commands.

For more information, see: https://wiki.haskell.org/Cabal/NewBuild
```

This PR just apply the suggestions from that warning but I think we
should consider expose this command as a configurable variable.


danieltdt:
Cool. I'll check it out and update this PR.

PR #2325, opened 2019-03-02
set correct Rust span end column

JeanMertz:
The Rust compiler returns the first column that is _not_ part of the
current span as `column_end`, while Ale expects `end_col` to signify
the last column of the span.

**Compiler message**

screenshot 2019-03-02 at 16 06 24

**Ale (before)**

screenshot 2019-03-02 at 16 06 06

**Ale (after)**

screenshot 2019-03-02 at 16 12 49


w0rp:
Cheers! :beers:

Issue #2318, opened 2019-02-26
clang-format for nvcc (cuda files)


akshit-sharma:
Clang-format currently works with c and cpp files.

I tried running in the terminal and it also works with cuda files. 

Ale as far as I saw only supports nvcc on cuda files. 
Making clang-format also work with cuda files would be helpful.

akshit-sharma:
Added in #2327.

PR #2327, opened 2019-03-03
clang-format supported with nvcc (cuda files)

akshit-sharma:
Added support for clang-format to be used with nvcc (cuda files). Issue #2318 

Just as ale-cpp-clangformat uses ale-c-clangformat for options. Similarly the ale-cuda-clangformat uses ale-c-clangformat for options.


w0rp:
Cheers! :beers:

PR #2332, opened 2019-03-05
Add lsp_config support for rls

n8henrie:
This is a first draft, based on how it was implemented for pyls. Still needs
tests. My first contribution to a vim project, so I'm still sorting things out.

Currently *not* working with `clippy_preference`, which is the main reason I'm interested in adding the feature.


w0rp:
Cheers! :beers:

PR #2334, opened 2019-03-06
change default action when can't find the root path for go's bingo linter

chenjie199234:
The bingo linter has a bug.
In default,when the FindProjectRoot function can't find the root path in ale_linters/go/bingo.vim,it will return empty,and this will make bingo stop.
This pull request change the default action.When can't find the root path like before,it will not return empty but use current buffer's dir as root path.Then,the bingo can start at anytime.
There is still another problem with bingo linter.
The hint message show's "Token file does not exist..."
I don't known how to do with this.

### DIFF
--- a/ale_linters/go/bingo.vim
+++ b/ale_linters/go/bingo.vim
@@ -17,7 +17,24 @@ function! ale_linters#go#bingo#FindProjectRoot(buffer) abort
         let l:mods = ':h:h'
     endif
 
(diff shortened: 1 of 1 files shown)


w0rp:
I think the root detection here could cause problems for some people. I recommend using the `b:ale_lsp_root` option instead to set the project path appropriately for your projects. See `:help b:ale_lsp_root`.

PR #2337, opened 2019-03-06
make options for credo configurable

### LINKED ISSUE #2335: Make options for credo configurable
Currently in file the [credo linter](https://github.com/w0rp/ale/blob/master/ale_linters/elixir/credo.vim) it is hard coded to use `mix credo suggest --format=flycheck --read-from-stdin %s`.  It would be great if we could set an option like `let ale_elixir_credo_strict_mode=1` this would default to 0 to not change how folks are using credo currently.

This will then give give back additional messages regarding software design (D) and convert it to (I) for ALE to display.
```
test/support/conn_helpers.ex:1:11: R: Modules should have a @moduledoc tag.
test/support/data_case.ex:82:5: D: Nested modules could be aliased at the top of the invoking module.
```
I will look into doing this unless someone beats me to it.  I have very limited knowledge of vim programming, but what a good way to learn.

Thoughts?

adam-e-trepanier:
Add credo --strict option

If a user sets 'let g:ale_elixir_credo_strict=1' it will run credo with    --strict instead of suggest.  The default (0) is to run as suggest.

ToC and doc file updated however I can't seem to get the tests to pass for the ToC, not sure what I am doing wrong.  Any help would be appreciated.

Closes #2335


adam-e-trepanier:
Found this issue.  @w0rp please let me now if you require any changes.

ale_linters/elixir/credo.vim
```diff
@@ -40,8 +40,15 @@ endfunction
...
     let l:project_root = ale#handlers#elixir#FindMixProjectRoot(a:buffer)
 
-    return ale#path#CdString(l:project_root)
-    \ . ' mix help credo && mix credo suggest --format=flycheck --read-from-stdin %s'
+    if get(g:, 'ale_elixir_credo_strict', 0)
+        let l:credo_cmd = ' mix help credo && mix credo --strict --format=flycheck --read-from-stdin %s'
+
+        return ale#path#CdString(l:project_root) . l:credo_cmd
+    else
+        let l:credo_cmd = ' mix help credo && mix credo suggest --format=flycheck --read-from-stdin %s'
```
w0rp: Instead of replacing the whole string, replace just the one part of it we need to change based on the setting. Have a look at some of the other linter files.

adam-e-trepanier: On it.  adding tests also.

adam-e-trepanier: @w0rp Fixed.  Moved option generation to function to return correct value and concat that into the string command.  Also added `tests/command_callback/test_elixir_credo.vader` file.  I had a hell of a time getting it to work as you can see from the commits.  Please let me know if you have suggestions to make it better/correct, first Vader test I have ever written.

w0rp:
Cheers! :beers:

PR #2340, opened 2019-03-07
make `gotype` return all errors

svanharmelen:
When using `gotype` without the `-e` option, it will only output the first 10 errors. When working on a larger package that often means that those 10 errors are in other files then the one that you are currently working on which then seems to indicate that there are no errors.

By adding the `-e` flag, all errors will be returned and shown properly in the file that you are working on.

### DIFF
--- a/ale_linters/go/gotype.vim
+++ b/ale_linters/go/gotype.vim
@@ -6,7 +6,7 @@ function! ale_linters#go#gotype#GetCommand(buffer) abort
         return ''
     endif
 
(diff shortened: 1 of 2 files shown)


svanharmelen:
@w0rp just updated the tests to match the updated command. Sorry for the delay on that!

w0rp:
Cheers! :beers:

Sorry for the delay. I was busy and on holiday.

svanharmelen:
No worries! Thanks 👍

PR #2355, opened 2019-03-12
 Fix parsing the third part of version string 

przepompownia:


przepompownia:
Reason: vint does not work after setting `color: true` in `~/.vintrc` because `ale` cannot add `--no-color` to command line (version `0.3.19` was recognized as `0.3.1`, which is too early to using this option).

w0rp:
Cheers! :beers:

PR #2357, opened 2019-03-13
Fixed URL decode on Windows

aldn:
l:encoded_path does not contain ':' so regexp match will fail and substitute() never called. The end result thus would still be '/C:/foo/bar' that the initial code presumably wanted to fix.

### DIFF
--- a/autoload/ale/path.vim
+++ b/autoload/ale/path.vim
@@ -205,10 +205,12 @@ function! ale#path#FromURI(uri) abort
         let l:encoded_path = a:uri
     endif
 
+    let l:decoded_path = ale#uri#Decode(l:encoded_path)
+
     " If the path is like /C:/foo/bar, it should be C:\foo\bar instead.
-    if l:encoded_path =~# '^/[a-zA-Z]:'
-        let l:encoded_path = substitute(l:encoded_path[1:], '/', '\\', 'g')
+    if l:decoded_path =~# '^/[a-zA-Z]:'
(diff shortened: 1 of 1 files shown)


w0rp:
I don't think this is right. I think `file:///C%3A/foo/bar/baz.tst` should be converted to `/C:/foo/bar/baz.tst`. Otherwise it's not possible to represent `/C:/foo` in Linux.

aldn:
At the moment GoToDefinition is broken on Windows because /C:/foo/bar is not a valid file (vim creates an empty buffer).
Judging by the comment the initial code was there to exactly adress this issue:
```
    " If the path is like /C:/foo/bar, it should be C:\foo\bar instead.
```
If we want to have equally supported /C:/foo/bar on Linux and C:\foo\bar on Windows then the URL decoder might need to have additional platform checks.

w0rp:
The URI received should be `file:///C:/foo/bar/baz.tst`, which will be translated to `C:\foo\bar\baz.tst`. The colon in the URI should not be encoded.

aldn:
With ccls I am receiving file:///C%3A/Users/....
Checked by hacking into `ale#definition#HandleLSPResponse` `l:item.uri`

w0rp:
Open a bug for that project and get them to output URIs with the colon in them. They shouldn't be encoding the colon in the drive letter part of the path.

aldn:
Ok thanks

w0rp:
On second thought, I'll just accept it and add a platform check. They should still output a colon without encoding it.

w0rp:
Done.

PR #2364, opened 2019-03-15
Fix ale_linters#php#langserver#GetProjectRoot() 

przepompownia:
At this moment it is only a start point to discussion.


w0rp:
See the issue for my comments.

PR #2367, opened 2019-03-16
WIP: Implement API to query for document symbols

andreypopp:
This only exposes a single function

```
ale#document_symbol#List(callback)
```

which queries first available LSP with `testDocument/documentSymbol`
request.

@w0rp Do you think ALE could expose just LSP methods and not UI? I have
PoC quality FZF bindings to that which I'm using. I can see how this can
be inline with exposing completion to deoplete (and other completoers).

TODO:

- [ ] Tests
- [ ] Docs
- [ ] Make sure callback is called even if no servers are available or
      no server support the request.
- [ ] Provide sync API

### DIFF
--- a/autoload/ale/document_symbol.vim
+++ b/autoload/ale/document_symbol.vim
@@ -1,0 +1,55 @@
+" Author: Andrey Popp <8mayday@gmail.com>
+" Description: Document Symbol support for LSP linters.
+
+let s:document_symbol_map = {}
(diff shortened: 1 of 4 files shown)


w0rp:
I think we should do both. Offer some basic way to display information, and also some functions for doing more complex things with the results.

w0rp:

PR #2377, opened 2019-03-21
Add a linter for clojure using clj-kondo

liquidz:
A minimal and opinionated linter for Clojure code that sparks joy.
https://github.com/borkdude/clj-kondo


w0rp:
Cheers! :beers:

PR #2387, opened 2019-03-24
Add support for latexindent

riley-martine:
Fist time writing vimscript and vader, so check real close that I actually did the right thing. Works to run latexindent on my machine, and tests pass.


w0rp:
Cheers! :beers:

PR #2394, opened 2019-03-26
feat: fecs support for js/html/css lint and format

harttle:
FECS is a lint tool for checking and fixing HTML/CSS/JavaScript code styles. I've been using the modified version locally, it works just fine. But I find that the ale-dev doc says:

> ALE should check files with as many tools as possible by default, except where they cause security issues or make excessive use of resources on modern machines.

I would like to make it public. It's widely used here, hope the following links helpful:

Github: https://github.com/ecomfe/fecs
Wiki: https://github.com/ecomfe/fecs/wiki/CLI

I passed the tests locally but there's some warnings for the document checking, but it seems not related. BTW, thank you guys for this great repo, I migrate from syntastic with no regret.

### DIFF
--- a/autoload/ale/fixers/fecs.vim
+++ b/autoload/ale/fixers/fecs.vim
@@ -1,9 +1,6 @@
 " Author: harttle <yangjvn@126.com>
 " Description: Apply fecs format to a file.
 
-call ale#Set('html_fecs_executable', 'fecs')
-call ale#Set('html_fecs_use_global', get(g:, 'ale_use_global_executables', 0))
-
 function! ale#fixers#fecs#Fix(buffer) abort
     let l:executable = ale#handlers#fecs#GetExecutable(a:buffer)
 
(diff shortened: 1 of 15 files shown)


autoload/ale/fixers/fecs.vim
```diff
@@ -0,0 +1,17 @@
+" Author: harttle <yangjvn@126.com>
+" Description: Apply fecs format to a file.
+
```
w0rp: You'll need to set defaults for the options here too, as the fixer might be used by the linter is loaded.

doc/ale.txt
```diff
@@ -49,6 +49,7 @@ ALE supports the following key features for linting:
 6. Using |echo| to show error messages when the cursor moves.
 7. Setting syntax highlights for errors.
 
+
```
w0rp: You can remove the blank line here.

autoload/ale/fixers/fecs.vim
harttle: Note that I called `ale#handlers#fecs#GetExecutable` instead of  `ale#node#FindExecutable`, and the defaults are already set in the `autoload/ale/handlers/fecs.vim`. So it should be OK I guess? Or do I miss something?

doc/ale.txt
harttle: Got it

harttle:
along with `test/handler/test_fecs_handler.vader`, I added `test/fixers/test_fecs_fixer_callback.vader` and `test/command_callback/test_fecs_command_callback.vader` for format and check respectively.

autoload/ale/fixers/fecs.vim
w0rp: You are right, that should work.

w0rp:
Cheers! :beers:

PR #2406, opened 2019-03-30
Pyflakes is supported + FAQ update

mafrosis:
Just came across the project so added a couple of things that would have been helpful for me :)

I was going to add "How do I enable particular linters?", but it meant re-indexing the whole FAQ and the answer is basically the same as the "disable linters" question.. Let me know if you think it might be worthwhile and I'll add the commit


w0rp:
Cheers! :beers:

PR #2410, opened 2019-04-01
Copy paste sbtserver to java fixes #2404

### LINKED ISSUE #2404: The sbtserver linter should work for java as well
Use case is for mixed java/scala projects. I'll try to get around to doing this eventually, should be effectively the same as when I added sbtserver for scala. Should I approach it as a copy/paste style task, or is there a technique for using the same linter between different languages that I should use?

(Also I'm curious if anyone uses the sbtserver feature besides me hah)

Ophirr33:
Is there a better way to add this besides copy/pasting the scala one?


w0rp:

PR #2411, opened 2019-04-02
Call rstcheck with --config option

### LINKED ISSUE #2144: Implement ale_rst_rstcheck_options
The rstcheck linter supports several command line options that allow controlling the report:
```
> rstcheck --help
usage: rstcheck [-h] [-r] [--report level] [--ignore-language language]
                [--ignore-messages messages] [--ignore-directives directives]
                [--ignore-substitutions substitutions] [--ignore-roles roles]
                [--debug] [--version]
                files [files ...]

Checks code blocks in reStructuredText. Sphinx is enabled.

positional arguments:
  files                 files to check

optional arguments:
  -h, --help            show this help message and exit
  -r, --recursive       run recursively over directories
  --report level        report system messages at or higher than level; info,
                        warning, error, severe, none (default: info)
  --ignore-language language, --ignore language
                        comma-separated list of languages to ignore
  --ignore-messages messages
                        python regex that match the messages to ignore
  --ignore-directives directives
                        comma-separated list of directives to ignore
  --ignore-substitutions substitutions
                        comma-separated list of substitutions to ignore
  --ignore-roles roles  comma-separated list of roles to ignore
  --debug               show messages helpful for debugging
  --version             show program's version number and exit
```
I would find it useful if there were an option such as  g:ale_rst_rstcheck_options to allow usage of these options.

### LINKED ISSUE #2408: rstcheck cannot use config files
As described  in https://github.com/myint/rstcheck/issues/56, ALE cannot use any config files for `rstcheck`. As far as I can tell, this is a shortcoming in `rstcheck` itself, and I've proposed the addition of a `--config` option to `rstcheck`. If accepted, `ale_linters/rst/rstcheck.vim` would need a minor modification to make use of that new option.

Enabling the proper use of config files for `rstcheck` would also would also address #2144: Any of the options that could be passed via a command line flag can also be set in a config file (and a config file is probably a better solution as it allows to make these settings problem-specific).

goerz:
The --config option is set to the original location of the rst file that
is being linted. This means that `rstcheck` will search for a config
file walking up from that location, instead of the location of the
temporary file that ALE generates.

Closes #2408. The resolution is based on recent changes in `rstcheck`,
cf. https://github.com/myint/rstcheck/issues/56.
There is a new option `g:ale_rst_rstcheck_use_project_config` that
enables this feature. This is so that anyone using an older
version <= 3.3.1 of `rstcheck`, which does not have the `--config`
options, is still able to run the linter. Since
https://github.com/myint/rstcheck/pull/57 has been merged,
any newer version of `rstcheck` should work out of the box.

The value `g:ale_rst_rstcheck_use_project_config` is 1 ("on") by
default, because I think people would generally expect their project
configuration to be taken into account in the default configuration.

A new option `g:ale_rst_rstcheck_options` is also provided. This
closes #2144. However, the main intended useage of this option is to
pass `--debug` to `rstcheck` (showing which config file is being used).

The documentation has been updated with a new
`ale-restructuredtext-rstcheck` section.

### DIFF
--- a/ale_linters/rst/rstcheck.vim
+++ b/ale_linters/rst/rstcheck.vim
@@ -1,6 +1,12 @@
-" Author: John Nduli https://github.com/jnduli
+" Authors:
+"   John Nduli https://github.com/jnduli,
+"   Michael Goerz https://github.com/goerz
 " Description: Rstcheck for reStructuredText files
 "
+
+call ale#Set('rst_rstcheck_options', '')
+call ale#Set('rst_rstcheck_use_project_config', 1)
+
 
 function! ale_linters#rst#rstcheck#Handle(buffer, lines) abort
     " matches: 'bad_rst.rst:1: (SEVERE/4) Title overline & underline
@@ -23,8 +29,15 @@ function! ale_linters#rst#rstcheck#Handle(buffer, lines) abort
 endfunction
 
 function! ale_linters#rst#rstcheck#GetCommand(buffer) abort
+    let l:dir = expand('#' . a:buffer . ':p:h')
+    let l:exec_args = ' ' . ale#Var(a:buffer, 'rst_rstcheck_options')
+    if ale#Var(a:buffer, 'rst_rstcheck_use_project_config')
+      let l:exec_args .= ' --config '. "'".l:dir."'"
+    endif
+
     return ale#path#BufferCdString(a:buffer)
     \   . 'rstcheck'
+    \   . l:exec_args
     \   . ' %t'
 endfunction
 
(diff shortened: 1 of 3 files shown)


goekce:
@goerz why did you close this PR before your PR was merged?

goerz:
I don't have the time or interest to pursue this further. If someone else wants to revive this, they'd be welcome to

w0rp:
Sometimes someone doesn't have the time to finish something themselves. Someone else can open a pull request if they can finish something.

Issue #2301, opened 2019-02-16
Text in the status line is squashed onto one line


anowlcalledjosh:
When I'm writing Rust, the errors in the status line get concatenated into one line:

![image](https://user-images.githubusercontent.com/9433472/52899916-aeb7d380-31e7-11e9-9f0a-9fb823791b4e.png)

but it's kind of confusing to have to work out where the first line of the error message ends and the other lines begin every time I want to fix an error. Compare with `:ALEDetail`:

![image](https://user-images.githubusercontent.com/9433472/52899942-f6d6f600-31e7-11e9-85f5-a2aa2b146fd0.png)

Would it be possible to separate the lines of the error message somehow in the status line? Even just putting a space between them would be an improvement, or it would be neat if it were possible to apply different styles to different parts of the error message (e.g. display the first line in bold, or everything except the first line in a dimmer colour).

RyanSquared:
I've seen this happen enough times with various Rust things that I think we should put up a handler for all Rust messages to be wrapped in this - i.e. if there's three lines, just put in a `: ` like what one of the tools does internally. This makes it more consistent from a UI perspective as all linters would give the same output.

w0rp:
Putting a space between them makes sense. They have to be put into one line and truncated so they don't create "Press Enter to continue" prompts as you move your cursor around, but we can add a space.

jethrosun:
> Putting a space between them makes sense. They have to be put into one line and truncated so they don't create "Press Enter to continue" prompts as you move your cursor around, but we can add a space.

This sounds great. Would it be trivial to implement? If so I can try to tackle it and push a PR.

I also want to point out that I think we need ALEDetail/virtualtext/status line msg to have a way to format them automatically anyway.

w0rp:
This shouldn't be hard to fix, split the lines and join them together again. Save different text in the `text` and `detail` keys.

w0rp:
Now newlines are replaced with spaces in the messages we echo.

PR #2425, opened 2019-04-08
improve the lsp diagnostic message format

jethrosun:
Here I am trying to improve the LSP diagnostic message format via replacing the line breaks with spaces (see #2301). This fix will affect the message in the status line, virtual text and the `ALEDetail`. 

However, I don't think this is what we desire: now the `ALEDetail` no longer has the line breaks. I assume the virtual text region is somehow binded with the `ALEDetail`? Any input on it? 

Regards,

### DIFF
--- a/autoload/ale/lsp/response.vim
+++ b/autoload/ale/lsp/response.vim
@@ -27,14 +27,8 @@ function! ale#lsp#response#ReadDiagnostics(response) abort
 
     for l:diagnostic in a:response.params.diagnostics
         let l:severity = get(l:diagnostic, 'severity', 0)
-        " we want to preserve the line breaks for `detail` but replace with
-        " spaces for text
-        let l:diagnostic.detail = l:diagnostic.message
-        let l:diagnostic.text = substitute(l:diagnostic.message, "\n", ' ', 'g')
-        let l:diagnostic.text= l:diagnostic.message
-
         let l:loclist_item = {
-        \   'text': l:diagnostic.text,
+        \   'text': substitute(l:diagnostic.message, "\n", ' ', 'g'),
         \   'type': 'E',
         \   'lnum': l:diagnostic.range.start.line + 1,
         \   'col': l:diagnostic.range.start.character + 1,
@@ -69,14 +63,14 @@ function! ale#lsp#response#ReadDiagnostics(response) abort
             \   ':' . (val.location.range.start.character + 1) .
             \   ":\n\t" . val.message
             \})
-            let l:loclist_item.detail = l:diagnostic.detail . "\n" . join(l:related, "\n")
+            let l:loclist_item.detail = l:diagnostic.message . "\n" . join(l:related, "\n")
         endif
 
         if has_key(l:diagnostic, 'source')
             let l:loclist_item.detail = printf(
             \   '[%s] %s',
             \   l:diagnostic.source,
-            \   l:diagnostic.text
+            \   l:diagnostic.message
             \)
         endif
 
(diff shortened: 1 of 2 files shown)


autoload/ale/lsp/response.vim
```diff
@@ -27,6 +27,7 @@ function! ale#lsp#response#ReadDiagnostics(response) abort
 
     for l:diagnostic in a:response.params.diagnostics
         let l:severity = get(l:diagnostic, 'severity', 0)
+        let l:diagnostic.message = substitute(l:diagnostic.message, "\n", ' ', 'g')
```
w0rp: Instead of modifying the incoming Dictionary, set the value with newlines in the `detail` key, and the value with newlines replaced in the `text` key. Add tests for this.

jethrosun:
I believe the code actually works but I am still failing the test. Am I missing something here?

autoload/ale/lsp/response.vim
```diff
@@ -27,8 +27,13 @@ function! ale#lsp#response#ReadDiagnostics(response) abort
 
     for l:diagnostic in a:response.params.diagnostics
         let l:severity = get(l:diagnostic, 'severity', 0)
+        " we want to preserve the line breaks for `detail` but replace with
+        " spaces for text
+        let l:diagnostic.detail = l:diagnostic.message
+        let l:diagnostic.text = substitute(l:diagnostic.message, "\n", ' ', 'g')
```
w0rp: Don't modify the incoming object here, just adjust the objects we're creating.

jethrosun: Do you mean something like:
```
        let l:diagnostic.text = l:diagnostic.message
        let l:diagnostic.text = substitute(l:diagnostic.text, "\n", ' ', 'g')
```

I don't think this passes the unit test. I am not very good at Vimscript so sorry about these problems.

w0rp: You can pass `substitute(l:diagnostic.message, "\n", ' ', 'g')` to the text key below without modifying the object, and use `l:diagnostic.message` below that as before.

jethrosun:
I just convert this PR to use the simple approach as suggested but the test is still failing. Do you have any insight?

test/lsp/test_read_lsp_diagnostics.vader
```diff
@@ -109,6 +109,26 @@ Execute(ale#lsp#response#ReadDiagnostics() should include sources in detail):
...
+  \     'lnum': 10,
+  \     'col': 15,
+  \     'end_lnum': 12,
+  \     'end_col': 22,
+  \   }
+  \ ],
+  \ ale#lsp#response#ReadDiagnostics({'params': {'uri': 'filename.ts', 'diagnostics': [
+  \   {
+  \     'range': Range(9, 14, 11, 22),
+  \     'message': 'cannot borrow `cap` as mutable more than once at a time\n\nmutable borrow starts here in previous iteration of loop',
```
w0rp:
```suggestion
  \     'message': "cannot borrow `cap` as mutable more than once at a time\n\nmutable borrow starts here in previous iteration of loop",
```

Here's the problem with your test. Escapes like `\n` are only applied for double quoted strings.

autoload/ale/lsp/response.vim
```diff
@@ -28,7 +28,7 @@ function! ale#lsp#response#ReadDiagnostics(response) abort
     for l:diagnostic in a:response.params.diagnostics
         let l:severity = get(l:diagnostic, 'severity', 0)
         let l:loclist_item = {
-        \   'text': l:diagnostic.message,
+        \   'text': substitute(l:diagnostic.message, "\n", ' ', 'g'),
```
w0rp:
```suggestion
        \   'text': substitute(l:diagnostic.message, '\(\r\n\|\n\|\r\)', ' ', 'g'),
```

I think you should handle Windows style CRLF newlines too. For example:

```vim
:echo substitute("one\ntwo\r\nthree\rfour", '\(\r\n\|\n\|\r\)', ' ', 'g')
```

The string for the regex has to be single quoted here. Add tests for `\r` by itself and `\r\n`.

w0rp:
Cheers! :beers:

PR #2430, opened 2019-04-11
Support $GO111MODULE with Go tooling

eliath:
Go allows the user to specify the $GO111MODULE environment variable
which changes the default behavior of go's built-in module system.

Golang documentation:
https://github.com/golang/go/wiki/Modules#how-to-use-modules

### DIFF
--- a/ale_linters/go/staticcheck.vim
+++ b/ale_linters/go/staticcheck.vim
@@ -8,17 +8,18 @@ function! ale_linters#go#staticcheck#GetCommand(buffer) abort
     let l:filename = expand('#' . a:buffer . ':t')
     let l:options = ale#Var(a:buffer, 'go_staticcheck_options')
     let l:lint_package = ale#Var(a:buffer, 'go_staticcheck_lint_package')
+    let l:env = ale#go#EnvString(a:buffer)
 
     " BufferCdString is used so that we can be sure the paths output from
     " staticcheck can be calculated to absolute paths in the Handler
     if l:lint_package
         return ale#path#BufferCdString(a:buffer)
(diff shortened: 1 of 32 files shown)


ale_linters/go/staticcheck.vim
```diff
@@ -3,21 +3,25 @@
...
     let l:filename = expand('#' . a:buffer . ':t')
     let l:options = ale#Var(a:buffer, 'go_staticcheck_options')
     let l:lint_package = ale#Var(a:buffer, 'go_staticcheck_lint_package')
+    let l:go111module = ale#Var(a:buffer, 'go_staticcheck_GO111MODULE')
 
     " BufferCdString is used so that we can be sure the paths output from
     " staticcheck can be calculated to absolute paths in the Handler
     if l:lint_package
         return ale#path#BufferCdString(a:buffer)
+        \   . (empty(l:go111module) ? '' : 'GO111MODULE="' . l:go111module . '" ')
```
w0rp: Use the `ale#Env` function generate the syntax for setting environment variables here. The syntax is different on Windows. Update the tests with a `has('win32')` check for running tests on Windows.

I think the variable should be named `ale_go_go111module`, and it should be used in all of the Go commands ALE runs which can read the environment variable.

eliath: Added the function `ale#go#EnvString()` to make it simple to support ALE configuration of other Go-specific environment variables in the future.

Added the `EnvString` in each Go linter command callback and fixer callback.
Updated tests for all Go tools accordingly

w0rp:
Cheers! :beers:

PR #2432, opened 2019-04-11
Let rust rls linter find the default rust toolchain automatically

superobertking:
This patch comes from this [comment](https://github.com/w0rp/ale/issues/1877#issuecomment-482096001).

As `rls` and `clippy` are [not always available](https://rust-lang.github.io/rustup-components-history/index.html) for the latest nightly rust build, nightly rust users typically needs to specify another version of nightly rust in order to get `rls` working. The version string looks somewhat like `nightly-2019-03-05` instead of just `nightly`. Therefore, the default value of `g:ale_rust_rls_toolchain` `nightly` usually just doesn't work and the documentation also did not point out the use of customizable toolchain names.  In fact, `rls` actually allows users to omit the toolchain name so that it will automatically find the default toolchain set by `rustup`, and the linter actually supports this behavior but it was not reflected in the documentation. Therefore, setting the default value as `''` is more suitable for most cases. In addition, stable rust users also won't need to change this variable by hand, and rls linter can work out-of-the-box.

I have read `:help ale-dev`. This patch does not seem to need additional tests.

### DIFF
--- a/ale_linters/rust/rls.vim
+++ b/ale_linters/rust/rls.vim
@@ -2,7 +2,7 @@
 " Description: A language server for Rust
 
 call ale#Set('rust_rls_executable', 'rls')
-call ale#Set('rust_rls_toolchain', 'nightly')
+call ale#Set('rust_rls_toolchain', '')
 call ale#Set('rust_rls_config', {})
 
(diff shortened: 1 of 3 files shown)


w0rp:
Cheers! :beers:

PR #2433, opened 2019-04-13
Remove otherproject#util#Double from d.vim

belka-ew:
I hope I don't misunderstand it but `otherproject#util#Double` looks suspicious, I don't see it used anywhere and neovim complains about it:

> Error detected while processing /home/belka/.local/share/nvim/plugged/ale/autoload/ale/d.vim
> line 6
> E746: Function name does not match script file name: otherproject#util#Double


w0rp:
Cheers! :beers: I think I must have accidentally put that in the file.

Issue #2434, opened 2019-04-13
Getting rid of line breaks in virtualtex


jethrosun:
In the current implementation (see [here](https://github.com/w0rp/ale/blob/master/autoload/ale/virtualtext.vim#L84) ), the virtualtext is using the `detail` field to show the error message. This should be `text` instead because we don't want the line breaks.

jethrosun:
For example, the current virtualtext message looks like this:
<img width="1642" alt="Screen Shot 2019-04-15 at 9 55 38 AM" src="https://user-images.githubusercontent.com/15823053/56138393-d5cf1f00-5f64-11e9-8c1e-13fad98d5f46.png">

I think it is better if we can lose the line breaks there as well:
<img width="1552" alt="Screen Shot 2019-04-15 at 9 56 24 AM" src="https://user-images.githubusercontent.com/15823053/56138451-f5fede00-5f64-11e9-9c1e-e114c25aab60.png">

w0rp:
The `text` key should be used there instead, then.

jethrosun:
I think so. Do you want me to create a PR? should be only one line of code...

w0rp:
Sure, go for it. Thanks for fixing the other issue. :+1:

jethrosun:
You are welcome. I enjoyed using ALE :)

PR #2441, opened 2019-04-15
fix 2434: use text field instead of the detail field for virtualtext

jethrosun:
We actually want to use `text` instead of `detail` in virtualtext because the formal handles the line breaks nicely. For detail please see #2434

### DIFF
--- a/autoload/ale/virtualtext.vim
+++ b/autoload/ale/virtualtext.vim
@@ -81,7 +81,7 @@ function! ale#virtualtext#ShowCursorWarning(...) abort
     call ale#virtualtext#Clear()
 
     if !empty(l:loc)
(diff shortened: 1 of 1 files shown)


autoload/ale/virtualtext.vim
```diff
@@ -81,7 +81,7 @@ function! ale#virtualtext#ShowCursorWarning(...) abort
     call ale#virtualtext#Clear()
 
     if !empty(l:loc)
-        let l:msg = get(l:loc, 'detail', l:loc.text)
+        let l:msg = get(l:loc, 'text', l:loc.text)
```
w0rp:
```suggestion
        let l:msg = l:loc.text
```

You can just do this instead.

w0rp:
Cheers! :beers:

PR #2444, opened 2019-04-17
Better phpstan default configuration

jdeniau:
- If a config file is found, then do not set the default level to `4`, the level should be in the configuration file (except if you force it of course).
- Mark phpstan error as error, not warning, as phpstan does not have warning, only errors.
- Use the `--no-progress` parameter

Thank you !

### DIFF
--- a/ale_linters/php/phpstan.vim
+++ b/ale_linters/php/phpstan.vim
@@ -13,8 +13,9 @@ function! ale_linters#php#phpstan#GetCommand(buffer, version) abort
     \   : ''
 
     let l:level =  ale#Var(a:buffer, 'php_phpstan_level')
+    let l:config_file_exists = ale#path#FindNearestFile(a:buffer, 'phpstan.neon')
 
-    if empty(l:level) && !filereadable('phpstan.neon')
+    if empty(l:level) && empty(l:config_file_exists)
         " if no configuration file is found, then use 4 as a default level
         let l:level = '4'
     endif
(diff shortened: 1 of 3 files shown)


test/handler/test_phpstan_handler.vader
```diff
@@ -68,10 +66,57 @@ Execute(Output for .inc files should be parsed correctly):
...
-  \     'type': 'W'
+  \     'type': 'E'
   \   }
   \ ],
   \ ale_linters#php#phpstan#Handle(bufnr(''), [
-  \   ' 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%',
   \   '/phpstan-test-files/foo/test.inc:9:Access to an undefined property Test::$var.',
   \])
+
+Execute(Test phpstan command output):
```
w0rp: Put these tests in a separate file in the `test/command_callback` directory. They may already be one. Use the special assertion commands documented in `:help ale-tests` instead.

ale_linters/php/phpstan.vim
```diff
@@ -12,14 +12,25 @@ function! ale_linters#php#phpstan#GetCommand(buffer, version) abort
     \   ? ' -c ' . l:configuration
     \   : ''
 
+    let l:level =  ale#Var(a:buffer, 'php_phpstan_level')
+
+    if empty(l:level) && !filereadable('phpstan.neon')
```
w0rp: You should use the `ale#path#FindNearestFile` function used in other files to find this. This code will only work if the cwd for Vim contains this file.

jdeniau: You are right, thanks !

jdeniau:
@w0rp changes made

w0rp:
Cheers! :beers:

PR #2445, opened 2019-04-18
Add option to show msg id when pylint is used

katsuya-horiuchi:
With this change, `let g:ale_python_pylint_use_msg_id = 1` can be used to tweak the output of pylint.

This would allow you to show message id instead of human-friendly format (See #852 for discussion).

I've never written vimscript before and basically have no idea what I'm doing, so couldn't write test.
I'd appreciate it if someone can add a test for this.


katsuya-horiuchi:
Sorry to have taken some time. Could you check the test I've added?

w0rp:
Cheers! :beers:

PR #2446, opened 2019-04-18
Add g:ale_sh_shellcheck_change_directory

fnichol:
This change adds a new `g:ale_sh_shellcheck_change_directory` option to control whether or not to change directory to where the source is located before running `shellcheck`. It defaults to `1`, identical to the other linter directory-related options (I took heavy direction from the Python linters here) and so preserves the current behavior of ALE.

This is to help ShellCheck users that have a larger source base with sub-directories and CI that prefers to run from the root of a repository. As ShellCheck determines the relative directory in the `shellcheck source=path/to/src.sh` to be relative to the current working directory, it's hard to to have a value that satisfies both a *vim/ALE and a Makefile/TravisCI setup. The option here can be used as a first step to harmonize these two cases while any future features/changes in ShellCheck develop (see koalaman/shellcheck#769 and koalaman/shellcheck#539 for more context).

---

Thanks so much for your hard work on ALE. I know I'm not alone, and I know maintainers don't hear this nearly often enough, but tools like this really make a huge difference in daily developer life. ALE has made a massive difference for me and how I work, so thank you! 🍺

### DIFF
--- a/ale_linters/sh/shellcheck.vim
+++ b/ale_linters/sh/shellcheck.vim
@@ -10,6 +10,7 @@ call ale#Set('sh_shellcheck_exclusions', get(g:, 'ale_linters_sh_shellcheck_excl
 call ale#Set('sh_shellcheck_executable', 'shellcheck')
 call ale#Set('sh_shellcheck_dialect', 'auto')
 call ale#Set('sh_shellcheck_options', '')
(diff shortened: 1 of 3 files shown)


w0rp:
I finally had a chance to look at this, and it looks good to me. Cheers! :beers:

PR #2448, opened 2019-04-19
Add support for Vim's tagstack to ALEGoToDefinition

### LINKED ISSUE #1236: Request: add to tag stack when using ALEGoToDefinition OR add a ALEGoBackFromDefinition
Vim shortcuts pair `<C-]> and <C-t>` nicely. With ALEGoToDefinition, there isn't a nice equivalent jump back command (e.g. `<C-t>` ALEGoBackFromDefinition or `<C-t>`) which would be nice to have.

reedriley:
Fixes #1236.

I added tests - but to be fair, this feature won't work on versions of Vim earlier than [8.1.0519](https://github.com/vim/vim/commit/f49cc60aa802862c595ff619dccc11271633a94b) and the test automation script doesn't test on anything that recent.

The feature is designed to degrade gracefully; and I'm open to suggestions on how to best clarify the version requirement in the docs.

### DIFF
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -3,6 +3,9 @@
 
 let s:go_to_definition_map = {}
 
+" Enable automatic updates of the tagstack
+let g:ale_update_tagstack = get(g:, 'ale_update_tagstack', 1)
+
 " Used to get the definition map in tests.
 function! ale#definition#GetMap() abort
     return deepcopy(s:go_to_definition_map)
@@ -17,6 +20,18 @@ function! ale#definition#ClearLSPData() abort
     let s:go_to_definition_map = {}
 endfunction
 
+function! ale#definition#UpdateTagStack() abort
+    let l:should_update_tagstack = exists('*gettagstack') && exists('*settagstack') && g:ale_update_tagstack
+
+    if l:should_update_tagstack
+        let l:from = [bufnr('%'), line('.'), col('.'), 0]
+        let l:tagname = expand('<cword>')
+        let l:winid = win_getid()
+        call settagstack(l:winid, {'items': [{'from': l:from, 'tagname': l:tagname}]}, 'a')
+        call settagstack(l:winid, {'curidx': len(gettagstack(l:winid)['items']) + 1})
+    endif
+endfunction
+
 function! ale#definition#HandleTSServerResponse(conn_id, response) abort
     if get(a:response, 'command', '') is# 'definition'
     \&& has_key(s:go_to_definition_map, a:response.request_seq)
@@ -27,6 +42,7 @@ function! ale#definition#HandleTSServerResponse(conn_id, response) abort
             let l:line = a:response.body[0].start.line
             let l:column = a:response.body[0].start.offset
 
+            call ale#definition#UpdateTagStack()
             call ale#util#Open(l:filename, l:line, l:column, l:options)
         endif
     endif
@@ -51,6 +67,7 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
             let l:line = l:item.range.start.line + 1
             let l:column = l:item.range.start.character + 1
 
+            call ale#definition#UpdateTagStack()
             call ale#util#Open(l:filename, l:line, l:column, l:options)
             break
         endfor
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -24,10 +24,12 @@ function! ale#definition#UpdateTagStack() abort
     let l:should_update_tagstack = exists('*gettagstack') && exists('*settagstack') && g:ale_update_tagstack
 
     if l:should_update_tagstack
-        let l:from = [bufnr('%'), line('.'), col('.'), 0]
+        " Grab the old location (to jump back to) and the word under the
+        " cursor (as a label for the tagstack)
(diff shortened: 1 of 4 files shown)


reedriley:
I modified the `Dockerfile` and `./run-tests` script; and there are issues with my proposed unit test.  I'll put up a corrected version shortly.

autoload/ale/definition.vim
```diff
@@ -17,6 +20,18 @@ function! ale#definition#ClearLSPData() abort
     let s:go_to_definition_map = {}
 endfunction
 
+function! ale#definition#UpdateTagStack() abort
+    let l:should_update_tagstack = exists('*gettagstack') && exists('*settagstack') && g:ale_update_tagstack
+
+    if l:should_update_tagstack
+        let l:from = [bufnr('%'), line('.'), col('.'), 0]
```
w0rp: You can use the filename, line, and column we already have to determine this.

autoload/ale/definition.vim
```diff
@@ -17,6 +20,18 @@ function! ale#definition#ClearLSPData() abort
...
+        let l:tagname = expand('<cword>')
+        let l:winid = win_getid()
```
w0rp: Get the window ID from the buffer number, instead of the current window.

autoload/ale/definition.vim
```diff
@@ -17,6 +20,18 @@ function! ale#definition#ClearLSPData() abort
     let s:go_to_definition_map = {}
 endfunction
 
+function! ale#definition#UpdateTagStack() abort
+    let l:should_update_tagstack = exists('*gettagstack') && exists('*settagstack') && g:ale_update_tagstack
+
+    if l:should_update_tagstack
+        let l:from = [bufnr('%'), line('.'), col('.'), 0]
+        let l:tagname = expand('<cword>')
```
w0rp: You might have to wait until the buffer is opened first, and grab the word second.

autoload/ale/definition.vim
reedriley: Assuming you're referring to `l:filename`, `l:line` and `l:column` defined in `ale#definition#HandleTSServerResponse` and `ale#definition#HandleLSPResponse` - these are the targets for the new location.  The values I'm pushing on to the stack are based on the old location.

I'll change the variable names to be a bit more clear.  :)

autoload/ale/definition.vim
reedriley: Same as above - the values I'm pushing on to the stack are based on the old location.  So I don't think I need to wait for the buffer to open; assuming I've understood this comment correctly?

autoload/ale/definition.vim
reedriley: This sounds like it's probably *really* relevant and important - but I'm afraid I'm not an expert on the Vim functions involving windows and buffers.  Any chance you can provide some pointers to the relevant docs?

reedriley:
> We should add tests for this. I'll have to update the Docker image so the 8.1 version we build is new enough to contain this feature.

Agreed.  The test I added is fairly minimal; but passes when I build a custom Docker image and update the run-tests script to use a local image.  I'm happy to include the Dockerfile change as well, if you'd like.

GoingFast:
@w0rp @reedriley doesn't seem to be working for me. :ALEGoToDefinition and when I <C-t> then it says tagstack is empty. So right now I use <C-o>

reedriley:
> @w0rp @reedriley doesn't seem to be working for me. :ALEGoToDefinition and when I C-t then it says tagstack is empty. So right now I use C-o

What version of Vim are you running?

wrachwal:
@w0rp  @reedriley I have the same problem (tagstack is empty). I am using Vim 8.1. Fresh ALE, commit * 27146ade 2019-05-30

reedriley:
@wrachwal: To be clear, I actually meant to ask which version of `Vim`.  The tagstack manipulation functions the diff uses were added in Vim [8.1.0519](https://github.com/vim/vim/commit/f49cc60aa802862c595ff619dccc11271633a94b).

wrachwal:
Thanks, after upgrade using `ppa:jonathonf/vim` it works fine.

**OLD** VIM - Vi IMproved 8.1 (2018 May 18, compiled Nov  6 2018 09:45:53)
Included patches: 1-513

**NEW** VIM - Vi IMproved 8.1 (2018 May 18, compiled May 27 2019 13:46:23)
Included patches: 1-1408

dmchoiboi:
Is there any plan to provide this functionality for NeoVim? I'm on the latest NeoVim version (v0.3.7) and I'm getting the same 'tagstack is empty' mentioned above

rislah:
Yep its not working on neovim.

hourliert:
Same issue here. It is not working on neovim 0.3.8. Can we make it work for neovim by any chance? Thanks in advance!

dmchoiboi:
@hourliert @rislah I just checked NeoVim's source, and it appears that their [patch for vim-8.1.0519](https://github.com/neovim/neovim/pull/9951) hasn't made it out to a stable release yet. According  to that pull request, the update is targeted to be out in the 0.4 release. I've tested out the [nightly release](https://github.com/neovim/neovim/releases/tag/nightly) which contains the fix and verified that this tagstack feature works correctly.

PR #2466, opened 2019-04-29
Include plugin description in local help line

alerque:
The recommended format for _vim's internal help files_ is "&lt;tag&gt; &lt;for vim version&gt; &lt;last change&gt;", (see `:help help-writing` but this format is not parsed the same way for plugins. For plugins the recommended format includes a description of the plugin such as "&lt;tag&gt; &lt;description&gt;". See `:help write-local-help` for the different template.


w0rp:
Cheers! :beers:

PR #2473, opened 2019-05-04
Add typescript-language-server support

soywod:
Added support of https://github.com/theia-ide/typescript-language-server, for both javascript and typescript. I'm using it since a while and I thought it could be useful to add them here :)

### DIFF
--- a/ale_linters/javascript/tls.vim
+++ b/ale_linters/javascript/tls.vim
@@ -1,0 +1,17 @@
+" Author: soywod <clement.douin@posteo.net>
+" Description: typescript-language-server integration for ALE
+
+call ale#Set('javascript_tls_executable', 'typescript-language-server')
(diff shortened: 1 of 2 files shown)


soywod:
> Looks fine to me. Add some tests for the linters. See `:help ale-tests`. Should we try to find `tsconfig.json` first, instead of `package.json`?

It's pointless, because you can't have both javascript and tsconfig. I mean, if you have a tsconfig, you should use typescript instead of javascript.

w0rp:
You could look for `jsconfig.json`.

soywod:
It's not that common in web development, but yes it could be checked before `package.json`. I will update soon.

omeid:
So what is blocking this?

w0rp:
The thing blocking this was the lack of tests. If someone wants to open a new PR with tests, please do. If the author comes back, we can reopen the PR.

PR #2491, opened 2019-05-10
Add phpstan autoload option

mpesari:
This PR adds the `-a` option to ale's phpstan call. I used to `let g:ale_php_phpstan_configuration = 'phpstan.neon -a autoload.php'`, but that was broken recently in commit 5f64f8dc5799d9ad93544d24fb780428cd07dbbe due to option escaping. So I just added this option and it works for me.

I could probably write a simple test to verify that the option is used, but not sure if it's necessary. What do you think?


w0rp:
Cheers! :beers:

PR #2495, opened 2019-05-12
Fall back to vim's tags in ALEGoToDefinition

liskin:
This makes it practical to globally map `<C-]>` and `<C-W><C-]>` to
`ALEGoToDefinition` and have it automatically choose between ale's LSP and
vim's `:tag`.

> comment above and the documentation referenced in it first. Write tests!
> 

I'm not sure if this change is useful for other people and I'm not sure
I implemented it at the correct place, so I'm presenting it as a Work in
Progress. If you think this is a useful addition, I'll add tests and
update documentation. Hope it's okay this way. :-)

### DIFF
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -137,8 +137,13 @@ function! ale#definition#GoTo(options) abort
     for l:linter in ale#linter#Get(&filetype)
         if !empty(l:linter.lsp)
             call s:GoToLSPDefinition(l:linter, a:options, 'definition')
+
+            return
         endif
     endfor
+
+    " Fall back to vim's tags
+    call ale#util#Tag(a:options)
 endfunction
 
 function! ale#definition#GoToType(options) abort
--- a/autoload/ale/util.vim
+++ b/autoload/ale/util.vim
@@ -114,6 +114,30 @@ function! ale#util#Open(filename, line, column, options) abort
     normal! zz
 endfunction
 
+" Jump to tag under cursor (using CTRL-]) respecting options['open_in']
+" (see above).
+function! ale#util#Tag(options) abort
+    let l:open_in = get(a:options, 'open_in', 'current-buffer')
+
+    if l:open_in is# 'tab'
+        call ale#util#Execute('tab split')
+    elseif l:open_in is# 'horizontal-split'
+        call ale#util#Execute('split')
+    elseif l:open_in is# 'vertical-split'
+        call ale#util#Execute('vsplit')
+    endif
+
+    try
+        call ale#util#Execute("normal! \<c-]>")
+    catch
+        if l:open_in isnot# 'current-buffer'
+            call ale#util#Execute('close')
+        endif
+
+        execute 'echohl ErrorMsg | echomsg v:exception | echohl None'
+    endtry
+endfunction
+
 let g:ale#util#error_priority = 5
 let g:ale#util#warning_priority = 4
 let g:ale#util#info_priority = 3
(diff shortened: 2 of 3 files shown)


w0rp:
This isn't the purpose of the command. It's purely for jumping to a definition known to a language server, and not what Vim thinks the right location is.

liskin:
@w0rp Thanks for the feedback. Is there a better way to bind `<C-]>` to ALEGoToDefinition with a fallback to tags? I read all of ale-api and there doesn't seem to be a function that tells me whether ALEGoToDefinition will do something. Shall I just call into ale's internals or do you think this usecase warrants adding something to ale?

liskin:
Anyway, the "Force sort locale in check-supported-tools-tables" commit is useful, I think. Shall I create a separate PR for that alone?

w0rp:
Yes, please create a separate pull request for that. That should be useful.

I can't think of a way to fall back on Vim tags off the top of my head, but I think it would be sensible if there was some way to easily do that, without it being something that ALE does automatically.

liskin:
PR here: https://github.com/w0rp/ale/pull/2496

Re fall back to tags: I dug a bit deeper into ale sources last night and realized that it's a bit more complicated than I expected — the LSP could fail to start, it can lack the "definition" capability, and we're not going to find out synchronously. So maybe I'm approaching this from the wrong direction. If I can hook into `s:UpdateCapabilities` and remap `<C-]>` buffer-locally if it provides definition capability, that would likely be much more robust. I'll give it some more time later. Thanks for the pointers so far.

w0rp:
I'm not entirely convinced this is the right approach yet, but the best idea might emit a `User` autocmd event for when jumping to a definition fails or succeeds, and then you could respond to that with whatever code you want. This would be similar to how ALE uses `silent doautocmd <nomodeline> User ALELintPost` after linting is done so you can do whatever you want after linting is done.

w0rp:
Another idea is to provide a function which is part of ALE's public API, which lets you set callbacks. I'm not sure that's the best idea yet either.

liskin:
I finally got to this last week and managed to hack something together without touching ale at all. In a nutshell:

 * hook into `autocmd User ALELintPost` (just so that I don't need to think about ordering of my project/filetype-specific BufRead autocmds that enable/disable linters; `ALELintPost` guarantees everything's been set up)
* issue `ale#lsp_linter#StartLSP` with custom callback
* call `ale#lsp#HasCapability` and set buffer-local variable indicating whether the fallback to tags is necessary
* bind `<C-]>` to a function that checks the buffer-local variable and either invokes `:ALEGoToDefinition` or `normal! \<C-]>`

It does use two internal ale functions but I don't really mind fixing it if/when it breaks in the future. If you think some public/stable API for this usecase is desirable, I can try to implement one of your suggestions, but I'll be just as happy to leave it as it is. :-)

Full code here: https://github.com/liskin/dotfiles/commit/a93cba488f40860f2bfe86a110a652caf9510b34#diff-28ad5e44396f2bd10662ebb5b7a57b83

toupeira:
@liskin thanks for that snippet, I adapted it to fallback to https://github.com/misterbuckley/vim-definitive, works great!

@w0rp it would be nice to have some of this built-in somehow, e.g. a buffer-local variable or helper function to check for active LSP capabilities.

liskin:
@toupeira I'm glad that it's useful for others! Do note that I've had to modify it a bit since then: https://github.com/liskin/dotfiles/commit/03a28922ac88000047fd34f4a803cedf43176c25#diff-28ad5e44396f2bd10662ebb5b7a57b83

ALELintPost wasn't being invoked by some LSPs. Took me a while to figure out what's going on and I forgot the details, unfortunately. :-(

toupeira:
@liskin yes I used the version from your `master` :grinning:

PR #2496, opened 2019-05-13
Force sort locale in check-supported-tools-tables

liskin:
Otherwise it reports that the list isn't sorted properly if user's LANG
is different.


w0rp:
Cheers! :beers:

PR #2501, opened 2019-05-13
[lsp] Improve default root directory for Texlab

rliang:
Now defaults to the directory of the file.


w0rp:
This default could cause problems. This would cause ALE to start a language server process for each directory you open.

rliang:
I see. The more fundamental problem here seems to be that LaTeX does not have the concept of a "project".

Maybe use ".git" folders then?

w0rp:
Yeah. You can also override any of the functions for determining the project root automatically with the `ale_lsp_root` setting.

w0rp:
Closed due to inactivity.

PR #2502, opened 2019-05-14
Change Python project root dir detection for flake8 configuration

ericvw:
The official configuration files for `flake8` are `.flake8`, `tox.ini`,
and `setup.cfg`.

After investigation, it is safe to remove `flake8.cfg` as it appears to
only exist as a typo in other tooling documentation (e.g.,
`python-language-server`).

Even though no linters automatically read `.flake8rc`, it is kept in
case projects may be using it for detecting the projects root directory.

http://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations

### DIFF
--- a/autoload/ale/python.vim
+++ b/autoload/ale/python.vim
@@ -25,7 +25,7 @@ function! ale#python#FindProjectRootIni(buffer) abort
         \|| filereadable(l:path . '/tox.ini')
         \|| filereadable(l:path . '/mypy.ini')
         \|| filereadable(l:path . '/pycodestyle.cfg')
-        \|| filereadable(l:path . '/flake8.cfg')
+        \|| filereadable(l:path . '/.flake8')
         \|| filereadable(l:path . '/.flake8rc')
         \|| filereadable(l:path . '/pylama.ini')
         \|| filereadable(l:path . '/pylintrc')
--- a/doc/ale-python.txt
+++ b/doc/ale-python.txt
@@ -29,7 +29,7 @@ ALE will look for configuration files with the following filenames. >
   tox.ini
   mypy.ini
   pycodestyle.cfg
-  flake8.cfg
+  .flake8
   .flake8rc
   pylama.ini
   pylintrc
--- a/test/test_python_find_project_root.vader
+++ b/test/test_python_find_project_root.vader
@@ -1,0 +1,11 @@
+Before:
+  call ale#test#SetDirectory('/testplugin/test')
+
+After:
+  call ale#test#RestoreDirectory()
+
+Execute(Detect root of Python project with .flake8 correctly):
+  call ale#test#SetFilename('python-test-files/python-package-project/package-name/module.py')
+  AssertEqual
+  \ ale#path#Simplify(g:dir . '/python-test-files/python-package-project'),
+  \ ale#python#FindProjectRoot(bufnr(''))


autoload/ale/python.vim
```diff
@@ -25,8 +25,7 @@ function! ale#python#FindProjectRootIni(buffer) abort
         \|| filereadable(l:path . '/tox.ini')
         \|| filereadable(l:path . '/mypy.ini')
         \|| filereadable(l:path . '/pycodestyle.cfg')
-        \|| filereadable(l:path . '/flake8.cfg')
```
w0rp: `python-language-server` discovers paths by looking for `flake8.cfg`, so I'd keep this. It's mentioned here: https://pypi.org/project/python-language-server/

autoload/ale/python.vim
```diff
@@ -25,8 +25,7 @@ function! ale#python#FindProjectRootIni(buffer) abort
...
-        \|| filereadable(l:path . '/.flake8rc')
```
w0rp: A number of projects and linter plugins search for `.flake8rc`, so I'd keep this too.

autoload/ale/python.vim
```diff
@@ -25,8 +25,7 @@ function! ale#python#FindProjectRootIni(buffer) abort
...
+        \|| filereadable(l:path . '/.flake8')
```
w0rp: You can move this further up the list, as this is the officially supported filename.

autoload/ale/python.vim
ericvw: It appears to be a typo in the README of `python-language-server`.

https://github.com/palantir/python-language-server/blob/96e08d85635382d17024c352306c4759f124195d/pyls/config/flake8_conf.py#L10

The commit which introduced the original introduced the change for reading `flake8` but documented it incorrectly (https://github.com/palantir/python-language-server/commit/3c1f18361179f3a687bdef6209e136a10bbf043f).

The `python-language-server` has limited contributions to the project to those who have contributed in the past — I am unable to open an issue nor a pull request to address this discrepancy 😞 .

autoload/ale/python.vim
ericvw: I searched the history of the `flake8` project and it has never supported `.flake8rc`, from what I can tell.

I conducted a hasty [Google search: ".flake8rc"](https://www.google.com/search?q=".flake8rc"&oq=".flake8rc") and identified the following (potential) projects that may be using this are:

- https://github.com/AtomLinter/linter-flake8
- https://github.com/flycheck/flycheck
- https://github.com/w0rp/ale

Ale is using this in the documentation and within this function.

For `linter-flake8`, it looks like it is another documentation bug (https://github.com/AtomLinter/linter-flake8/blob/875c273d3affe253478b804a51441688a0b2d03e/package.json#L27), from search https://github.com/AtomLinter/linter-flake8/search?q=flake8rc&unscoped_q=flake8rc.

For `flycheck`, it appears they have a `flycheck-flake8rc`-specific configuration file documented in (https://github.com/flycheck/flycheck/blob/47174a12fd84a685f36019632838c73b4813b66d/doc/languages.rst).

w0rp:
Leave `.flake8rc` in place. `flake8` has never recognised it, but there are projects out there that use it. If `flake8.cfg` isn't used in any project, we can remove it.

ericvw:
> Leave `.flake8rc` in place. `flake8` has never recognised it, but there are projects out there that use it.

I'm curious more than anything: What other linters have you seen that use it?  I haven't been able to find anything concrete that would affect the Python linters which ale calls out to.

I'll leave `.flake8rc` in place as requested and will update the PR once I have a test in place for detecting `.flake8`.  Greatly appreciate the quick review feedback :).

w0rp:
No linters automatically read it, but the location of the file is used to figure out where the root of a project might be.

ericvw:
@w0rp, I have updated the PR with the changes requested and added a new test suite for future iterations of detecting a Python project's root directory.

w0rp:
Cheers! :beers:

PR #2506, opened 2019-05-15
WIP: Add cppcheck includes when modified buffer or no compile_commands.json

ckoehler:
This can probably be expanded to look at -I directives from compile_commands.json to make it more robust. See comments in the code.

I don't know how to modify a buffer from a Vader Execute block, so those tests are failing right now. You may have a better idea.


w0rp:
You can use `:set modified` to mark a buffer as modified.

Is there any danger in always setting `-I` with the directory name? Does it make the results less accurate? If so, your solution is a good one. Paired with the code I had for improving how `compile_commands.json` files are found, this should be good.

ckoehler:
I'm not sure it'd make a difference. I think cppcheck just ignores -I when --project is given.
Doing a `:set modified` or `set modified` does not seem to work in the test...

ckoehler:
Actually, after setting modified, do I need to call ALE again? I've tried 

```
set modified
ALELint
``` 

and 

```
set modified
call ale#Queue(0)
```

but neither seem to work. I'd appreciate the help, if you want that last test. I don't have time to pursue this much further.
Let me know if you figure it out, or whether I can just remove the last test. I'll then clean up the pull request.

ale_linters/c/cppcheck.vim
```diff
@@ -18,10 +28,21 @@ function! ale_linters#c#cppcheck#GetCommand(buffer) abort
...
     \   : ''
 
+    " if the buffer is modified or we haven't found compile_commands.json,
+    " include the file's directory.
+    if l:modified || empty(l:compile_commmands_path)
+      " Get path to this buffer so we can include it into cppcheck with -I
+      " This could be expanded to get more -I directives from the compile
+      " command in compile_commands.json, if it's found.
+      let l:buffer_path = fnamemodify(bufname(a:buffer), ':p:h')
+      let l:buffer_path_include = ' -I'.l:buffer_path
```
w0rp:
```suggestion
      let l:buffer_path_include = ' -I' . ale#Escape(l:buffer_path)
```

Escape the path, in case it contains spaces, etc.

ckoehler: I'll escape it. What do you mean by duplicate code? The GetCommand function for C and CPP is very similar, do you mean that? Where would this common function live?

ckoehler:
I escaped the paths. 

Can you clarify what you think should be a separate function, and where it should live?

w0rp:
You can create a function in `autoload/ale/handlers/cppcheck.vim` that both versions of the linter can use. Make sure to update the code here to take the recent changes I've made for using `build/compile_commands.json` into account.

ckoehler:
Alright, refactored, and all but that one test (where I `set modified`) pass. I don't know enough about Vader to figure this one out, so I'll leave it up to you from here. Thanks!

w0rp:
@bartlibert Does this look okay to you?

bartlibert:
Looks fine to me. Nice improvement.

w0rp:
I fixed the test and merged this via the commandline now.

PR #2509, opened 2019-05-16
Add support for Erlang dialyzer

AntoineGagne:
This PR adds support for the Erlang tool `dialyzer` which type check Erlang modules.

### DIFF
--- a/ale_linters/erlang/dialyzer.vim
+++ b/ale_linters/erlang/dialyzer.vim
@@ -16,7 +16,7 @@ function! ale_linters#erlang#dialyzer#FindPlt() abort
 endfunction
 
 function! ale_linters#erlang#dialyzer#GetPlt(buffer) abort
-    return ale#Var(a:buffer, 'erlang_plt_file')
+    return ale#Var(a:buffer, 'erlang_dialyzer_plt_file')
 endfunction
 
 function! ale_linters#erlang#dialyzer#GetExecutable(buffer) abort
@@ -63,8 +63,8 @@ endfunction
 
 let g:ale_erlang_dialyzer_executable =
 \   get(g:, 'ale_erlang_dialyzer_executable', 'dialyzer')
-let g:ale_erlang_plt_file =
-\   get(g:, 'ale_erlang_plt_file', ale_linters#erlang#dialyzer#FindPlt())
+let g:ale_erlang_dialyzer_plt_file =
+\   get(g:, 'ale_erlang_dialyzer_plt_file', ale_linters#erlang#dialyzer#FindPlt())
 
 call ale#linter#Define('erlang', {
 \   'name': 'dialyzer',
--- a/ale_linters/erlang/dialyzer.vim
+++ b/ale_linters/erlang/dialyzer.vim
@@ -8,7 +8,10 @@ endfunction
 function! ale_linters#erlang#dialyzer#FindPlt(buffer) abort
     let l:rebar3_profile = ale_linters#erlang#dialyzer#GetRebar3Profile(a:buffer)
     let l:plt_file_directory = ale#path#FindNearestDirectory(a:buffer, '_build' . l:rebar3_profile)
-    let l:plt_file = split(globpath(l:plt_file_directory, '/*_plt'), '\n')
+
+    if !empty(l:plt_file_directory)
+        let l:plt_file = split(globpath(l:plt_file_directory, '/*_plt'), '\n')
+    endif
 
     if !empty(l:plt_file)
         return l:plt_file[0]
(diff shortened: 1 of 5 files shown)


doc/ale-erlang.txt
```diff
@@ -3,6 +3,27 @@ ALE Erlang Integration                                     *ale-erlang-options*
 
 
 ===============================================================================
+dialyzer                                                  *ale-erlang-dialyzer*
+
+g:ale_erlang_syntaxerl_executable            *g:ale_erlang_dialyzer_executable*
```
w0rp: There's a typo here.

ale_linters/erlang/dialyzer.vim
```diff
@@ -0,0 +1,74 @@
+" Author: Autoine Gagne - https://github.com/AntoineGagne
+" Description: Define a checker that runs dialyzer on Erlang files.
+
+function! ale_linters#erlang#dialyzer#FindPlt() abort
+    let l:plt_file = split(globpath('_build', '**/*_plt'), '\n')
```
w0rp: Do Erlang projects always or often have `_build` directories? You should use a function like `ale#path#FindNearestDirectory` to find a directory based on where the buffer is. The use of `globpath` with `**` could cause major performance issues. I've seen bug reports where using `globpath` to search through directories in large directories has slowed Vim significantly.

doc/ale-erlang.txt
```diff
@@ -3,6 +3,27 @@ ALE Erlang Integration                                     *ale-erlang-options*
...
+                                             *b:ale_erlang_dialyzer_executable*
+  Type: |String|
+  Default: `'dialyzer'`
+
+  This variable can be changed to specify the dialyzer executable.
+
+
+g:ale_erlang_plt_file                                   *g:ale_erlang_plt_file*
```
w0rp: Name this `ale_erlang_dialyzer_plt_file`, or move the documentation for this into a general section above. Have a look a the C or Python documentation for examples.

ale_linters/erlang/dialyzer.vim
```diff
@@ -0,0 +1,74 @@
...
+        let l:match = matchlist(l:line, l:pattern)
+
+        if len(l:match) != 0
+            let l:code = l:match[2]
+
+            call add(l:output, {
+            \   'lnum': str2nr(l:match[1]),
+            \   'lcol': 0,
+            \   'text': l:code,
+            \   'type': 'W'
```
w0rp: Are all of the problems warnings?

AntoineGagne: The tool doesn't provide any information as to whether it is a warning or an error so I chose to make the reported information into warnings.

ale_linters/erlang/dialyzer.vim
AntoineGagne: The `_build` directory is not always present. It is present when the `rebar3` build tool is used (it is the most used build system in the ecosystem). 

I used the wildcard `**` because I didn't know which profile was used when generating the `*_plt` file but I changed the code so a profile can be configured thus discarding the need of having to use the wildcard to recursively search.

ale_linters/erlang/dialyzer.vim
```diff
@@ -0,0 +1,88 @@
...
+endfunction
+
+function! ale_linters#erlang#dialyzer#GetExecutable(buffer) abort
+    return ale#Var(a:buffer, 'erlang_dialyzer_executable')
+endfunction
+
+function! ale_linters#erlang#dialyzer#GetCommand(buffer) abort
+    let l:command = fnameescape(ale_linters#erlang#dialyzer#GetExecutable(a:buffer))
+    \   . ' -n'
+    \   . ' --plt ' . fnameescape(ale_linters#erlang#dialyzer#GetPlt(a:buffer))
```
w0rp: Use `ale#Escape()` for escaping command line arguments. `fnameescape` is only intended for use with `:execute`.

ale_linters/erlang/dialyzer.vim
```diff
@@ -0,0 +1,88 @@
...
+    \   . ' -Wunmatched_returns'
+    \   . ' -Werror_handling'
+    \   . ' -Wrace_conditions'
+    \   . ' -Wunderspecs'
+    \   . ' %s'
```
w0rp: If this checks the file on disk, set `'lint_file': 1` below, so the linters is only run when the file is changed on disk.

ale_linters/erlang/dialyzer.vim
```diff
@@ -0,0 +1,88 @@
...
+" Description: Define a checker that runs dialyzer on Erlang files.
+
+function! ale_linters#erlang#dialyzer#GetRebar3Profile(buffer) abort
+    return ale#Var(a:buffer, 'erlang_dialyzer_rebar3_profile')
+endfunction
+
+function! ale_linters#erlang#dialyzer#FindPlt(buffer) abort
+    let l:rebar3_profile = ale_linters#erlang#dialyzer#GetRebar3Profile(a:buffer)
+    let l:plt_file_directory = ale#path#FindNearestDirectory(a:buffer, '_build' . l:rebar3_profile)
+    let l:plt_file = split(globpath(l:plt_file_directory, '/*_plt'), '\n')
```
w0rp: Check if `l:plt_file_directory` is empty before running `globpath`.

w0rp:
Cheers! :beers:

PR #2511, opened 2019-05-16
Attempt to resurrect nvim highlight

andreypopp:
Second attempt at #1274 prev #2169

This one makes sure we don't pass negative line numbers to nvim API so it
doesn't errors. Also update Dockerfile with nvim 0.3.5 and added more tests.


andreypopp:
Needs updated w0rp/ale image, otherwise tests run with an older nvim

w0rp:
#2638 has been merged now, which should address this. Thanks for your work on this!

PR #2520, opened 2019-05-21
Fix eclipselsp workspace config

### LINKED ISSUE #2519: eclipselsp fails to correctly initialize/sync gradle projects.
Reported at eclipse/eclipse.jdt.ls#967 but this seems to be an issue with ALE eclipselsp linter and not the language server itself.

When working with gradle projects, ALE fails to auto-complete classes of third party dependencies.

hsanson:
Fixes command line used to start eclipselsp to set workspace correctly. Also adds a conf variable to set a specific eclipse workspace path.

Fixes #2519

### DIFF
--- a/ale_linters/java/eclipselsp.vim
+++ b/ale_linters/java/eclipselsp.vim
@@ -5,6 +5,7 @@ let s:version_cache = {}
 
 call ale#Set('java_eclipselsp_path', ale#path#Simplify($HOME . '/eclipse.jdt.ls'))
 call ale#Set('java_eclipselsp_config_path', '')
+call ale#Set('java_eclipselsp_workspace_path', '')
 call ale#Set('java_eclipselsp_executable', 'java')
 
 function! ale_linters#java#eclipselsp#Executable(buffer) abort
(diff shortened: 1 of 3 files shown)


ale_linters/java/eclipselsp.vim
```diff
@@ -76,6 +77,16 @@ function! ale_linters#java#eclipselsp#CommandWithVersion(buffer, version_lines,
...
 endfunction
 
+function! ale_linters#java#eclipselsp#WorkspacePath(buffer) abort
+    let l:wspath = ale#Var(a:buffer, 'java_eclipselsp_workspace_path')
+
+    if !empty(l:wspath)
+        return l:wspath
+    endif
+
+    return ale#path#Simplify(ale#java#FindProjectRoot(a:buffer) . '/..')
```
w0rp: You can use `return ale#path#Dirname(ale#java#FindProjectRoot(a:buffer))`. It gets the dirname for a path and handles the path being empty.

GrimKriegor:
@hsanson Thank you so much for this.

I have spent hours and hours trying to figure out what was wrong with my setup and right after I posted [this](https://github.com/eclipse/eclipse.jdt.ls/issues/967#issuecomment-494080895) you came up with a solution right away.

Just tested pulling these changes and opening up a Java+Gradle project and can confirm Gradle gets triggered, the dependencies get downloaded and third-party libraries now have auto-completion and hover support.

Keep up the good work!

w0rp:
You'll have to rebase this now. While you're at it, add `ale#Escape()` to all of the paths used in the command, or the command can fail if any of the paths contain spaces, etc.

ale_linters/java/eclipselsp.vim
hsanson: Changed to use Dirname.

hsanson:
@GrimKriegor thanks for reporting. This allows me to find out issues that require attention.

hsanson:
@w0rp rebased and added ale#Escape() to all paths in the command.

w0rp:
I'll merge this, though I have a question. If the workspace path is empty, and we pass `-data ''` to the command, will it still work? If it doesn't, we should probably check if the workspace path is empty and return `''` so no command is run instead.

w0rp:
Cheers! :beers:

PR #2523, opened 2019-05-21
Allow running eclipselsp as installed by system package on GNU/Linux

GrimKriegor:
This is a continuation of https://github.com/w0rp/ale/pull/2514

Thanks for the feedback on the last PR and please allow me to suggest a different approach based on the suggestions presented there.

I removed the code that creates the `~/.jdtls` config directory and replaced it with an optional parameter that takes a path to the config directory. If this parameter is not present then ALE keeps the previous behavior of looking for the config directory within eclipselsp installation directory.
I've also added a mention to this in the documentation.
This way ALE has no side-effects and does no configuration for eclipselsp but still allows the users to have the configuration directory wherever they wish.

Is it acceptable like this?

Please note that the main goal of this PR is to have ALE working with eclipselsp when it is installed by a distro package. As is, ALE assumes that the config dir is within the installation dir. That can't be the case with a distro package because eclipselsp is installed to a non-writable dir like `/usr/share/java/jdtls/`.

Should I attempt adding any particular tests to cover these additions?

This implementation was inspired by https://github.com/w0rp/ale/pull/2520.

Thank you for your time.

### DIFF
--- a/ale_linters/java/eclipselsp.vim
+++ b/ale_linters/java/eclipselsp.vim
@@ -32,18 +32,36 @@ function! ale_linters#java#eclipselsp#JarPath(buffer) abort
         return l:files[0]
     endif
 
+    " Search jar file within system package path
+    let l:files = globpath('/usr/share/java/jdtls/plugins', 'org.eclipse.equinox.launcher_\d\.\d\.\d\d\d\.*\.jar', 1, 1)
+
+    if len(l:files) == 1
+        return l:files[0]
+    endif
+
     return ''
 endfunction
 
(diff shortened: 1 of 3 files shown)


GrimKriegor:
Thanks a lot for the feedback, would the test included in the commit above be acceptable and desirable?

w0rp:
That's good. The Windows test fails because the test path needs to be simplified for Windows, but I'll fix that. I also noticed that the paths in the command aren't escaped, so I'll fix those.

w0rp:
Cheers! :beers:

PR #2529, opened 2019-05-23
Add a terraform linter

maxwell-k:
This linter uses the check functionality built into terraform. ALE
already has a fixer using `terraform fmt` but this doesn't provide error
messages. ALE already has a linter using `tflint` but this requires an
extra application to be installed.

For example this linter will give a warning that ! is an illegal
character in the line below:

    variable "example" !{}

This linter runs the buffer through the command below and parses the
output:

    terraform fmt -no-color -check=true -

This commit includes a basic implementation, documentation and tests.
The only option is to control which executable is run.

Tested with:

    $ terraform -version
    Terraform v0.11.13


w0rp:
Cheers! :beers:

PR #2539, opened 2019-05-27
Feature/add ant support

### LINKED ISSUE #2538: Java Ant classpath
Detecting the Java classpath is necessary for properly linting a Java file.

We already are already able to autodetect classpath from [Maven](#361) and [Gradle](#745), do you  think it would be possible to do the same for Apache Ant? I find Ant to be minimalist and easy to use for quick projects. Neomake [supports](https://github.com/neomake/neomake/pull/1544) Ant configuration via a simple task that prints the classpath-- maybe we could do something similar.

I've never written anything substantial in VimScript, but with a few pointers, I could also look into implementing this. Or if there's anyone who's a Vim Wizard and would be willing to code this out, I would be very appreciative!

ndrewtl:
This PR fixes #2538 and patches in support for inferring classpath from Ant `build.xml` files. The code is based on the equivalent [Neomake](https://github.com/neomake/neomake/pull/1544) functionality and is written to resemble the Gradle classpath-inferrer. Tests are written.


ndrewtl:
Build is failing on Windows build, I think due to path errors.

In order to get one of the tests to work, I need to 'trick' the container into thinking that an Ant executable exists. I do this by editing the `$PATH` variable to point at an `ant` stub I created. I'm not sure how Windows handles paths, so I don't know how to accomplish the same thing on for it.
I also don't have a Windows machine. Could I have some help in making this test also work on Appveyor?

test/ant-test-files/non-ant-project/src/several/namespace/layers/A.java
```diff
@@ -0,0 +1,2 @@
+package several.namespace.layers;
+public class A{}
```
w0rp: You can delete all of the `.java` files and the files that need to exist for the purposes of tests can be 0 bytes in size. The contents of the files don't matter.

test/ant-test-files/bin/ant
```diff
@@ -0,0 +1,4 @@
+# /bin/bash
+# this file exists to provide an 'executable' to run ant tests
+# it doesn't need to do anything-- it just needs to exist
+echo ':)'
```
w0rp: The test for this fails in the Windows tests because it's actually trying to find `ant.exe` in path. Add a 0 byte `ant.exe` file in this directory too, and I think it should work.

test/ant-test-files/non-ant-project/src/several/namespace/layers/A.java
ndrewtl: Deleted the Java files in a02926ff87cd9b8a94ce70e005b52e34100bc9a3

test/ant-test-files/bin/ant
ndrewtl: Zeroed out the `ant` executable, and added an `ant.exe` file in d1b90eaadc2c15035262d271ecbbe0943498d932

w0rp:
Cheers! :beers:

PR #2541, opened 2019-05-28
fix neovim missing setbufline

jwkvam:
This fixes https://github.com/w0rp/ale/issues/1739 for neovim head by adding a check for setbufline as well 😬 

I do not have a test, I may be able to write one given much more time but I haven't looked into it.


w0rp:
Cheers! :beers: Thanks for pointing this out.

PR #2542, opened 2019-05-28
Improve java checkstyle command configuration.

hsanson:
The g:ale_java_checkstyle_options only worked with absolute paths making
it difficult to have a per project configuration. This MR introduces a
new g:ale_java_checkstyle_config configuration. This option can be an
absolute path or a filename. When a filename is set then the nearest
file that matches it in the current buffer folder or any of its parents
will be used.

Note: The above is a breaking change since it replaces the
g:ale_java_checkstyle_options variable with a different one.

Also a new configuration g:ale_java_checkstyle_executable is added. This
allows users to create more specialized launch scripts (#1305) to invoke
checkstyle.

### DIFF
--- a/ale_linters/java/checkstyle.vim
+++ b/ale_linters/java/checkstyle.vim
@@ -1,6 +1,9 @@
 " Author: Devon Meunier <devon.meunier@gmail.com>
 " Description: checkstyle for Java files
 
+call ale#Set('java_checkstyle_executable', 'checkstyle')
+call ale#Set('java_checkstyle_config', 'google_checks.xml')
+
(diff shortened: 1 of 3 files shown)


hsanson:
Added the _options setting back. If defined I use this setting to build the command and ignore the new ones. If this option is defined is expected to have '-c' so no need to check for it. Also mentioned in the docs that the other settings are preferred over this one.

w0rp:
I rewrote this in the latest commit. It should support absolute paths for the configuration file in the new `_config` option, relative paths, prefer `-c` in `_options` if the configuration file is set in the options, and the executable is now configurable.

PR #2543, opened 2019-05-28
Fix java checkstyle handler.

hsanson:
The checkstyle handler is capable of parsing the new and old output
formats. Unfortunately there are some particular output messages that
matched both the new and old regular expressions:

   [WARN] whatever:11:7: WhitespaceAround: ''if'' is not followed by whitespace. [WhitespaceAround]

This caused ALE to report extra errors since the message was being
matched twice, once as a warning and another (incorrect) old formatted
error.

This MR fixes this by stopping any parsing using the old format regexp
is any errors of the new format are correcly parsed. There is no reason
to expect checkstyle to output both styles in the same report.

### DIFF
--- a/ale_linters/java/checkstyle.vim
+++ b/ale_linters/java/checkstyle.vim
@@ -17,6 +17,10 @@ function! ale_linters#java#checkstyle#Handle(buffer, lines) abort
         \})
     endfor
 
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

PR #2549, opened 2019-05-31
Add API for custom LSP requests

### LINKED ISSUE #2474: Add API for custom LSP requests
Hi!
I was wondering if it would be possible to allow users to make custom requests to the LSP servers, either by adding a public and documented way to register custom callbacks or, much better imho, to expose a public function with an interface like

```viml
ale#lsp#request(server, method, parameters, callback)
```

All other major vim LSP clients expose a function of this kind (e.g. [vim-lsp](https://github.com/prabirshrestha/vim-lsp/blob/fe5057641cadba27caa26df74efa896e4255308e/autoload/lsp.vim#L731), [vim-lsc](https://github.com/natebosch/vim-lsc/blob/9616bcb14c0128a6722806dafe2b6f325fa0c26f/autoload/lsc/server.vim#L98), [LanguageClient-neovim](https://github.com/autozimu/LanguageClient-neovim/blob/0ed9b69dca49c415390a8317b19149f97ae093fa/autoload/LanguageClient.vim#L626), [coc.nvim](https://github.com/neoclide/coc.nvim/blob/eb747ceb75cf022f02d0916e820c7ba1a357135f/plugin/coc.vim#L38)), and this allows other plugins to use data from LSP servers without needing to reimplement an LSP client, but rather relying on the public API of some existing LSP client.

As a practical example, recently I have been developing [vim-ccls](https://github.com/m-pilia/vim-ccls), a small plugin to navigate symbol hierarchies in C/C++ (provided by the [ccls](https://github.com/MaskRay/ccls) language server) that works on top of coc.nvim, vim-lsp, vim-lsc, and LanguageClient-neovim. I also added [experimental support for ALE](https://github.com/m-pilia/vim-ccls/blob/7a1f3bed873800209e33a40e9ac0be0935d39d78/autoload/ccls/lsp.vim#L119-L139), but the way I did it is brittle, since it relies on some not-really-public part of the API, e.g. [ale#lsp#GetConnections()](https://github.com/w0rp/ale/blob/7f0954b89ef7c43e5fb7b6490665e9814c8205b2/autoload/ale/lsp.vim#L65), and relies on some knowledge of the internal implementation of ALE that may change at any time.

There are more popular examples of plugins that could benefit from this, the first example that comes to my mind is [vista.vim](https://github.com/liuchengxu/vista.vim), that implements a tag-bar using either vim-lsp, coc.nvim, or LanguageClient-neovim as LSP back-end.

I would see this as a reasonable approach until vim/neovim implement some built-in LSP API (I think the neovim community is [already working](https://github.com/neovim/neovim/pull/6856) on something in that direction).

Last but not least, I could help by implementing this and opening a PR, if the idea is acceptable.

m-pilia:
Implement a function `ale#lsp_linter#SendRequest` that allows to send
custom LSP requests to an enabled LSP linter.

Resolves #2474

### DIFF
--- a/autoload/ale/lsp.vim
+++ b/autoload/ale/lsp.vim
@@ -5,6 +5,9 @@
 let s:connections = get(s:, 'connections', {})
 let g:ale_lsp_next_message_id = 1
 
+" A Dictionary to track one-shot callbacks for custom LSP requests
+let s:custom_callbacks = get(s:, 'custom_callbacks', {})
+
 " Given an id, which can be an executable or address, and a project path,
 " create a new connection if needed. Return a unique ID for the connection.
 function! ale#lsp#Register(executable_or_address, project, init_options) abort
@@ -296,10 +299,19 @@ function! ale#lsp#HandleMessage(conn_id, message) abort
     " responses.
     if l:conn.initialized
         for l:response in l:response_list
-            " Call all of the registered handlers with the response.
-            for l:Callback in l:conn.callback_list
-                call ale#util#GetFunction(l:Callback)(a:conn_id, l:response)
-            endfor
+            if has_key(l:response, 'id') && has_key(s:custom_callbacks, l:response.id)
+                " Response to a custom request, call the registered one-shot handler.
+                try
+                    call s:custom_callbacks[l:response.id](l:response)
+                finally
+                    call remove(s:custom_callbacks, l:response.id)
+                endtry
+            else
+                " Call all of the registered handlers with the response.
+                for l:Callback in l:conn.callback_list
+                    call ale#util#GetFunction(l:Callback)(a:conn_id, l:response)
+                endfor
+            endif
         endfor
     endif
 endfunction
@@ -523,6 +535,18 @@ function! ale#lsp#Send(conn_id, message) abort
     call s:SendMessageData(l:conn, l:data)
 
     return l:id == 0 ? -1 : l:id
+endfunction
+
+" Send a custom request to an LSP server.
+" The given callback is called on response.
+function! ale#lsp#SendCustomRequest(conn_id, message, Callback) abort
+    let l:id = ale#lsp#Send(a:conn_id, a:message)
+
+    if l:id > 0
+        let s:custom_callbacks[l:id] = a:Callback
+    endif
+
+    return l:id
 endfunction
 
 " Notify LSP servers or tsserver if a document is opened, if needed.
--- a/autoload/ale/lsp.vim
+++ b/autoload/ale/lsp.vim
@@ -5,9 +5,6 @@
 let s:connections = get(s:, 'connections', {})
 let g:ale_lsp_next_message_id = 1
 
-" A Dictionary to track one-shot callbacks for custom LSP requests
-let s:custom_callbacks = get(s:, 'custom_callbacks', {})
-
 " Given an id, which can be an executable or address, and a project path,
 " create a new connection if needed. Return a unique ID for the connection.
 function! ale#lsp#Register(executable_or_address, project, init_options) abort
@@ -299,19 +296,10 @@ function! ale#lsp#HandleMessage(conn_id, message) abort
     " responses.
     if l:conn.initialized
         for l:response in l:response_list
-            if has_key(l:response, 'id') && has_key(s:custom_callbacks, l:response.id)
-                " Response to a custom request, call the registered one-shot handler.
-                try
-                    call s:custom_callbacks[l:response.id](l:response)
-                finally
-                    call remove(s:custom_callbacks, l:response.id)
-                endtry
-            else
-                " Call all of the registered handlers with the response.
-                for l:Callback in l:conn.callback_list
-                    call ale#util#GetFunction(l:Callback)(a:conn_id, l:response)
-                endfor
-            endif
+            " Call all of the registered handlers with the response.
+            for l:Callback in l:conn.callback_list
+                call ale#util#GetFunction(l:Callback)(a:conn_id, l:response)
+            endfor
         endfor
     endif
 endfunction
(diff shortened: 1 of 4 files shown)


autoload/ale/lsp_linter.vim
```diff
@@ -413,3 +413,29 @@ endfunction
...
+        throw 'Linter "' . a:linter_name . '" not found!'
+    endif
+
+    let l:linter = l:linter_list[0]
+    let l:executable_or_address = ''
+
+    if l:linter.lsp is# 'socket'
+        let l:executable_or_address = ale#linter#GetAddress(a:buffer, l:linter)
+    else
+        let l:executable_or_address = ale#linter#GetExecutable(a:buffer, l:linter)
```
w0rp: The results of these could be deferred objects, which are Dictionary values for computing the executable or address based on some other command, etc. We'll have to handle those here.

autoload/ale/lsp_linter.vim
```diff
@@ -413,3 +413,29 @@ endfunction
...
+    endif
+
+    let l:root = ale#util#GetFunction(l:linter.project_root)(a:buffer)
+    let l:conn_id = l:executable_or_address . ':' . l:root
```
w0rp: The connection ID should be computed somewhere in `lsp.vim` and returned here instead. That way this code will continue to work if the format changes.

autoload/ale/lsp_linter.vim
```diff
@@ -413,3 +413,29 @@ endfunction
...
+    let l:linter = l:linter_list[0]
+    let l:executable_or_address = ''
+
+    if l:linter.lsp is# 'socket'
+        let l:executable_or_address = ale#linter#GetAddress(a:buffer, l:linter)
+    else
+        let l:executable_or_address = ale#linter#GetExecutable(a:buffer, l:linter)
+    endif
+
+    let l:root = ale#util#GetFunction(l:linter.project_root)(a:buffer)
```
w0rp: The project root could just be a string, so handle that like we do elsewhere.

autoload/ale/lsp.vim
```diff
@@ -296,10 +299,19 @@ function! ale#lsp#HandleMessage(conn_id, message) abort
...
+                    call s:custom_callbacks[l:response.id](l:response)
+                finally
+                    call remove(s:custom_callbacks, l:response.id)
+                endtry
+            else
+                " Call all of the registered handlers with the response.
+                for l:Callback in l:conn.callback_list
+                    call ale#util#GetFunction(l:Callback)(a:conn_id, l:response)
+                endfor
+            endif
```
w0rp: In order to avoid making this part of the code slower, let's keep this code how it is, and set up a callback which handles the message with the custom callbacks dictionary in the `lsp_linter.vim` file instead. That way we won't have to check for custom callbacks on every response when you never send custom requests.

autoload/ale/lsp_linter.vim
m-pilia: Removed, thanks!

autoload/ale/lsp_linter.vim
m-pilia: Removed, thanks!

autoload/ale/lsp_linter.vim
m-pilia: I didn't realise that, thanks for pointing it out! Removed!

autoload/ale/lsp.vim
m-pilia: Done, I added a callback.

m-pilia:
Thanks for the review comments! I didn't realise I could re-use the logic of `StartLSP`.  If I am not overlooking anything, now it should be better.

autoload/ale/lsp_linter.vim
```diff
@@ -407,9 +410,50 @@ endfunction
...
+        throw 'Linter "' . a:linter_name . '" not found!'
+    endif
+
+    let l:linter = l:linter_list[0]
+
+    if empty(l:linter.lsp)
+        throw 'Linter "' . a:linter_name . '" does not support LSP!'
+    endif
+
+    let l:message = [0, a:method, a:parameters]
```
w0rp: I just realised that we'll need one more parameter for the value that is `0` here. That value is `1` when sending notifications, and JSON-RPC spec requires that notifications are sent without the `id` parameter used for other messages.

Maybe we could replace `method, parameters`, above with `message`, and document the format of the messages as `[is_notification, method, parameters]` in the help text. That will match the format returned by functions in `autoload/ale/lsp/message.vim`, some of which we might decide to expose later.

w0rp: We should make the callback optional for notifications, but required for messages that aren't notifications, as it shouldn't be stored in a Dictionary and won't be used. No responses are sent for notifications.

m-pilia: Good point, I did not think of this use case. That also requires to make the handler argument optional, since no response is expected when a notification is sent.

autoload/ale/lsp_linter.vim
```diff
@@ -420,24 +420,25 @@ endfunction
 
 function! s:HandleLSPResponseToCustomRequests(conn_id, response) abort
     if has_key(a:response, 'id')
-    \&& has_key(s:custom_callbacks_map, a:response.id)
-        let l:Callback = remove(s:custom_callbacks_map, a:response.id)
-        call l:Callback(a:response)
+    \&& has_key(s:custom_handlers_map, a:response.id)
+        let l:Handler = remove(s:custom_handlers_map, a:response.id)
+        call l:Handler(a:response)
```
m-pilia: I thought using `Handler` here instead of `Callback` may help to make the code a bit more readable, since there is another layer of callbacks above it that are already called `Callback`s.

w0rp:
Cheers! :beers:

PR #2551, opened 2019-06-02
Use JSON output for ESLint and fix tsserver column

laino:
This makes ale use ESLint's JSON formatter/reporter instead of the default line-based one. Besides being cleaner, this has the added benefit of giving us native ``end_col`` values:

image
_Old vs. New_

I have kept the old parser around, because the ``standard`` linter [does not support JSON output yet](https://github.com/standard/standard/issues/1286), and uses the ESLint output parser internally. The ``xo`` linter has been migrated to the JSON parser though.

The new handler is at ``ale#handlers#eslint#HandleJSON``, while the old line-based parser is still at ``ale#handlers#eslint#Handle``.

I have [added tests](https://github.com/w0rp/ale/pull/2551/commits/79d1b99067878bbec129cfd29a7c9587c276dec6#diff-e9f1d2cc3405f58a5e784ea80c733159R1) 1:1 as they existed for the old parser, except this time for JSON output of course.

There was a test that dealt with something that is only relevant for the old parser, which is for ESLint adding "Error/" in front of some error codes. ESLint doesn't do this with JSON output, so that test has no equivalent.

Additionally tsserver appears to report ``[a, b[`` ranges instead of ``[a,b]``, so I fixed that off-by-one error as well. This is also visible in the screenshot above (red error).


ale_linters/typescript/xo.vim
```diff
@@ -11,7 +11,7 @@ endfunction
 function! ale_linters#typescript#xo#GetCommand(buffer) abort
     return ale#Escape(ale_linters#typescript#xo#GetExecutable(a:buffer))
     \   . ale#Pad(ale#Var(a:buffer, 'typescript_xo_options'))
-    \   . ' --reporter unix --stdin --stdin-filename %s'
+    \   . ' --reporter json --stdin --stdin-filename %s'
```
w0rp: The reporter was updated here, but not the handler below.

autoload/ale/lsp/response.vim
```diff
@@ -90,7 +90,7 @@ function! ale#lsp#response#ReadTSServerDiagnostics(response) abort
         \   'lnum': l:diagnostic.start.line,
         \   'col': l:diagnostic.start.offset,
         \   'end_lnum': l:diagnostic.end.line,
-        \   'end_col': l:diagnostic.end.offset,
+        \   'end_col': l:diagnostic.end.offset - 1,
```
w0rp: I'm not sure if this is right. I'll check it with some errors in my codebase.

ale_linters/typescript/xo.vim
laino: Yup. I must've forgotten it when I added the "old" code back in.

I'll fix this stuff tomorrow.

laino:
@w0rp Have you been able to confirm the correctness of the tsserver fix on your end?

w0rp:
I checked it now and it works. Cheers! :beers:

PR #2558, opened 2019-06-04
Fix javalsp command callback.

hsanson:
The command used to invoke the language server is missing some options
to include additional java modules. Without these modules the server
was not working properly.

The correct command can be found in a `launcher` script on the same
directory the `java` executable for the language server is found.

This commit changes the docs to prefer the launcher script over the java
executable. For backward compatibility it also fixes the command
invocation in case the java executable is configured.

### DIFF
--- a/ale_linters/java/javalsp.vim
+++ b/ale_linters/java/javalsp.vim
@@ -1,7 +1,7 @@
 " Author: Horacio Sanson <https://github.com/hsanson>
 " Description: Support for the Java language server https://github.com/georgewfraser/vscode-javac
 
-call ale#Set('java_javalsp_executable', 'java')
+call ale#Set('java_javalsp_executable', 'launcher')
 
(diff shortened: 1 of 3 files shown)


hsanson:
@w0rp I thought about this for a while and cannot really find a way to prevent getting the wrong binary . The only solution I can see is to force users to explicitly set a value for `g:ale_java_javalsp_executable`.

When installing the java-language-server manually there is no pre-defined location so it can be anywhere on the users home dir and when installing it via VSCode plugin there is no binary at all.  Not sure how vscode invokes the language server.

Falling back to `java` is not good either. it must be the `java` binary generated when compiling the language server. It cannot be the system installed java. Therefore, this also must be explicitly set to the absolute path of the language server provided binaries.

w0rp:
Cheers! :beers:

PR #2561, opened 2019-06-05
Add javalsp lsp options

hsanson:
This MR adds two new configuration options `g:ale_java_javalsp_external_deps` and `g:ale_java_javalsp_classpaths` that are analog to the `java.externalDependencies` and `java.classPath` configurations available in the [vscode/settings.json](https://github.com/georgewfraser/java-language-server#settings) file.

With these is possible to add Android deps to the configuration and get auto-completion for Android classes.
 
Moreover these options can be set to a vim function that provides means  to dynamically update the dependencies. The only concern here is if the function is executed in background by ALE or not. Retrieving dependencies using gradle/mnv can take a long time on large projects.

FYI: @DonnieWest @GrimKriegor this may be of interest for you.

### DIFF
--- a/ale_linters/java/javalsp.vim
+++ b/ale_linters/java/javalsp.vim
@@ -2,9 +2,36 @@
 " Description: Support for the Java language server https://github.com/georgewfraser/vscode-javac
 
 call ale#Set('java_javalsp_executable', '')
+call ale#Set('java_javalsp_external_deps', [])
+call ale#Set('java_javalsp_classpaths', [])
 
 function! ale_linters#java#javalsp#Executable(buffer) abort
     return ale#Var(a:buffer, 'java_javalsp_executable')
+endfunction
+
+function! ale_linters#java#javalsp#Deps(buffer) abort
+    let l:deps = ale#Var(a:buffer, 'java_javalsp_external_deps')
+
+    return type(l:deps) is v:t_list
+    \ ? l:deps
+    \ : function(l:deps)(a:buffer)
+endfunction
+
+function! ale_linters#java#javalsp#Paths(buffer) abort
+    let l:paths = ale#Var(a:buffer, 'java_javalsp_classpaths')
+
+    return type(l:paths) is v:t_list
+    \ ? l:paths
+    \ : function(l:paths)(a:buffer)
+endfunction
+
+function! ale_linters#java#javalsp#Options(buffer) abort
+    return {
+    \ 'java': {
+    \    'externalDependencies': ale_linters#java#javalsp#Deps(a:buffer),
+    \    'classPath': ale_linters#java#javalsp#Paths(a:buffer)
+    \ }
+    \}
 endfunction
 
 function! ale_linters#java#javalsp#Command(buffer) abort
@@ -38,4 +65,5 @@ call ale#linter#Define('java', {
 \   'command': function('ale_linters#java#javalsp#Command'),
 \   'language': 'java',
 \   'project_root': function('ale#java#FindProjectRoot'),
+\   'lsp_config': function('ale_linters#java#javalsp#Options')
 \})


DonnieWest:
@hsanson thanks for the mention and the contribution - this looks very interesting :)

GrimKriegor:
@hsanson here you are saving the day once again ahah.
Thanks a lot for this!
I am in the process of getting `javalsp` working with my setup on GNU/Linux so that I can try this out and report back.
If any Arch user comes across this, [here is](https://aur.archlinux.org/packages/java-language-server/) a recently baked AUR package for `javalsp`.

w0rp:
I'll wait for #2558 to be finished first.

GrimKriegor:
This is most excellent.
By populating `g:ale_java_javalsp_external_deps` with a cleaned up output of `gradle ${project}:dependencies` I am getting auto-completion for third party libraries.

```
let g:ale_java_javalsp_external_deps = [
  \ "io.reactivex.rxjava2:rxandroid:2.0.1",
  \ "io.reactivex.rxjava2:rxjava:2.0.1",
(snip)
```
As seen here:
third-party-autocomp

Out of curiosity, what is the most elegant way to include the core Android libraries?
Using `ale_java_javalsp_external_deps` makes things easy, but I am unsure how to add the core Android libs without resorting to setting `ale_java_javalsp_classpaths`.
I have tried using both, but as expected, the `classpaths` parameter overrides whatever effect `external_deps` has.

Probably populating the classpath from scratch, like `vim-android` does for `$CLASSPATH` would be the way to go, but I am wondering if there is a simpler way.

Once again, thank you for your time and awesome contributions.

hsanson:
@GrimKriegor this PR and my recent changes to vim-android are the first steps in my plan to enable java/android development in vim.

This PR allows to set a function name for g:ale_java_javalsp_classpaths and  vim-android has functions that coincidentally return and array of classpath dependencies for [gradle](https://github.com/hsanson/vim-android/blob/master/autoload/gradle.vim#L465) and [android]( https://github.com/hsanson/vim-android/blob/master/autoload/android.vim#L208) projects.

I still need to change some things on vim-android to get it working along with ALE. At this moment vim-android conflicts with ALE because both manipulate the loclist and the signs. Also need to evaluate the best place to execute gradle#sync() to extract dependencies that can take a long time to complete, etc. 

Possibly on the weekend I can continue with the experiments. Hopefully full porting of vim-android to ALE won't be necessary, that would take a long time to complete.

GrimKriegor:
@hsanson amazing work as always. Thanks a lot!

I have been looking at `vim-android` searching for a way to use those arrays of classpath dependencies to somehow populate the `ale_java_javalsp_classpaths` variable. I am extremely happy to learn `vim-android` is progressing in this direction.

While my Vimscript is poor, I will attempt crafting a crude function to generate something `ale_java_javalsp_classpaths` can handle, for the time being.

hsanson:
@GrimKriegor adding the following to your vim configuration should do:

```vim
 " Greatly speeds up dependency caching on subsequent runs.
let g:gradle_daemon=1 

 " Disable automatic gradle sync. Leave this to ALE. Requires latest vim-android
let g:gradle_sync_on_load=0 

 " Disable signs so it does not interfere with ALE.
let g:gradle_show_signs=0 

" Function that syncs gradle deps and returns an array with all gradle/android classpath dependencies.
" Requires latest vim-android plugin.
function! LoadDeps(buffer) abort
  call gradle#sync()
  return extend(gradle#classPaths(), android#classPaths())
endfunction

" Configure ALE javalsp linter to use the above method to load gradle/android dependencies.
" Requires PR https://github.com/w0rp/ale/pull/2561
let g:ale_java_javalsp_classpaths = 'LoadDeps'


" Autocommand to update deps when build.gradle is edited.
augroup GradleGroup
  autocmd!
  au BufWrite build.gradle cal gradle#sync()
augroup END
```

Still some dependencies are not found like "android.support.v7.app.AppCompatActivity". Not sure why.

GrimKriegor:
Thanks @hsanson, excellent work!
Completion is working for me with the configurations above.
However, it would appear that makes `LoadDeps` and therefore `gradle#sync()` run every time I type in a short word in insert mode, making Vim hang for a few seconds making insert mode unusable for me. I assume this is by design so the `ale_java_javalsp_classpaths` remains up to date.
Maybe this could help. https://github.com/hsanson/vim-android/issues/70
Or maybe I am doing something wrong.
I am considering populating `ale_java_javalsp_classpaths` with the `gradle` methods from `vim-android`, on buffer initialization, for the time being, like so:

**~/.vim/after/ftplugin/java.vim**
```
(...)
call gradle#sync()
let b:ale_java_javalsp_classpaths = extend(gradle#classPaths(), android#classPaths())
(...)
```

What are your thoughts?

hsanson:
@GrimKriegor interesting, I did not realize this was happening. Not sure why this is invoked on every key stroke. Maybe @w0rp can help here, I am not too familiar with ALE internals.

On my machine the LoadDeps method is also being called on every keystroke but it seems to be on the background and my vim does not hang.

As a workaround you may check for the cache status and run gradle#sync() only if the cache is not set:

```vim
function! LoadDeps(buffer) abort
  if !gradle#isGradleDepsCached()
    call gradle#sync()
  endif
  return extend(gradle#classPaths(), android#classPaths())
endfunction
```

hsanson:
@w0rp this is what I am trying to achieve. Be able to set an array of dependencies and inject them to ALE.  Unfortunately I cannot see a way to trigger the "didChangeConfiguration" command to pass these dependencies to the LSP server. Only on initialization this command is sent to LSP server.

There is a method `ale#lsp#UpdateConfig()` that does this but requires the channel connection id `l:conn_id` as first argument that is not easy to obtain as far as I can see. 

If there was a method to update the LSP config that requires only the buffer and the configuration dictionary would be great. This allows for dynamically updating the configuration by external means (e.g. other plugins).

w0rp:
We could add a command for manually sending a message for the configuration being changed. Messages are sent automatically if the Dictionary has changed since the last time other actions were taken, like requesting completion results, etc.

Worry about that later. For now, just make the whole Dictionary configurable.

GrimKriegor:
@hsanson, that workaround is working quite well for me, thanks!

GrimKriegor:
@hsanson btw:

>On my machine the LoadDeps method is also being called on every keystroke but it seems to be on the background and my vim does not hang.

Could it be because you are using Neovim?

hsanson:
@w0rp I created a separate PR that allows to set a simple dictionary (#2600) deprecating this PR. Still not sure how to dynamically set the configuration and trigger the didChangeConfiguration command on the language server but trust it is or will be possible.

PR #2563, opened 2019-06-06
prettier: Support experimental languages (Handlebars)

dcyriller:
In case the filetype has an exprimental support in Prettier, the --parser flag has to be set manually.

I'm not 100% sure if it's relevant or not to tie ale to an experimental flag in Prettier. But the patch is done, so I leave it to your judgement @w0rp 🙂.

For the time being, I'm using this workaround: https://github.com/w0rp/ale/issues/2247#issuecomment-457660222

Would close https://github.com/w0rp/ale/issues/2247

### DIFF
--- a/autoload/ale/fixers/prettier.vim
+++ b/autoload/ale/fixers/prettier.vim
@@ -41,36 +41,49 @@ function! ale#fixers#prettier#ApplyFixForVersion(buffer, version) abort
 
     " Append the --parser flag depending on the current filetype (unless it's
     " already set in g:javascript_prettier_options).
-    if empty(expand('#' . a:buffer . ':e')) && match(l:options, '--parser') == -1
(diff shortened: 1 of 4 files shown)


w0rp:
Actually, that doesn't work. Because of the compound filetype and `html` being matched first, it selected the wrong parser. In that case, just use the first filetype check to set `l:parser` to `'glimmer'`, and check `!exists('l:parser')`.

dcyriller:
And what do you say of switching `html` and `handlebars` lines in `l:prettier_parsers` definition (with a comment)? I liked your solution being so light.

dcyriller:
> And what do you say of switching html and handlebars lines in l:prettier_parsers definition (with a comment)? I liked your solution being so light.

This doesn't work. Your suggestion is implemented.

w0rp:
That's much better. Cheers! :beers:

Issue #2450, opened 2019-04-20
Add R Markdown (.Rmd) support with R lintr


cjber:
The implementation of the  current linter and fixer for .R scripts I believe would both work with .Rmd files as they currently are. Both of which use the `library('styler')` lint I believe (from [here](https://github.com/tvatter/ale/pull/1)). For example I am using this function:

```
function! Rlint()
        let filename = expand("%:r:t")
        let rcmd "require('styler');
            \styler::style_file(\"" . filename . ".Rmd\")"
        call g:SendCmdToR(rcmd)
endfunction
```
with `'djoshea/vim-autoread'` to auto update the file and `'jalvesaq/Nvim-R'` for the r console connection which works perfectly well with my .Rmd files containing both markdown and R code within chunks. 

I am very new to vim (<1 week) and this is my first post to github so apologies if I make any mistakes or if this seems an overly simple request.

tvatter:
@cjber is totally correct and I can try a PR for that shortly.

For styler, it's just about adding to the list of suggested filetypes in the registry right?

tvatter:
Well, better late than never. I had completely forgotten about that, my bad.

statquant:
Hello, sorry to come back to this but are we supposed to have linting for rmd files ?
For me it works in .r files but not .rmd (with or without a R chunk)

tvatter:
For me, it works. Stupid question, but you have the updated version from ALE (post-June 6, i.e., the date of my PR's merge) right?

tvatter:
So, I just noticed something: in the PR, I used `rmarkdown` as the filetype. Right now, when testing in vim, I openened a `Rmd` file, and the lintr/styler didn't work because it was detected with and `rmd`. Doing `:set filetype=rmarkdown` activated ALE for the file.

tvatter:
To prevent this, I just added `autocmd BufRead,BufNewFile *.Rmd set filetype=rmarkdown` to my `init.vim`.

statquant:
Ah ok, problem is i think 99.99% of people using lintr in vim are using the the nvimr plugin that might expect ft=rmd ...

statquant:
Copied your PR to create another one with rmd

tvatter:
Goog idea, I also use nvim-r and didn't thing of that ^^

tvatter:
Can you please let me know if/when the PR is accepted?

w0rp:
I've merged that now. Thanks!

PR #2564, opened 2019-06-06
Make rmarkdown files work with styler and lintr

tvatter:
See #2450


cjber:
I have been working with rnoweb (.rnw) files recently and they work basically the same as R Markdown so might be worth adding as well if you were happy to.

tvatter:
I don't agree. For rnoweb files, I would rather have the linters/fixers related to latex.

cjber:
Ah of course, I do agree, hadn't thought about that

autoload/ale/linter.vim
```diff
@@ -13,6 +13,7 @@ let s:default_ale_linter_aliases = {
 \   'Dockerfile': 'dockerfile',
 \   'csh': 'sh',
 \   'plaintex': 'tex',
+\   'rmarkdown': 'r',
```
w0rp: Update the documentation as mentioned above, and I will merge this.

tvatter: Added as you suggested, I think that the travis failure is outside of my control:

https://travis-ci.com/w0rp/ale/jobs/206776131#L450

w0rp:
Cheers! :beers:

tvatter:
Thanks!

PR #2565, opened 2019-06-06
Run xml linters on xsd and xslt files

Tharre:
Both xsd and xslt are by definition written in XML, and thus the same
linter(s) can be run to check them for well-formedness.


w0rp:
Cheers! :beers:

PR #2567, opened 2019-06-09
Add support for reorder-python-imports fixer

theevocater:
isort is great, but I've personally prefer reorder-python-imports. The tool has a focus on smaller diffs than isort. reorder-python-imports is also a little smarter than isort which is nice.

I've more or less copied the implementation for this from the existing python fixers. I've also included the same tests they all have for finding the executable and ensuring options are passed.

I've run ./run-tests, and made sure everything passes vint (what a cool tool!).


w0rp:
Cheers! :beers:

PR #2568, opened 2019-06-09
Remove unnecessary asterisk from doc/ale-reasonml

h-michael:
ale-reasonml.txt contains two "must" tags, so we can't create helptags.
And I think we shouldn't use the general words like "must" for helptags.

[related commit](https://github.com/w0rp/ale/commit/507f164a09d2cb3c99983424f9d186d6f4081820)


w0rp:
Cheers! :beers:

PR #2574, opened 2019-06-10
In README, more efficient git clones

enterprisey:
Users don't need the entire git history to use the plugin, so don't download it


w0rp:
Cheers! :beers:

w0rp:
This is a good idea, and I just had another one. I can't remove the images from git history without rewriting all of the commits and ruining everyone's day, but I can move the images to a GitHub issue and then that would reduce the download size greatly.

PR #2577, opened 2019-06-11
Fix checkstyle default configuration.

hsanson:
Checkstyle xml configuration is mandatory and not providing one causes
the tool to fail with the following error:

    Must specify a config XML file.

Checkstyle itself contains a default configuration as part of its
assests named `/google_checks.xml`. Invoking checkstyle with this config
works even if such file does not exists in the file system:

    checkstyle -c /google_checks.xml

This should be the default invocation to allow ALE to use checkstyle
with zero configuration.

Also when a user sets `g:ale_java_checkstyle_config` option, ALE should
use it to invoke checkstyle even if such file does not exists in the
filesystem. This is because checkstyle is able to use configuration files
within JAR files defined in the CLASSPATH. The default `/google_checks.xml`
is an example of such configuration available within a JAR resource.

### DIFF
--- a/ale_linters/java/checkstyle.vim
+++ b/ale_linters/java/checkstyle.vim
@@ -2,7 +2,7 @@
 " Description: checkstyle for Java files
 
 call ale#Set('java_checkstyle_executable', 'checkstyle')
-call ale#Set('java_checkstyle_config', 'google_checks.xml')
+call ale#Set('java_checkstyle_config', '/google_checks.xml')
 call ale#Set('java_checkstyle_options', '')
 
 function! ale_linters#java#checkstyle#Handle(buffer, lines) abort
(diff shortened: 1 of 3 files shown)


ale_linters/java/checkstyle.vim
```diff
@@ -39,11 +39,17 @@ function! ale_linters#java#checkstyle#Handle(buffer, lines) abort
     return l:output
 endfunction
 
+function! s:GetConfig(buffer, config) abort
+    let s:file = ale#path#FindNearestFile(a:buffer, a:config)
```
w0rp: We can probably avoid checking the filesystem at all if the configuration file starts with `/`.

hsanson: Indeed.

w0rp:
Cheers! :beers:

PR #2578, opened 2019-06-11
Elm: Update link to compiler repository

andys8:
Updating the docs because the link was referring to the previous archived repository. It should be `elm/compiler`.


w0rp:
Cheers! :beers:

PR #2586, opened 2019-06-13
Csc linter

hernot:

Hi has been some time since adding mcsc linter meanwhile plenty has change on mono side. The mcs compiler which is utilized by mcs and mcsc linters is not further developed in favour for the csc. The latter differs in the following three points output to stdout options prefixed by / instead of - and no /syntax so no csc based counter part to the ale-cs-mcs linterr. 

I start the pull request now to have Travis check if my assumption that exempt from the above everything already existent for mcsc can be reused for csc. I will need some iterations and thus let you know when ready for final review.

Besides mcsc and csc got additional pattrn to capture messages not specific to source file, which were not captured before,


hernot:
I'm not sure what is wrong and from Travis logs I do fail to deduce what to change to make Travis checks running. 

1) How to run tests locally without docker?
2) Would need your support to figure what is wrong

I have first updated my forks master to upstream master and than created from fork master a new branch to not get in conflict with the previous one. Hopefully nothing went wrong there. Stuck need help of somebody to pull me back on road.

hernot:
> Remove the test code in `sign.vim`. That's probably what's causing Travis to break. Add tests for the changes to the other linter.

Aaaaaaaggggrrrrrrrr fucing echos. Can we have there a check report or somehting telling that it failse cause of echo not captured inside buffers or would that be a feature request against the checks test package

Thans again i have now the test output and can continue fixing what is really improper.

hernot:
Ok than you very much again.
Tests added
Checks passed
Please let me know if i should change any further before squasching my branch and doing the final commit and push to finish  pullrequest.

w0rp:
Cheers! :beers:

PR #2588, opened 2019-06-14
Handle powershell unexpected token with newline

zigford:
This is a small fix to the powershell handler when 2 newlines are directly after an unexpected token. Without the fix Ale spits out some nast callback errors.

Test included to capture it in future.

Example code that breaks the handler

```
{
    "" "

    }
````

### CODE REVIEW

File: ale_linters/powershell/powershell.vim [hunk 1]
```diff
@@ -49,6 +49,11 @@ function! ale_linters#powershell#powershell#Handle(buffer, lines) abort
                 let l:matchcount = 1
             endif
 
+            " If the match is 0, it was a failed match
+            " probably due to an unexpected token which
+            " contained a newline. Reset matchcount and
+            " continue to next set of matches
+            if ! l:match[1] | let l:matchcount = 0 | continue | endif
```
w0rp: Write `if !empty(l:match[1])`, the `l:item` assignment below, and put `let l:matchcount = 0` in `else`. No need to put an if statement all in one line for script files like this, and no need to `continue` when there's not a lot of code below.

w0rp:
Cheers! :beers:

Issue #2510, opened 2019-05-16
Too much CFLAGS parsed


fx-carton:
I updated ALE, and I got problems with CFLAGS again. I'm guessing it is caused by #2131, because some filenames (not options) are kept.
In my case, the command line included `-pedantic filename.c`, which are two arguments but are treated as one by the parser.
This is really a problem in my case, because for some reason having the input twice in the command line (`filename.c` and `< /tmp/.../filename.c`) makes gcc and thus vim hang. This makes ALE unusable, as vim hangs as soon as the file is open or modified.

It seems to me that a whitelist approach would be better than a blacklist.
I would suggest reverting that commit, and add `-isystem` as a known flag (which was the issue of that PR).

Here is the relevant part of `ALEInfo`:
```
(finished - exit code 0) ['/bin/bash', '-c', 'cd /home/user/somedir && make -n']
<<<OUTPUT STARTS>>>
cc -O2 -Wall -std=c99 -pedantic    prog.c libexample.a   -o prog
<<<OUTPUT ENDS>>>
(started) ['/bin/bash', '-c', '''gcc'' -S -x c -fsyntax-only -iquote ''/home/user/somedir'' -O2 -Wall -std=c99 -pedantic prog.c libexample.a -Wall -std=c89 -pedantic - < ''/tmp/vwPopcc/4/prog.c''']
```

As a note, it also seems that #2131 removed a test I added in #2168 that I think would have catched the issue.

I don't have a lot of free time right now, so I'll just downgrade ALE for the time being. If I get a chance to fix this, I'll make a PR.

w0rp:
Parsing C compiler flags is incredibly complicated. The only practical solution is to continue to add more test cases until all of the important cases are covered. Please feel free to submit a pull request to fix this, which also supports other existing valid use cases.

fx-carton:
I think a whitelist would allow parsing the different kinds of arguments correctly, by having several cases:
- the include flags (`-I`, `-isystem`) which have one argument that is a filename (and might need adjusting)
- some flags that have an argument which is not a filename (eg. `-D`)
- some flags without arguments (eg. warning flags, `-pedantic` and probably a lot more)

If I get some time to spend on this, I'll look at the arguments using `gcc`'s man page, filter the unwanted ones and classify the others into the categories listed above.

w0rp:
Sure, let me know how it goes.

fx-carton:
I do agree that it is extremely complicated and need more tests.

[...]

PR #2590, opened 2019-06-14
Fix cflags parsing (#2510, #2265)

### LINKED ISSUE #2265: C/C++ compiler flag -MMD creates -.d file
Vim version: 8.1
OS: macOS Mojave 10.14.3

I am using the flag `-MMD` to generate dependency files to determine when a header has been changed. Since more compiler flags are now being parsed this flag causes a file with the name of `-.d` to be generated in the working directory.

I believe this is related to #2131 (possibly #2168 or #2194).

This flag should be ignored since it doesn't provide any useful linting information.

Thanks!

fx-carton:
This PR fixes the cflags parsing problem I had in #2510: some filenames were passed as arguments. I believe it should fix #2265 as well.

1) The splitting is now done in ale#c#ShellSplit, which is intended to split the command line closely to how a shell would do.
2) Flags are selected following a whitelist of arguments. I went through GCC's man page to select which flags made sense to be parsed. I believe the others either don't need to be passed for syntax checking or cannot be passed (because of output files or non-wanted behaviour).
3) Added two tests for flags we want or don't want to be passed.

Ping for testing: @0mco @davidvandebunte @ranebrown

### DIFF
--- a/test/test_c_flag_parsing.vader
+++ b/test/test_c_flag_parsing.vader
@@ -321,3 +321,37 @@ Execute(ParseCFlags should handle parenthesis and quotes):
   \     . '-Dtest2=''(` `)'' file2.o '
   \     . '-Dtest3=`(" ")` file3.o '
   \ )
+
+Execute(CFlags we want to pass):
+  AssertEqual
+  \ '-I ' . ale#Escape(ale#path#Simplify(g:dir. '/test_c_projects/makefile_project/inc'))
+  \ . ' -I ' . ale#Escape(ale#path#Simplify(g:dir. '/test_c_projects/makefile_project/include'))
+  \ . ' -iquote ' . ale#Escape(ale#path#Simplify(g:dir. '/test_c_projects/makefile_project/incquote'))
+  \ . ' -isystem ' . ale#Escape(ale#path#Simplify(g:dir. '/test_c_projects/makefile_project/incsystem'))
+  \ . ' -idirafter ' . ale#Escape(ale#path#Simplify(g:dir. '/test_c_projects/makefile_project/incafter'))
+  \ . ' -Dmacro=value -D macro2 -Bbdir -B bdir2'
+  \ . ' -iprefix prefix -iwithprefix prefix2 -iwithprefixbefore prefix3'
+  \ . ' -isysroot sysroot --sysroot=test --no-sysroot-suffix -imultilib multidir'
+  \ . ' -Wsome-warning -std=c89 -pedantic -pedantic-errors -ansi'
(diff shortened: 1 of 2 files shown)


test/test_c_flag_parsing.vader
```diff
@@ -14,48 +14,37 @@ Execute(The CFlags parser should be able to parse include directives):
   call ale#test#SetFilename('test_c_projects/makefile_project/subdir/file.c')
 
   AssertEqual
-  \ '-I' . ale#path#Simplify(g:dir. '/test_c_projects/makefile_project/subdir'),
+  \ '-I' . ' ' . ale#Escape(ale#path#Simplify(g:dir. '/test_c_projects/makefile_project/subdir')),
   \ ale#c#ParseCFlagsFromMakeOutput(bufnr(''), ['gcc -Isubdir -c file.c'])
 
-  AssertEqual
-  \ '-isystem ' . '/usr/include/dir',
-  \ ale#c#ParseCFlagsFromMakeOutput(bufnr(''), ['gcc -isystem /usr/include/dir -c file.c'])
```
w0rp: Shouldn't this still work?

test/test_c_flag_parsing.vader
```diff
@@ -14,48 +14,37 @@ Execute(The CFlags parser should be able to parse include directives):
...
-
-Execute(ParseCFlags should ignore -c and -o):
-  call ale#test#SetFilename('test_c_projects/makefile_project/subdir/file.c')
-
-  AssertEqual
-  \ '-I' . ale#path#Simplify(g:dir. '/test_c_projects/makefile_project/subdir'),
-  \ ale#c#ParseCFlagsFromMakeOutput(bufnr(''), ['gcc -Isubdir -c file.c -o a.out'])
```
w0rp: Shouldn't we still test that `-c` and `-o` aren't included in the command?

fx-carton:
My bad. It was caused by the revert.

w0rp:
@0mco @davidvandebunte @ranebrown What do you think?

ranebrown:
That fixes #2265 for me. Thanks!

0mco:
It works for me. But since this just considered `GCC`'s options, I still personally prefer the blacklist approach.

fx-carton:
I prefer the whitelist option, because there are a lot of options that should be removed. And gcc/clang may add more unwanted options in future versions, so we would have to be more careful about that.

However, we could add an option for additional whitelisted flags, which would enable easy configuring for other compilers and/or newer versions without changing the source code.

w0rp:
Okay, I'll merge this now. Thanks for this!

PR #2591, opened 2019-06-15
Raise deoplete source rank to 1000

blahgeek:
The deoplete source rank of 100 is far too low for a language specific completion source. The builtin "around" completion source which "collects candidates around the cursor" even have a higher rank of 300. [LanguageClient-neovim](https://github.com/autozimu/LanguageClient-neovim/blob/next/rplugin/python3/deoplete/sources/LanguageClientSource.py#L13) also set this to 1000.


w0rp:
Cheers! :beers:

PR #2599, opened 2019-06-19
Enable C++ autocompletion on '::' and '->'

parkovski:
Small change to make autocompletion happen in C++ after `::` and `->`.


w0rp:
Cheers! :beers:

w0rp:
Hello again! I just merged #2601 which updated the patterns for Deoplete. Do you think we should also add `->` to the patterns there? I have a hunch that we should.

delphinus:
Yes. The pattern in `ale.py` should be updated when `s:trigger_character_map` is updated. I will push PR.

PR #2606, opened 2019-06-19
Default executable for javalsp is empthy string.

hsanson:
This fixes documentation for javalsp to match actual implementation.


w0rp:
Cheers! :beers:

PR #2614, opened 2019-06-22
Add Sorbet ruby linter and fixer

rintaun:
Adds support for the [Sorbet](https://sorbet.org) type checker for Ruby that was just released publicly. I basically used the `rubocop` linter as a guide, along with a couple other LSP linters, so I'm not 100% sure that I did everything (or anything? 😂) in the best way -- especially the tests.

Thanks!

### DIFF
--- a/ale_linters/ruby/sorbet.vim
+++ b/ale_linters/ruby/sorbet.vim
@@ -1,0 +1,23 @@
+call ale#Set('ruby_sorbet_executable', 'srb')
+call ale#Set('ruby_sorbet_options', '')
+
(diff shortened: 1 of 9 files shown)


w0rp:
Cheers! :beers:

will:
@rintaun Thanks for this, I've been using it a lot. I was wondering though why you're using `--disable-watchman` ? I don't know much about watchman but maybe it could make the sorbet lsp pick up changes faster if it was on?

rintaun:
@will Honestly, I don't remember exactly, but as I recall it was necessary for it to work at all. That may have changed, as this was made quite a while ago, and I haven't been keeping up with the development of Sorbet.

PR #2616, opened 2019-06-24
languagetools: stop auto-appending --autoDetect

pappasam:
* I've updated the relevant tests and run them; they pass

This pull request fixes an issue where Ale prevents the user from overriding a broken setting in languagetools (--autoDetect). --autoDetect conflicts with a local
command line usage of languagetools with the ngram data installed
locally. In my case, I've downloaded "en" ngrams with support for the English language.

If you want to retain current behavior, you can change the default value for the languagetools command. That said, since it appears to be somewhat broken, it's probably better to just default to languagetools with no options (as this PR currently implements).

Design-wise, auto-appending anything to a command is a problem because it prevents
users from overriding the appended configuration. There was no way to
remove --autoDetect from the executed command without this commit /
updating the plugin source code. If a user wants autoDetect they are free to place it in their bash script (or other path-included script).

Example command (in a bashscript in my path called languagetools):

    #!/bin/bash
    java -jar ~/java/LanguageTool-4.5/languagetool-commandline.jar \
      --languagemodel $HOME/Data/ngrams \
      ${@}

By placing "--autoDetect" before my filename, languagetools would only
return linting messages unassociated with ngrams (which are necessary
for checking "their" versus "there" errors). Once I manually removed
this line, everything works correctly.

## Example with ngrams and NO autoDetect

image

**notice the second suggestion for "their" versus "there"**

## Example with ngrams and autoDetect

image

**notice there is only 1 suggestion**


pappasam:
@w0rp the _options variables has been created. I've updated the tests and the docs accordingly. Please let me know if there's anything else required.

abers:
I don't use ngrams but experiencing issue with --autoDetect resulting in the following output:

`[main] DEBUG com.optimaize.langdetect.LanguageDetectorImpl - ==> [DetectedLanguage[en:0.9999988934696109
]]
18:56:09.429 [main] DEBUG com.optimaize.langdetect.LanguageDetectorImpl - ==> [DetectedLanguage[en:0.9999990562851635
]]
18:56:09.429 [main] DEBUG com.optimaize.langdetect.LanguageDetectorImpl - ==> [DetectedLanguage[en:0.9999984259788106
]]
18:56:09.429 [main] DEBUG com.optimaize.langdetect.LanguageDetectorImpl - ==> [DetectedLanguage[en:0.9999962700310778
]]`

Changing '--autoDetect' to -'-language en-GB' in the handler fixes the issue.

Konfekt:
It seems that `LanguageTool` checks for `ngram` errors or alike by default when `--language``` is passed, but not `--autoDetect` as there are less error messages using the latter.

In the meanwhile, you could work around this issue by my simple [plug-in](https://github.com/Konfekt/vim-langtool) (together with, say, [Makejob](https://git.danielmoch.com/vim-makejob/about/) to run the LanguageTool grammar check in the background) where all parameters passed can be entirely overridden.

pappasam:
> I don't use ngrams but experiencing issue with --autoDetect resulting in the following output:
> 
> `[main] DEBUG com.optimaize.langdetect.LanguageDetectorImpl - ==> [DetectedLanguage[en:0.9999988934696109 ]] 18:56:09.429 [main] DEBUG com.optimaize.langdetect.LanguageDetectorImpl - ==> [DetectedLanguage[en:0.9999990562851635 ]] 18:56:09.429 [main] DEBUG com.optimaize.langdetect.LanguageDetectorImpl - ==> [DetectedLanguage[en:0.9999984259788106 ]] 18:56:09.429 [main] DEBUG com.optimaize.langdetect.LanguageDetectorImpl - ==> [DetectedLanguage[en:0.9999962700310778 ]]`
> 
> Changing '--autoDetect' to -'-language en-GB' in the handler fixes the issue.

@w0rp it seems like the comments in this thread are pointing out that --autoDetect may be a bad default. That said, this PR makes it easily override-able. If there is nothing else to do, would we be able to merge this soon? I'd like to stop using my personal fork for daily work, if possible.

PR #2617, opened 2019-06-24
Fix crystal-lang non file-tied message handling

robindar:
Some messages of the crystal compiler are not tied to a file.
This causes a 'Key not present in Dictionnary' error (E716).
For the record, the json output on ```require "./nonexistent.cr"```
is the following :
```json
  [
  { "file":"/tmp/file.cr", "line":1, "column":1, "size":0,
    "message":"while requiring \"./nonexistent.cr\"" },
  { "message":"can't find file './nonexistent.cr' relative to '/tmp'" }
  ]
```
The second message does not have line/column attributes.


robindar:
Sorry about the long wait, I've been quite busy lately. I added the check you recommended. The test fails without the patch, and passes with it.

w0rp:
Cheers! :beers:

PR #2618, opened 2019-06-24
Update Rust cargo linter to better integrate with Clippy

JeanMertz:
Two changes:

* It was confusing (and not mentioned in the documentation) that `rust_cargo_clippy_options` required you to start the string with `--` (such as `-- -D warning`). It now accepts both forms (so also `-D warning`).
* For some reason (AFAIK incorrectly) all cargo-specific flags were only supported for `cargo check`, while these are also valid for `cargo build` and `cargo clippy`. This now works as expected.

Aside from the unit tests, I tested both changes in my local Rust project, and they worked as expected.


w0rp:
Cheers! :beers:

PR #2625, opened 2019-06-28
Update README section about running linters on save

ericdwang:
The default for `g:ale_lint_on_insert_leave` was recently changed to 1,
so it now needs to be explicitly set to 0 to run linters only when files
are saved.


w0rp:
Cheers! :beers:

PR #2627, opened 2019-06-29
Add asyncomplete.vim Support

DonnieWest:
This PR adds support for `asyncomplete.vim`

There may be another, better approach here but it basically works by creating an API for `ALE` to provide completions to other tooling via a callback. This theoretically opens up support for other tools like `ncm2` to also add support for `ALE` in the future with `asyncomplete.vim` being a test case for this approach.

This is _probably_ a rather naive approach and I'm super open to modifying this based on any feedback you might have.

Ref: https://github.com/prabirshrestha/asyncomplete.vim/issues/150

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -269,7 +269,7 @@ function! ale#completion#Show(result) abort
 endfunction
 
 function! ale#completion#GetAllTriggers() abort
-    return s:trigger_character_map
+    return deepcopy(s:trigger_character_map)
 endfunction
 
 function! s:CompletionStillValid(request_id) abort
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -264,22 +264,12 @@ function! ale#completion#Show(result) abort
     endif
 
     if l:source is# 'ale-callback'
-        call ale#completion#CallbackComplete()
+        call b:CompleteCallback(b:ale_completion_result)
     endif
-endfunction
-
-function! ale#completion#CallbackComplete() abort
-    call b:asyncomplete_callback(b:ale_completion_result)
 endfunction
 
 function! ale#completion#GetAllTriggers() abort
     return s:trigger_character_map
-endfunction
-
-function! ale#completion#RequestCallbackCompletions(callback) abort
-    let b:asyncomplete_callback = a:callback
-
-    call ale#completion#GetCompletions('ale-callback')
 endfunction
 
 function! s:CompletionStillValid(request_id) abort
@@ -574,12 +564,25 @@ endfunction
 
 " This function can be used to manually trigger autocomplete, even when
 " g:ale_completion_enabled is set to false
-function! ale#completion#GetCompletions(source) abort
+function! ale#completion#GetCompletions(...) abort
+    let l:source = get(a:000, 0, '')
+    let l:options = get(a:000, 1, {})
+
+    if len(a:000) > 2
+        throw 'Too many arguments!'
+    endif
+
+    let l:CompleteCallback = get(l:options, 'callback', v:null)
+
+    if l:CompleteCallback isnot v:null
+        let b:CompleteCallback = l:CompleteCallback
+    endif
+
     let [l:line, l:column] = getpos('.')[1:2]
 
     let l:prefix = ale#completion#GetPrefix(&filetype, l:line, l:column)
 
-    if a:source is# 'ale-automatic' && empty(l:prefix)
+    if l:source is# 'ale-automatic' && empty(l:prefix)
         return 0
     endif
 
@@ -592,7 +595,7 @@ function! ale#completion#GetCompletions(source) abort
     \   'prefix': l:prefix,
     \   'conn_id': 0,
     \   'request_id': 0,
-    \   'source': a:source,
+    \   'source': l:source,
     \}
     unlet! b:ale_completion_result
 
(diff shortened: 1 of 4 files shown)


DonnieWest:
That's what I get for not running the tests locally, apologies :joy:

w0rp:
I haven't read the code yet, but I like the idea. :+1: I'll go through a backlog of messages and get around to this.

DonnieWest:
@w0rp following up on this, please let me know if there's anything I can do to help get this merged in :+1:

ranebrown:
If it helps out any I gave this a try and it seems to be working.

It might be a good idea to modify the documentation to not use priority as an example when registering the source since that option was removed from asyncomplete in v2. (See prabirshrestha/asyncomplete.vim#139)

autoload/ale/completion.vim
```diff
@@ -262,6 +262,24 @@ function! ale#completion#Show(result) abort
...
     endif
+
+    if l:source is# 'ale-callback'
+        call ale#completion#CallbackComplete()
+    endif
+endfunction
+
+function! ale#completion#CallbackComplete() abort
+    call b:asyncomplete_callback(b:ale_completion_result)
+endfunction
```
w0rp: Remove this function, and put the call for the callback inline in the function above.

autoload/ale/completion.vim
```diff
@@ -262,6 +262,24 @@ function! ale#completion#Show(result) abort
...
+
+function! ale#completion#GetAllTriggers() abort
+    return s:trigger_character_map
```
w0rp: This function returns a reference to the Dictionary, which will be modified by the asyncomplete source. That will be difficult to maintain. You can apply `deepcopy` here to fix this.

```suggestion
    return deepcopy(s:trigger_character_map)
```

autoload/ale/completion.vim
```diff
@@ -262,6 +262,24 @@ function! ale#completion#Show(result) abort
...
+endfunction
+
+function! ale#completion#RequestCallbackCompletions(callback) abort
+    let b:asyncomplete_callback = a:callback
+
+    call ale#completion#GetCompletions('ale-callback')
```
w0rp: I think instead of adding a wrapper around the `GetCompletions` function, add `...` to `GetCompletions` for varargs, and you can do the following.

```vim
let l:options = get(a:000, 0, {})

if len(a:000) > 1
    throw 'Too many arguments!'
endif

let l:Callback = get(l:options, 'callback', v:null)

if l:Callback isnot v:null
    let b:CompleteCallback = l:Callback
endif
```

You can then use `call ale#completion#GetCompletions('ale-callback', {'callback': CALLBACK_HERE})` instead below.

You should probably name funcref variables in PascalCase, as Vim typically doesn't let you name funcref variables as lowercase. I recommend using a more generic callback variable name here, for supporting other completion plugins later.

autoload/asyncomplete/sources/ale.vim
```diff
@@ -0,0 +1,25 @@
+function! asyncomplete#sources#ale#get_source_options(...) abort
+    return extend(extend({
+    \     'name': 'ale',
+    \     'completor': function('asyncomplete#sources#ale#completor'),
+    \     'whitelist': ['*'],
+    \     'triggers': asyncomplete#sources#ale#get_triggers(),
+    \ }, a:0 >= 1 ? a:1 : {}), {'refresh_pattern': '\k\+$'})
```
w0rp: Split this into a variable with the first `extend` call, and `return extend(l:that_variable, {'refresh_pattern': '\k\+$'})`. That'll be easier to follow.

autoload/asyncomplete/sources/ale.vim
```diff
@@ -0,0 +1,25 @@
...
+endfunction
+
+function! asyncomplete#sources#ale#get_triggers() abort
+    let l:triggers = ale#completion#GetAllTriggers()
+    let l:triggers['*'] = l:triggers['<default>']
+
+    return l:triggers
+endfunction
+
+function! asyncomplete#sources#ale#completor(opts, ctx) abort
```
w0rp: Use English names like `options`, `context`, and `keywords`.

autoload/asyncomplete/sources/ale.vim
```diff
@@ -0,0 +1,25 @@
...
+    let l:kw = matchstr(a:ctx.typed, '\w\+$')
+    let l:kwlen = len(l:kw)
+    let l:startcol = a:ctx.col - l:kwlen
```
w0rp: You can move the `len` expression here. No need for the variable.

test/completion/test_completion_events.vader
```diff
@@ -333,6 +333,27 @@ Execute(b:ale_completion_info should be set up correctly for other sources):
   \ b:ale_completion_info
   Assert !exists('b:ale_completion_result')
 
+Execute(b:ale_completion_info should be set up correctly when requesting completions via callback):
+  let b:ale_completion_result = []
+  call setpos('.', [bufnr(''), 3, 14, 0])
+  function! Callback() abort
+    echo 'Called'
+  endfunction
```
w0rp: Add a `delfunction` line in `After`, or this function will continue to exist when other tests are run. Use a more descriptive name for the test callback, in case it happens to clash with a test function name later.

DonnieWest:
Fixed up everything according to your comments. Not sure how the build failing is the result of anything I did here though https://travis-ci.com/dense-analysis/ale/jobs/222171865

w0rp:
I'm back again. :+1:  I'll see if I can fix the tests here and merge this.

w0rp:
Cheers! :beers:

DonnieWest:
Thank you!

jgonera:
If asyncomplete/ALE integration stopped working for you in ALE 3.0.0 then replace `g:ale_completion_tsserver_autoimport` with `g:ale_completion_autoimport`.

PR #2638, opened 2019-07-10
Use Neovim API for highlights when available

### LINKED ISSUE #1274: Implement support for NeoVim's more robust hightlighting API

frangio:
Fixes #1274.

I tried to keep the changes as simple as possible by just defining a wrapper around `nvim_buf_add_highlight` that emulates the `matchaddpos` interface.

I'm having trouble testing this. I wrapped the function calls in autoload functions to be able to mock them. But the original function still seems to be used in some of the tests. I don't understand the exact sequence of `runtime`, `let`, `unlet`, etc that I'm supposed to be running for these mocks to work. Is there an explanation of this pattern somewhere that I could read? Thanks!

### DIFF
--- a/autoload/ale/highlight.vim
+++ b/autoload/ale/highlight.vim
@@ -26,6 +26,12 @@ endif
 let s:MAX_POS_VALUES = 8
 let s:MAX_COL_SIZE = 1073741824 " pow(2, 30)
 
+let s:has_nvim_highlight = exists('*nvim_buf_add_highlight') && exists('*nvim_buf_clear_namespace')
+
+if s:has_nvim_highlight
+    let s:ns_id = nvim_create_namespace('ale_highlight')
+endif
+
 function! ale#highlight#CreatePositions(line, col, end_line, end_col) abort
     if a:line >= a:end_line
         " For single lines, just return the one position.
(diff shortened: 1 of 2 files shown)


frangio:
I've been playing with a bit more, and I really can't get the autoload redefinitions to work...

Here's an example test that looks to me like it should work but it doesn't.

```
Before:
  runtime autoload/ale/highlight.vim

  function! ale#highlight#nvim_buf_add_highlight(buffer, ns_id, hl_group, line, col_start, col_end)
    return "test"
  endfunction

Execute(Uses mocked autoload function):
  let l:result = ale#highlight#nvim_buf_add_highlight('', '', '', '', '', '')
  AssertEqual "test" l:result
```

w0rp:
Thanks for trying to bring this back again. I'll have a deeper look later.

w0rp:
Sorry I've been unavailable for a while! This PR is near the top of my list when I finally have the time to come back. I have around 140ish emails to look through when I start having free time again, which could be any time between tonight and some time next week.

w0rp:
@andreypopp It's finally back in master again. :+1:  If there are any issues with it, I'll prioritise fixing them.

andreypopp:
Thanks @w0rp

PR #2643, opened 2019-07-14
Add Deoplete's input_patterns for cpp

delphinus:
ref #2599 

`input_patterns` in `ale.py` should be updated when `s:trigger_character_map` in `completion.vim` have been updated.


w0rp:
Cheers! :beers:

PR #2644, opened 2019-07-15
Allow use of quickfix window for ALEFindReferences

### LINKED ISSUE #1759: quickfix list support for ALEFindReferences and commands for navigating through references
If I've got a list of references then there's a decent chance that I'll want to cycle through them.

Suggest using the quickfix list on the grounds that: that's what `cscope` and `YouCompleteMe` both do in analogous situations, so it's probably what people are used to.

jamesob:
Fixes #1759.

Introduces new configuration `g:ale_use_quickfix_findref` which
defaults to 0.

Sort of a proof of concept - this is my first vimscript foray so I'm
sure the code is garbage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -635,6 +635,9 @@ let g:ale_keep_list_window_open = 1
 You can also set `let g:ale_list_vertical = 1` to open the windows vertically
 instead of the default horizontally.
 
+To present `ALEFindReferences` results in the quickfix window, use
+`let g:ale_use_quickfix_findref = 1`.
+
 <a name="faq-jsx-stylelint-eslint"></a>
 
 ### 5.xii. How can I check JSX files with both stylelint and eslint?
--- a/autoload/ale/references.vim
+++ b/autoload/ale/references.vim
@@ -1,5 +1,7 @@
 let s:references_map = {}
 
+let g:ale_use_quickfix_findref = get(g:, 'ale_use_quickfix_findref', 0)
+
 " Used to get the references map in tests.
 function! ale#references#GetMap() abort
     return deepcopy(s:references_map)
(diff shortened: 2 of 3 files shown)


autoload/ale/references.vim
```diff
@@ -51,18 +53,44 @@ function! ale#references#HandleLSPResponse(conn_id, response) abort
...
-                \ 'filename': ale#path#FromURI(l:response_item.uri),
-                \ 'line': l:response_item.range.start.line + 1,
-                \ 'column': l:response_item.range.start.character + 1,
-                \})
+                let l:col = l:response_item.range.start.character + 1
+                let l:lnum = l:response_item.range.start.line + 1
+                let l:filename = ale#path#FromURI(l:response_item.uri)
+
+                if g:ale_use_quickfix_findref
+                    let l:cwd = getcwd()
```
w0rp: Have a look at the custom linter output which suggests an alternative to `getcwd()` that works if you change buffers before an asynchronous callback is fired.

autoload/ale/references.vim
```diff
@@ -51,18 +53,44 @@ function! ale#references#HandleLSPResponse(conn_id, response) abort
...
+                    call add(l:item_list, {
+                    \ 'filename': substitute(l:filename, l:cwd . '/', "", ""),
+                    \ 'lnum': l:lnum,
+                    \ 'col': l:col,
+                    \ 'text': system('sed "' . l:lnum . 'q;d" ' . l:filename),
```
w0rp: You'll have to do something other than a `system()` call here. This will slow down Vim a lot, and it won't work on Windows.

README.md
```diff
@@ -635,6 +635,9 @@ let g:ale_keep_list_window_open = 1
 You can also set `let g:ale_list_vertical = 1` to open the windows vertically
 instead of the default horizontally.
 
+To present `ALEFindReferences` results in the quickfix window, use
+`let g:ale_use_quickfix_findref = 1`.
```
w0rp: Instead of making this a global option, make this an option for the command. Have a look at the existing option below, and uses of the `ale#args#Parse` in the codebase for other commands in `symbol.vim` and `loclist_jumping.vim`. We could make the option in this case `'quickfix'`, so we can use it like `:ALEFindReferences -quickfix`.

oguzbilgic:
Any updates on this @jamesob ? 👍

MrAwesome:
Also very interested in getting this!

PR #2648, opened 2019-07-18
Add support for `solc` Solidity compiler

karlb:


w0rp:
Cheers! :beers:

I'll add a test for the executable and command.

Sorry for the delay. I'm slowly going through a backlog of emails until it's down to 0.

PR #2652, opened 2019-07-20
Adding support for LSP `window/showMessage` method

suoto:
Text received via this methods is added to the message stream.

Implementation notes:

* Use variable `g:ale_lsp_show_message_format` to tweak the format; default is `'%severity%:%linter%: %text%'` (this value includes all options implemented)
* `window/showMessage` can have severity "log", which seems the same as `window/logMessage`. Since I could not find a logging mechanism, I chose to ignore them. An alternative would be to log like Info, but that might lead to flooding the UI
* I've hard coded the highlights for error and warning severities to `ErrorMsg` and `WarningMsg` respectively, please advise if there's an ALE specific highlight group that's better suited

### DIFF
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -11,7 +11,7 @@ let g:ale_linters_ignore = get(g:, 'ale_linters_ignore', {})
 let g:ale_disable_lsp = get(g:, 'ale_disable_lsp', 0)
 
 " LSP window/showMessage format
-let g:ale_lsp_show_message_format = get(g:, 'ale_lsp_show_message_format', '%severity%:%linter%: %text%')
+let g:ale_lsp_show_message_format = get(g:, 'ale_lsp_show_message_format', '%severity%:%linter%: %s')
 " FIXME(suoto): Can use severity to display only errors by default.
 " Valid values mimic LSP definitions (error, warning and information; log is
 " never shown)
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -12,7 +12,6 @@ let g:ale_disable_lsp = get(g:, 'ale_disable_lsp', 0)
 
 " LSP window/showMessage format
 let g:ale_lsp_show_message_format = get(g:, 'ale_lsp_show_message_format', '%severity%:%linter%: %s')
-" FIXME(suoto): Can use severity to display only errors by default.
 " Valid values mimic LSP definitions (error, warning and information; log is
 " never shown)
 let g:ale_lsp_show_message_severity = get(g:, 'ale_lsp_show_message_severity', 'error')
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -12,6 +12,10 @@ let g:ale_disable_lsp = get(g:, 'ale_disable_lsp', 0)
 
 " LSP window/showMessage format
 let g:ale_lsp_show_message_format = get(g:, 'ale_lsp_show_message_format', '%severity%:%linter%: %text%')
+" FIXME(suoto): Can use severity to display only errors by default.
+" Valid values mimic LSP definitions (error, warning and information; log is
+" never shown)
+let g:ale_lsp_show_message_severity = get(g:, 'ale_lsp_show_message_severity', 'error')
 
 let s:lint_timer = -1
 let s:getcmdwintype_exists = exists('*getcmdwintype')
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -5,6 +5,7 @@
 " Strings used for severity in the echoed message
 let g:ale_echo_msg_error_str = get(g:, 'ale_echo_msg_error_str', 'Error')
 let g:ale_echo_msg_info_str = get(g:, 'ale_echo_msg_info_str', 'Info')
+let g:ale_echo_msg_log_str = get(g:, 'ale_echo_msg_log_str', 'Log')
 let g:ale_echo_msg_warning_str = get(g:, 'ale_echo_msg_warning_str', 'Warning')
 " Ignoring linters, for disabling some, or ignoring LSP diagnostics.
 let g:ale_linters_ignore = get(g:, 'ale_linters_ignore', {})
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -10,6 +10,9 @@ let g:ale_echo_msg_warning_str = get(g:, 'ale_echo_msg_warning_str', 'Warning')
 let g:ale_linters_ignore = get(g:, 'ale_linters_ignore', {})
 let g:ale_disable_lsp = get(g:, 'ale_disable_lsp', 0)
 
+" LSP window/showMessage format
+let g:ale_lsp_show_message_format = get(g:, 'ale_lsp_show_message_format', '%severity%:%linter%: %text%')
+
 let s:lint_timer = -1
 let s:getcmdwintype_exists = exists('*getcmdwintype')
 
--- a/autoload/ale/lsp/window.vim
+++ b/autoload/ale/lsp/window.vim
@@ -39,6 +39,7 @@ function! ale#lsp#window#formatString(format, args) abort
             throw 'Invalid argument ''' . l:key . '''. Arguments must follow ' .
             \ 'pattern [a-zA-Z][a-zA-Z0-9_]*'
         endif
+
         let l:string = substitute(l:string, '\V%' . l:key . '%', '\=l:value', 'g')
     endfor
 
@@ -52,6 +53,11 @@ function! ale#lsp#window#showMessage(linter_name, format, params) abort
     let l:message = a:params.message
     let l:type = a:params.type
 
+    if l:type is# s:LSP_MESSAGE_TYPE_LOG
+        " Discard log severity for now
+        return
+    endif
+
     " Common formatting arguments
     let l:format_args = {'linter': a:linter_name, 'text': l:message}
 
@@ -63,11 +69,6 @@ function! ale#lsp#window#showMessage(linter_name, format, params) abort
         let l:format_args['severity'] = g:ale_echo_msg_warning_str
         let l:Handler = funcref('s:echoWarning')
     elseif l:type is# s:LSP_MESSAGE_TYPE_INFORMATION
-        let l:format_args['severity'] = g:ale_echo_msg_info_str
-        let l:Handler = funcref('s:echoInfo')
-    elseif l:type is# s:LSP_MESSAGE_TYPE_LOG
-        " TODO: If/when there's logging, we can handle it here. Until then,
-        " handle it as a regular info
         let l:format_args['severity'] = g:ale_echo_msg_info_str
         let l:Handler = funcref('s:echoInfo')
     endif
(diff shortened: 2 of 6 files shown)


autoload/ale/lsp/window.vim
```diff
@@ -0,0 +1,83 @@
...
+
+" Text is going to be echo'ed by ale#util#Execute, so any single quote the
+" text has must be reescaped
+function! s:escapeQuotes(text) abort
+    return substitute(a:text, '''', '''''', 'g')
+endfunction
+
+function! s:echoError(text) abort
+    call ale#util#Execute(
+    \ 'redraw | echohl ErrorMsg | echomsg ''' . s:escapeQuotes(a:text) . ''' | echohl None')
```
w0rp: Remove `redraw`, as I don't think we should force a redraw on every message.

remove the `echohl` commands for now.

I recommend replacing all of these echoes with calls for the `ale#util#ShowMessage` function, which will show larger messages in preview windows. Long messages cannot be echoed in Vim 8 from an asynchronous context.

autoload/ale/lsp/window.vim
```diff
@@ -0,0 +1,83 @@
...
+        return
+    endif
+
+    " Common formatting arguments
+    let l:format_args = {'linter': a:linter_name, 'text': l:message}
+
+    " Severity will depend on the message type
+    if l:type is# s:LSP_MESSAGE_TYPE_ERROR
+        let l:format_args['severity'] = g:ale_echo_msg_error_str
+        let l:Handler = funcref('s:echoError')
```
w0rp: Prefer `function` over `funcref`. `funcref` is only needed when a function is redefined later.

autoload/ale/lsp_linter.vim
```diff
@@ -130,6 +130,11 @@ function! ale#lsp_linter#HandleLSPResponse(conn_id, response) abort
         call s:HandleLSPErrorMessage(l:linter_name, a:response)
     elseif l:method is# 'textDocument/publishDiagnostics'
         call s:HandleLSPDiagnostics(a:conn_id, a:response)
+    elseif l:method is# 'window/showMessage'
+        call ale#lsp#window#showMessage(
+        \ s:lsp_linter_map[a:conn_id],
+        \ g:ale_lsp_show_message_format,
+        \ a:response.params)
```
w0rp:
```suggestion
        \ a:response.params
      )
```

Put the closing parenthesis on its own line, and use 4 space indenting for Vim source files.

autoload/ale/lsp_linter.vim
```diff
@@ -130,6 +130,11 @@ function! ale#lsp_linter#HandleLSPResponse(conn_id, response) abort
         call s:HandleLSPErrorMessage(l:linter_name, a:response)
     elseif l:method is# 'textDocument/publishDiagnostics'
         call s:HandleLSPDiagnostics(a:conn_id, a:response)
+    elseif l:method is# 'window/showMessage'
+        call ale#lsp#window#showMessage(
```
w0rp: I think we should probably add a setting, which might be disabled by default, for showing these messages. I imagine they could be annoying to a lot of users.

test/lsp/test_handling_window_requests.vader
```diff
@@ -0,0 +1,113 @@
...
+  \  {'type':1,'message':'an error'})
+
+  AssertEqual
+  \ ['redraw | ' .
+  \  'echohl ErrorMsg | ' .
+  \  'echomsg ''Error:some_linter: an error'' | ' .
+  \  'echohl None'],
+  \ g:expr_list
+
+
```
w0rp: Too many blank lines throughout a lot of the file.

autoload/ale/lsp/window.vim
suoto: Having a thought about this, I actually wonder what's the best way to display those messages. As a user, I'd expect something non intrusive (e.g. doesn't requires me to press enter for long messages) that keeps the info until I think it's best to look at it, quite similar to `ALEDetail` but keeping old messages. This is food for thought now, I'm happy to hear suggestions.

I'll switch to `ale#util#ShowMessage` and see how that behaves, although that will be skewed since I don't use that many language servers regularly.

autoload/ale/lsp/window.vim
suoto: Not needed anymore, not using `s:echo[Error|Warning|Info]` in favor of `ale#util#ShowMessage`

autoload/ale.vim
```diff
@@ -10,6 +10,13 @@ let g:ale_echo_msg_warning_str = get(g:, 'ale_echo_msg_warning_str', 'Warning')
 let g:ale_linters_ignore = get(g:, 'ale_linters_ignore', {})
 let g:ale_disable_lsp = get(g:, 'ale_disable_lsp', 0)
 
+" LSP window/showMessage format
+let g:ale_lsp_show_message_format = get(g:, 'ale_lsp_show_message_format', '%severity%:%linter%: %s')
+" FIXME(suoto): Can use severity to display only errors by default.
+" Valid values mimic LSP definitions (error, warning and information; log is
+" never shown)
+let g:ale_lsp_show_message_severity = get(g:, 'ale_lsp_show_message_severity', 'error')
```
suoto: Added a severity setting that defaults to `error`, which seems more suitable than on/off. Let me know if the on/off is still preferred.

autoload/ale/lsp_linter.vim
suoto: https://github.com/dense-analysis/ale/pull/2652/files#r309254007

doc/ale.txt
```diff
@@ -1181,6 +1181,43 @@ b:ale_loclist_msg_format                             *b:ale_loclist_msg_format*
...
+  This variable defines the format that messages received from an LSP will
+  have. The following sequences of characters will be replaced.
+
+    `%s`           - replaced with the message text
+    `%linter%`     - replaced with the name of the linter
+    `%severity%`   - replaced with the severity of the message
+
+  To configure the strings used on the `%severity%` field, see |g:ale_echo_msg_format|.
+
+  The echo message format can also be configured separately for each buffer,
```
suoto: This format is not actually configurable per buffer (I kept the text on purpose). Not sure if it even makes sense, this is not a channel for the server to flag anything specific regarding a file. I'd suggest having a dict to configure the format per server or file type but that looks like too much hassle for too little benefit really.

autoload/ale/lsp/window.vim
```diff
@@ -0,0 +1,68 @@
+" Author: suoto <andre820@gmail.com>
```
w0rp: Move this code to a file like `autoload/ale/lsp_window.vim`. Everything in `ale/lsp/` and `ale/lsp.vim` should be only for implementing communication with language servers, and all of the code for how ALE handles the messages should be outside of that.

autoload/ale/lsp/window.vim
```diff
@@ -0,0 +1,68 @@
...
+let s:LSP_MESSAGE_TYPE_INFORMATION = 3
+let s:LSP_MESSAGE_TYPE_LOG = 4
+
+" Translate strings from the user config to a number so we can check
+" severities
+let s:CFG_TO_LSP_SEVERITY = {
+\   'error': s:LSP_MESSAGE_TYPE_ERROR,
+\   'warning': s:LSP_MESSAGE_TYPE_WARNING,
+\   'information': s:LSP_MESSAGE_TYPE_INFORMATION,
+\   'log': s:LSP_MESSAGE_TYPE_LOG
```
w0rp: Add another severity level which disables the messages completely, in case people don't want to see them. The numeric value can be `0`. I'm happy with the default being `'error'`.

autoload/ale.vim
```diff
@@ -10,6 +10,13 @@ let g:ale_echo_msg_warning_str = get(g:, 'ale_echo_msg_warning_str', 'Warning')
 let g:ale_linters_ignore = get(g:, 'ale_linters_ignore', {})
 let g:ale_disable_lsp = get(g:, 'ale_disable_lsp', 0)
 
+" LSP window/showMessage format
+let g:ale_lsp_show_message_format = get(g:, 'ale_lsp_show_message_format', '%severity%:%linter%: %s')
+" FIXME(suoto): Can use severity to display only errors by default.
```
w0rp: You can remove this FIXME comment.

autoload/ale/lsp/window.vim
```diff
@@ -0,0 +1,68 @@
...
+    " Severity will depend on the message type
+    if l:type is# s:LSP_MESSAGE_TYPE_ERROR
+        let l:format_args['%severity%'] = g:ale_echo_msg_error_str
+    elseif l:type is# s:LSP_MESSAGE_TYPE_WARNING
+        let l:format_args['%severity%'] = g:ale_echo_msg_warning_str
+    elseif l:type is# s:LSP_MESSAGE_TYPE_INFORMATION
+        let l:format_args['%severity%'] = g:ale_echo_msg_info_str
+    endif
+
+    call ale#util#ShowMessage(ale#lsp#window#formatString(a:format, l:format_args))
```
w0rp: Remove the `formatString` function and use `substitute` here for specifically the strings above, as that's all we'll ever use. Always replace `%severity%` with something.

autoload/ale/lsp/window.vim
```diff
@@ -0,0 +1,68 @@
...
+" - details: dict containing linter name and format (g:ale_lsp_show_message_format)
+" - params: dict with the params for the call in the form of {type: number, message: string}
+function! ale#lsp#window#HandleShowMessage(linter_name, format, params) abort
+    let l:message = a:params.message
+    let l:type = a:params.type
+
+    " Discard log severity for now
+    if l:type is# s:LSP_MESSAGE_TYPE_LOG
+        return
+    endif
```
w0rp: Because you have the setting for controlling the log level, you can show log level messages if the level is set that high, now that the default is to only show error messages.

autoload/ale/lsp_linter.vim
suoto: Closing this (being discussed elsewhere in this PR)

autoload/ale/lsp/window.vim
suoto: Done!

autoload/ale/lsp/window.vim
suoto: Done. Also made 'info' an alias for 'information' so the user can set either. (Can remove if sticking strictly to LSP naming is preferred)

autoload/ale.vim
suoto: Done

autoload/ale/lsp/window.vim
suoto: Ok, will also add a new config, `g:ale_echo_msg_log_str`, to configure the string

autoload/ale/lsp/window.vim
suoto: Done

suoto:
Done with the last comments; kept discussions open on items that need an OK.

w0rp:
Cheers! :beers:

Sorry for the delay. I will eventually get to everything. I'll update the documentation too to mention the `'disabled'` option.

PR #2654, opened 2019-07-22
add buildifier fixer for Bazel

camja014:
Adds a `buildifier` fixer for Bazel build and Starlark files.

Includes some basic documentation and tests.

Let me know if anything needs changing/fixing.


autoload/ale/fix/registry.vim
```diff
@@ -32,6 +32,11 @@ let s:default_registry = {
 \       'suggested_filetypes': ['javascript', 'css', 'html'],
 \       'description': 'Apply fecs format to a file.',
 \   },
+\   'buildifier': {
+\       'function': 'ale#fixers#buildifier#Fix',
+\       'suggested_filetypes': ['bzl'],
```
aignas: This tool can also be used for simple `starlark` scripts, which have a `star` extension.

test/fixers/test_buildifier_fixer_callback.vader
```diff
@@ -0,0 +1,24 @@
...
+  silent cd command_callback
+  let g:dir = getcwd()
+
+After:
+  Restore
+
+  call ale#test#RestoreDirectory()
+
+Execute(The buildifier callback should return the correct default values):
+  call ale#test#SetFilename('bzl_paths/BUILD')
```
aignas: Consider checking that we work out of the box with:
* `bzl_paths/BUILD.bazel`
* `bzl_paths/rules.bzl`

which are also a common extensions, which could work with this.

If you decide to add support for starlark scripts then consider doing the same for `.star` files.

joprice:
@camja014 would be awesome to see this merged. I can pitch in if you need a hand!

autoload/ale/fix/registry.vim
camja014: I've haven't seen a Starlark script with a `star` extension. I've only ever seen the `bzl` extension. And at least on Bazel 0.29.1, `bzl` is the only supported file extension for Starlark scripts. Anything else throws an error.

test/fixers/test_buildifier_fixer_callback.vader
camja014: I modified my unit tests so that multiple different file types associated with Bazel are tested. 

At least, _I think_ that's what I'm testing. My knowledge of vimscript is next to non-existent, and I don't know the ale API at all, so they may not be correct :grin:

autoload/ale/fix/registry.vim
aignas: I was thinking about `starlark.vim` file type detection in https://github.com/cappyzawa/starlark.vim/blob/master/ftdetect/starlark.vim

I have seen starlark scripts with a `star` extension in a few places, but cannot give any links right now. :)

camja014:
Any way this could get merged in the near future?

heynemann:
Can we move this forward? It would be awesome to have this work automagically like everything else in ale :)

matawed:
Review comments were marked as resolved. Can we please re-open this, so that it eventually gets reviewed and merged. Please, :pray:

PR #2659, opened 2019-07-25
Use the psalm executable with the LS option

greg0ire:
This is the simplest way of getting the psalm language server to run
now.

### DIFF
--- a/ale_linters/php/psalm.vim
+++ b/ale_linters/php/psalm.vim
@@ -1,7 +1,7 @@
 " Author: Matt Brown <https://github.com/muglug>
 " Description: plugin for Psalm, static analyzer for PHP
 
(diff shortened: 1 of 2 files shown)


greg0ire:
I'm really new to vimL and Vader and have little to no idea what I've been doing.

muglug:
It was added 5 months ago - there's a slight risk of breaking changes, but (on the other hand) people _should_ be using the latest version of Psalm (which is still on the same major version as when the plugin was released)

greg0ire:
Haha yeah! It's ripe now!

w0rp:
I only just noticed that the `_executable` option isn't properly named. It should be `g:ale_php_psalm_executable`, instead of `g:ale_psalm_langserver_executable`. Update your settings now if you use that option. I'm about to change that.

PR #2672, opened 2019-08-01
Fix broken ouptut of eslint when there are depricated rules

myshov:
When there is DeprecationWarning in the output of eslint, then plugin won't work. For example I stumbled upon on this rule `parserOptions.ecmaFeatures.experimentalObjectRestSpread` that caused broken linting experience:

```
(node:25580) [ESLINT_LEGACY_OBJECT_REST_SPREAD] DeprecationWarning: The 'parserOptions.ecmaFeatures.experimentalObjectRestSpread' option is deprecated. Use 'parserOptions.ecmaVersion' instead. (found in "airbnb-base")
```

### DIFF
--- a/autoload/ale/handlers/eslint.vim
+++ b/autoload/ale/handlers/eslint.vim
@@ -44,7 +44,7 @@ function! ale#handlers#eslint#GetCommand(buffer) abort
 
     return ale#node#Executable(a:buffer, l:executable)
     \   . (!empty(l:options) ? ' ' . l:options : '')
-    \   . ' -f json --stdin --stdin-filename %s'
+    \   . ' -f json --stdin --stdin-filename %s 2> /dev/null'
 endfunction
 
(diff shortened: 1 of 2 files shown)


autoload/ale/handlers/eslint.vim
```diff
@@ -44,7 +44,7 @@ function! ale#handlers#eslint#GetCommand(buffer) abort
 
     return ale#node#Executable(a:buffer, l:executable)
     \   . (!empty(l:options) ? ' ' . l:options : '')
-    \   . ' -f json --stdin --stdin-filename %s'
+    \   . ' -f json --stdin --stdin-filename %s 2> /dev/null'
```
w0rp: This will break the linter on Windows, and also some of the other parsing for detecting and reporting errors from ESLint.

I recommend finding some other way to fix this. Maybe the error parser can be updated with tests including example output which causes the problem.

w0rp:

PR #2677, opened 2019-08-03
Remove `/bin/ls` assumption from check-toc test.

davidtwco:
On some systems, notably NixOS, there is no `/bin/ls` and thus this test can fail unnecessarily on those systems. This PR uses `/usr/bin/env ls` which resolves the issue.


w0rp:
Cheers! :beers:

PR #2681, opened 2019-08-06
Use cache for clj-kondo linter

mynomoto:
This is a very small change, I'm not sure about how to test it if tests are necessary for the command executed for linting. Let me know if this is missing something.


w0rp:
I have merged this now. Sorry for the delay.

Issue #2687, opened 2019-08-10
sh's shell linter doesn't work


chenjie199234:
### Problem:no warning or error sign in localist or quickfix
<details>
<summary>MyConfig</summary>

```
	let g:ale_enabled=1
	let g:ale_sign_error='>x'
	let g:ale_sign_warning='>!'
	let g:ale_set_loclist=1
	let g:ale_set_quickfix=0
	let g:ale_open_list=0
	"lint
	let g:ale_lint_on_text_changed=0
	let g:ale_lint_on_insert_leave=0
	let g:ale_lint_on_enter=1
	let g:ale_lint_on_save=1
	let g:ale_lint_on_filetype_changed=1
	let g:ale_linters={
		\ 'c':['clang','clangd'],
		\ 'cpp':['clang','clangd'],
		\ 'go':['gopls','gofmt','golint','govet'],
		\ 'javascript':['eslint'],
		\ 'proto':['protoc-gen-lint'],
		\ 'lua':['luac'],
		\ 'sh':['language_server','shell'],
		\ }
	let g:ale_linters_explict=1
	"fix
	let g:ale_fix_on_save=1
	let g:ale_fixers={
		\ '*':['remove_trailing_lines','trim_whitespace'],
		\ 'c':['clang-format','uncrustify'],
		\ 'cpp':['clang-format','uncrustify'],
		\ 'go':['gofmt','goimports','gomod'],
		\ 'javascript':['eslint'],
		\ }
	"lsp
	if $GO111MODULE!="on"
		let g:ale_lsp_root={
			\ 'gopls':$GOPATH,
			\ }
	endif
	let g:ale_completion_enabled=1
	let g:ale_completion_delay=100
	let g:ale_completion_max_suggestions=100
```
</details>
<details>
<summary>MyTestCode</summary>

```
#!/bin/bash
echo 1bc"
1223b
kbdqj
```
</details>
<details>
<summary>AleInfo</summary>

```
ALEInfo
 Current Filetype: sh
Available Linters: ['language_server', 'shell', 'shellcheck']
  Enabled Linters: ['language_server', 'shell']
 Suggested Fixers:
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'shfmt' - Fix sh files with shfmt.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

let g:ale_sh_shell_default_shell = 'bash'
 Global Variables:

[...]

PR #2688, opened 2019-08-11
force bash use en_US.UTF-8 lang

chenjie199234:
fix https://github.com/dense-analysis/ale/issues/2687
if system lang is zh_CN.UTF-8,bash will output chinese and ale will not show any warning or error sign.
maybe other linters also have this problem.
in this linter force  bash to use en_US.UTF-8 will fix this problem


zoonfafer:
Hi @chenjie199234 , another approach would be to fix the regexp pattern used to parse the outputs (#2741), which does not favour one language over the other.  See if it works better for you.

PR #2700, opened 2019-08-15
clangcheck: Add -fno-color-diagnostics (closes #2188)

### LINKED ISSUE #2188: Have -fno-color-diagnostics default for clangcheck
When I tried out Ale with clangcheck and clangtidy, I got these escape codes (color codes) in the warnings from clangcheck showing in the quickfix pane. Making it impossible to navigate to that file.

I found out that adding the line below fixed the problem:
let g:ale_cpp_clangcheck_options = '--extra-arg="-fno-color-diagnostics"'
Using clang-check from LLVM version 6.0.0

So my question is - shouldn't that option be default?

Cheers,

Daniel

elebow:
Also change to the modern `--extra-arg` syntax.

clang has supported `-fno-color-diagnostics` for at least ten years: https://github.com/llvm-mirror/clang/commit/a46c71abb1dc2f9758ad1abe3378534fe138e802

---

This works on my machine, but I don't use clang often so a second tester would be appreciated.

### DIFF
--- a/ale_linters/cpp/clangcheck.vim
+++ b/ale_linters/cpp/clangcheck.vim
@@ -20,7 +20,7 @@ function! ale_linters#cpp#clangcheck#GetCommand(buffer) abort
     " being generated. These are only added if no build directory can be
     " detected.
     return '%e -analyze %s'
-    \   . (empty(l:build_dir) ? ' -extra-arg -Xclang -extra-arg -analyzer-output=text' : '')
+    \   . (empty(l:build_dir) ? ' --extra-arg=-Xclang --extra-arg=-analyzer-output=text --extra-arg=-fno-color-diagnostics': '')
     \   . ale#Pad(l:user_options)
     \   . (!empty(l:build_dir) ? ' -p ' . ale#Escape(l:build_dir) : '')
 endfunction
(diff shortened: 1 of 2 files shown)


ale_linters/cpp/clangcheck.vim
```diff
@@ -20,7 +20,7 @@ function! ale_linters#cpp#clangcheck#GetCommand(buffer) abort
     " being generated. These are only added if no build directory can be
     " detected.
     return '%e -analyze %s'
-    \   . (empty(l:build_dir) ? ' -extra-arg -Xclang -extra-arg -analyzer-output=text' : '')
+    \   . (empty(l:build_dir) ? ' --extra-arg="-Xclang" --extra-arg="-analyzer-output=text" --extra-arg="-fno-color-diagnostics"': '')
```
w0rp: I reckon it's worth removing the double quotes here, as double quotes can sometimes break commands on Windows. Processing double quotes in a shell in Windows is the responsibility of the command instead of the shell, which causes issues where you sometimes end up passing `"foo"` to a Windows program instead of `foo`.

elebow: Double quotes removed.

w0rp:
Cheers! :beers:

Issue #1519, opened 2018-04-23
Implement renaming symbols via LSP and tsserver


w0rp:
LSP and tsserver support operations for renaming things, and ALE should support these operations.

mwilliammyers:
I am interested in tackling this...

wavded:
This would be super handy to have. Any plans still @w0rp ? I would prefer to just use ale than have another lang server plugin that supports this.

w0rp:
The plan is to just eventually implement it somehow.

eugene-bright:
LanguageClient-neovim implements this feature. It's a subject to some 'code sharing'.
https://github.com/autozimu/LanguageClient-neovim/issues/35#issuecomment-288731403

w0rp:
It sure does. It's not ALE. I'll implement this eventually, or someone else can try.

eugene-bright:
May we get the feature by low price by passing LSP connection to LanguageClient-neovim functions? Raw ugly but possible working idea.

mwilliammyers:
I have an old [commit](https://github.com/mwilliammyers/ale/commit/5eb8d3) that adds support for the lsp and tsserver renaming protocols. 

I haven't added support for actually performing the rename in vim and probably won't have time to work on it for a while, but I have rebased my branch onto upstream so I could open a work in progress PR or someone is welcome to use [my commit](https://github.com/mwilliammyers/ale/commit/5eb8d3) as a basis...

w0rp:
Thank you for sharing that. Someone might pick it up and finish the job. :+1:

Risto-Stevcev:
I tried getting this to work a while back but it doesn't seem to rename anything:
https://github.com/Risto-Stevcev/ale/commit/a94bf9b095d8a542f17c9ceeece311d2ab283077

This is such a critical IDE feature, it would be nice to get this in

jeremija:
I made rename work for TypeScript by starting off of @mwilliammyers's commit (thanks!). It's on my fork on the [feature/renaming-symbols](https://github.com/jeremija/ale/compare/feature/renaming-symbols) branch. Here's the link to the main [commit](https://github.com/jeremija/ale/commit/2e6e77a87b2931a4966112fa531177360cc5d6b4). I also merged `upstream/master` into it to keep it up to date.

[...]

PR #2709, opened 2019-08-17
Add ALERename (tsserver & LSP), ALEOrganizeImports (tsserver) and auto import support (tsserver)

jeremija:
Related to #1519, 

### ALERename [new_name]

Tested for `tsserver`, and should work for any LSP that has `rename` capability, but I currently do not have a LSP project to test it on. Currently only supports renaming of symbols (no file rename).

### ALEOrganizeImports

Works with `tsserver` and will organize all import statements when called.

### Auto Imports

Currently works with `tsserver`. The `ts@completions` message now included an optional parameter: `includeExternalModuleExports`. If such a module is returned, the necessary code actions will be stored in the `user_data` variable of the completion item.

To support both `ale-manual` and `ale-auto-automatic` completion triggers, a new `augroup ALECompletionActions` has been added.

This functionality can be disabled by setting `g:ale_completion_tsserver_autoimport` to `0`.

---

A new function was added for handling code actions: `ale#code_action#HandleCodeAction`. I've tested this extensively on a TypeScript project. I would appreciate feedback, especially on the implementation of this function. There might be a better way to edit the buffer without resorting to visual mode.


w0rp:
Nice work! :+1: I've got a long list of emails to get through, and I'll take a look when I can.

autoload/ale/lsp/tsserver_message.vim
```diff
@@ -37,11 +37,14 @@ function! ale#lsp#tsserver_message#Geterr(buffer) abort
 endfunction
 
 function! ale#lsp#tsserver_message#Completions(buffer, line, column, prefix) abort
+    let l:external = get(g:, 'ale_completion_tsserver_autoimport', 1)
```
w0rp: Turn this off by default instead. That what current behaviour won't change, and we'll continue to show more refined results. Compute the option in the `completion.vim` file instead. These functions should not check any settings, they are just for getting the messages.

autoload/ale/lsp/tsserver_message.vim
```diff
@@ -77,3 +80,23 @@ function! ale#lsp#tsserver_message#Quickinfo(buffer, line, column) abort
     \   'file': expand('#' . a:buffer . ':p'),
     \}]
 endfunction
+
+function! ale#lsp#tsserver_message#Rename(buffer, line, column) abort
+    return [0, 'ts@rename', {
+    \   'line': a:line,
+    \   'offset': a:column,
+    \   'file': expand('#' . a:buffer . ':p'),
+    \   'arguments': {'findInComments': v:true, 'findInStrings': v:true}
```
w0rp: Make finding names in comments and strings optional, and turn both off by default. They can be fuzzy and hit some false positives.

autoload/ale/rename.vim
```diff
@@ -0,0 +1,193 @@
...
+                endfor
+
+                call add(l:changes, {
+                \   'fileName': l:filename,
+                \   'textChanges': l:text_changes,
+                \})
+            endfor
+
+            if empty(l:changes)
+                call ale#util#Execute('echom ''Could not rename.''')
```
w0rp: If it's easy to print the symbol name the user was trying to rename here, include that. Just "Could not rename" sounds weird.

autoload/ale/code_action.vim
```diff
@@ -0,0 +1,73 @@
...
+            call ale#util#Execute('buffer ' . l:buf)
+        else
+            call ale#util#Execute('edit ' .  l:file_code_edit.fileName)
+            let l:buf = ale#util#GetBufferNumber(l:file_code_edit.fileName)
+        endif
+
+        let l:initial_pos = getpos('.')
+        let l:line_diff = 0
+
+        for l:code_edit in reverse(copy(l:file_code_edit.textChanges))
```
w0rp: Try what I was working on for applying changes to file instead: https://github.com/dense-analysis/ale/commit/ca2026bd13346cb7e1afeecb54dbec3da0bd5151

1. Search for an open buffer, and use `getbufline` to read it if the buffer exists. If the filename isn't open in a buffer, use `readfile()` to read the file.
2. Work out how to add in the text changes by slicing lines and then columns.
3. Write the list of lines back to the file again.

That should work pretty well. There may be some bug with it I hadn't spotted yet, but I had a series of tests against it that worked.

You can add that code in, and exclude the other code I had for context menus for now. I eventually plan on supporting pretty much all LSP code actions through either the right click context menu, or through a console menu. I might throw in a command for applying code actions.

autoload/ale/completion.vim
```diff
@@ -321,13 +324,21 @@ function! ale#completion#ParseTSServerCompletionEntryDetails(response) abort
...
         \   'word': l:suggestion.name,
         \   'kind': l:kind,
         \   'icase': 1,
         \   'menu': join(l:displayParts, ''),
         \   'info': join(l:documentationParts, ''),
-        \})
+        \}
+
+        if has_key(l:suggestion, 'codeActions')
+            let l:result.user_data = json_encode({
```
w0rp: Try removing `json_encode` here and `json_decode` below. I think you can store `Dictionary` values too.

autoload/ale/completion.vim
```diff
@@ -451,10 +462,11 @@ function! ale#completion#HandleTSServerResponse(conn_id, response) abort
     let l:command = get(a:response, 'command', '')
 
     if l:command is# 'completions'
+        let l:completions = ale#completion#ParseTSServerCompletions(a:response)
         let l:names = ale#completion#Filter(
         \   l:buffer,
         \   &filetype,
-        \   ale#completion#ParseTSServerCompletions(a:response),
+        \   l:completions,
```
w0rp: You can change these lines back to how they were.

autoload/ale/completion.vim
```diff
@@ -670,6 +691,29 @@ function! ale#completion#Queue() abort
...
+function! ale#completion#HandleUserData(completed_item) abort
+    let l:source = get(get(b:, 'ale_completion_info', {}), 'source', '')
+
+    if l:source isnot# 'ale-automatic' && l:source isnot# 'ale-manual'
+        return
+    endif
+
+    let l:user_data_json = get(a:completed_item, 'user_data', '')
+
+    if l:user_data_json is# ''
```
w0rp: You can use `empty(l:user_data_json)` to check for any empty value.

autoload/ale/completion.vim
```diff
@@ -670,6 +691,29 @@ function! ale#completion#Queue() abort
...
+        return
+    endif
+
+    let l:user_data = json_decode(l:user_data_json)
+
+    if has_key(l:user_data, 'codeActions')
+        for l:code_action in l:user_data.codeActions
```
w0rp:
```suggestion
        for l:code_action in get(l:user_data, 'codeActions' [])
```

You can do this, and then you don't need the `if` statement.

autoload/ale/completion.vim
```diff
@@ -670,6 +691,29 @@ function! ale#completion#Queue() abort
...
+            " echom 'l:code_action: ' .string(l:code_action)
```
w0rp: Remove this test code.

autoload/ale/code_action.vim
```diff
@@ -0,0 +1,73 @@
...
+    let l:changes = a:code_action.changes
+
+    for l:file_code_edit in l:changes
+        let l:buf = ale#util#GetBufferNumber(l:file_code_edit.fileName)
+
+        if l:buf != -1
+            let l:existing_buffers[l:buf] = 1
+
+            if l:buf != l:current_buffer && ale#util#IsModified(l:buf)
+                call ale#util#Execute('echom ''Aborting action, file is unsaved''')
```
w0rp: I reckon just apply the changes anyway, even if buffers are modified. The changes will often be merged into files you're currently editing, and so they are likely to be modified. The chance of modifying something someone didn't want to save is rare. If we do want to not modify something for some reason, then don't change buffers for them, and don't apply any changes. Let them change buffers.

autoload/ale/util.vim
```diff
@@ -477,3 +477,23 @@ function! ale#util#FindItemAtCursor(buffer) abort
...
+
+" A wrapper function for setpos() so we can test calls for it.
+function! ale#util#SetPos(expr, list) abort
+    call setpos(a:expr, a:list)
+endfunction
+
+" This method exists to make mocking in tests easier
+function! ale#util#IsModified(buffer) abort
+    return getbufvar(a:buffer, '&mod')
+endfunction
```
w0rp: I reckon all of these wrappers should be removed. See my existing tests for applying changes. You can test actually modifying buffers, which is a much better test.

autoload/ale/completion.vim
```diff
@@ -678,10 +722,12 @@ function! ale#completion#Done() abort
     let s:last_done_pos = getpos('.')[1:2]
 endfunction
 
+augroup ALECompletionActions
+    autocmd CompleteDone * call ale#completion#HandleUserData(v:completed_item)
+augroup END
+
 function! s:Setup(enabled) abort
     augroup ALECompletionGroup
-        autocmd!
```
w0rp: Put this back again. Every completion group should start by clearing out previous `autocmd` lines.

autoload/ale/completion.vim
```diff
@@ -678,10 +722,12 @@ function! ale#completion#Done() abort
     let s:last_done_pos = getpos('.')[1:2]
 endfunction
 
+augroup ALECompletionActions
+    autocmd CompleteDone * call ale#completion#HandleUserData(v:completed_item)
```
w0rp: Put this logic into the existing `ale#completion#Done()` function instead. You can start passing `v:completed_item` to that.

autoload/ale/rename.vim
```diff
@@ -0,0 +1,193 @@
+let s:rename_map = {}
```
w0rp: Start the file with `Author` and `Description` lines to explain the purpose of the file and who did most of the work on it.

autoload/ale/rename.vim
```diff
@@ -0,0 +1,193 @@
...
+        if !has_key(a:response, 'result')
+            call ale#util#Execute('echom ''Could not rename.''')
+
+            return
+        endif
+
+        let l:workspace_edit = a:response.result
+
+        if !has_key(l:workspace_edit, 'changes')
+            call ale#util#Execute('echom ''Could not rename.''')
```
w0rp: Write a script function for printing this message, so you don't have to write the same message several times throughout the file.

autoload/ale/organize_imports.vim
```diff
@@ -0,0 +1,64 @@
+function! ale#organize_imports#HandleTSServerResponse(conn_id, response) abort
```
w0rp: Add `Author` and `Description` lines.

autoload/ale/organize_imports.vim
```diff
@@ -0,0 +1,64 @@
...
+        if !empty(l:linter.lsp)
+            call add(l:lsp_linters, l:linter)
+        endif
+    endfor
+
+    if !empty(l:lsp_linters)
+        for l:lsp_linter in l:lsp_linters
+            call s:OrganizeImports(l:lsp_linter)
+        endfor
+    endif
```
w0rp: Instead of appending entries to `lsp_linters` and then using them below, you can replace the call to `add` with `s:OrganizeImports` above.

doc/ale.txt
```diff
@@ -564,7 +568,6 @@ g:ale_completion_delay                                 *g:ale_completion_delay*
 
   See |ale-completion|
 
-
```
w0rp: Leave the double spacing in place. I tend to write two blank lines between each option.

doc/ale.txt
```diff
@@ -2507,6 +2518,16 @@ ALEHover                                                             *ALEHover*
 
   A plug mapping `<Plug>(ale_hover)` is defined for this command.
 
+ALEOrganizeImports                                         *ALEOrganizeImports*
+
+  Organize imports using tsserver. Currently not implemented for LSPs.
+
```
w0rp:
```suggestion


```

Write two blank lines between each option/command, etc.

autoload/ale/rename.vim
```diff
@@ -0,0 +1,193 @@
...
+
+    for l:linter in ale#linter#Get(&filetype)
+        if !empty(l:linter.lsp)
+            call add(l:lsp_linters, l:linter)
+        endif
+    endfor
+
+    if !empty(l:lsp_linters)
+        let l:prompt = 'new name: '
+        let l:new_name = a:0 ? join(a:000) : input(l:prompt, expand('<cWORD>'))
```
w0rp: Instead of supporting an optional argument for a new name in the command, always require the new name to be input with `input()` and the current name. Then you always know exactly what name ALE is going to try and rename, and it's easy to adjust the current name to something slightly different. See here: https://github.com/dense-analysis/ale/commit/ca2026bd13346cb7e1afeecb54dbec3da0bd5151#diff-c8e42b11378165eaf5fb1be0e53860e6R90

autoload/ale/rename.vim
```diff
@@ -0,0 +1,193 @@
...
+
+        if !empty(l:new_name)
+            for l:lsp_linter in l:lsp_linters
+                call s:ExecuteRename(l:lsp_linter, l:new_name)
+            endfor
+        endif
```
w0rp: Print messages when a symbol can't be renamed because the new name is empty, or when no servers are available for the action.

autoload/ale/completion.vim
jeremija: I thought so too, but I get `E731: using Dictionary as a String` when I remove the `json_encode`/`json_decode` functions :(

autoload/ale/completion.vim
jeremija: Roger, I removed this by accident.

autoload/ale/completion.vim
jeremija: I believe that `ale#completion#Done()` will not be called when automatic completion is disabled, and the completion is triggered manually (via a hotkey). That's why I created a new group. Perhaps I can modify the `s:Setup` to look like:

```vim
function! s:Setup(enabled) abort
    augroup ALECompletionGroup
        autocmd!
        if a:enabled
            autocmd TextChangedI * call ale#completion#Queue()
        endif
        if a:enabled || g:ale_completion_tsserver_autoimport
            autocmd CompleteDone * call ale#completion#Done(v:completed_item)
        endif
    augroup END

    if !a:enabled && !g:ale_completion_tsserver_autoimport
        augroup! ALECompletionGroup
    endif
endfunction
```

What do you think?

autoload/ale/code_action.vim
jeremija: I added this check because tsserver could return wrong ranges in case a buffer was modified, since it will read the files from the disk and not from the buffer. For example if somebody tries to rename the `value` symbol from `file2.ts`:

### file1.ts
```typescript
/* 1 */ class A {
/* 2 */   value: number
/* 3 */ }
```

### file2.ts
```typescript
/* 1 */ import {A} from './file1'
/* 2 */ const a = new A()
/* 3 */ console.log(a.value)
```

If `file1.ts` was modified and a line was added above line 2, then the result from `tsserver` would not match the actual contents of the buffer.

jeremija: Or are you saying the buffers should be saved before sending any requests to tsserver?

autoload/ale/rename.vim
jeremija: I implemented this change but since `feedkeys()` doesn't work for input prompt, I added a new util method `ale#util#Input(message, value)` and mocked that method in the test. Let me know if this works!

autoload/ale/code_action.vim
jeremija: Thanks, this was super helpful! I fixed some indexing bugs that I encountered with my tests and implemented post-edit cursor movement.

jeremija:
Thanks for all your comments! Your commit references were super helpful! This change turned out to be a lot bigger than I initially expected.

I believe I have addressed most your change requests, I'm just putting everything in one place since mot if the code comments above are now marked as `outdated`:

1. Remove newly added util methods; test buffer contents instead. I used `writefile` and `readfile` over `Given`/`Expect` blocks because this way I was able to test for modifications of multiple files.

2. Turn off autoimport by default

3. Turn off findInStrings and findInComments and make it optional

4. Remove unnecessary array creations/iterations, and commented out echom statements

5. Add more descriptive error messages for `ALERename` and a helper function for printing them

6. Always prompt user for new name when using `ALERename`

7. Fix spacing in `doc/ale.txt` and add descriptions of new config vars

8. Pass config variables as arguments when constructing tsserver messages, instrad of using them directly when creating messages.

9. Add more robust cursor movement when modifying modifying current buffer in `code_actions.vim`

10. Add Author and Description to new files.

Unfortunately, it looks like dictionaries cannot be used for `user_data`, so I had to keep `json_encode`/`json_decode`.

---

I only have a few more questions/comments:

1. What to do if some of the files to be changed have been modified? See my comment [here](https://github.com/dense-analysis/ale/pull/2709#discussion_r315025352)

2. What do you think about my [modification suggestions](https://github.com/dense-analysis/ale/pull/2709#discussion_r315023925) for `ALECompletionGroup`? to make it work with manually triggered completion?

3. I had to add `ale#util#Input()` to be able to test a function that calls `input()`  Otherwise the tests would hang and I was unable to enter the name in the test by using `feedkeys()`.

cirias:
Nice work! But I found autoimport doesn't work when I use deoplete. Is it possible to support that?

jeremija:
Hi @cirias, I do not use `deoplete` so I can't test this. Could you try to add a check for `deoplete` completion source on this line: https://github.com/dense-analysis/ale/blob/61cfb3fefb0ebd8654be452046bd2ba24025311f/autoload/ale/completion.vim#L721

it should look like:

```vim
    if l:source isnot# 'ale-automatic' && l:source isnot# 'ale-manual' && l:source isnot# 'deoplete'
```

Let me know if this works!

cirias:
I tried. But it still doesn't work. Also I found the problem is, when `g:ale_completion_enabled` is 0, it doesn't list any external symbols as candidates. As long as `let g:ale_completion_enabled = 1`, it works correct. Hope that helps you debug

jeremija:
Perhaps we should move this conversation to a new issue. I just installed and configured deoplete to use ale and I believe I was able to reproduce the issue.

Can you confirm that the same happens to you:

a) deoplete is configured to use ale
b) external symbols show up the first time I enter insert mode and start typing. auto import works for external symbols listed if I add the change in the comment above
c) external symbols stop showing up the next time I enter insert mode

w0rp:
Importing external symbols isn't currently designed to work with Deoplete. That will require some additional work. Open an issue for that.

thisjeremiah:
Excited that this plugin could replace our usage of `nvim-typescript`!

I have noticed that the autoimport feature does not work when autocomplete is manually executed via `<c-o><c-x>`. Is that the intended behavior?

jeremija:
@thisjeremiah What configuration are you using to trigger completion via `<c-o><c-x>`?

This is my configuration for `<C-Space>` and it works for me:

```vim
imap <C-Space> <Plug>(ale_complete)
```

thisjeremiah:
I've actually just been using completion via `<enter>` or `<esc>`, which successfully fills out the term selected in the popup menu, but the autoimport does not execute.

I set up your mapping and it worked, probably because it is calling `ale_complete` directly. I'll probably use that for now. Thanks!

In general, I'm trying to reduce the steps needed to autoimport an already written term. The `:TSImport` command from `nvim-typescript` had a few less steps since that command could be run from normal mode with the term selected, whereas this command has to run in insert mode and manually execute `ale_complete`.

heyrict:
Glad that ale is on the road to support `CodeAction`! I used to depend heavily on YCM's `FixIt` command when developing typescript projects and this will finally remove YCM from my `YCM + ALE` stack!

One thing (though not bug) I found inconvenient is that, You don't know which file the symbol you are importing from. e.g. When I want to import `compose` from `react-redux`, it shows `compose` from several other packages which are hard to differenciate.

Screenshot from 2019-09-19 19-34-33

w0rp:
If anyone knows enough about Deoplete to make it work, please create a pull request.

I plan on adding code action support with Vim menus later, when I finally have the time. I have some experimental code for that in a branch.

I did notice one error with renaming in other files I saw once, but I haven't been able to repeat it since.

jeremija:
I believe I found a fix for Deoplete, see #2779. @cirias can you test this?

jeremija:
@heyrict The autoimport description should now be visible on `master` since 7b38e97 commit. See #2780 for more details.

PR #2710, opened 2019-08-17
Allow user to set their own completion symbols within ALE

DonnieWest:
Not sure if this is a desired change, but I had it running locally for a bit and figured others could benefit from it. Having nerd-font esque symbols in completions is pretty nice :)

I'm willing to make any changes necessary to get this in, but I might need some insight on how to test it since I'm not super sure how to test if the symbols have been replaced in the pum.

Thanks!

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -39,6 +39,17 @@ let s:LSP_COMPLETION_COLOR_KIND = 16
 let s:LSP_COMPLETION_FILE_KIND = 17
 let s:LSP_COMPLETION_REFERENCE_KIND = 18
 
+let g:ale_completion_symbols = get(g:, 'ale_completion_symbols', {
+\ '<default>': 'v',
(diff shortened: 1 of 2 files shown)


delphinus:
I'm curious on this. Can you put screenshots in completing with this feature?

autoload/ale/completion.vim
```diff
@@ -433,7 +429,7 @@ function! ale#completion#ParseLSPCompletions(response) abort
 
         call add(l:results, {
         \   'word': l:word,
-        \   'kind': l:kind,
+        \   'kind': ale#completion#GetCompletionSymbols(get(l:item, 'kind', '')),
```
w0rp: `l:item.kind` is  a number, as defined above, and there are a few more in the spec now too. https://microsoft.github.io/language-server-protocol/specification

Maybe we should map those numbers to strings that are easier for users to configure the new setting with.

PR #2719, opened 2019-08-20
Remove texlab rust rewrite link

kevinywlui:
It seems like the texlab rust rewrite has completed so the link is now broken and not necessary.


w0rp:
Cheers! :beers:

PR #2720, opened 2019-08-20
Python linter pylint generate --indent-string opt

ghthor:
# Description

I've added an additional feature to pylint command `let g:ale_python_pylint_auto_indent_string` that when enabled will generate the value for the pylint option `--indent-string` from the buffers `expandtab,shiftwidth,tabstop` values. I modeled this off the implementation for the [shfmt fixer](https://github.com/dense-analysis/ale/blob/master/autoload/ale/fixers/shfmt.vim#L8) which does something similar.

# Motivation

To reduce the sources of truth for this value down to one place.

From

```vimscript
let g:ale_python_pylint_options = '--indent-string="  "'
autocmd FileType python set expandtab tabstop=2 shiftwidth=2
```

To

```vimscript
let g:ale_python_pylint_auto_indent_string = 1
autocmd FileType python set expandtab tabstop=2 shiftwidth=2
```

### DIFF
--- a/ale_linters/python/pylint.vim
+++ b/ale_linters/python/pylint.vim
@@ -7,6 +7,7 @@ call ale#Set('python_pylint_use_global', get(g:, 'ale_use_global_executables', 0
 call ale#Set('python_pylint_change_directory', 1)
 call ale#Set('python_pylint_auto_pipenv', 0)
 call ale#Set('python_pylint_use_msg_id', 0)
+call ale#Set('python_pylint_auto_indent_string', 0)
 
 function! ale_linters#python#pylint#GetExecutable(buffer) abort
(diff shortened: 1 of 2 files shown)


RyanSquared:
I'm not sure whether it's a good idea to make a change like this without adding a relevant option. Either way, the tests are failing because they're not expecting the option. Do you mind fixing that?

ghthor:
I'd rather have this enabled by default, as the `shfmt` fixer is. But I'm happy with upgrading this into an option!

ghthor:
Looking for a review here, not trying to nag, just trying to communicate clearly that this is completed from my perspective and ready for review.

ghthor:
@RyanSquared Can you review this for merging?

PR #2721, opened 2019-08-21
clangtidy: ensure *.h header files are treated as C++ (filetype cpp)

### LINKED ISSUE #1608: clang-tidy treats .h header files as c headers
## clang-tidy not working properly for c++ header files

**VIM version**

NVIM v0.2.2
Build type: Release

Operating System: macOS 10.13.4

### :ALEInfo

```
 Current Filetype: cpp
Available Linters: ['clang', 'clangcheck', 'clangtidy', 'cppcheck', 'cpplint', 'flawfinder', 'gcc']
   Linter Aliases:
'gcc' -> ['g++']
  Enabled Linters: ['clangtidy']
 Linter Variables:

let g:ale_cpp_clangtidy_checks = ['*']
let g:ale_cpp_clangtidy_executable = 'clang-tidy'
let g:ale_cpp_clangtidy_options = ''
......
```

## What went wrong

When current file is a C++ header file with extension ‘.h', ALE can recognize it is a C++ header file. But the clang-tidy seems to treat it as a C header file. As it find error like "'string' file not found" and "unknown type name 'namespace'".

If I rename the header file with extension '.hpp', then the clang-tidy can work properly. Do anyone has ideas to solve this problem?

## Reproducing the bug

1. Install clang-tidy
2. Open C++ header file with extension '.h'

richyfish:
Fixes #1608

Including "-x c++" option such that clang-tidy treats .h files as C++.  

This PR only applies to `cpp` filetype - the default for .h files in vim.

### DIFF
--- a/ale_linters/cpp/clangtidy.vim
+++ b/ale_linters/cpp/clangtidy.vim
@@ -25,6 +25,14 @@ function! ale_linters#cpp#clangtidy#GetCommand(buffer) abort
     " Get the options to pass directly to clang-tidy
     let l:extra_options = ale#Var(a:buffer, 'cpp_clangtidy_extra_options')
 
+    if expand('#' . a:buffer . ':e') is? 'h'
+        " for .h header files, passing -x to compiler to force cpp
+        if !empty(l:options)
+            let l:options .= ' '
+        endif
+        let l:options .= '-x c++'
+    endif
+
     return '%e'
     \   . (!empty(l:checks) ? ' -checks=' . ale#Escape(l:checks) : '')
     \   . (!empty(l:extra_options) ? ' ' . ale#Escape(l:extra_options) : '')
--- a/ale_linters/cpp/clangtidy.vim
+++ b/ale_linters/cpp/clangtidy.vim
@@ -30,6 +30,7 @@ function! ale_linters#cpp#clangtidy#GetCommand(buffer) abort
         if !empty(l:options)
             let l:options .= ' '
         endif
+
         let l:options .= '-x c++'
     endif
 
--- a/test/command_callback/test_clang_tidy_command_callback.vader
+++ b/test/command_callback/test_clang_tidy_command_callback.vader
@@ -63,7 +63,7 @@ Execute(The build directory should be ignored for header files):
 
   AssertLinter 'clang-tidy',
   \ ale#Escape('clang-tidy')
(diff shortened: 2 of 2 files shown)


AbigailBuccaneer:
Passing `c` or `c++` isn't strictly the right thing to do for headers at all; `c-header` or `c++-header` is correct. I just made a draft pull request, #2724, that fixes this for the four `clang` linters. It adds an `ale#c#GetLanguage(buffer)` function that could be used here.

richyfish:
@AbigailBuccaneer  - thanks, good suggestion.  I will try with `c++-header` for this PR.

richyfish:
Hi @AbigailBuccaneer  - upon testing clang-tidy with 'c++-header'  it appears not to be working yielding 'Compile command not found', but with `c++` it is working. 


```
~/projects/cpp$ ls
Makefile  build  compile_commands.json  header.h  header.h.pch  header.pch  main.cpp  main.plist  other.cpp  other.plist

~/projects/cpp$ clang-tidy -checks='*' header.h -p compile_commands.json   -- -x c++
1 error generated.
Error while processing /home/docker/projects/cpp/header.h.
/home/docker/projects/cpp/header.h:4:23: error: expected ';' after top level declarator [clang-diagnostic-error]
static const int x = 0
                      ^
                      ;

~/projects/cpp$ clang-tidy -checks='*' header.h -p compile_commands.json   -- -x c++-header
Skipping /home/docker/projects/cpp/header.h. Compile command not found.

~/projects/cpp$

```
I'm not sure what the subtle differences are between using -x c++ vs c++-header - are you able to explain what they are ?

AbigailBuccaneer:
The primary example of differences I'm aware of is only a semicolon away from yours, and is mentioned in #2724: `static const int x = 0;` gives a `-Wunused-const-variable` warning when compiled with `-x c` or `-x c++`, but not `-x c-header` or `-x c++-header`.

I don't know how `clang-tidy` handles checking headers, now I think about it - there's no entry for `header.h` in the `compile_commands.json` you pass in. How does it get any compile flags? Does it find a corresponding translation unit, or just pick some other translation unit to use, or something else?

richyfish:
I'm closing this PR because of compatibility conflicts with C: 

cpp is the default filetype for .h files in Vim. This means when you have a .h header file open under a C project (not C++) then ALE will think this file is -x c++ which is clearly wrong.

Instead I think the fix should introduce a new variable that switches on this behaviour for c++ header files.

PR #2724, opened 2019-08-22
clang: pass appropriate language for header files

AbigailBuccaneer:
Some C/C++ constructs cause errors in translation units, but not in header files. For example, the following:

```c
static const int x = 0;
```

...fails to compile with `clang -x c -Wall -Werror -fsyntax-only`, but succeeds with `clang -x c-header -Wall -Werror -fsyntax-only`.

This commit updates the Clang linters to use a buffer's filetype and extension to pass the appropriate language as `-x`. This fixes false-positive errors when linting header files.

I've left this marked as a draft for now as it doesn't include any tests - I've yet to figure out how to write tests, and I'm out of free time for today. This is my first time writing any Vimscript more complicated than a `.vimrc` so if there are any coding guidelines I've ran afoul of please let me know!

### DIFF
--- a/ale_linters/c/clang.vim
+++ b/ale_linters/c/clang.vim
@@ -9,7 +9,8 @@ function! ale_linters#c#clang#GetCommand(buffer, output) abort
 
     " -iquote with the directory the file is in makes #include work for
     "  headers in the same directory.
-    return '%e -S -x c -fsyntax-only'
+    return '%e -S -fsyntax-only'
+    \   . ' -x ' . ale#c#GetLanguage(a:buffer)
     \   . ' -iquote ' . ale#Escape(fnamemodify(bufname(a:buffer), ':p:h'))
     \   . ale#Pad(l:cflags)
     \   . ale#Pad(ale#Var(a:buffer, 'c_clang_options')) . ' -'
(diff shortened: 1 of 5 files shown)


w0rp:
This code isn't needed. If the filetype is already set appropriately, everything should work.

AbigailBuccaneer:
@w0rp what's the appropriate filetype for C/C++ headers? There's no separate "C header" or "C++ header" filetype that I'm aware of. The point of this PR is that C++ headers should be treated differently from C++ translation units (and C headers differently from C translation units), despite having the same Vim filetype.

w0rp:
You have to set `filetype` setting in Vim to whatever filetype you want. ALE doesn't detect filetypes, it runs linters based on whatever filetype it was given.

AbigailBuccaneer:
Yes, and my point is that there is no such thing as a separate Vim filetype for C/C++ headers - they use the same `c` and `cpp` filetypes as translation units. If I did invent my own filetypes for them then ALE would no longer lint them. But they require special handling.

w0rp:
You have to set the `filetype` appropriately based on the project. It could be `cpp` for `.h` files in a C++ project, `c` for `.h` files in a C project, Objective C, etc.

AbigailBuccaneer:
@w0rp I think you've misunderstood the issue here. The project I encountered this bug in uses `.hpp` files, which are correctly given the `cpp` filetype by default. This isn't about changing or ignoring the filetype, or about a mixup between the `c` and `cpp` filetypes - in fact, this pull request relies on a buffer's filetype.

I've just updated the issue description to clarify what I'm talking about.

Issue #1398, opened 2018-03-05
Flow 0.66 changed the error messages format


ahmedelgabri:
Flow 0.66 changed the error messages format [more here](https://medium.com/flow-type/better-flow-error-messages-for-the-javascript-ecosystem-73b6da948ae2) which now makes ale messages not very useful

w0rp:
The changelog claims that existing plugins should continue to work, until you pass a flag to use the new JSON format.

ahmedelgabri:
But that's not the case when I see the errors from ale here is my ale [config](https://github.com/ahmedelgabri/dotfiles/blob/master/vim/.vim/after/plugin/ale.vim) & this is an example:

**Flow `0.53.1`**

<img width="1005" alt="flow 0 53 1" src="https://user-images.githubusercontent.com/63876/37039997-1bc955e8-2159-11e8-9cf1-99795e334334.png">

**Flow `0.66`**

<img width="1003" alt="flow 0 66" src="https://user-images.githubusercontent.com/63876/37040018-26b3fcba-2159-11e8-976b-9eae1bac8a96.png">

w0rp:
Sounds like they fucked up, then. I might have a look when I get the time, or someone else can create a pull request for it. I believe some version detection is already down for the linter, so it shouldn't be too hard to fix.

ahmedelgabri:
@TrySound the error messages are not complete, if you check he article I linked to flow now with have something like footnotes or markers. And they are not always going be shown in full now, here is an example, check the second error. It's not as clear & maybe even gets cut off now

<img src="https://cdn-images-1.medium.com/max/2000/1*BAYW1TlbDM9VhVYCoo2azQ.jpeg" >

TrySound:
I'd say before they weren't full too

ahmedelgabri:
Ok, then maybe I'm missing something. But after the flow update some messages didn't seem to make much sense, I'll keep an eye on it again & see.

loyd:
Also `ALEDetail` is useless now:
```
References: [1] [2]
```

w0rp:
Feel free to submit a pull request to fix it.

choonkending:
Ale newbie here. 

Is there any chance we could try to format the error message output by ale to be multiline and as descriptive as command line (My error messages appear in 1 long line now, and they can be difficult to read)? 

[...]

PR #2725, opened 2019-08-22
"Fixed" missing references in flow linter

sdeleon28:
Some of the output from Flow is [swallowed by ale](https://github.com/dense-analysis/ale/issues/1398). This makes it a little bit better. While the ideal thing would be to show the entire region delimited by the `loc` key in the extra messages (which is what flow outputs when used in human readable format), I've only added the first line of them. The more complete solution would be to read the referenced file, showing the range of lines and highlighting the bit where Flow is blowing up.


w0rp:
Looks like a good idea to me. Would you like to add a test for this?

Sorry for the delay.

PR #2742, opened 2019-08-31
clangtidy treat .h files as C++ when g:cpp_clangtidy_h_is_hpp

### LINKED ISSUE #1608: clang-tidy treats .h header files as c headers

richyfish:
Fixes #1608 

Introduced new variable for clangtidy `g:cpp_clangtidy_h_is_hpp`.  This enables behaviour for  clangtidy to treat the .h header file as C++. 

Since cpp is the default filetype for .h files in vim, I have introduced this variable to avoid "normal" C header (.h) files being treated as C++ which could yield incorrect linter results.

### DIFF
--- a/ale_linters/cpp/clangtidy.vim
+++ b/ale_linters/cpp/clangtidy.vim
@@ -13,6 +13,10 @@ call ale#Set('cpp_clangtidy_options', '')
 call ale#Set('cpp_clangtidy_extra_options', '')
 call ale#Set('c_build_dir', '')
 
+" Set this to enable treating .h header files as C++
+call ale#Set('cpp_clangtidy_h_is_hpp', 0)
+
+
 function! ale_linters#cpp#clangtidy#GetCommand(buffer) abort
     let l:checks = join(ale#Var(a:buffer, 'cpp_clangtidy_checks'), ',')
     let l:build_dir = ale#c#GetBuildDirectory(a:buffer)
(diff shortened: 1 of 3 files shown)


w0rp:
I'm gonna close this and make it so the C++ version of the clang-tidy linter tells `clang-tidy` that `.h` files are C++ files. Which checks you run should match the filetype in Vim.

PR #2622, opened 2019-06-26
feat(template-lint): Read from stdin

dcyriller:
## Description

This PR proposes to leverage a new feature introduced in ember-template-lint v1.6: reading from stdin.

Currently, ale copies the file to a temporary file and lints it from there. But this approach has a few issues regarding the configuration files (especially `.prettierrc.js`).

Issue #2227, opened 2019-01-20
Local Prettier config .prettierrc ignored for prettier-eslint fixer


jondkinney:
## Information

**VIM version**

NVIM v0.3.1
Build type: Release

Operating System: macOS Mojave 10.14.1

## What went wrong

Attempting to use `prettier-eslint` doesn't load the local `.prettierrc` config file. So when you save, prettier only formats with the defaults.

```
let g:ale_fixers = {
\   'javascript': ['prettier_eslint'],
\}

let g:ale_fix_on_save = 1
```

## Reproducing the bug

Save the file, and prettier will run, but with the default settings. Then ESLint will run `--fix` on the result.

I think that this is similar to https://github.com/w0rp/ale/issues/983 and if you look at the bottom of the ALEInfo below, you can see that `--stdin-filename` is _not_ being passed to `prettier-eslint`. So if that's the flag you need for it to infer the local `.prettierrc` file and/or traverse up the path until it finds one, then that'd be the issue, I think.

Looking back at lots of issues and commits, it seems that we can specify the fixers as both `prettier` and `eslint` (shown below) to achieve the same end results. Is that correct? Or are we missing out on some of the nuance of `prettier-eslint` by letting ALE do the passing of the prettier formatted output to eslint's `--fix`?

```
let g:ale_fixers = {
\   'javascript': ['prettier', 'eslint'],
\}
```

### :ALEInfo

```
 Current Filetype: javascript.jsx
Available Linters: ['eslint', 'flow', 'flow-language-server', 'jscs', 'jshint', 'standard', 'tsserver', 'xo']
  Enabled Linters: ['eslint']
 Suggested Fixers: 
  'eslint' - Apply eslint --fix to a file.
  'importjs' - automatic imports for javascript
  'prettier' - Apply prettier to a file.
  'prettier_eslint', 'prettier-eslint' - Apply prettier-eslint to a file.
  'prettier_standard', 'prettier-standard' - Apply prettier-standard to a file.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'standard' - Fix JavaScript files using standard --fix
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'xo' - Fix JavaScript files using xo --fix.
 Linter Variables:

let g:ale_javascript_eslint_executable = 'eslint'
let g:ale_javascript_eslint_options = ''
let g:ale_javascript_eslint_suppress_eslintignore = 0
[...]

Issue #1648, opened 2018-06-11
prettier-eslint can not find eslint dependency installed locally


roneyrao:
## Information

**VIM version**

PASTE JUST THE FIRST TWO LINES OF `:version` HERE.
VIM - Vi IMproved 8.1 (2018 May 17, compiled May 24 2018 14:10:12)
macOS version

Operating System: WHAT OS WERE YOU USING?
macOS
### :ALEInfo

PASTE OUTPUT OF `:ALEInfo` HERE. YOU CAN TRY `:ALEInfoToClipboard`.
(finished - exit code 7) ['/bin/bash', '-c', 'cd ''/Users/roneyrao/web/react_demo/src/configs'' && ''prettier-eslint'' --eslint-config-path ''/Users/roneyrao/web/react_demo/.eslintrc'' --stdin-filepath ''/Users/roneyrao/web/react_demo/src/configs/menus.js'' --stdin < ''/var/folders/qt/yllps6vs68986j5t3jx_16c80000gn/T/vyHKVFm/998/menus.js''']

## What went wrong

WRITE WHAT WENT WRONG HERE.
The actual error is locally installed eslint plugin can not be found. The reason is `prettier-eslint-cli` finds `--eslint-config-path` based on where itself is located.

## Suggestion
`--eslint-config-path` is redudant here, `prettier-eslint` will handle this correctly.

## Reproducing the bug

Steps for repeating the bug:

1. Write a list of steps.
eslint and plugins are installed locally in a project, then these plugins will not be found when call `ALEFix`.

2. Otherwise nobody will fix the bug.

w0rp:
What's the bug? Have you installed `prettier-eslint` globally and a plugin locally? You should install both of them locally, or both of them globally.

Issue #1425, opened 2018-03-15
.prettierignore file not used when fixing on save


cmvanb:
## Information
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Mar 14 2018 23:02:54)
MS-Windows 64-bit console version

Operating System: Windows 10 64bit

### :ALEInfo
 Current Filetype: javascript
Available Linters: ['eslint', 'flow', 'jscs', 'jshint', 'standard', 'xo']
  Enabled Linters: ['eslint', 'flow', 'jscs', 'jshint', 'standard', 'xo']
 Linter Variables:

let g:ale_javascript_eslint_executable = 'eslint'
let g:ale_javascript_eslint_options = ''
let g:ale_javascript_eslint_suppress_eslintignore = 0
let g:ale_javascript_eslint_suppress_missing_config = 0
let g:ale_javascript_eslint_use_global = 0
let g:ale_javascript_flow_executable = 'flow'
let g:ale_javascript_flow_use_global = 0
let g:ale_javascript_flow_use_home_config = 0
let g:ale_javascript_jscs_executable = 'jscs'
let g:ale_javascript_jscs_use_global = 0
let g:ale_javascript_jshint_executable = 'jshint'
let g:ale_javascript_jshint_use_global = 0
let g:ale_javascript_standard_executable = 'standard'
let g:ale_javascript_standard_options = ''
let g:ale_javascript_standard_use_global = 0
let g:ale_javascript_xo_executable = 'xo'
let g:ale_javascript_xo_options = ''
let g:ale_javascript_xo_use_global = 0
 Global Variables:

let g:ale_cache_executable_check_failures = 0
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = 100
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = 50
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 1
let g:ale_fixers = {'javascript': ['prettier']}
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
[...]

PR #2745, opened 2019-09-01
fixers/stylelint: enhance `stylelint` fixer

charlesbjohnson:
👋 

This upgrades the `stylelint` fixer. A few things have been done:

1. The fixer now supports `let g:ale_stylelint_options` for custom flags
2. The fixer now supports stdin by virtue of `--stdin-filename` (available in [6.3.0](https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md#630))
3. The fixer now changes the working directory before running `stylelint`

(much of this is inspired from [`fixers/prettier`](https://github.com/dense-analysis/ale/blob/master/autoload/ale/fixers/prettier.vim))

I believe the combination of 2 & 3 may resolve a few issues since it allows any `stylelint` plugins to find project configuration files as expected (similar to #2622). This _may_ fix:
- ~#2227~ Probably won't fix, seems unrelated to `stylelint`
- #1425 Possibly partial fix, assuming `prettier` is being used with `stylelint`
- ~#1648~ Probably won't fix, seems unrelated to `stylelint`


charlesbjohnson:
Please hold on review, I'm running into a bug where the error output is actually replacing the buffer when using stdin.

charlesbjohnson:
According to https://github.com/stylelint/stylelint/issues/3633, it seems to only happen when a non-CSS `--syntax` is used.

Without any `stylelint` configuration for SCSS:

```sh
❯ echo '$color:red; .foo { &-bar { color:$color } }' | npx stylelint --fix --stdin-filename foo.css
$color:red; .foo { &-bar { color:$color } }
```

With `stylelint` configuration for SCSS (ie. [`stylelint-scss`](https://github.com/kristerkari/stylelint-scss) with `scss/dollar-variable-colon-newline-after`) and valid syntax:

```sh
❯ echo '$color:red; .foo { &-bar { color:$color } }' | npx stylelint --fix --stdin-filename foo.css
$color:
red; .foo { &-bar { color:$color } }
```

With `stylelint` configuration for SCSS and invalid syntax:

```sh
❯ echo '.foo- {' | npx stylelint --fix --stdin-filename foo.css
.foo- {
}
```

As such, I've made a change to force a `.css` extension after the stdin filename. So far it seems to resolve the issue for me, but perhaps better to let it bake for some time 🍞 first.

w0rp:
Does this work acceptably well for other file formats?

charlesbjohnson:
I have not yet tested with other file formats since I've found that this workaround still isn't a good enough solution. I'm still occasionally running into issues where `stylelint` dumps a stacktrace to stdout and blows away my buffer.

Let me know how you want to proceed. My guess is that it's either a bug in `stylelint` or there's some flag that I'm unaware of. I have yet to take a look at how `stylelint` plugins for other editors are handling this.

w0rp:
If the `--stdin-filename` argument can cause some problems, maybe take that away for now and keep just the other changes which shouldn't cause any issues, and bring that back later when the issues have been resolved.

charlesbjohnson:
> take that away for now and keep just the other changes which shouldn't cause any issues

Done. Two things worth noting:

1. It seems that using a temporary file is still an issue for configuration detection. With [`stylelint-prettier`](https://github.com/prettier/stylelint-prettier), for example, `prettier` is unable to detect a project configuration file since it starts trying to find one from the file being formatted. This is essentially the same thing as #2622, and it will be resolved when we can use `--stdin-filename` without issue.

2. I believe I misjudged the issues I linked in the PR description. The behavior in those issues is similar, though, and I think the fix would be similar to what we're doing here but just for a different tool (ie `eslint` vs `stylelint`). These changes probably won't fix them.

> bring that back later when the issues have been resolved

Sounds good, stay tuned 📺

Thanks!

PR #2749, opened 2019-09-03
add languagetool commandline jar support and add according doc.

cyyever:


autoload/ale/handlers/languagetool.vim
```diff
@@ -2,18 +2,26 @@
 " Description: languagetool for markdown files
 "
 call ale#Set('languagetool_executable', 'languagetool')
+call ale#Set('languagetool_commandline_jar', 'languagetool-commandline.jar')
```
w0rp: Make this an empty string by default, or otherwise determine a valid path to the .jar.

doc/ale.txt
```diff
@@ -2001,6 +2004,13 @@ g:ale_languagetool_executable                   *g:ale_languagetool_executable*
 
   The executable to run for languagetool.
 
+g:ale_languagetool_commandline_jar              g:ale_languagetool_commandline_jar
+                                                b:ale_languagetool_commandline_jar
+
+  Type: |String|
+  Default: `'languagetool-commandline.jar'`
+
+  The commandline jar for languagetool.
```
w0rp: Explain that using this option will be mean `_executable` has no effect.

autoload/ale/handlers/languagetool.vim
cyyever: make it empty by default.

cyyever:
@w0rp I have added test and doc.

w0rp:
I thought about this, and realised it's actually not as useful as just setting the `_executable` option to the java executable, and specifying the jar in the `_options` variable. That lets you do the same thing, but also lets you pick different Java versions. I recommend doing that instead.

cyyever:
> I thought about this, and realised it's actually not as useful as just setting the `_executable` option to the java executable, and specifying the jar in the `_options` variable. That lets you do the same thing, but also lets you pick different Java versions. I recommend doing that instead.

if we use option to store jar path, then we lack a way to normalize the path, especially on Windows.

PR #2750, opened 2019-09-03
Elm language server params update

andys8:
## Changes

* There is a new param 'elmAnalyseTrigger' (change|save|never)
* Empty path default to let the language server search for binary

## Note

Although `--stdio` argument is now default, it'll still be passed for compatibility with people using older version of the language server,

## Settings documentation

https://github.com/elm-tooling/elm-language-server#server-settings


andys8:
Tested it and it works fine with `@elm-tooling/elm-language-server@1.4.0`.

andys8:
`elm-language-server` itself is still in early active development. So using older versions isn't encouraged.

The changes should be compatible, though.

PR #2751, opened 2019-09-03
Add an option to reuse the elixir mix build directory

pgr0ss:
The mix build directory is random, so linting requires a fresh build
every time which can take a while. This adds an option to reuse the
build directory so repeated linting is much faster.

I didn't see any elixir tests and I wasn't quite sure where to start with them.

### DIFF
--- a/ale_linters/elixir/mix.vim
+++ b/ale_linters/elixir/mix.vim
@@ -2,6 +2,9 @@
 " Author: colbydehart - https://github.com/colbydehart
 " Description: Mix compile checking for Elixir files
 
+call ale#Set('elixir_mix_random_build_dir', 1)
+call ale#Set('elixir_mix_build_dir', '/tmp/vim_mix_build')
+
 function! ale_linters#elixir#mix#Handle(buffer, lines) abort
     " Matches patterns like the following:
     "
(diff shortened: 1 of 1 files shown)


dewyze:
Using this (`let g:ale_elixir_mix_random_build_dir = 0`) felt about twice as fast. Definitely a big improvement.

ale_linters/elixir/mix.vim
```diff
@@ -32,11 +35,13 @@ endfunction
 function! ale_linters#elixir#mix#GetCommand(buffer) abort
     let l:project_root = ale#handlers#elixir#FindMixProjectRoot(a:buffer)
 
-    let l:temp_dir = ale#command#CreateDirectory(a:buffer)
+    let l:build_dir = ale#Var(a:buffer, 'elixir_mix_random_build_dir')
+    \   ? ale#command#CreateDirectory(a:buffer)
+    \   : ale#Var(a:buffer, 'elixir_mix_build_dir')
```
w0rp: Could you add code to create the other directory if it doesn't exist already?

Add tests for this.

pgr0ss: I think mix will create the directory for you. I haven't had to explicitly create it in my testing.

I didn't see any tests for this plugin. I'm not quite sure where to start. Do you think tests are a blocker for merging this?

w0rp: Yeah, you should add some tests for this. Have a look at some existing tests for examples, and `:help ale-dev` for documentation.

PR #2753, opened 2019-09-04
Map mypy "note" messages to info-level lints

### LINKED ISSUE #2704: mypy notes aren't shown
ALE shows `error`s from mypy, but not `note`s. This firstly makes it hard to debug some issues, since mypy often emits a note immediately after an error with more explanation of the types involved or a link to an explanation of the error (and these notes aren't even visible in `:ALEDetail`), and secondly means that adding `reveal_type` to something doesn't show the type of the expression in the editor – you have to go and run mypy by hand to see the result.

It would be nice if mypy's notes could be shown in the editor as well as its errors.

Hnasar:
ALE shows errors from mypy, but not notes. This firstly makes it hard to
debug some issues, since mypy often emits a note immediately after an
error with more explanation of the types involved or a link to an
explanation of the error (and these notes aren't even visible in
:ALEDetail), and secondly means that adding reveal_type to something
doesn't show the type of the expression in the editor – you have to go
and run mypy by hand to see the result.

We map "note" messages to ALEInfo highlight level groups because that
means when the user runs :lwin they'll see the expanded note if there's
also an error on the line, and if there's no error they'll just see the
ALEInfo highlight group.

Fixes #2704

### DIFF
--- a/ale_linters/python/mypy.vim
+++ b/ale_linters/python/mypy.vim
@@ -51,7 +51,7 @@ function! ale_linters#python#mypy#Handle(buffer, lines) abort
     " Lines like these should be ignored below:
     "
     " file.py:4: note: (Stub files are from https://github.com/python/typeshed)
-    let l:pattern = '\v^([a-zA-Z]?:?[^:]+):(\d+):?(\d+)?: (error|warning): (.+)$'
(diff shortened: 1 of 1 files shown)


w0rp:
I'll add an implementation of this now with an option `g:ale_python_mypy_show_notes`, which can be set to `0` to disable notes. I think some people will want to turn the notes off.

PR #2769, opened 2019-09-14
2737 pipenv deep

### LINKED ISSUE #2737: Empty Pipfile files are created with Pipenv
## Information

**VIM version**

VIM - Vi IMproved 8.1 (2018 May 18, compiled Jul 21 2019 22:54:00) 

Operating System: Mac

## What went wrong

Empty Pipfile files are created alongside Python modules that are three subfolders deep from the projet's Pipfile.

See https://github.com/pypa/pipenv/issues/3908

## Reproducing the bug

1. Put in your config: `let g:ale_python_auto_pipenv = 1`
2. Create three layers, e.g. `a/b/c/foo.py`
3. Edit and save `foo.py`
4. Observe that an empty `a/b/c/Pipfile` is created

### :ALEInfo

Relevant section:

```
(finished - exit code 0) ['/usr/local/bin/zsh', '-c', 'cd ''[redacted]/api/api/task'' && ''pipenv'' run flake8 --format=default --stdin-display-name ''[redacted]/api/api/task/
connection.py'' - < ''/var/folders/ss/2cch6pjs7fqf55xztqstklh40000gn/T/vBtf28V/25/connection.py'''] 
```

### Context

Pipenv will look for a `Pipfile` up to `PIPENV_MAX_DEPTH` (defaults to 3) up. If it does not find one, it will create an empty `Pipfile`. 

A simple fix would be to use `PIPENV_PIPFILE` to provide an explicit path to the `Pipfile`.

richyfish:
Fixes #2737 

When `g:ale_python_auto_pipenv = 1 `

By default `pipenv` only searches for the `Pipfile` up 3 levels, if it does not find the `Pipfile` then it will create a new one when `pipenv run` is invoked.  This may be unexpected behaviour for ALE users and also slows down the linting process.

To facilitate this in ALE I've added `g:ale_python_auto_pipenv_maxdepth` (or `b:ale_python_auto_pipenv_maxdepth`) which when defined will set `PIPENV_MAX_DEPTH` during the call to `pipenv run`.  If the variable is left undefined (default) then existing behaviour is maintained.


w0rp:
If setting an environment variable for pipenv fixes this, then let's just add documentation describing how to set the environment variable instead. No need to set this in every command.

PR #2771, opened 2019-09-16
Use Makefile output with clang-tidy when useful

gpanders:
In the case where neither a build directory nor a compile_commands.json
file is found, use the output of `make -n` to provide options to
clang-tidy.

If a compile_commands.json file is found, then `empty(l:build_dir)` will be false, so the behavior is unchanged. This will only affect instances where there is not a compile_commands.json file present.


gpanders:
> Could you add a test for this? See existing tests for other C linters, and check out `:help ale-dev`.

Done!

w0rp:
Cheers! :beers:

Issue #2761, opened 2019-09-10
govet linter broken with Go 1.13


majewsky:
In Go 1.12.9, govet reports errors like this (excerpt from `ALEInfo`):

```
(executable check - success) go
(finished - exit code 2) ['/bin/zsh', '-c', 'cd ''/x/src/github.com/sapcc/keppel/internal/api/keppel'' &&  go vet  .']
<<<OUTPUT STARTS>>>
# github.com/sapcc/keppel/internal/api/keppel [github.com/sapcc/keppel/internal/api/keppel.test]
./accounts_test.go:45:8: not enough arguments in call to NewAPI
        have (keppel.AuthDriver, keppel.NameClaimDriver, *keppel.DB)
        want (keppel.AuthDriver, keppel.NameClaimDriver, *keppel.DB, keppel.Auditor)
<<<OUTPUT ENDS>>>
```

This works fine with latest ALE (tested with https://github.com/dense-analysis/ale/commit/6e18c03d80c323e740f87103fc05955b5c61b54e). However, in Go 1.13, I don't get a mark for this error in the UI. The output now looks like this:

```
(executable check - success) go
(finished - exit code 2) ['/bin/zsh', '-c', 'cd ''/x/src/github.com/sapcc/keppel/internal/api/keppel'' &&  go vet  .']
<<<OUTPUT STARTS>>>
# github.com/sapcc/keppel/internal/api/keppel
vet: ./accounts_test.go:45:20: too few arguments in call to NewAPI
<<<OUTPUT ENDS>>>
```

I suspect that the issue is that `go vet` started prefixing its error messages with `vet: `.

mrvik:
Same issue here with `go version go1.13 linux/amd64`

rsheasby:
I'm having the exact same issue. Can confirm the issue is the `vet: ` in front of the output. Changing the go binary to a shell script that trims the `vet: ` from output works perfectly. Should be a quick fix, but I can't for the life of me figure out where the output parsing occurs. The source code is a maze

doesnotcommit:
fwiw, `go vet` output parsing occurs here: https://github.com/dense-analysis/ale/blob/61cfb3fefb0ebd8654be452046bd2ba24025311f/autoload/ale/handlers/go.vim#L11

aj3423:
Same problem here,  is there any quick workaround for now ?

majewsky:
You can apply this patch to your local copy of ale:
```patch
[...]

PR #2776, opened 2019-09-19
fix broken govet linter with go 1.13

doesnotcommit:
This is an attempt to address #2761. It's ugly, but it works. Would be happy to know if there's a better way.

### DIFF
--- a/autoload/ale/handlers/go.vim
+++ b/autoload/ale/handlers/go.vim
@@ -8,7 +8,7 @@
 " Description: moved to generic Golang file from govet
 
 function! ale#handlers#go#Handler(buffer, lines) abort
-    let l:pattern = '\v^([a-zA-Z]?:?[^:]+):(\d+):?(\d+)?:? ?(.+)$'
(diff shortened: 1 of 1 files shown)


autoload/ale/handlers/go.vim
```diff
@@ -8,7 +8,7 @@
 " Description: moved to generic Golang file from govet
 
 function! ale#handlers#go#Handler(buffer, lines) abort
-    let l:pattern = '\v^([a-zA-Z]?:?[^:]+):(\d+):?(\d+)?:? ?(.+)$'
+    let l:pattern = '\v^v*e*t*:*\s*([a-zA-Z]?:?[^:]+):(\d+):?(\d+)?:? ?(.+)$'
```
w0rp: `*` repeats the previous character/expression 0 or more times, and there's no need to match on several of these characters. Try `%(vet:)?` for optionally matching on a non-capturing group.

Add tests with the new message format.

johnny88: I was looking into this but had a hard time trying to figure out how to test this. @w0rp do you know of any similar tests that I could look at?

w0rp:
Superseded by #3191.

PR #2779, opened 2019-09-20
Add TypeScript autoimport support for deoplete

jeremija:
This should add support for TypeScript autoimport in deoplete.

This also fixes a bug with the current ALE source for Deoplete since ALE limits the number of completions to `g:ale_completion_max_suggestions` (default `50`). When `window.` is typed, there will be a lot more results than `50`, so most will be filtered out. Because of this, when `window.Syn` is typed `SyntaxError` will not be shown. I've tried increasing the limit to a higher number, but then TypeScript completions become really slow (possibly because of completion entry details requests), so I've set the `is_volatile = True` variable to ALE source.

From deoplete [docs](https://github.com/Shougo/deoplete.nvim/blob/master/doc/deoplete.txt):

```
is_volatile
		(Bool)				(Optional)
		If it is True, the source depends on the user input. It means
		that if this flag is set to False, deoplete will cache
		gather_candidates results and will not call gather_candidates
		on each input change. Only on_post_filter method will be
		called on each input change (if implemented).

		Default: False
```

It would be great if somebody actively using deoplete (and TypeScript) would be able to test this before merging!


w0rp:
Thanks for this! I'll ask around on IRC to see if anyone wants to try this out.

cirias:
You are awesome! Just tested and it worked for me!

thisjeremiah:
I tested it and it worked when I removed `self.is_volatile = True` from the deoplete plugin code. If I kept the line, the autocomplete options wouldn't populate.

These are my deoplete settings: 

```vim
Plug 'Shougo/deoplete.nvim'
let g:deoplete#enable_at_startup = 1
set completeopt-=preview
let g:deoplete#max_abbr_width = -1
let g:deoplete#max_menu_width = -1
let g:deoplete#ignore_sources = {}
let g:deoplete#ignore_sources._ = [
  \ 'tags', 'buffer', 'around', 'local',
  \ 'file', 'member', 'omni', 'link',
  \ 'LanguageClient'
\]
autocmd FileType * call deoplete#custom#option({
  \ 'auto_complete': v:true,
  \ 'auto_complete_delay': 0,
  \ 'smart_case': v:true,
  \ 'min_pattern_length': 1
\ })
autocmd FileType * call deoplete#custom#option('sources', {
\ '_': ['ale']
\})

```

Maybe it one of these settings that is causing the different behavior?

jeremija:
Just tested your config and I am unable to reproduce this. Which version of deoplete are you running?

jeremija:
@thisjeremiah I took another look at this, this time I ran it in a clean neovim environment in Docker:

```vimscript
call plug#begin('~/.config/nvim/plugged')
  Plug 'jeremija/ale', { 'branch': 'feature/autoimport-deoplete' }
  Plug 'Shougo/deoplete.nvim', { 'do': ':UpdateRemotePlugins' }
call plug#end()

augroup FiletypeGroup
    autocmd!
    au BufNewFile,BufRead *.ts set filetype=typescript
augroup END

let g:ale_completion_tsserver_autoimport = 1

let g:deoplete#enable_at_startup = 1
autocmd FileType * call deoplete#custom#option('sources', {
\ '_': ['ale']
\})
```

I added some `echom` statements for logging and it looks like deoplete keeps calling `ale#completion#GetCompletions` in an infinite loop wehn `self.is_volatile` is set to `True` (`l:source` is always set to `deoplete`). See this video for more information:

deoplete

I've already noticed this before and commented about it [here](https://github.com/dense-analysis/ale/pull/2709#issuecomment-532165783):

> On a side note, I noticed that the ale#completion#GetCompletionResult function frequently gets called many, many (too many?) times from ale.py.

jeremija:
@thisjeremiah can you test now?

w0rp:
How recently was the newer callback system added?

jeremija:
It seems that a pretty recent change introduced problems for some people. See:

- https://github.com/Shougo/deoplete.nvim/issues/1006#issuecomment-526797857
-https://github.com/Shougo/deoplete.nvim/issues/1013

The `is_async` has been marked as deprecated 23 days ago according to [git blame](https://github.com/Shougo/deoplete.nvim/blame/master/doc/deoplete.txt#L1230).

My latest commit message has some more details.

jeremija:
Also relevant https://gitter.im/Shougo/deoplete.nvim?at=5d69f10438499c13a67411e0

w0rp:
If it's a very recent change, then we should still support `is_async` for some time, as many people won't bother to update Deoplete for a while.

Maybe we can detect if the newer feature exists or not, and use one thing or the other.

jeremija:
Yeah, it looks like the latest changes in this PR will work only since a very recent commit in Deoplete:

```
commit 782fdaf0671d3acc0d70920ed7ca521cfdd5578e
Date:   Thu Sep 19 21:12:30 2019 +0900
```

thisjeremiah:
Yes, I updated deoplete and tested with your updated branch. It works for me now!

w0rp:
Okay, I think some detection of the Deoplete version is needed, and maybe we can only enable support for autoimports in Deoplete if the Deoplete version is new enough.

jeremija:
I agree, I'm just not sure what would be the best way to check for this, since the old autocomplete functionality is buggy in the current version of deoplete (is_async callbacks sometimes work, and sometimes do not because it calls `gather_candidates` in an endless loop of `is_async` and `is_refresh` events). It does not seem to matter if autoimport is enabled or not, but it has something to do with the `is_volatile` flag.

AFAIK, there is no way to retrieve Deoplete version from vim/python - the only place I found it was in `doc/deoplete.txt`.

If anybody has an idea on how to reliably check for the new feature (or the bug), I'd love to hear it!

w0rp:
I think we can check `exists('deoplete#auto_complete')` in Vim, via the same mechanism for calling a command.

jeremija:
> I think we can check exists('deoplete#auto_complete') in Vim, via the same mechanism for calling a command.

Unfortunately, this function was in deoplete before Shougo/deoplete.nvim@782fdaf0671d3acc0d70920ed7ca521cfdd5578e, which is the most recent commit that my fixes work with :/

w0rp:
Maybe there's something else we can use to detect if Deoplete is new enough.

laino:
> Unfortunately, this function was in deoplete before [Shougo/deoplete.nvim@782fdaf](https://github.com/Shougo/deoplete.nvim/commit/782fdaf0671d3acc0d70920ed7ca521cfdd5578e), which is the most recent commit that my fixes work with :/

Using deoplete at that version together with your branch gives me working auto-import functionality.

w0rp:
At this point, I'll merge this, and if anyone has problems they can update Deoplete. The version which works with this is old enough now.

w0rp:
Cheers! :beers:

PR #2786, opened 2019-09-23
569 support vim sign group and priority

hsanson:
Add sign-group and sign-priority support to ALE. With sign-group ALE can manipulate the signs without risk of conflicts with other plugins that also set signs and with priority users are able to adjust ALE signs priority to override or be overridden by other plugins signs.

### DIFF
--- a/autoload/ale/sign.vim
+++ b/autoload/ale/sign.vim
@@ -147,24 +147,49 @@ function! ale#sign#GetSignName(sublist) abort
     return 'ALEErrorSign'
 endfunction
 
+function! s:GroupCmd() abort
+  if has('nvim-0.4.0') || v:version >= 801
+    return ' group=ale '
+  else
+    return ' '
+  endif
+endfunction
+
 " Read sign data for a buffer to a list of lines.
 function! ale#sign#ReadSigns(buffer) abort
     redir => l:output
-        silent execute 'sign place buffer=' . a:buffer
+        silent execute 'sign place ' . s:GroupCmd()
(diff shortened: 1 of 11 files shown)


hsanson:
@w0rp this pull request seems to break the tests somehow and my understanding on how these run is not enough to find the issue. 

If I run the tests (e.g. ./run-tests) with the master branch all works ok but when running the tests with the changes of this PR, the tests hang forever after completing the vader tests:

```
.....
  Success/Total: 2/2                                                                                                   
Success/Total: 2180/2180 (assertions: 3090/3090)                                                                       
Elapsed time: 57.295863 sec.   <--- hangs here forever --->
```

autoload/ale/sign.vim
```diff
@@ -361,6 +390,7 @@ function! ale#sign#GetSignCommands(buffer, was_sign_set, sign_map) abort
     if l:is_dummy_sign_set && !g:ale_sign_column_always
         call add(l:command_list, 'sign unplace '
         \   . g:ale_sign_offset
+        \   . s:GroupCmd()
```
w0rp: The use of `priority` in the unplace command here might be what's breaking things. Same as above.

hsanson: Thanks, the problem was that sign-group and sign-priority are supported on vim 8.1.0614 onward but the tests use 8.1.0519 causing all sign commands to fail. I changed the version test to include the patch version and everything seems to be working.

Also had to modify all tests to also consider the sign group.

w0rp:
Cheers! :beers:

PR #2791, opened 2019-09-26
Add definition of c/clangd's language as C

### LINKED ISSUE #2785: c/clangd.vim not defining the language as c
## Information

**VIM version**

NVIM v0.4.2
Build type: Release

Operating System: Arch Linux

## What went wrong

clangd for `.c` files isn't defining the language as `C` so it gives `C++` warnings/errors.
https://github.com/dense-analysis/ale/blob/master/ale_linters/c/clangd.vim#L8
clang for `.c` files does define it, it's just adding the option `-x c`
https://github.com/dense-analysis/ale/blob/master/ale_linters/c/clang.vim#L12

I changed line 8 in clangd.vim to:
`return '%e -x c' . ale#Pad(ale#Var(a:buffer, 'c_clangd_options'))` and I stopped getting the `C++` warnings/errors.

RageKnify:
Fixes #2785

### DIFF
--- a/ale_linters/c/clangd.vim
+++ b/ale_linters/c/clangd.vim
@@ -5,7 +5,7 @@ call ale#Set('c_clangd_executable', 'clangd')
 call ale#Set('c_clangd_options', '')
 
 function! ale_linters#c#clangd#GetCommand(buffer) abort
-    return '%e' . ale#Pad(ale#Var(a:buffer, 'c_clangd_options'))
+    return '%e -x c' . ale#Pad(ale#Var(a:buffer, 'c_clangd_options'))
 endfunction
 
 call ale#linter#Define('c', {
--- a/test/command_callback/test_c_clangd_command_callbacks.vader
+++ b/test/command_callback/test_c_clangd_command_callbacks.vader
@@ -4,7 +4,7 @@ Before:
   Save &filetype
   let &filetype = 'c'
 
-  let b:command_tail = " -x c"
+  let b:command_tail = ' -x c'
 
 After:
   unlet! b:command_tail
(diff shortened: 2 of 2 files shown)


RageKnify:
The [clangd test](https://github.com/dense-analysis/ale/blob/master/test/command_callback/test_c_clangd_command_callbacks.vader) needs to be updated, I'm looking at the [clang test](https://github.com/dense-analysis/ale/blob/master/test/command_callback/test_c_clang_command_callbacks.vader) as a guide.

w0rp:
Cheers! :beers:

khalsah:
@w0rp I believe this change should be reverted as it appears to have broken clangd support on c files. I expect the disappearance of C++ warnings mentioned in #2785 is due to clangd no long running, rather than it correcting the file type.

Running against the latest master of ale, and clangd 9, with only the clangd linter enabled, I observed that clangd was not reporting any errors, warnings, or completions in my project. In attempting to track down I cause I noticed this change and when I removed the `-x c` flags from the clangd linter definition functionality was restored.

I tested clangd versions 7, 8, and 9 to verify this was not simply a change in behavior across versions. In each case, when run from the command line with `clangd -x c` the command exited with errors of the form:

```
clangd: Unknown command line argument '-x'.  Try: 'clangd --help'
clangd: Did you mean '  -h'?
clangd: Unknown command line argument 'c'.  Try: 'clangd --help'
```

RageKnify:
I assumed `clangd` had the `-x` option like `clang`, since it doesn't it seems like it's intended to be used solely for `c++`.

khalsah:
@RageKnify I'm new to clangd myself, but my understanding is that clangd expects either a `compile_commands.json` or a `compile_flags.txt` file in the project to configure those details: https://clang.llvm.org/extra/clangd/Installation.html#project-setup (This also allows it to correctly handle projects with a mix of settings across files).

w0rp:
Okay, I'll revert that change now.

w0rp:
I have reverted that now.

PR #2793, opened 2019-09-27
Update to use config file for clj-kondo if exists

liquidz:
I added support to use clj-kondo config file.


ale_linters/clojure/clj_kondo.vim
```diff
@@ -1,6 +1,15 @@
 " Author: Masashi Iizuka <liquidz.uo@gmail.com>
 " Description: linter for clojure using clj-kondo https://github.com/borkdude/clj-kondo
 
+function! ale_linters#clojure#clj_kondo#GetCommand(buffer) abort
+    let l:conf_dir = ale#path#FindNearestDirectory(a:buffer, '.clj-kondoo')
+    let l:conf_option = empty(l:conf_dir)
+          \ ? ''
```
RageKnify: The `\` should come before the 4 spaces, like in lines 28-34.

liquidz: @RageKnify Thanks! I fixed it.

test/command_callback/test_clj_kondo_command_callback.vader
```diff
@@ -0,0 +1,21 @@
...
+  unlet! g:config_path
+  call ale#assert#TearDownLinterTest()
+
+Execute(The default command should be correct):
+  AssertLinter 'clj-kondo',
+  \ 'clj-kondo  --lint %t'
+
+Execute(The configuration file should be automatically detected):
+  let g:config_path = ale#path#Simplify(
+  \ g:dir . /clj_kondo_paths/.clj-kondo
```
RageKnify: Is this missing the `'`s from before the first `/` up to the end?

liquidz:
@w0rp Thanks for your navigation!
I added a test.

PR #2803, opened 2019-10-02
Update highest phpstan level

diegosouza:
[According to the phpstan doc](https://github.com/phpstan/phpstan#rule-levels), currently the highest/strictest level is 7. I've updated the documentation + the vader tests.


w0rp:
Cheers! :beers:

PR #2804, opened 2019-10-02
Adds hdl_checker LSP support

suoto:
Adds support for https://github.com/suoto/hdl_checker. This tool is not a linter per se, but a wrapper around some tools out there to make a dev life's easier, loosely similar to parsing a C Makefile output to auto detect gcc flags and include paths.

The rationale is VHDL/Verilog/SystemVerilog projects are a pain to setup, this tool tries to automate this with educated guesses. There's some overlap with VHDL/Verilog/SystemVerilog tools supported by ALE but, as of now, using any of verilog/vlog.vim, vhdl/ghdl.vim, vhdl/vcom.vim or vhdl/xvhdl.vim will require quite a lot of workarounds, which imho are too time consuming for anything with more than a couple of files.

Documentation is very much WIP at the moment.


ale_linters/vhdl/hdl_checker.vim
```diff
@@ -0,0 +1,70 @@
...
+    return l:command
+endfunction
+
+" To allow testing
+function! ale_linters#vhdl#hdl_checker#GetInitOptions(buffer) abort
+    return {'project_file': ale#Var(a:buffer, 'hdl_checker_config_file')}
+endfunction
+
+" Setup is identical for VHDL and Verilog/SystemVerilog
+for s:language in ['vhdl', 'verilog']
```
w0rp: Instead of defining both in the same `vhdl` directory, move the code for this to an `autoload/handler` file, and wrap this call to `Define` inside a function that accepts the language as an argument.

suoto: Cool, should've looked at how other linters tackled this!

Done 👍

autoload/ale/handlers/hdl_checker.vim
```diff
@@ -0,0 +1,70 @@
...
+
+call ale#Set('hdl_checker_executable', 'hdl_checker')
+call ale#Set('hdl_checker_config_file', has('unix') ? '.hdl_checker.config' : '_hdl_checker.config')
+call ale#Set('hdl_checker_options', '')
+
+" Use this as a function so we can mock it on testing. Need to do this because
+" test files are inside /testplugin (which refers to the ale repo), which will
+" always have a .git folder
+function! ale#handlers#hdl_checker#IsDotGit(path) abort
+    return isdirectory(a:path) && ! empty(a:path)
```
w0rp: Check `!empty()` first, so we don't need to check if an empty string is a directory first. That will probably be faster.

autoload/ale/handlers/hdl_checker.vim
```diff
@@ -0,0 +1,70 @@
...
+    if ale#handlers#hdl_checker#IsDotGit(l:project_root)
+        return fnamemodify(l:project_root, ':h:h')
+    endif
+
+    " As a fallback, use the path of the current buffer
+    return expand('#' . a:buffer . ':p:h')
+endfunction
+
+function! ale#handlers#hdl_checker#GetCommand(buffer) abort
+    let l:command = ale#Var(a:buffer, 'hdl_checker_executable') . ' --lsp'
```
w0rp: I just noticed, you should apply `ale#Escape()` to the executable path here, in case the path contains spaces, etc.

hughesjs:
Hi, what's the status of this? This looks like a valuable integration to have.

w0rp:
I'll merge this now. This is just one we forgot about.

pidgeon777:
Hello, any update on this merge?

w0rp:
Maybe you're commenting on the wrong pull request. I merged this one a while ago.

Issue #2806, opened 2019-10-03
ale-cpp-clangd compile commands


murfalo:
Currently, as far as I can tell, when using the clangd linter for C++, the compile_commands.json file is not passed to clangd's -compile-commands-dir flag.  It would be really helpful to add this feature in a way similar to how the clang-tidy linter is currently automatically using the -p flag.

My current workaround for this is just symlinking compile_commands.json to the root of the source tree which is not really ideal.

PR #2807, opened 2019-10-03
Added automatic compile-commands-dir arg to ale_cpp_clangd linter

murfalo:
Really quick fix for the issue I created just a minute ago (#2806)


murfalo:
I updated the ale-c-clangd linter to match the ale-cpp-clangd linter and added a test case for automatic detection of compile commands.  My last commit a1fab59 has a fix for certain variables leaking out of the test and causing mayhem.  I'm sure this fix is not perfect, but I'm unfortunately a total noob when it comes to VimScript so I'm not immediately sure how to do it better.

w0rp:
Cheers! :beers:

PR #2819, opened 2019-10-07
Add the possiblity to add extra psalm options

greg0ire:

Maybe not at first but you can't prove it :stuck_out_tongue:


w0rp:
Cheers! :beers:

Sorry about the delay. I'll get to everything when I'm able.

w0rp:
I just noticed that all of the options for `psalm` aren't properly named, so I'm going to change them all to `ale_php_psalm_...`. Update your settings if you use those options.

PR #2823, opened 2019-10-10
Do not save after ALERename and ALEOrganizeImports

### LINKED ISSUE #2811: Using `ALEOrganizeImports` and `g:ale_completion_tsserver_autoimport` writes the buffer unexpectedly
Using `ALEOrganizeImports` or `g:ale_completion_tsserver_autoimport` causes the buffer to automatically be written.  Since they only modify the current buffer, it would be preferable if it did not automatically write the file in the process, in case the user decides to just `:q` and doesn't realize that the file has actually changed.

@jeremija

jeremija:
Closes #2811

### DIFF
--- a/autoload/ale/code_action.vim
+++ b/autoload/ale/code_action.vim
@@ -112,7 +112,7 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
     if a:should_save
         call ale#util#Writefile(l:buffer, l:lines, a:filename)
     else
-        call ale#fix#SetBufferContents(l:buffer, l:lines)
+        call ale#util#SetBufferContents(l:buffer, l:lines)
     endif
 
     if l:is_current_buffer
--- a/autoload/ale/code_action.vim
+++ b/autoload/ale/code_action.vim
@@ -1,7 +1,7 @@
 " Author: Jerko Steiner <jerko.steiner@gmail.com>
 " Description: Code action support for LSP / tsserver
 
-function! ale#code_action#HandleCodeAction(code_action) abort
+function! ale#code_action#HandleCodeAction(code_action, should_save) abort
     let l:current_buffer = bufnr('')
     let l:changes = a:code_action.changes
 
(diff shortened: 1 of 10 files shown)


autoload/ale/fix.vim
```diff
@@ -4,40 +4,15 @@ call ale#Set('fix_on_save_ignore', {})
 " Vim doesn't let you modify hidden buffers.
 function! ale#fix#ApplyQueuedFixes(buffer) abort
     let l:data = get(g:ale_fix_buffer_data, a:buffer, {'done': 0})
-    let l:has_bufline_api = exists('*deletebufline') && exists('*setbufline')
 
-    if !l:data.done || (!l:has_bufline_api && a:buffer isnot bufnr(''))
+    if !l:data.done
```
w0rp: This change will break how fixing files works, because we need to avoid the `remove()` call below to keep the data until the buffer comes back into focus again, because older versions of Vim without the `bufline` API don't let you modify buffers other than the current buffer. You'll need to keep the check here still.

autoload/ale/code_action.vim
```diff
@@ -106,10 +109,17 @@ function! ale#code_action#ApplyChanges(filename, changes) abort
         call remove(l:lines, -1)
     endif
 
-    call ale#util#Writefile(l:buffer, l:lines, a:filename)
+    if a:should_save
+        call ale#util#Writefile(l:buffer, l:lines, a:filename)
+    else
+        call ale#fix#SetBufferContents(l:buffer, l:lines)
```
w0rp: Move the function to the `util.vim` file so the `code_action.vim` code doesn't load the `fix.vim` code, if `:ALEFix` is never used.

autoload/ale/fix.vim
jeremija: The check is still there, i just moved it to the `ale#fix#SetBufferContents`.

jeremija: I thought that the `remove()` call was somewhere else. I've added that check back!

chaucerbao:
Just wanted to check in on this PR, and whether it was ready to merge.
@jeremija @w0rp

jeremija:
I believe I've addressed all required changes. I've been using this branch since and it's working well for me, but I'm happy to address any issues if there are any.

w0rp:
I have merged this now. Sorry for the delay!

jeremija:
Thanks 👍

PR #2829, opened 2019-10-13
Fix TCP server config example.

rcorre:
The docs for the `address` parameter of `Define` say:

> This argument must only be set if the `lsp` argument
> is set to `'socket'`.


w0rp:
Cheers! :beers:

PR #2838, opened 2019-10-16
Switch from style to transformers

tvatter:
The styler package writers recommend using the `transformers` argument rather than the `style` argument for more flexibility: https://github.com/r-lib/styler/pull/500.

This was also addressed in this issue: https://github.com/r-lib/styler/issues/340.

It does not change the behavior for most users, but one can now do something like use `ale_r_styler_options` to tweak specific elements of the fixer (e.g., do `tidyverse_style(strict = FALSE)` to change from the default `strict = TRUE`) rather than having to change the whole style (e.g. change from `tidyverse_style` to something completely different). With the current code, only the later was allowed, and the former wasn't possible. With the PR, both are allowed.

PS: Sorry for all the commits. My fork wasn't clean. It would be great if you could squash everything when merging.


tvatter:
Any news on that? That's a straightforward modification for a fixer that already exist...

w0rp:
Cheers! :beers:

Sorry for the delay. I catch up on emails now and then.

tvatter:
No problem, thanks a lot!

PR #2844, opened 2019-10-18
Allow code actions to work on callback based sources

DonnieWest:
This enables code actions on callback based sources such as `asyncomplete` so everyone can participate with the fun :)


w0rp:
Cheers! :beers:

PR #2845, opened 2019-10-18
Fix tsserver not returning details for items with empty source

DonnieWest:
Prior to #2709 `tsserver` based completions would include details on more items, whereas now there are fewer items with details.

This is caused by including an empty string as the source when communicating with the `tsserver`, which confuses it and forces it to return less information about completed items.

This resolves the issue, and fixes the requisite tests


w0rp:
Cheers! :beers:

PR #2846, opened 2019-10-18
Exclude undetailed items

DonnieWest:
Allows the user to exclude items without details when completing with the `tsserver`. This can drastically clean up completions in situations where `tsserver` just returns junk data

Relies on #2845

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -16,6 +16,7 @@ let g:ale_completion_delay = get(g:, 'ale_completion_delay', 100)
 let g:ale_completion_excluded_words = get(g:, 'ale_completion_excluded_words', [])
 let g:ale_completion_max_suggestions = get(g:, 'ale_completion_max_suggestions', 50)
 let g:ale_completion_tsserver_autoimport = get(g:, 'ale_completion_tsserver_autoimport', 0)
+let g:ale_completion_tsserver_remove_items_without_detail = get(g:, 'ale_completion_tsserver_remove_items_without_detail', 0)
 
 let s:timer_id = -1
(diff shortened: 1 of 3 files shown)


w0rp:
I think this is a bad idea, as this will exclude any variables that don't have documentation, for example.

DonnieWest:
@w0rp Is there a different version of this PR that you would accept? I notice that anything `tsserver` is unsure about is returned with a type of `warning`, could I filter those out instead?

w0rp:
Can you provide an example? Why would completion results include warnings?

DonnieWest:
Sure! If you create a file called `test.js` with the following contents:

```
const foo = {};
foo.
// ^ trigger completions here
```

You'll see a completion of `foo` as a property of the `foo` variable. Like so:

Screen Shot 2019-10-18 at 9 48 48 AM

If you log what that comes in as from `tsserver` here as `l:suggestion`: https://github.com/dense-analysis/ale/blob/master/autoload/ale/completion.vim#L300

You get `{'name': 'foo', 'kindModifiers': '', 'sortText': '4', 'kind': 'warning'}`

I'd bet I could filter on that somehow and remove these junk completions from `tsserver`

DonnieWest:
Also of note: these code changes listed here resolves it entirely for me. Undetailed completion items tend to be invalid on my projects, but I can totally see some edge case where that's not the case

w0rp:
Yes, that looks like the thing to do. I found the issue being similarly fixed in this Emac plugin here. https://github.com/ananthakumaran/tide/pull/257

DonnieWest:
@w0rp should I put it behind a variable or should we strip them by default?

PR #2847, opened 2019-10-18
Allow user to customize items

DonnieWest:
Continues from #2710

Allows the user to set their own completion fields

Relies on #2845


DonnieWest:
@w0rp looks like it's failing on `vint` although the syntax is correct. Any advice for making it pass?

w0rp:
The error message printed says there's no `scriptencoding`, so add one. It's complaining because there a Unicode characters in the file without declaring the script as using UTF-8.

test/completion/test_tsserver_completion_parsing.vader
```diff
@@ -36,23 +36,23 @@ Execute(TypeScript completion details responses should be parsed correctly):
   \     'word': 'abc',
   \     'menu': '(property) Foo.abc: number',
   \     'info': '',
-  \     'kind': 'f',
+  \     'kind': 'v',
```
w0rp: These tests should still pass, as we should keep the existing defaults.

test/completion/test_lsp_completion_parsing.vader
```diff
@@ -1,7 +1,7 @@
 After:
   unlet! b:ale_completion_info
 
-Execute(Should handle Rust completion results correctly):
+Execute(uhould handle Rust completion results correctly):
```
w0rp: typo

test/completion/test_tsserver_completion_parsing.vader
DonnieWest: Do we want to maintain strict backward compatibility? 

Seems to me we'd want to make these roughly match the kinds that vim has built in:

```
	v	variable
	f	function or method
	m	member of a struct or class
	t	typedef
	d	#define or macro
```

In particular, since this is a `property` it would probably map best to `m` (not `v` as it is right now, sorry). Same as the Rust types above. They probably map best to `t`, right?

I'm happy to maintain compat though or try this new logical mapping

test/completion/test_lsp_completion_parsing.vader
DonnieWest: Good catch

DonnieWest:
Sorry for the noise. Still having trouble getting `vint` to pass on both this PR and #2849

I'm not sure what I'm missing here :/

test/completion/test_tsserver_completion_parsing.vader
DonnieWest: @w0rp sorry to bug you, but still waiting for a definitive response here :)

DonnieWest:
@w0rp got the linter passing :)

Once I get feedback on the completion items, I should be good to go

DonnieWest:
AppVeyor issue appears to be caused by `html-beautifier` as noted in the other PR

DonnieWest:
@w0rp looks like the `html-beautifier` fix also fixed things over here. Just waiting for feedback on whether we want to actually retain backwards compat or if we're fine changing some types returned by default during completion.

Thanks :)

w0rp:
Cheers! :beers:

DonnieWest:
Thanks!

PR #2849, opened 2019-10-18
Allow the user to remove warnings from completions

DonnieWest:
Continues from #2846 

Switches to a solution that filters out suggestions that are warnings, and only if a global variable is set

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -298,7 +298,9 @@ function! ale#completion#ParseTSServerCompletions(response) abort
     let l:names = []
 
     for l:suggestion in a:response.body
-        if g:ale_completion_tsserver_remove_warnings == 0 || l:suggestion.kind isnot# 'warning'
+        let l:kind = get(l:suggestion, 'kind', '')
+
+        if g:ale_completion_tsserver_remove_warnings == 0 || l:kind isnot# 'warning'
             call add(l:names, {
             \ 'word': l:suggestion.name,
             \ 'source': get(l:suggestion, 'source', ''),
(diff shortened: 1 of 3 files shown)


DonnieWest:
@w0rp Finally figured out the linter error on this project. The AppVeyor build failure appears to be related to `html-beautify` and not part of this project

```
[00:01:26]     (1/1) [EXECUTE] (X) {'command': 'html-beautify  -'} should be equal to {'command': '''html-beautify''  -'}
```

DonnieWest:
@w0rp Thanks for merging that fix to the HTML beautifier! Rebased off of master and it looks like all tests pass now :)

DonnieWest:
Is there anything I can do to help get this merged in? :)

w0rp:
Do you have any real world examples of what these warnings are? I wonder if we ever need them. It sounds weird to have a warning in completion results.

DonnieWest:
@w0rp we've already had this discussion here: https://github.com/dense-analysis/ale/pull/2846 and I offered an example there.

Of particular note, you found other in-the-wild tools that do the same thing to reduce noise in the completion options

w0rp:
Oh yeah, thanks for pointing that out. I forgot about that, as that was ages ago now.

I'm just going to merge this now, and we can think about making this the default if anyone wants to later. I've never encountered these completion results when working on TypeScript code yet, so I'm not sure how you get them to appear, but it's good to have an option to get rid of them. It looks like something I'd never want.

w0rp:
Cheers! :beers:

w0rp:
I know why I've never seen this. I think it's because you have to be using `tsserver` with JavaScript code, but nearly 90% of my front end codebase now is TypeScript.

PR #2851, opened 2019-10-19
Updated Luacheck to be project aware

Zulsorai:
Updated luacheck to be able to find and lint a directory containing _.luacheckrc_.
Luacheck is also able to parse the range of an error using the `--ranges` option.
When the option `lua_luacheck_lint_project` is disabled ale is able to return to the previous behavior of linting the buffer rather than the file on disk.

Tests and docs have been updated to reflect changes.

### DIFF
--- a/ale_linters/lua/luacheck.vim
+++ b/ale_linters/lua/luacheck.vim
@@ -3,10 +3,22 @@
 
 call ale#Set('lua_luacheck_executable', 'luacheck')
 call ale#Set('lua_luacheck_options', '')
+call ale#Set('lua_luacheck_lint_project', 1)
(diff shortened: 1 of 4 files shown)


RyanSquared:
```
ale_linters/lua/luacheck.vim:15 Trailing whitespace
```

Tests reported there's a trailing whitespace on the file. I'll be able to test this PR out soon, but can you get that fixed as well?

Zulsorai:
That should be everything fixed.

lithammer:
It would be nice to revive this. I'm running into this a lot that I have to `cd` into the correct directory first or else `luacheck` will complain.

lithammer:
CC @Zulsorai @w0rp

PR #2858, opened 2019-10-21
Provide configuration option to run flake8 from project root.

kalekseev:
Option `per-file-ignores` was introduced in flake8 version 3.7.0.
It allows to ignore specific errors in specific files using glob syntax.
For example `per-file-ignores = src/generated/*.py:F401` will
ignore `F401` error in all python files in `src/generated`.
Thus ale have to run flake8 from project root where .flake8 config
is placed otherwise glob won't match linted file.

Not sure how that should be handled in ale style but I'm ready to follow on this pr.

### DIFF
--- a/ale_linters/python/flake8.vim
+++ b/ale_linters/python/flake8.vim
@@ -4,7 +4,7 @@
 call ale#Set('python_flake8_executable', 'flake8')
 call ale#Set('python_flake8_options', '')
 call ale#Set('python_flake8_use_global', get(g:, 'ale_use_global_executables', 0))
-call ale#Set('python_flake8_change_directory', 1)
+call ale#Set('python_flake8_change_directory', 'project')
 call ale#Set('python_flake8_auto_pipenv', 0)
 
 function! s:UsingModule(buffer) abort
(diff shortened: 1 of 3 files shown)


kalekseev:
@w0rp please take a look, that feature can be useful

ale_linters/python/flake8.vim
```diff
@@ -38,10 +38,25 @@ function! ale_linters#python#flake8#RunWithVersionCheck(buffer) abort
     \)
 endfunction
 
+function! ale_linters#python#flake8#GetCdString(buffer) abort
+    let l:change_directory = ale#Var(a:buffer, 'python_flake8_change_directory')
+
+    if l:change_directory == 2
```
w0rp, line 44: We can do better than `2`. Let's add strings for the existing options and new option, and also continue to support `1` and `0`. Make sure to update the documentation, including the documented type.

Let's make this new setting the new default, and make sure we fall back on setting the CWD to the directory the file is in, in the Vader tests.

w0rp (changes requested):

kalekseev:
Hi @w0rp, I updated the pr with changes you requested.

> and make sure we fall back on setting the CWD to the directory the file is in, in the Vader tests.

that implicitly tested by unrelated tests, because project root is default now and by default tests run on a file that doesn't have project mark in the parent dirs. If you prefer to have an explicit test for that case let me know.

w0rp (approved):

w0rp:
Cheers! :beers:

rafikdraoui:
Thanks! I've been using something similar to what's in this PR for a while now. I'm happy to be able to simplify my vimrc and use the default ale linter again!

The fact that `per-file-ignores` doesn't handle paths properly is a bug that is being tracked by `flake8`: https://gitlab.com/pycqa/flake8/-/issues/517.

PR #2863, opened 2019-10-25
Implement commandline options for cfn_python_lint

benjyiw:
Implemented the cloudformation_cfnlint_options option in order to allow setting commandline options for cfn-lint. Added documentation for the option. Used terminology that excluded "_python_" so that it would match naming convention standard of `<language>_<lintername>_options` without additional underscores.

🍻


0xMH:
@hsanson Hay man sorry for bothering. But is it possible for this to get through and get merged?

Issue #2824, opened 2019-10-11
Vale allow to pass additional options


jghauser:
It would be useful if there was an option like `g:ale_vale_options` that allows you to pass extra options to vale. Basically the same as exists already for languagetool and write-good.

PR #2865, opened 2019-10-26
Add vale options

riley-martine:
Closes https://github.com/dense-analysis/ale/issues/2824

Are there additional tests that should be added for this? If so, what should be tested, and are there examples of other similar tests you can point me at?


hsanson:
@riley-martine you can create a command callback test for the linter. There are lots of examples inside the `test/command_callback` folder. The `test_ibm_openapi_validator_command_callback.vader` is a simple example.

PR #2869, opened 2019-10-30
Fix clang-tidy for out-of-tree builds

clktmr:
clang-tidy prints file paths relative to the build directory if
compile-commands.json is used via the '-p' option.  This causes ALE to
fail resolving the filename if the build directory is any other than
pwd.

This patch prepends the build directory to the filenames to get the
absolute path.


clktmr:
I don't think the failed build is related to my changes. Can you rerun it?

PR #2870, opened 2019-10-30
Clean up excess whitespace in deoplete completion menu

tomtomjhj:
This patch simply removes the extra whitespace in the completion menu when passing the completion data to deoplete. I didn't add any tests because it doesn't change the behavior significantly. But I'll try to add one if you think it's needed.


w0rp:
I merged #2779 now, so you'll have to update this. Add a unit test for this in the Python unit test file if you can.

Sorry for the long delay.

PR #2871, opened 2019-11-01
Fix regural expression for cppcheck. issues 2507

nigredon1991:
In new version cppcheck change error format by default.
Change Regexp for parsing this and change tests.

New version cppcheck(1.89)
```
$ cppcheck --template=gcc --enable=style ../learn_c/test.c 
Checking ../learn_c/test.c ...
../learn_c/test.c:6:3: warning: %d in format string (no. 1) requires 'int' but the argument type is 'const char *'. [invalidPrintfArgType_sint]
  printf("%d", "More then 2 input value");
  ^
```
Old version(1.69):
```
$ ./cppcheck --template=gcc --enable=style ../learn_c/test.c 
Checking ../learn_c/test.c...
../learn_c/test.c:6: warning: %d in format string (no. 1) requires 'int' but the argument type is 'const char *'.
```
Now it works for both.

### DIFF
--- a/autoload/ale/handlers/cppcheck.vim
+++ b/autoload/ale/handlers/cppcheck.vim
@@ -43,9 +43,11 @@ endfunction
 
 function! ale#handlers#cppcheck#HandleCppCheckFormat(buffer, lines) abort
     " Look for lines like the following.
-    "
-    " [test.cpp:5]: (error) Array 'a[10]' accessed at index 10, which is out of bounds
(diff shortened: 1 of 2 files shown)


nigredon1991:
Ok, I will do it.

nigredon1991:
Done.

nigredon1991:
My work is done better by others. Close the request.
https://github.com/dense-analysis/ale/pull/3030

PR #2872, opened 2019-11-01
Properly extract all error messages from `nix-instantiate --parse`

Ma27:
This command is actually able to detect invalid variable references,
however those aren't displayed in my Vim. The problem here is that the
regex expects a comma between the error message and the error's
location. However this isn't the case for invalid references:

    error: undefined variable 'bar' at /home/ma27/Projects/nixos-config/test.nix:1:11

Right now the `,` at the end of the string is directly stripped as
optional character groups `(,?)` would've been ignored by the regex
engine as it was interpreted as part of the second match group `(.+)`.


Ma27:
@hsanson @w0rp any chance to take a look at this? :)

PR #2887, opened 2019-11-07
Add nimcheck end_col options

tamago324:
Thank you for the nice plugin!

`end_col` is not used in highlight of error part in `nimcheck`, so `end_col` is used.
I thought it would be easier to highlight with `end_col`.

I am the first to send a pull request for the source code, so I may need help.


tamago324:
Thank you for merging !

PR #2889, opened 2019-11-09
Default errorview in pwsh7 now concise

zigford:
In the upcoming powershell7, the default error view is now 'concise' which isn't as easy to parse. I'm working on another change to do most of the parsing in powershell, but this is an effective fix that works across all powershell versions until that's ready.


w0rp:
Cheers! :beers:

PR #2890, opened 2019-11-09
Add nimpretty fixer for nim-lang

nhanb:
Worked on my machine (tm). I did write some tests but do tell me if they're not enough.


w0rp:
Cheers! :beers:

PR #2894, opened 2019-11-10
Use stdin instead of lint_file for nimcheck

tamago324:
This pull request gives you two options:
* `g:ale_lint_on_text_changed`
* `g:ale_lint_on_insert_leave`

I often use `g: ale_lint_on_insert_leave = 1`.
But nimcheck couldn't


w0rp:
Are there any disadvantages to checking code via stdin, such as checking for type errors or anything with import lines, or does this just work?

tamago324:
Cancel this pull request because the result was different when using A and not 😭

RyanSquared:
That's strange. What was the difference? Could it possibly be a bug in nimcheck?

tamago324:
I'll check if it's a bug in nimcheck

tamago324:
Sorry for being very late.

I have asked [nimforum](https://forum.nim-lang.org/t/5534) a question.
As a result, it **should not be read from stdin**.
So this pull-request is closed.

Araq is also the author of nim.

w0rp:
Okay, thanks for looking into this.

Issue #2891, opened 2019-11-09
ALE does not show errors for cppcheck


themightyoarfish:
## Information

**VIM version**
```
VIM - Vi IMproved 8.1 (2018 May 18, compiled May  2 2019 12:07:21)
macOS version
Included patches: 1-1189
Compiled by rasmus@Falcon-9.local
Huge version without GUI.  Features included (+) or not (-):
+acl               +extra_search      +mouse_netterm     -tag_old_static
+arabic            -farsi             +mouse_sgr         -tag_any_white
+autocmd           +file_in_path      -mouse_sysmouse    -tcl
+autochdir         +find_in_path      +mouse_urxvt       +termguicolors
-autoservername    +float             +mouse_xterm       +terminal
-balloon_eval      +folding           +multi_byte        +terminfo
+balloon_eval_term -footer            +multi_lang        +termresponse
-browse            +fork()            -mzscheme          +textobjects
++builtin_terms    -gettext           +netbeans_intg     +textprop
+byte_offset       -hangul_input      +num64             +timers
+channel           +iconv             +packages          +title
+cindent           +insert_expand     +path_extra        -toolbar
-clientserver      +job               -perl              +user_commands
+clipboard         +jumplist          +persistent_undo   +vartabs
+cmdline_compl     +keymap            +postscript        +vertsplit
+cmdline_hist      +lambda            +printer           +virtualedit
+cmdline_info      +langmap           +profile           +visual
+comments          +libcall           -python            +visualextra
+conceal           +linebreak         +python3           +viminfo
+cryptv            +lispindent        +quickfix          +vreplace
-cscope            +listcmds          +reltime           +wildignore
+cursorbind        +localmap          +rightleft         +wildmenu
+cursorshape       -lua               -ruby              +windows
+dialog_con        +menu              +scrollbind        +writebackup
+diff              +mksession         +signs             -X11
+digraphs          +modify_fname      +smartindent       -xfontset
-dnd               +mouse             +startuptime       -xim
-ebcdic            -mouseshape        +statusline        -xpm
+emacs_tags        +mouse_dec         -sun_workshop      -xsmp
+eval              -mouse_gpm         +syntax            -xterm_clipboard
+ex_extra          -mouse_jsbterm     +tag_binary        -xterm_save
   system vimrc file: "$VIM/vimrc"
     user vimrc file: "$HOME/.vimrc"
 2nd user vimrc file: "~/.vim/vimrc"
[...]

PR #2909, opened 2019-11-25
Add option to prevent node from outputting deprecation warnings

velrest:
Since the output of eslint should not contain anything more than linting errors, we should be able to pass node custom options to f.e. disable deprecation warnings.

Solution for https://github.com/dense-analysis/ale/issues/2891

### DIFF
--- a/autoload/ale/handlers/eslint.vim
+++ b/autoload/ale/handlers/eslint.vim
@@ -85,7 +85,7 @@ endfunction
 
 function! s:parseJSON(buffer, lines) abort
     try
-        let l:parsed = json_decode(a:lines[-1])
+        let l:parsed = json_decode(a:lines[0])
     catch
         return []
     endtry
--- a/autoload/ale/node.vim
+++ b/autoload/ale/node.vim
@@ -30,13 +30,18 @@ endfunction
 " function, the executable string will be escaped when returned by this
 " function.
 "
+" Node parameters can be passed as optional function params.
+"
 " The executable is only prefixed for Windows machines
-function! ale#node#Executable(buffer, executable) abort
+function! ale#node#Executable(buffer, executable, ...) abort
+    "Dont put deprecation warnings into the output
+    let l:node_options = join(get(a:, '000', []) + ['--no-deprecation'], ' ')
+
     if has('win32') && a:executable =~? '\.js$'
         let l:node = ale#Var(a:buffer, 'windows_node_executable_path')
-
-        return ale#Escape(l:node) . ' ' . ale#Escape(a:executable)
+    else
+        let l:node = 'node'
     endif
 
-    return ale#Escape(a:executable)
+    return ale#Escape(l:node) . ' ' . ale#Escape(l:node_options) . ' ' . ale#Escape(a:executable)
 endfunction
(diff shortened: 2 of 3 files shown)


elebow:
I think we always want to prevent warning output from nodejs, for all linters. We could probably do something like this:

```diff
diff --git a/autoload/ale/node.vim b/autoload/ale/node.vim
index 69060122..929be499 100644
--- a/autoload/ale/node.vim
+++ b/autoload/ale/node.vim
@@ -35,7 +35,9 @@ function! ale#node#Executable(buffer, executable) abort
     if has('win32') && a:executable =~? '\.js$'
         let l:node = ale#Var(a:buffer, 'windows_node_executable_path')
 
-        return ale#Escape(l:node) . ' ' . ale#Escape(a:executable)
+        return ale#Escape(l:node)
+        \   . ' --no-deprecation '
+        \   . ale#Escape(a:executable)
     endif
 
     return ale#Escape(a:executable)
```

But that's a partial solution, only for certain situations on Windows. Most of the time, node binaries are not invoked through `node`, but rather through an executable script with a `#!/usr/bin/node` hashbang.

In that case, there is no opportunity to pass command-line arguments to the node interpreter, but we may be able to use the `NODE_NO_WARNINGS` environment variable. I'm not sure exactly where this should be set in ALE, or even if it's a good approach.

An example, just to illustrate the concept—this is obviously not correct:

```diff
diff --git a/autoload/ale/engine.vim b/autoload/ale/engine.vim
index 491d3c2e..be33af53 100644
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -450,6 +450,8 @@ function! s:RunJob(command, options) abort
         return 0
     endif
 
+    let $NODE_NO_WARNINGS = 1
+
     let l:executable = a:options.executable
     let l:buffer = a:options.buffer
     let l:linter = a:options.linter
```

elebow:
Or actually, we _could_ invoke `node` every time, like this PR.

This might be all we need:

```diff
diff --git a/autoload/ale/node.vim b/autoload/ale/node.vim
index 69060122..997a7e51 100644
--- a/autoload/ale/node.vim
+++ b/autoload/ale/node.vim
@@ -34,9 +34,11 @@ endfunction
 function! ale#node#Executable(buffer, executable) abort
     if has('win32') && a:executable =~? '\.js$'
         let l:node = ale#Var(a:buffer, 'windows_node_executable_path')
-
-        return ale#Escape(l:node) . ' ' . ale#Escape(a:executable)
+    else
+        let l:node = 'node'
     endif
 
-    return ale#Escape(a:executable)
+    return ale#Escape(l:node)
+    \   . ' --no-deprecation '
+    \   . ale#Escape(a:executable)
 endfunction
```

There's probably an edge case there on non-Unixy platforms, but I think that general approach would work fine.

velrest:
Is there a better way to fix the test than prefixing every Assert that executes a node command with `node --no-deprecation`?

elebow:
It looks like https://github.com/dense-analysis/ale/pull/2910 would be a more reliable solution to the problem, since it works even if users override the eslint executable.

If you were going to fix the tests in this PR, then yes you would have to mass replace the asserted values.

velrest:
Okay thanks. Closing since there is already a solution.

PR #2919, opened 2019-12-04
Allow headers to participate in the project.

bogado:
I don't think there's a good reason to disable unconditionally the use of the compile database when editing headers. It's an unexpected behavior and at least on my use case here it seems to break the linting on headers.

There are no tests, I could not find the tests that are testing that, and I can understand if that is denied on this grounds. Unfortunately I have little time to dig into how to run and add tests, sorry.


w0rp:
I've pushed a commit now which has this with tests for it.

PR #2920, opened 2019-12-05
Fix prettier_standard to respect the configuration file

ConradIrwin:
Before this change, prettier_standard would run and ignore any
.prettierrc, now it will respect the configuration of the file being
linted.

This change relies on prettier-standard 16.1.0 for the --stdin-filepath
flag, but is backward compatible: older versions of prettier-standard
will ignore the unknown flag and continue to run with no configuration
file.


ConradIrwin:
@w0rp ping on behalf of the stale-bot (I have tests and documentation :D!).

w0rp:
Cheers! :beers:

PR #2924, opened 2019-12-05
verilog: Add filename to vlog linter output

patrick96:
I have customized the options for `vlog` to lint all my verilog files because files depend on each other. But now the warnings in other files are always displayed in the current buffer.

For that reason I have extended the handler to also match the filename and include it in the output.

`./run-tests` works fine and it also works in my project now.


patrick96:
@w0rp Is there anything that I missed in this PR?

w0rp:
Cheers! :beers:

Issue #2922, opened 2019-12-05
Support popup windows for ALEHover and ALEDetail


dotdash:
Using the popup feature you can get a balloon like experience without having to use the mouse.

A proof of concept patch is below, but I'm not experienced enough in either vimscript or ale's code specifically to provide a fully fleshed out PR.

```diff
diff --git autoload/ale/cursor.vim autoload/ale/cursor.vim
index 8c331c5..edb99e2 100644
--- autoload/ale/cursor.vim
+++ autoload/ale/cursor.vim
@@ -136,8 +136,7 @@ function! s:ShowCursorDetailForItem(loc, options) abort
 
     let s:last_detailed_line = line('.')
     let l:message = get(a:loc, 'detail', a:loc.text)
-    let l:lines = split(l:message, "\n")
-    call ale#preview#Show(l:lines, {'stay_here': l:stay_here})
+    call ale#util#ShowMessage(l:message)
 
     " Clear the echo message if we manually displayed details.
     if !l:stay_here
diff --git autoload/ale/util.vim autoload/ale/util.vim
index 99cd856..cf46136 100644
--- autoload/ale/util.vim
+++ autoload/ale/util.vim
@@ -21,15 +21,7 @@ function! ale#util#ShowMessage(string) abort
         call ale#preview#CloseIfTypeMatches('ale-preview.message')
     endif
 
-    " We have to assume the user is using a monospace font.
-    if has('nvim') || (a:string !~? "\n" && len(a:string) < &columns)
-        execute 'echo a:string'
-    else
-        call ale#preview#Show(split(a:string, "\n"), {
-        \   'filetype': 'ale-preview.message',
-        \   'stay_here': 1,
-        \})
-    endif
+    call popup_atcursor(split(a:string, "\n"), {'border':[], 'moved': 'any'})
 endfunction
 
 " A wrapper function for execute, so we can test executing some commands.
```

RyanSquared:
related, closed by #2925 

commenting to get ref added to #2925

w0rp:
I looked at the PR, and it's a good idea.

w0rp:
Looking at #3341, do we need popup windows when we can have floating windows? There are so many "similar, but a little bit different" Vim and NeoVim features now.

pusewicz:
vim-lsp is using the popup and it's so much nicer than balloons. Would love to get the support for it!

w0rp:
[...]

PR #2828, opened 2019-10-12
Add option to show hover messages in preview.

akhilman:
Add new option 'ale_hover_to_preview' to show hover messages
in preview window.

Main reason is that I want to read the function signature while editing the code.

Possible solution for #2317

TODO: Add this option to documentation.

PR #2925, opened 2019-12-05
Add popup support to ALEHover and ALEDetails

ian-howell:
Addresses #2922.

This causes the behavior of `ALEHover` and `ALEDetails` to change when `g:ale_set_balloons = 1`, causing each command to output information to a balloon rather than using the preview window.

It doesn't appear that the modified functions have tests, and I'm honestly not sure how to go about testing such a change. If anyone has any suggestions, I'll write them.

@dotdash This change is merely a slight change to the patch you proposed in #2922. If you'd like to take ownership of this change, I will gladly abandon this PR.

### DIFF
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -137,8 +137,8 @@ function! s:ShowCursorDetailForItem(loc, options) abort
     let s:last_detailed_line = line('.')
     let l:message = get(a:loc, 'detail', a:loc.text)
 
-    if g:ale_set_balloons == 1 && has('balloon_eval')
-        call ale#util#ShowMessage(l:message)
+    if g:ale_set_popups == 1
+        call ale#popup#Show(l:message)
     else
         let l:lines = split(l:message, "\n")
         call ale#preview#Show(l:lines, {'stay_here': l:stay_here})
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -136,8 +136,12 @@ function! s:ShowCursorDetailForItem(loc, options) abort
 
     let s:last_detailed_line = line('.')
     let l:message = get(a:loc, 'detail', a:loc.text)
-    let l:lines = split(l:message, "\n")
-    call ale#preview#Show(l:lines, {'stay_here': l:stay_here})
+    if g:ale_set_balloons == 1
+        call ale#util#ShowMessage(l:message)
+    else
+        let l:lines = split(l:message, "\n")
+        call ale#preview#Show(l:lines, {'stay_here': l:stay_here})
+    endif
 
     " Clear the echo message if we manually displayed details.
     if !l:stay_here
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -136,7 +136,8 @@ function! s:ShowCursorDetailForItem(loc, options) abort
 
     let s:last_detailed_line = line('.')
     let l:message = get(a:loc, 'detail', a:loc.text)
-    if g:ale_set_balloons == 1
+
+    if g:ale_set_balloons == 1 && has('balloon_eval')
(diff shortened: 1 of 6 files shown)


autoload/ale/cursor.vim
```diff
@@ -136,8 +136,13 @@ function! s:ShowCursorDetailForItem(loc, options) abort
 
     let s:last_detailed_line = line('.')
     let l:message = get(a:loc, 'detail', a:loc.text)
-    let l:lines = split(l:message, "\n")
-    call ale#preview#Show(l:lines, {'stay_here': l:stay_here})
+
+    if g:ale_set_balloons == 1 && has('balloon_eval')
```
dotdash: Popups are more general and separate from the balloon feature. The feature check needs to check for `textprop` instead of `balloon_eval`.

There should probably also be a separate option, since at least I want popups, but no balloons as I find the mouse hover feature rather irritating.

dotdash:
Hey, thanks for picking this up :+1:

I'd rather have you push it across the finish line than take ownership if that's fine with you.

ian-howell:
I'm going to continue working on this when I find some free time. I'm planning on taking an approach similar to the strategy in https://github.com/dense-analysis/ale/pull/2828.

I'll squash all the commits when I'm finished.

ian-howell:
@w0rp I've added the `g:ale_hover_to_popup` option as well as various functionality checks.

I know that #2828 is very much fresh off the burner, but I think it may be a good idea to merge `g:ale_hover_to_popup` and `g:ale_hover_to_preview` to a single string value, perhaps `g:ale_hover_window_type`. That would reduce the complexity of having one option override the other.

Otherwise, this PR should be good to go - I just need to squash and clean up the PR and commit messages.

Let me know your thoughts.

sbromberger:
Is there any chance this might be merged soon? I'd really like to use popups instead of the preview window. Thanks.

Ktan32:
Hi! First of all, thanks to @ian-howell for this awesome feature. I've tried this branch locally and it looks very promising.

Is there anything blocking this PR from being merged to master? If yes, I'd like to help to speed things up. Right now PR looks abandoned without reason (if possible).

dotdash:
@ian-howell since this actually good to go, it might be a good idea to remove the WIP prefix and then ping the maintainer(s?) again

ian-howell:
@w0rp This should be ready to go

Ktan32:
@w0rp can you take a look please? Is there something to do left to accept the PR?

plugin/ale.vim
```diff
@@ -125,9 +125,15 @@ let g:ale_close_preview_on_insert = get(g:, 'ale_close_preview_on_insert', 0)
 " This flag can be set to 0 to disable balloon support.
 let g:ale_set_balloons = get(g:, 'ale_set_balloons', has('balloon_eval') && has('gui_running'))
 
+" This flag can be set to 0 to disable popup support.
+let g:ale_set_popups = get(g:, 'ale_set_popups', has('popupwin'))
```
w0rp, line 129: Let's remove this setting and add a separate setting for showing `:ALEDetail` messages in popups, and let's turn it off by default. Most of the text I look at for details will fill the screen horizontally.

Update the documentation for `g:ale_cursor_detail` to mention the popup behaviour. It mentions only the `preview-window` at the moment.

It might be easier if you split your pull request into two parts. First add popup support for hover, and then add popup support for detailed messages for problems.

w0rp:
I'm interested in adding popup support for mouseover hovers, which could be the default. I think `:ALEDetail` should still default to the preview window. I also want to add support for handling markdown and converting it into Vim syntax highlighting inside of the popups.

w0rp:
I'd like to see a new pull request which implements just one of these two uses of popups with tests. I recall from reading this that one part was mostly good, but I wasn't sure about the other part.

PR #2932, opened 2019-12-12
Add ale_javascript_eslint_fix_options to control eslint fixer behavior

vais:
Before I write any tests, is this something you'd be open to merging? Thanks...


w0rp:
I might consider a patch like this the option has the following behavior.

1. A new option just for the fixer version of ESLint command exists, and we make sure both JavaScript and TypeScript are supported.
2. The new option defaults to `v:null`.
3. The new option completely replaces the `_options` variable when set to a String.
4. We default to the current `_options` variable when the new option is `v:null`.

If we implement the options that way, you'll be able to both add and take away options from ESLint when running fixers.

PR #2936, opened 2019-12-17
vlog|vcom: Add temporary working directory

RyanSquared:
@engr248 this works, right?


RyanSquared:
I am unsure of how to fix tests that include a tmpdir.

w0rp:
For testing commands with a temporary directory, call the function to get the command directly, and use `subsittute()` to replace the temporary directory names with a stub w can test for.

Issue #2787, opened 2019-09-24
ESLint 6 couldn't find the plugin


kevinoid:
## Information

**VIM version**

    VIM - Vi IMproved 8.1 (2018 May 18, compiled Jun 15 2019 16:41:15)
    Included patches: 1-875, 878, 884, 948, 1046, 1365-1368, 1382, 1401

Operating System: Debian GNU/Linux (testing)

## What went wrong

Using ESLint 6, which implements [RFC 2018-simplified-package-loading](https://github.com/eslint/rfcs/blob/master/designs/2018-simplified-package-loading/README.md), if Vim edits a JavaScript file from a directory outside the project root (i.e. directory containing `node_modules`) and an applicable `.eslintrc` references plugins or formatters which are not installed globally, an error such as the following is produced:

    Oops! Something went wrong! :(
    
    ESLint: 6.4.0.
    
    ESLint couldn't find the plugin "eslint-plugin-node".
    
    (The package "eslint-plugin-node" was not found when loaded as a Node module from the directory "/tmp/b".)
    
    It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:
    
        npm install eslint-plugin-node@latest --save-dev
    
    The plugin "eslint-plugin-node" was referenced from the config file in "../a/.eslintrc.json".
    
    If you still can't figure out the problem, please stop by https://gitter.im/eslint/eslint to chat with the team.

## Reproducing the bug

1. Create a project using `eslint-plugin-node`
```sh
cd "$(mktemp -d)"
npm init -y
npm install --save-dev eslint eslint-plugin-node
echo '{"extends":"plugin:node/recommended"}' > .eslintrc.json
```
2. Run `vim index.js` - `:ALEInfo` shouldn't show any ESLint errors.
3. Run `cd .. && vim "$OLDPWD/index.js` - `:ALEInfo` should show ESLint errors similar to the above.

<details>
<summary><strong>:ALEInfo</strong></summary>

     Current Filetype: javascript
    Available Linters: ['eslint', 'fecs', 'flow', 'flow-language-server', 'jscs', 'jshint', 'standard', 'tsserver', 'xo']
      Enabled Linters: ['eslint']
     Suggested Fixers: 
      'eslint' - Apply eslint --fix to a file.
      'fecs' - Apply fecs format to a file.
      'importjs' - automatic imports for javascript
      'prettier' - Apply prettier to a file.
[...]

PR #2937, opened 2019-12-17
Run ESLint from project root dir where possible

kevinoid:
As discussed in #2787, ESLint 6 loads all plugins/configs/parsers relative to the project root which, by default, is the directory in which ESLint is invoked, as described in [ESLint RFC 2018-simplified-package-loading].

Therefore, ALE should run ESLint from the project root, when possible, so that dependencies will load.  This PR makes the change to do so.

[ESLint RFC 2018-simplified-package-loading]: https://github.com/eslint/rfcs/blob/master/designs/2018-simplified-package-loading/README.md

Fixes: #2787

Thanks for reviewing,
Kevin

### DIFF
--- a/autoload/ale/handlers/eslint.vim
+++ b/autoload/ale/handlers/eslint.vim
@@ -42,7 +42,18 @@ function! ale#handlers#eslint#GetCommand(buffer) abort
 
     let l:options = ale#Var(a:buffer, 'javascript_eslint_options')
 
-    return ale#node#Executable(a:buffer, l:executable)
+    " ESLint 6 loads plugins/configs/parsers from the project root
+    " By default, the project root is simply the CWD of the running process.
+    " https://github.com/eslint/rfcs/blob/master/designs/2018-simplified-package-loading/README.md
+    " https://github.com/dense-analysis/ale/issues/2787
+    " Identify project root from presence of node_modules dir.
+    " Note: If node_modules not present yet, can't load local deps anyway.
+    let l:modules_dir = ale#path#FindNearestDirectory(a:buffer, 'node_modules')
+    let l:project_dir = !empty(l:modules_dir) ? fnamemodify(l:modules_dir, ':h:h') : ''
+    let l:cd_command = !empty(l:project_dir) ? ale#path#CdString(l:project_dir) : ''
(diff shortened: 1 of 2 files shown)


w0rp:
Cheers! :beers:

kevinoid:
If not, feel free to ping me and I'll take another shot at it.

Thanks for reviewing and merging. :beers:

eliOcs:
Lint is working in my project but fix isn't, after debugging I found that this change should also be included in the eslint fixer for it to properly work in my project.

kevinoid:
Thanks for the heads-up @eliOcs.  Apparently it was reported in #3094 and #3095, but I created #3096 before realizing.  (Which may be useful, since it includes updates to the tests and is ready to merge.)

autoload/ale/handlers/eslint.vim
```diff
@@ -42,7 +42,18 @@ function! ale#handlers#eslint#GetCommand(buffer) abort
...
 
-    return ale#node#Executable(a:buffer, l:executable)
+    " ESLint 6 loads plugins/configs/parsers from the project root
+    " By default, the project root is simply the CWD of the running process.
+    " https://github.com/eslint/rfcs/blob/master/designs/2018-simplified-package-loading/README.md
+    " https://github.com/dense-analysis/ale/issues/2787
+    " Identify project root from presence of node_modules dir.
+    " Note: If node_modules not present yet, can't load local deps anyway.
+    let l:modules_dir = ale#path#FindNearestDirectory(a:buffer, 'node_modules')
+    let l:project_dir = !empty(l:modules_dir) ? fnamemodify(l:modules_dir, ':h:h') : ''
```
joeldodge79, line 52: @kevinoid - for better or worse I'm in a monorepo (using Yarn Workspaces) that has the eslint stuff at the repo root but each "package" has its own package.json and node_modules/ as well (for other dependencies, they just don't have eslint). see [my comment here](https://github.com/dense-analysis/ale/issues/3143#issuecomment-652452362)

seems like this line is cd'ing me into the wrong directory. I wonder if we could change the logic to use the modules_dir that has the executable defined as the project dir?

PR #2941, opened 2019-12-20
Add a language option "-l" to known filetypes for uncrustify

sankhesh:
This change adds a `-l <language>` parameter to the final uncrustify
command that is executed via ale_fix().

ALE passes the current buffer to uncrustify via an input stream without
providing a language parameter. This causes an issue where some of the
uncrustify configuration options are ignored and the output is different
than expected.

This change introduces a global `g:ale_c_uncrustify_language_mappings`
that is a dictionary of mappings between vim filetypes and uncrustify
recognized languages.

### DIFF
--- a/autoload/ale/fixers/uncrustify.vim
+++ b/autoload/ale/fixers/uncrustify.vim
@@ -3,10 +3,24 @@
 
 call ale#Set('c_uncrustify_executable', 'uncrustify')
 call ale#Set('c_uncrustify_options', '')
+call ale#Set('c_uncrustify_language_mappings', {
+\   'c'      : 'c',
+\   'cpp'    : 'cpp',
+\   'objc'   : 'oc',
+\   'objcpp' : 'oc+',
+\   'cs'     : 'cs',
+\   'java'   : 'java'
+\ })
 
 function! ale#fixers#uncrustify#Fix(buffer) abort
     let l:executable = ale#Var(a:buffer, 'c_uncrustify_executable')
     let l:options = ale#Var(a:buffer, 'c_uncrustify_options')
+    let l:language_mappings = ale#Var(a:buffer, 'c_uncrustify_language_mappings')
+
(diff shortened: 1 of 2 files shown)


sankhesh:
@w0rp This is a relatively non-invasive fix for an issue I've been seeing. Can this be reviewed and merged?

w0rp:
I'll keep this tab open and have a look later. I've slowly been merging pull requests that are basically good to go recently. At a quick glance, it looks like this is something I didn't merge a while ago because it needs tests. Generally all pull requests should come with tests.

Have a look at `:help ale-dev` for documentation on how to write tests for ALE.

autoload/ale/fixers/uncrustify.vim
```diff
@@ -3,10 +3,26 @@
...
     let l:executable = ale#Var(a:buffer, 'c_uncrustify_executable')
     let l:options = ale#Var(a:buffer, 'c_uncrustify_options')
+    let l:language_mappings = ale#Var(a:buffer, 'c_uncrustify_language_mappings')
+
+    if type(l:language_mappings) is v:t_dict
+        if has_key(l:language_mappings, &filetype)
+            let l:options = l:options
+            \   . ' -l ' . l:language_mappings[&filetype]
+        endif
+    endif
```
w0rp, line 25:
```suggestion
    if type(l:language_mappings) is v:t_dict && has_key(l:language_mappings, &filetype)
        let l:options .= ' -l ' . l:language_mappings[&filetype]
    endif
```

You can simplify this.

sankhesh:
Thank you for reviewing @w0rp. I'll make the suggested change and add a test.

sankhesh:
@w0rp I'm a little lost with the testing suite. The tests that I see for fixers are primarily(IMHO) just testing if the fixer can be run on empty files. Could you please point me to an example where the test provides input text and compares the fixer output with an expected result?

Issue #2905, opened 2019-11-21
Completion seems to trigger ALEDocumentation


RobinClowers:
## Information

**VIM version**

VIM - Vi IMproved 8.1 (2018 May 18, compiled Feb 19 2019 12:07:03)
macOS version

Operating System: macOS 10.14.6

## What went wrong

I recently updated all my plugins, including ALE, and now when I trigger completion with `C-n`, the ALEDocumentation command seems to run, it opens a scratch buffer with the docs for the top completion option. I've search for other folks having this problem and tried searching my plugins for that mapping, but nothing seems to come up.

## Reproducing the bug
 
Using deoplete.nvim and ale.vim

1. Open a typescript file
2. Trigger completion with `C-n`
3. Observe the documentation opening in a scratch preview buffer

### :ALEInfo

<details><summary>Show ALEInfo</summary>
<p>
 Current Filetype: typescript
Available Linters: ['eslint', 'standard', 'tslint', 'tsserver', 'typecheck', 'xo']
  Enabled Linters: ['eslint', 'standard', 'tslint', 'tsserver', 'typecheck', 'xo']
 Suggested Fixers: 
  'eslint' - Apply eslint --fix to a file.
  'prettier' - Apply prettier to a file.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'tslint' - Fix typescript files with tslint --fix.
  'xo' - Fix JavaScript/TypeScript files using xo --fix.
 Linter Variables:

let g:ale_typescript_standard_executable = 'standard'
let g:ale_typescript_standard_options = ''
let g:ale_typescript_standard_use_global = 0
let g:ale_typescript_tslint_config_path = ''
let g:ale_typescript_tslint_executable = 'tslint'
let g:ale_typescript_tslint_ignore_empty_files = 0
let g:ale_typescript_tslint_rules_dir = ''
let g:ale_typescript_tslint_use_global = 0
let g:ale_typescript_tsserver_config_path = ''
let g:ale_typescript_tsserver_executable = 'tsserver'
let g:ale_typescript_tsserver_use_global = 0
let g:ale_typescript_xo_executable = 'xo'
let g:ale_typescript_xo_options = ''
let g:ale_typescript_xo_use_global = 0
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
[...]

Issue #2463, opened 2019-04-28
Window content is moved when the preview window pops ups


superDross:
## Information

**VIM version**
VIM - Vi IMproved 8.1 (2018 May 18, compiled Mar 27 2019 09:36:25)
Included patches: 1-1140

Operating System: Ubuntu 18.04

## What went wrong

A preview window pops up showing some documentation when selecting a suggested completion. This causes the text to move in the main window, especially when the preview window is repeatedly opening and closing or changing height. This behaviour occurs while autocompleting with pyls but not with tsserver.

#### Unwanted Behaviour
```
+--------------+              +--------------+
| a            |              | c  (cursor)  |
| b            |              | d            |
| c  (cursor)  |     -->      | e            |
| d            |              +--------------+
| e            |              |  (preview)   |
+--------------+              +--------------+
```
[A gif showing this behaviour with pyls](https://terminalizer.com/view/4a651f63845)

#### Desired Behaviour

```
+--------------+              +--------------+
| a            |              | a            |
| b            |              | b            |
| c  (cursor)  |     -->      | c  (cursor)  |
| d            |              +--------------+
| e            |              |  (preview)   |
+--------------+              +--------------+
```
[A gif showing this behaviour with tsserver](https://terminalizer.com/view/b009b5d6844)

#### Python File
Current Filetype: python
Available Linters: ['bandit', 'flake8', 'mypy', 'prospector', 'pycodestyle', 'pydocstyl
e', 'pyflakes', 'pylama', 'pylint', 'pyls', 'pyre', 'vulture']
  Enabled Linters: ['pyls']
 Suggested Fixers:
  'add_blank_lines_for_python_control_statements' - Add blank lines before control stat
ements.
  'autopep8' - Fix PEP8 issues with autopep8.
  'black' - Fix PEP8 issues with black.
  'isort' - Sort Python imports with isort.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every lin
e.
  'yapf' - Fix Python files with yapf.
 Linter Variables:
let g:ale_python_pyls_auto_pipenv = 0
let g:ale_python_pyls_config = {}
let g:ale_python_pyls_executable = 'pyls'
let g:ale_python_pyls_use_global = 0
 Global Variables:
let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
[...]

PR #2942, opened 2019-12-22
Allow popup to be used instead of preview in completeopt

PsiPhire:
Some language servers send additional info (e.g. documentation, see https://github.com/dense-analysis/ale/issues/2905) along with auto-completion which is displayed in a preview window or not displayed at all, depending on whether `preview` is present in `completeopt`. For certain language severs (e.g. `pyls`), the `preview` option causes the window content to move which is very annoying (see https://github.com/dense-analysis/ale/issues/2463). However, the user may still want to see the extra info by using the `popup` option for `completeopt`, but since `completeopt` is overwritten by `ALE` before calling the auto-complete, this is not currently possible. This PR makes `ALE` respect the `popup` setting if the user has set it.

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -261,6 +261,8 @@ function! s:ReplaceCompletionOptions() abort
 
         if &l:completeopt =~# 'preview'
             let &l:completeopt = 'menu,menuone,preview,noselect,noinsert'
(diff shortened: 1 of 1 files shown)


PsiPhire:
Since this is such a minor change, I haven't added any tests.

w0rp:
Nice and simple.

Cheers! :beers:

Issue #2965, opened 2020-01-11
Lint error is not displayed with the latest markdownlint-cli


koketani:
## Information

**VIM version**

```text
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Dec 12 2019 16:51:54)
macOS version
```

Operating System:  

`Darwin 18.7.0 Darwin Kernel Version 18.7.0: Sun Dec  1 18:59:03 PST 2019; root:xnu-4903.278.19~1/RELEASE_X86_64 x86_64 i386`

## What went wrong

Lint error by [markdownlint](https://github.com/DavidAnson/markdownlint) is not displayed.

## Reproducing the bug

1. Install the latest [markdownlint-cli](https://github.com/igorshubovych/markdownlint-cli).
```shell-session
koketani:tmp$ npm install -g markdownlint-cli
(snipped)

koketani:tmp$ npm list -g markdownlint-cli
/Users/y-tsuji/.nvm/versions/node/v12.14.0/lib
└── markdownlint-cli@0.21.0
```

2. Write broken markdown file.
```shell-session
koketani:tmp$ cat hoge.md
a

koketani:tmp$ markdownlint hoge.md
hoge.md:1 MD041/first-line-heading/first-line-h1 First line in file should be a top level heading [Context: "a"]
```

3. Then you got the issue. (Expecting the error is displayed, but not...)

### :ALEInfo

```text
 Current Filetype: markdown
Available Linters: ['alex', 'languagetool', 'markdownlint', 'mdl', 'proselint', 'redpen', 'remark_lint', 'textlint', 'vale', 'writegood']
   Linter Aliases:
'remark_lint' -> ['remark-lint']
'writegood' -> ['write-good']
  Enabled Linters: ['markdownlint']
 Suggested Fixers: 
  'prettier' - Apply prettier to a file.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'textlint' - Fix text files with textlint --fix
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

let g:ale_markdown_remark_lint_executable = 'remark'
let g:ale_markdown_remark_lint_options = ''
let g:ale_markdown_remark_lint_use_global = 0
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = v:null
let g:ale_completion_enabled = 0
[...]

PR #2926, opened 2019-12-06
Fixed the markdownlint regex to use with the latest version

OhMyMndy:
This adds support for [DavidAnson/markdownlint](https://github.com/DavidAnson/markdownlint) the Javascript markdownlint which is inspired by the Ruby [markdownlint](https://github.com/markdownlint/markdownlint)

PR #2895, opened 2019-11-12
fix line regex pattern in markdownlint

davidsierradz:
I have:

```sh
npm install -g markdownlint-cli
markdownlint --version
0.19.0
```

The current parser for markdownlint doesn't work, this fixes it.

pinging @tylucaskelley the original contributor, thanks.

PR #2966, opened 2020-01-11
Fix 2965 - update markdownlint pattern

koketani:
#2965 

If ale needs support old pattern, please let me know.


koketani:
duplicated with #2895 or #2926

PR #2986, opened 2020-01-28
misc: change email address for @ndrewtl

ndrewtl:
This is kind of a peculiar reason for a PR, but I no longer control the email listed. I want to change it to avoid people getting the wrong email for me. Also, I still control the domain, but if at any point I don't, I want to put down in writing that if you get an email from this, it's not from me.


RyanSquared:
I'm merging this as I can verify that's the right email based on the original commit here: c6a5cbb3c737261bc46477733b3aa26203eb63d9 being "authored by" the new email. Thanks for telling us!

PR #2989, opened 2020-02-01
Add variable to allow users to change phpmd suffixes

bdlangton:
Building off #1799.

I like the idea of allowing the user to set suffixes or to use the current file's suffix as a fallback. However, I think it should still be limited to the current file having a PHP filetype. That way if the user doesn't set the suffixes variable it won't run on every kind of file they open. Let me know what you think.

### DIFF
--- a/test/command_callback/test_phpmd_command_callbacks.vader
+++ b/test/command_callback/test_phpmd_command_callbacks.vader
@@ -10,3 +10,18 @@ Execute(Custom executables should be used for the executable and command):
   AssertLinter 'phpmd_test',
   \ ale#Escape('phpmd_test')
   \ . ' %s text cleancode,codesize,controversial,design,naming,unusedcode --ignore-violations-on-exit %t'
+
+Execute(User provided options should be used):
+  let g:ale_php_phpmd_suffixes = 'module,inc'
+
+  AssertLinter 'phpmd',
+  \ ale#Escape('phpmd')
+  \ . ' %s text cleancode,codesize,controversial,design,naming,unusedcode --suffixes module,inc --ignore-violations-on-exit %t'
+
+Given php(An empty PHP file):
+Execute(Current file extension should be used as the suffix):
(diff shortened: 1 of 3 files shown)


test/command_callback/test_phpmd_command_callbacks.vader
```diff
@@ -10,3 +10,18 @@ Execute(Custom executables should be used for the executable and command):
...
+  \ ale#Escape('phpmd')
+  \ . ' %s text cleancode,codesize,controversial,design,naming,unusedcode --suffixes module,inc --ignore-violations-on-exit %t'
+
+Given php(An empty PHP file):
+Execute(Current file extension should be used as the suffix):
+  let extension = expand('#' . bufnr('') . ':e')
+
+  AssertLinter 'phpmd',
+  \ ale#Escape('phpmd')
+  \ . ' %s text cleancode,codesize,controversial,design,naming,unusedcode --suffixes ' . extension . ' --ignore-violations-on-exit %t'
```
RyanSquared: Why is `--suffixes` used here? I assume that if the option wasn't set that it wouldn't be used at all?

RyanSquared:
There's a couple of formatting issues as was caught by the --linters-only linter. You can see what exactly in Travis. I also left a comment regarding why `--suffixes` is used if the option isn't set.

test/command_callback/test_phpmd_command_callbacks.vader
bdlangton: This PR is to add --suffixes if the user provides a list of suffixes with `ale_php_phpmd_suffixes`. If they didn't set that variable but the current file is a php filetype, then -suffixes will be set with the extension of the current file. That is what this test case is testing. Only when the user doesn't provide `ale_php_phpmd_suffixes` and the current file isn't a php filetype will it not add a `--suffixes` check.

bdlangton:
Hello, just checking if this is worth still reviewing. I think it's worth adding this functionality.

PR #3007, opened 2020-02-14
Add Apple's swift-format as a linter

klaaspieter:
Adds [swift-format](https://github.com/apple/swift-format) as a linter.

There are a couple of trade-offs to be aware of:

1. This uses `swift run swift-format` when inside a Swift package. Using `swift run` means it'll build the package and then run `swift-format`. This can potentially take a while and I've seen it break when the build was interupted. The alternative is to look for swift-format in `./build/debug/swift-format` but it won't exist if the package hasn't been built.
2. Related to 1, if a package doesn't have swift-format as a dependency the linter will simply fail to run. Suggestions on how to imrove this and/or communicate this with the user are welcome 🙏 . I tried showing a generic error in `ale_linters#swift#swiftformat#Handle` but it doesn't seem to get called when `swift run` fails.
3. `swift-format` generally isn't installed in `$PATH` which means most users will have to set `swift_swiftformat_executable`. The version of `swift-format` that is installed by the default (outside of `$PATH`) is very out of date and is not compatible with this linter implementation.

### DIFF
--- a/ale_linters/swift/swiftformat.vim
+++ b/ale_linters/swift/swiftformat.vim
@@ -1,0 +1,62 @@
+" Author: Klaas Pieter Annema <https://github.com/klaaspieter>
+" Description: Support for swift-format https://github.com/apple/swift-format
+
+let s:default_executable = 'swift-format'
+call ale#Set('swift_swiftformat_executable', s:default_executable)
+
+function! ale_linters#swift#swiftformat#UseSwift(buffer) abort
+    let l:swift_config = ale#path#FindNearestFile(a:buffer, 'Package.swift')
+    let l:executable = ale#Var(a:buffer, 'swift_swiftformat_executable')
+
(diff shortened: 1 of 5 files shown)


klaaspieter:
I can't make any sense of the Travis failure. I get completely different failures locally (even on master) making it hard to fix them...

It also looks like there is (yet another) failure [on master](https://travis-ci.com/dense-analysis/ale/jobs/287386546)

klaaspieter:
Anything I can do to move this forward?

klaaspieter:
@w0rp fixed now. 😄

w0rp:
Cheers! :beers:

People might be confused by there being two different tools that sound the same. I guess that's a problem for Swift.

klaaspieter:
100% agree. SwiftFormat is a community project while swift-format is an Apple project. To make matters even more complicated the Apple project seems to largely be maintained by Google engineers 🤯 

Thanks for merging!

PR #3010, opened 2020-02-16
Alias ps1 filetype to powershell

kevinoid:
Thanks for writing a linter for PSScriptAnalyzer @zigford!  I was surprised that it isn't used by default with [vim-ps1] due to requiring manual configuration to alias the `ps1` filetype to `powershell`.  Is there a reason you chose not to include it in `s:default_ale_linter_aliases`?

Since [vim-ps1] is a popular (the only?) PowerShell ftplugin and there do not appear to be any other uses of the `ps1` filetype on vim.org, this seems like a safe and reasonable default.  This PR would add it, assuming there wasn't a specific reason to avoid it.

Thanks again,
Kevin

[vim-ps1]: http://www.vim.org/scripts/script.php?script_id=1327

### DIFF
--- a/autoload/ale/linter.vim
+++ b/autoload/ale/linter.vim
@@ -14,6 +14,7 @@ let s:default_ale_linter_aliases = {
 \   'csh': 'sh',
 \   'javascriptreact': ['javascript', 'jsx'],
 \   'plaintex': 'tex',
+\   'ps1': 'powershell',
 \   'rmarkdown': 'r',
 \   'rmd': 'r',
 \   'systemverilog': 'verilog',
--- a/doc/ale-powershell.txt
+++ b/doc/ale-powershell.txt
@@ -25,13 +25,6 @@ Installation
 
 Install PSScriptAnalyzer by any means, so long as it can be automatically
 imported in PowerShell.
-Some PowerShell plugins set the filetype of files to `ps1`. To continue using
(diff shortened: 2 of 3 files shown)


zigford:
Hi @kevinoid I used powershell over ps1 as I didn't agree on vim-ps1's decision to use `ps1` as the filetype name of powershell scripts. It doesn't fit with how every other language names their filetypes. Ie, python filetype is not named after it's extension, nor is ruby.

I've also made [https://github.com/zigford/vim-powershell](vim-powershell) in the hopes that others will contribute and we can have sane powershell + vim functionality.

Anyway, I digress. The proper answer is that I did not know about `s:default_ale_linter_aliases`, so I think this will be a welcome change and make the current powershell linting easier to discover in ale.

kevinoid:
Thanks @zigford!  Makes sense to me.  I'm looking forward to checking out [vim-powershell](https://github.com/zigford/vim-powershell).

@w0rp: I think it's safe to merge.  (If you approve, of course.)

kevinoid:
@w0rp I think this PR is good to merge and not stale.

w0rp:
Cheers! :beers:

kevinoid:
Yes indeed.  Thanks for reviewing and merging it!  :clinking_glasses:

PR #3013, opened 2020-02-18
Fixer: Black: Use the executable variable

ayroblu:
It wasn't using the variable but I think it should?

### DIFF
--- a/autoload/ale/fixers/black.vim
+++ b/autoload/ale/fixers/black.vim
@@ -14,7 +14,7 @@ function! ale#fixers#black#GetExecutable(buffer) abort
     endif
 
     let l:executable = ale#Var(a:buffer, 'python_black_executable')
-    
+
     return ale#python#FindExecutable(a:buffer, 'python_black', [l:executable])
 endfunction
 
--- a/autoload/ale/fixers/black.vim
+++ b/autoload/ale/fixers/black.vim
@@ -14,6 +14,7 @@ function! ale#fixers#black#GetExecutable(buffer) abort
     endif
 
     let l:executable = ale#Var(a:buffer, 'python_black_executable')
+    
     return ale#python#FindExecutable(a:buffer, 'python_black', [l:executable])
 endfunction
 
--- a/autoload/ale/fixers/black.vim
+++ b/autoload/ale/fixers/black.vim
@@ -12,8 +12,9 @@ function! ale#fixers#black#GetExecutable(buffer) abort
     \ && ale#python#PipenvPresent(a:buffer)
         return 'pipenv'
     endif
-
-    return ale#python#FindExecutable(a:buffer, 'python_black', ['black'])
+    
+    let l:executable = ale#Var(a:buffer, 'python_black_executable')
+    return ale#python#FindExecutable(a:buffer, 'python_black', [l:executable])
 endfunction
 
 function! ale#fixers#black#Fix(buffer) abort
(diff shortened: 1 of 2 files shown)


ayroblu:
@w0rp pls review

autoload/ale/fixers/black.vim
```diff
@@ -13,7 +13,9 @@ function! ale#fixers#black#GetExecutable(buffer) abort
         return 'pipenv'
     endif
 
-    return ale#python#FindExecutable(a:buffer, 'python_black', ['black'])
+    let l:executable = ale#Var(a:buffer, 'python_black_executable')
+
+    return ale#python#FindExecutable(a:buffer, 'python_black', [l:executable])
```
w0rp, line 18: This doesn't make sense. The string in the `List` before is the name to look for inside of virtualenv directories. ALE uses the global executable name if a virtualenv path cannot be found, or if you override searching in virtualenv with the `use_global` option. Almost all of the Python linters work this way.

ayroblu: @w0rp I'm not sure what I should be doing here. I want to support "cblack" instead of "black" as my executable. Is it just a matter of putting `[l:executable, 'black']` or something else I need to do?

w0rp: It looks like what you actually want to do is run `cblack`. If you want to use other executables, set `ale_python_black_executable` to the path to that executable, and you probably want to set `ale_python_black_use_global` to `1` to turn off the virtualenv logic ALE has for looking for `black` inside of `venv/bin` and so on. You don't need to change any code in ALE.

w0rp: The easiest way to use different settings for different projects is with `g:ale_pattern_options` or logic in your `ftplugin` files. You want to specify the full file system path to the executable.

ayroblu: @w0rp So to be clear, I can't use the venv cblack executable, and instead need to have a global absolute executable?

w0rp: Pretty much, yeah. You can set the absolute path to an executable in a virtualenv directory. Because this is a linter just for `black`, it doesn't look for `cblack`. The purpose of the `executable` settings in ALE is to set absolute paths to executables somewhere else on your system.

We could consider using just `['black', 'cblack']` in the code here, if `cblack` is common enough. From what I can tell, it looks like it's kind of a niche.

ayroblu: @w0rp Sorry last thing. this makes "python_black_executable" a completely useless variable and we should remove it then? I think this is what confused me when I first looked at this, there is the philosophical support for this, but not the wiring up

w0rp: Read `:help ale-integrations-local-executables`. For pretty much all Python and JavaScript tools, the locally installed executables are checked for first, before falling back on globally installed programs, so ALE can run different versions for each project. You can change one setting to change the behaviour everywhere.

It's not a "completely useless variable," you just haven't read the documentation to understand what the setting does.

ayroblu: I had a go at this again (this pr was made a long time ago). It appears the issue of using the variable according to the documentation is no longer the case. I'm guessing at the time it seemed that the variable was not hooked up and so it was not working for me. The code now is very different from the code at the time of this diff.

I'm not sure whether this is user error or a bug that was fixed in the mean time, but at the time I couldn't get the executable variable to behave as intended. This is no longer an issue, I apologise for the inconvenience

w0rp: I'm glad you got it working the way you want to. Don't worry about it.

PR #3017, opened 2020-02-23
Truncate tag stack before pushing

### LINKED ISSUE #2945: Tagstack is always appended to, behaves different to native tags
Currently ale doesn't replicate Vim's native tag behavior when using `:ALEGoToDefinition`. The way that the tagstack is updated is different.

Natively, `CTRL-T` will go up in the tagstack, decrementing the "current index". A subsequent `CTRL-]` will push to the tagstack at the position of the current index, replacing everything that came after this index.

Ale, however, always append at the end of the tagstack, ignoring the current index. This breaks `CTRL-T` because it will jump to the last definition that had been jumped to chronologically, rather than the one that was previously "current".

Unfortunately `settagstack` can only replace or append to the tagstack, so we have to manually replicate the behavior. (I've just opened https://github.com/vim/vim/issues/5405 to suggest to provide it natively.)

frangio:
Fixes #2945.

### DIFF
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -29,8 +29,19 @@ function! ale#definition#UpdateTagStack() abort
         let l:old_location = [bufnr('%'), line('.'), col('.'), 0]
         let l:tagname = expand('<cword>')
         let l:winid = win_getid()
-        call settagstack(l:winid, {'items': [{'from': l:old_location, 'tagname': l:tagname}]}, 'a')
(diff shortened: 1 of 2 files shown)


j-hui:
Not a maintainer, but this would be a nice feature to merge in. Is there any chance of reviving this?

j-hui:
This would be a nice feature to merge in. Is there any chance a maintainer would take a look at this?

PR #3025, opened 2020-03-04
Refactor the "s:LoadArgCount()" function

lifecrisis:
See commit message for details.

### DIFF
--- a/autoload/ale/util.vim
+++ b/autoload/ale/util.vim
@@ -343,7 +343,7 @@ function! s:LoadArgCount(function) abort
     endtry
 
     let l:match = matchstr(split(l:output, "\n")[0], '\v\([^)]+\)')[1:-2]
-    let l:arg_list = filter(split(l:match, ', '), "v:val isnot# '...'")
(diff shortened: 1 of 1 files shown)


autoload/ale/util.vim
```diff
@@ -336,18 +336,14 @@ function! ale#util#GetMatches(lines, patterns) abort
...
+    try
+        let l:output = execute('function a:function')
+    catch /E123/
         return 0
-    endif
+    endtry
 
     let l:match = matchstr(split(l:output, "\n")[0], '\v\([^)]+\)')[1:-2]
-    let l:arg_list = filter(split(l:match, ', '), 'v:val isnot# ''...''')
+    let l:arg_list = filter(split(l:match, ', '), "v:val isnot# '...'")
```
w0rp:
```suggestion
    let l:arg_list = filter(split(l:match, ', '), 'v:val isnot# ''...''')
```

I'll change this back, as I prefer it how it was. Single quote doubling for escaping is common, and double quote usage should be pretty rare.

w0rp:
Cheers! :beers:

PR #3029, opened 2020-03-06
ktlint: Explicitly pass editorconfig path, if found

dhleong:
This enables ktlint to respect [editorconfig](https://github.com/pinterest/ktlint#editorconfig) rules when the file has been
copied to a temporary directory


hsanson:
#3119 also allows support for editorconfig by using ktlint in stdin mode that I believe is a better approach.

dhleong:
Oh, nice! I'll close this PR in favor of yours.

PR #3043, opened 2020-03-10
Fix sqlfmt fixer

johannes-jansson:
ISSUE:
Running the sqlfmt fixer just replaced my current file with the text:
`unknown shorthand flag: 'w' in -w`

SOLUTION: 
Remove the -w flag, that no longer seems to be supported in sqlfmt.

### DIFF
--- a/autoload/ale/fixers/sqlfmt.vim
+++ b/autoload/ale/fixers/sqlfmt.vim
@@ -7,7 +7,6 @@ function! ale#fixers#sqlfmt#Fix(buffer) abort
 
     return {
     \   'command': ale#Escape(l:executable)
-    \       . ' -w'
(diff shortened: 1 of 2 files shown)


johannes-jansson:
I was just informed that there are two different parsers with a binary called sqlfmt, so this might not be the ideal way to solve my issue. I'll do some more research and modify or drop the PR.

adamsanghera:
Perhaps we can follow the pattern of `ale_sql_sqlfmt_options`,
and make `-w` the default option for purposes of backwards-compatibility?

Or maybe there's some precedent for naming collisions, that I'm not aware of. :)

w0rp:
Make the options configurable instead. I think `-w` is valid for the `sqlfmt` tool ALE is actually using.

Issue #2816, opened 2019-10-06
StandardJS "fix" doesn't work.


pav881:
Debian 10, last vim from git(vim/src/ make, sudo make install). Another "fixers", for example: "prettier" - works fine.

.bashrc
NPM_PACKAGES="${HOME}/.local/.npm"
export PATH="$NPM_PACKAGES/bin:$PATH"
export MANPATH="$NPM_PACKAGES/share/man:$(manpath)"

npm -g install standard
let g:ale_fixers = {'c': ['clang-format'], 'javascript': ['standard']}
I've tried ALEFix - no result.

Info:
Current Filetype: javascript
Available Linters: ['eslint', 'fecs', 'flow', 'flow-language-server', 'jscs', 'jshint', 'standard', 'tsserver', 'xo']
  Enabled Linters: ['standard', 'tsserver']
Suggested Fixers:
  'eslint' - Apply eslint --fix to a file.
  'fecs' - Apply fecs format to a file.
  'importjs' - automatic imports for javascript
  'prettier' - Apply prettier to a file.
  'prettier_eslint', 'prettier-eslint' - Apply prettier-eslint to a file.
  'prettier_standard', 'prettier-standard' - Apply prettier-standard to a file.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'standard' - Fix JavaScript files using standard --fix
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'xo' - Fix JavaScript/TypeScript files using xo --fix.
Linter Variables:

let g:ale_javascript_standard_executable = 'standard'
let g:ale_javascript_standard_options = ''
let g:ale_javascript_standard_use_global = 0
let g:ale_javascript_tsserver_config_path = ''
let g:ale_javascript_tsserver_executable = 'tsserver'
let g:ale_javascript_tsserver_use_global = 0
Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = v:null
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = v:null
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
[...]

PR #3046, opened 2020-03-12
Fix 2816 - Standard fix does not work.

hsanson:
The standard linter --fix fails if the file being input is not relative
to the project root (https://github.com/standard/standard/issues/1384).

This MR attempts to fix this by changing the command so the input file
is relative to the project root and the output is to a temporary file.

Preliminary tests with toy javascript projects seem to indicate this
works fine.


Frando:
Found this after being biten by #2816. I tested can confirm that this PR fixes  `:ALEFix` with StandardJS 14.3.3.

w0rp:
Cheers! :beers:

Issue #569, opened 2017-05-21
ale should override signs from gitgutter


codeHatcher:
Ale signs should take precedence over gitgutter, that's also how syntastic and gitgutter work together with syntastic taking precedence. the idea being as I fix my syntax errors, I'll be able to see my gitgutter signs. Especially important because I should be fixing syntax errors before committing code (which would clear the gitgutter signs)

Here is a related discussion on this for gitgutter and syntastic quite some time ago https://github.com/airblade/vim-gitgutter/issues/1

w0rp:
This follows on from #23, one of the earliest issues, where I made it so it doesn't remove gitgutter signs. This might be slightly harder to implement than for Syntastic, as ALE updates signs as you type. If anyone wants to try fixing this, give it a go.

codeHatcher:
Yup, I noticed that as well. Originally I was using Syntastic which is where I got use to the functionality. I happened to reinstall my dev environment and only now was wondering where the linter went. For now I can disable gitgutter until someone is able to come up with a fix or I move to Syntastic/Neomake.

@w0rp is it something that may not be possible with Ale? Helpful to know for deciding how long I should wait to switch over. Is there a priority level for what displays in the sign column?

w0rp:
I reckon it should be possible. Someone just needs to go ahead and do it.

codeHatcher:
@w0rp do you have a bounty bot you use, I wouldn't mind attaching a couple bucks to this issue

codeHatcher:
I'm wrong, my issue was because I was using emoji for the gutter for ale and I see there is another issue here https://github.com/w0rp/ale/issues/511 that references a problem with tmux 2.4. In order to get both my gitgutter and ale signs to display, I changed the ale sign to simple characters and now everything looks ok

![image](https://cloud.githubusercontent.com/assets/5360982/26287031/6988de4a-3e40-11e7-8abf-d5abb806c96f.png)

w0rp:
Cool. :+1:

tjkirch:
I still have this issue when using the default ">>" and "--" ASCII sign characters for ale.  Same thing if they're single characters, e.g. ">" and "-".  gitgutter overrides ale, but I definitely want errors to take priority over git status.

My workaround is to have ale show the location list on error, but I'd prefer the sign characters alone with the ability to open the location list when needed.

I'm using vim 8.0.1415, ale 1568bf8128, and vim-gitgutter 1e60667322, all the latest as of now, and no tmux.

[...]

PR #3050, opened 2020-03-15
Fix vim sign priority patch check

StarryLeo:
With Vim 8.2 released, the previous check method is not accurate enough.

Refer: https://github.com/dense-analysis/ale/issues/569#issuecomment-576613818

### DIFF
--- a/autoload/ale/sign.vim
+++ b/autoload/ale/sign.vim
@@ -23,7 +23,7 @@ let g:ale_sign_offset = get(g:, 'ale_sign_offset', 1000000)
 let g:ale_sign_column_always = get(g:, 'ale_sign_column_always', 0)
 let g:ale_sign_highlight_linenrs = get(g:, 'ale_sign_highlight_linenrs', 0)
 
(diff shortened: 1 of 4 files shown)


RyanSquared:
I'm all for this simplification but I'm not sure why the previous method was not accurate enough? It checks if the version is at least 8.1 and has the above patches, which is the same thing your PR should do, no?

StarryLeo:
> I'm all for this simplification but I'm not sure why the previous method was not accurate enough? It checks if the version is at least 8.1 and has the above patches, which is the same thing your PR should do, no?

See `:h v:version` may be able to answer the question.

hsanson:
@RyanSquared thanks for bringing up the question. I did some more testing and indeed the current check fails in vim  8.2:

```
echo has('patch614')  => 0
```

This means, I assume, that vim 8.2 has the sign priority feature added to it by means other than patch 614, therefore it does not include it?

In any case @StarryLeo fix properly works on vim version 8.1.0614 onwards. I tested this with different versions using w0rp/ale docker image.

StarryLeo:
@hsanson I think `has('patch614')` only check for current `v:version`, there is indeed no patch614 for vim8.2 now.

Issue #2339, opened 2019-03-07
golangci-lint doesn't work correctly when linting files only


svanharmelen:
## Information

**VIM version**

NVIM v0.3.4
Build type: Release

Operating System: macOS Mojave (10.14.3)

## What went wrong

Thanks for adding `golangci-lint` support to ALE @saschagrunert!

But I wonder if the current implementation works correctly? I prefer to only fill my location list with issues that correspond to the file that I'm currently working on (so using `let g:ale_go_golangci_lint_package=0` which is actually the default), but when using `golangci-lint` the `typecheck` is unable to correctly check types that are defined in other package files and returns unexpected type errors.

In `gometaliner` this works different and gives the expected and correct output, so I was hoping that it could be fixed for `golangci-lint` as well? For now I guess the only way to do this is by specifying `gotype` as a second linter in my ALE config, but it would be much nicer if that wouldn't be necessary.

## Reproducing the bug

let g:ale_linters = {
  \ 'go': ['golangci-lint']
  \'
let g:ale_go_golangci_lint_options = '
  \ --fast
  \'

1. Run ALE on a file that uses types defined in other files in the same package

svanharmelen:
In addition, I must say that running this linter on the whole package still seems to be pretty fast. But then your location list is filled with _all_ errors so jumping to the next error will then redirect you to completely unrelated files (in relation to the one you are working on).

I wonder if we should take the same approach as we did with `gometaliner` where the linter are executed on the whole package, but the output is parsed to only use the issues for your current file.

svanharmelen:
I created this issue https://github.com/golangci/golangci-lint/issues/420 to check if using the same approach we did with `gometalinter` is an option.

SparrowOchon:
An easy fix for this is to use golangci-lint v1.19.0, this allows the linter to work on singular files.

saschagrunert:
Yeah I think golangci-lint does not work at all with single files. :thinking:

svanharmelen:
Closing this one as I no longer use this plugin and the referenced issue (https://github.com/golangci/golangci-lint/issues/420) has been closed 6+ months ago as well...

ryanc414:
What fixed this issue for me was adding:

```
let g:ale_go_golangci_lint_package = 1
```

to my `init.vim`. Seems that it tells ALE to lint the whole package rather than just a single file.

PR #3074, opened 2020-03-25
support project-level golangci-lint linting

JeanMertz:
The new `g:ale_go_golangci_lint_scope` option supersedes `g:ale_go_golangci_lint_package` and supports three different modes:

* file
* module
* project

The default is `file`, and means `golangci-lint` lints the current file.

The `module` option inherits the behavior previously provided by `g:ale_go_golangci_lint_package`.

The new option `project` searches for the root of the Go project (either with a "go.mod" file, a ".git" directory, or within $GOPATH) and runs the linter from there.

This is helpful in cases such as https://github.com/dense-analysis/ale/issues/2339 where the `unused` lint will report errors if it runs against a module or file, as opposed to from the project root.

### DIFF
--- a/ale_linters/go/golangci_lint.vim
+++ b/ale_linters/go/golangci_lint.vim
@@ -3,25 +3,31 @@
 
 call ale#Set('go_golangci_lint_options', '--enable-all')
 call ale#Set('go_golangci_lint_executable', 'golangci-lint')
+call ale#Set('go_golangci_lint_scope', 'file')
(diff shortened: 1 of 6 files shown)


JeanMertz:
I'm not sure yet why the CI is failing.

Locally, the tests are passing (`./run-tests`), but both CI environments have the same test failing (related to creating a temporary `.git` directory). Could there be some race condition w.r.t. parallel tests?

Issue #2958, opened 2020-01-07
Addtional ^M characters on Windows


jdhao:
## Information

**Neovim version**

NVIM v0.5.0-172-g1ff5b60cb
Build type: RelWithDebInfo

Operating System: Windows 10 version 1803

## What went wrong

The ALE lint message on the status line has addtional `^M` character for Windows.

![image](https://user-images.githubusercontent.com/16662357/71887273-b7572680-3178-11ea-8c52-8a86191e0ebe.png)

## Reproducing the bug

minimal `init.vim`:

```vim
set rtp+=C:\Users\Administrator\AppData\Local\nvim-data\plugged\ale

" linters for different filetypes
let g:ale_linters = {
    \ 'python': ['pylint', 'flake8'],
    \ 'vim': ['vint']
\}

" Only run linters in the g:ale_linters dictionary
let g:ale_linters_explicit = 1

" Linter signs
let g:ale_sign_error = '❌'
let g:ale_sign_warning = '!'
```

Open a Python source file with the above `init.vim`:

```
nvim-qt -- -u init.vim test.py
```

You should be able to see that the warnning message on status line
has additional `^M` character.

I believe this is an issue with new line handling from linters on Windows. I
have previously been using Neomake for linting my code and haven't met this
issue.

### :ALEInfo

```
 Current Filetype: python
Available Linters: ['bandit', 'flake8', 'mypy', 'prospector', 'pycodestyle', 'pydocstyle', 'pyflakes', 'pylama', 'pylint', 'pyls', 'pyre', 'vulture']
  Enabled Linters: ['flake8', 'pylint']
 Suggested Fixers: 
  'add_blank_lines_for_python_control_statements' - Add blank lines before control statements.
  'autopep8' - Fix PEP8 issues with autopep8.
  'black' - Fix PEP8 issues with black.
  'isort' - Sort Python imports with isort.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'reorder-python-imports' - Sort Python imports with reorder-python-imports.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'yapf' - Fix Python files with yapf.
 Linter Variables:

let g:ale_python_auto_pipenv = 0
let g:ale_python_flake8_auto_pipenv = 0
let g:ale_python_flake8_change_directory = 1
let g:ale_python_flake8_executable = 'flake8'
let g:ale_python_flake8_options = ''
let g:ale_python_flake8_use_global = 0
[...]

PR #3078, opened 2020-03-26
Issue 2958: Addtional ^M characters on Windows

jgehrig:
**Issue #2958**

Windows may insert carriage return line endings, which ALE does not handle
well. These characters should not be displayed.

Adds a line to remove these characters for all messages.

Tests are still needed.

I'm not familiar with this code base. If this looks like a reasonable approach, let me know and I will write some tests.


jgehrig:
There is a typo in the commit message `Issue 2598`... I will fix it when/if tests are added.

Let me know if the approach here is reasonable.

w0rp (commented):
This looks okay to me. I know how to modify the tests quickly to cover this.

w0rp (approved):

w0rp:
Cheers! :beers;

jgehrig:
Thanks! :+1:

PR #3083, opened 2020-03-29
Remove --stdin from prettier config for prettier v2 change

rhysd:
From prettier v2, `--stdin` was removed.

https://github.com/prettier/prettier/pull/7668

This PR removes the flag when applying Prettier fixer. I'm not familiar with prettier-eslint and prettier_d. So this PR fixes only standard `prettier` CLI.


rhysd:
Can this PR have chance to be merged? Specifying `--stdin` causes warnings as unknown option.

Issue #1095, opened 2017-11-08
ale doesn't respect `.prettierignore` file


ahmedelgabri:
Ale prettier fixer doesn't respect `.prettierignore` file rules.

Example:

`.prettierignore`
```
foo.js
```

Running this commend on the CLI will respect `.prettierignore`

```sh
$ prettier --write foo.js # won't be formatted
```

While running the same command through ale will modify the file, I assume that's because ale has no knowledge about the paths in `.prettierignore` & it just modifies the buffer.  I dunno how this can be fixed to be honest, any ideas?

w0rp:
Does `prettier` support fixing files via stdin with flag like `--stdin-filename` yet? That's the reason why things like this don't work so well.

w0rp:
To support this, we'll probably have to read the ignore file and apply the globs ourselves to ignore things.

ahmedelgabri:
> Does `prettier` support fixing files via stdin with flag like `--stdin-filename` yet? That's the reason why things like this don't work so well.

I don't think so.

> To support this, we'll probably have to read the ignore file and apply the globs ourselves to ignore things.

Yeah I thought so, but then it might be too much work especially that paths in `.prettierignore` mostly are relative & you can have multiple of these files inside multiple directories etc...

w0rp:
At the very least you can use `g:ale_pattern_options` to change the fixers for particular files.

ahmedelgabri:
Something like this for example to disable the fixer?

```vim
  let g:ale_pattern_options = {
  \   '\.foo\.js$': {
  \       'ale_fixers': [],
  \   },
  \}
```

w0rp:
Replace `[]`, with something like `{'javascript': []}`.

ahmedelgabri:
Thanks!

ahmedelgabri:
As far as I understood from the `:help`, the regex will use `match()` to match agains the absolute path of the buffer. But for some reason, these don't work, even thought doing `:echo match(expand('%:p'), <string>)` works fine with each single one of these.

```vim
let g:ale_pattern_options = {
      \   '\.min\.(js\|css)$': {
      \       'ale_linters': {},
      \       'ale_fixers': {},
      \   },
      \   'jquery.*$': {
      \       'ale_linters': {},
      \       'ale_fixers': {},
      \   },
      \   'node_modules': {
      \       'ale_linters': {},
      \       'ale_fixers': {},
      \   },
      \   'Sites/forks': {
[...]

PR #3101, opened 2020-04-10
Set prettier working directory to where .prettierignore is

iclanzan:
Prettier does not use `.prettierignore` unless the current directory is the root where the `.prettierignore` file resides.

### DIFF
--- a/autoload/ale/fixers/prettier.vim
+++ b/autoload/ale/fixers/prettier.vim
@@ -97,8 +97,7 @@ function! ale#fixers#prettier#ApplyFixForVersion(buffer, version) abort
     " 1.4.0 is the first version with --stdin-filepath
     if ale#semver#GTE(a:version, [1, 4, 0])
         return {
-        \   'command': ale#path#BufferCdString(a:buffer)
-        \       . ale#Escape(l:executable)
+        \   'command': ale#Escape(l:executable)
         \       . (!empty(l:options) ? ' ' . l:options : '')
         \       . ' --stdin-filepath %s --stdin',
         \}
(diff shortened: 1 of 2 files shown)


RyanSquared:
related: #1095

RyanSquared:
I don't see the reason why --stdin-filename required CdString.

iclanzan:
@w0rp do you think we can merge this?

w0rp:
I don't think it's the right solution. Unless the CWD is set to something, it could be anything at all. For many linters, the CWD doesn't matter. If the CWD has to be the same as the directory that `.prettierignore` is in, we should use the `ale#path#FindNearestFile` function to find it and set the CWD to that directory. That way the fixer will always use the right directory even when people have configured Vim to set the CWD to other directories.

I have Vim configured to set the CWD automatically to the directory the current file I'm editing is in.

iclanzan:
That wouldn’t work though because many projects don’t have a `.prettierignore`.

w0rp:
We'd handle the path possibly being there or not. Here's just one example of another tool doing a similar thing.

```vim
function! ale_linters#cpp#cquery#GetProjectRoot(buffer) abort
    " Try to find cquery configuration files first.
    let l:config = ale#path#FindNearestFile(a:buffer, '.cquery')

    if !empty(l:config)
        return fnamemodify(l:config, ':h')
    endif

    " Fall back on default project root detection.
    return ale#c#FindProjectRoot(a:buffer)
endfunction
```

w0rp:
I've re-opened this. You can make the changes we need.

iclanzan:
Thanks for the guidance @w0rp, I think this is ready for another review!

w0rp (approved):

w0rp:
Cheers! :beers:

Issue #611, opened 2017-06-01
Support fixers which must look at files on disk


w0rp:
Some tools for fixing files might have to only work with files on disk in exactly the right path. ALE should support these tools. Here is what needs to be done.

1. Add a new option `fix_file` which when set to `1` will tell ALE that the file will be modified.
2. Before jobs for fixers with `fix_file` are run, write changes to the file so far with `writefile`.
3. After job for fixers with `fix_file` are finished, load the lines of input for the next fixer back from the file.
4. Fixers with `fix_file` set to `1` should be skipped when fixing is done from an event for fixing files while typing.

JeanMertz:
I know this is quite an old issue, so perhaps this has long been resolved.

However, I ran into a situation like this (?) today, when I wanted to use `rustfix` as a fixer for Rust in Ale.

Here's what I had to do to make this _somewhat_ work:

```vim
call ale#Set('rust_rustfix_executable', 'cargo fix')
call ale#Set('rust_rustfix_options', '--allow-dirty --allow-staged')

function! RustFix(buffer) abort
	let l:executable = ale#Var(a:buffer, 'rust_rustfix_executable')
	let l:options = ale#Var(a:buffer, 'rust_rustfix_options')

	return {
	\   'command': l:executable
	\       . (empty(l:options) ? '' : ' ' . l:options),
	\   'read_buffer': 0,
	\   'process_with': 'RustFixCallback',
	\}
endfunction

function! RustFixCallback(buffer, output) abort
	" edit
	return []
endfunction
```

```vim
let g:ale_fixers = {
	\ '*': ['remove_trailing_lines', 'trim_whitespace'],
	\ 'rust': ['rustfmt', 'rustfix'],
	\ }
```

```vim
augroup ALEPostFix
	autocmd!
	au User ALEFixPost edit
	" au User ALEFixPost checktime %
augroup END
```

The reason I needed this hack is that:

1. `cargo fix` (which is the command to run `rustfix`) works on an entire workspace, not just a single file
2. it also does not accept any input on stdin, nor does it produce any (usable) output on stdout.
3. I managed to make it apply the changes, and could see them on my filesystem, but Vim didn't reload the file, so the changes weren't visible after running `ALEFix`
4. If I used the (commented out) `edit` inside `RustFixCallback`, it would work, but if there was another fixer that still had to run, it would error with "The file was changed before fixing finished"
[...]

PR #3106, opened 2020-04-15
Elixir Mix Format: Execute in Mix project root

connorlay:
This PR improves the experience of using the `mix_format` fixer for Elixir by ensuring execution the `mix format` command in the root of the Mix project. This change guarantees that `mix format` will read the `.formatter.exs` file if present.

This implementation is how the [official documentation](https://hexdocs.pm/mix/Mix.Tasks.Format.html#module-when-to-format-code) suggests integrating `mix format` with editors.

### DIFF
--- a/autoload/ale/fixers/mix_format.vim
+++ b/autoload/ale/fixers/mix_format.vim
@@ -9,10 +9,13 @@ function! ale#fixers#mix_format#GetExecutable(buffer) abort
 endfunction
 
 function! ale#fixers#mix_format#GetCommand(buffer) abort
+    let l:project_root = ale#handlers#elixir#FindMixUmbrellaRoot(a:buffer)
     let l:executable = ale#Escape(ale#fixers#mix_format#GetExecutable(a:buffer))
     let l:options = ale#Var(a:buffer, 'elixir_mix_format_options')
 
-    return l:executable . ' format'
+    return ale#path#CdString(l:project_root)
+    \   . l:executable
+    \   . ' format'
     \   . (!empty(l:options) ? ' ' . l:options : '')
     \   . ' %t'
 endfunction
--- a/test/fixers/test_mix_format_fixer_callback.vader
+++ b/test/fixers/test_mix_format_fixer_callback.vader
@@ -18,7 +18,7 @@ Execute(The mix_format callback should return the correct default values):
   AssertEqual
   \ {
   \   'read_temporary_file': 1,
-  \   'command': 'cd ''.'' && '
+  \   'command': ale#path#CdString('.')
   \     . ale#Escape('xxxinvalid')
   \     . ' format %t',
   \ },
@@ -31,7 +31,7 @@ Execute(The mix_format callback should include the correct format options):
   AssertEqual
   \ {
   \   'read_temporary_file': 1,
-  \   'command': 'cd ''.'' && '
+  \   'command': ale#path#CdString('.')
   \     . ale#Escape('xxxinvalid')
   \     . ' format invalid_options %t',
   \ },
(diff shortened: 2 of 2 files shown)


connorlay:
```
Starting Vader: C:\testplugin\test\fixers\test_mix_format_fixer_callback.vader
    (1/2) [EXECUTE] The mix_format callback should return the correct default values
    (1/2) [EXECUTE] (X) {'read_temporary_file': 1, 'command': 'cd /d . && xxxinvalid format %t'} should be equal to {'read_temporary_file': 1, 'command': 'cd ''.'' && xxxinvalid format %t'}
    (2/2) [EXECUTE] The mix_format callback should include the correct format options
    (2/2) [EXECUTE] (X) {'read_temporary_file': 1, 'command': 'cd /d . && xxxinvalid format invalid_options %t'} should be equal to {'read_temporary_file': 1, 'command': 'cd ''.'' && xxxinvalid format invalid_options %t'}
  Success/Total: 0/2
```

Here are the relevant test failures in CI. I'll see if I can fix these.

dkarter:
Wow I ***REALLY*** need this fix! Thank you for submitting this PR!

I tried doing some testing on your branch but was unable to get it working locally, even though I can confirm the `cd` command is correct and it was finding the root of the umbrella app, for me that was not the issue. ALE still cannot pick up the specific configuration in nested directories, but running mix format directly from the command line does pick it up.

Without ALE:


With ALE (this branch + added `echom` messages for confirming paths):

connorlay:
> ALE still cannot pick up the specific configuration in nested directories, but running mix format directly from the command line does pick it up.

@dkarter Hey thanks for giving this a shot! To clarify, you have an umbrella project that has a `.formatter.exs` at the root, as well as in each child application? I'll see if I can reproduce this and fix it :smile:

dkarter:
@connorlay correct! That's what you get by default when you create a new umbrella project from `phx.new` - maybe you can use that in your testing:

```sh
mix phx.new umbrella_format_example --umbrella --no-ecto --no-webpack
```

image

This is _without_ your fix (ignores the `import_deps: [:phoenix]` inside of the `_web` directory which imports the `.formatter.exs` from `deps/phoenix/.formatter.exs` that has a `locals_without_parens` rule for `plug`, `get` etc):



This is _with_ your fix (deleted my plugin install dir and replaced my plug definition with `Plug 'connorlay/ale', { 'branch': 'mix-format-umbrella-support' }`):


This is with `mix format` directly (works as expected):



Thanks again! :)

connorlay:
@dkarter After poking through the [source](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/format.ex), it appears that `mix format $my_file_path` determines which `.formatter.exs` files to include based on the structure of the workspace the path is a member of. When fixing a buffer, ALE creates a temporary file, e.g. `/tmp/nvimVusxoL/4/router.ex`, and fixes _just_ that file. This temporary file does not include the rest of the workspace needed for `mix format` to pick up the appropriate `.formatter.exs` files. This explains why manually invoking `mix format` in your example works as expected.

I attempted to solve this by removing the file path argument entirely, but this stopped the buffer from updating post-fix.  I found this thread https://github.com/dense-analysis/ale/issues/611 that discusses a similar issue with "workspace aware" fixers. It seems that ALE, by design, does not really support "workspace aware" fixers.

dkarter:
@connorlay Thank you so much for taking the time to dig into this!!

That makes sense - other formatter plugins behave similarly (neoformat for example). I have been using vim-mix-format for the past few days and it is working well for me too.

I've actually had a similar issue with prettier recently not picking up workspace specific configurations, now that you verified ALE's limitations I might actually stop using it for formatting altogether. Maybe if I have time in a month or so, I'll look into forking vim-mix-format and making it into a more generic "workspace aware" formatting plugin.

Thanks again for all your efforts on this! :)

connorlay:
Closing in favor of https://github.com/mhinz/vim-mix-format for this use-case.

PR #2918, opened 2019-12-03
Fix tslint fixer not work issue (temporary file not found)

allex:
Fix tslint fixer not work issue (temporary file not found)

PR #3107, opened 2020-04-15
Fix tslint fixer not working issue (temporary file not found)

daliusd:
This is the same pull request as https://github.com/dense-analysis/ale/pull/2918 but with tests fixed.


daliusd:
@w0rp could you or someone review this one?

w0rp:
I've merged this now. It's worth noting that `tslint` is deprecated, and everyone will eventually switch to using `eslint` instead.

PR #3114, opened 2020-04-17
Broken message in pycodestyle (or any other Linters whose name include 'code')

aiotter:
When you `let g:ale_echo_msg_format = '[%severity%] %s (%linter%% code%)'` and open python script with pycodestyle available in ale, the echoed message is substituted in wrong way like `[Error] s (pyE501style code%)`.

This is because the pycodestyle includes 'code' in its name and it is substituted like this way:
1. ` [%severity%] %s (%linter%% code%)`
2. ` [Error] %s (%linter%% code%)` <- %severity% substitution
3. ` [Error] %s (pycodestyle% code%)` <- %linter% substitution
4. ` [Error] s (pyE501style code%)` <- %...code...% substitution

This PR resolves the problem by swapping the substitution order like this way:
1. ` [%severity%] %s (%linter%% code%)`
2. ` [%severity%] %s (%linter% E501)` <- %...code...% substitution
3. ` [Error] %s (%linter% E501)` <- %severity% substitution
4. ` [Error] %s (pycodestyle E501)` <- %linter% substitution
5. ` [Error] line too long (82 > 79 characters) (pycodestyle E501)` <- %s% substitution

### DIFF
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -258,9 +258,9 @@ function! ale#GetLocItemMessage(item, format_string) abort
 
     " Replace special markers with certain information.
     " \=l:variable is used to avoid escaping issues.
+    let l:msg = substitute(l:msg, '\v\%([^\%]*)code([^\%]*)\%', l:code_repl, 'g')
     let l:msg = substitute(l:msg, '\V%severity%', '\=l:severity', 'g')
     let l:msg = substitute(l:msg, '\V%linter%', '\=l:linter_name', 'g')
-    let l:msg = substitute(l:msg, '\v\%([^\%]*)code([^\%]*)\%', l:code_repl, 'g')
     " Replace %s with the text.
     let l:msg = substitute(l:msg, '\V%s', '\=a:item.text', 'g')
 
--- a/test/test_cursor_warnings.vader
+++ b/test/test_cursor_warnings.vader
@@ -75,6 +75,17 @@ Before:
   \       'type': 'E',
   \       'text': 'lowercase error',
   \     },
+  \     {
+  \       'lnum': 3,
+  \       'col': 5,
+  \       'bufnr': bufnr('%'),
+  \       'vcol': 0,
+  \       'linter_name': 'fakecodelinter',
+  \       'nr': -1,
+  \       'type': 'E',
+  \       'code': 'E001',
+  \       'text': 'This error does not exist',
+  \     },
   \   ],
   \ },
   \}
@@ -236,6 +247,14 @@ Execute(The %code% and %ifcode% should be removed when there's no code):
   call ale#cursor#EchoCursorWarning()
 
   AssertEqual 'Some information', GetLastMessage()
+
+Execute(The %code% should be formatted correctly when other variables contain 'code'):
+  let g:ale_echo_msg_format = '%s (%linter%% code%)'
+
+  call cursor(3, 5)
+  call ale#cursor#EchoCursorWarning()
+
+  AssertEqual 'This error does not exist (fakecodelinter E001)', GetLastMessage()
 
 Execute(The buffer message format option should take precedence):
   let g:ale_echo_msg_format = '%(code) %%s'


RyanSquared:
Instead of doing this is there a specific reason we actually accept this instead of just replacing `%code%`? Looking into it now.

RyanSquared:
I believe it was introduced in this commit: https://github.com/dense-analysis/ale/commit/584e0bc7f25563bf4ab3ae738b78d9d13a898f94#diff-980991cf1cdfc75ce97aaed938840d05

I see no specific reason to why it used a different format.

aiotter:
Unlike `severity` and `linter`, `code` has a chance to have no value. I think that is why `code` has different format.
If the `code` has the same format, you have less flexibility on how to show the message.

For example, if you want to show the message in the following way, you cannot make it with the same format as `linter` or `severity`:
1. `[Error (code: X001)] description here` (code has a value)
2. `[Error] description here` (code has no value)

ipkiss42:
The problem here is that `[Error] %s (pycodestyle% code%)` contains the `%...code...%` sequence twice. The second one should be replaced, but the first one is replaced instead.

I don't think that changing the order of substitutions is a correct fix, because there is no way to guarantee that replacing `%s` does not introduce more `%` signs, potentially creating more issues later on. So it should really be done last.

I proposed a different fix in #3116. It fixes the issue by requiring word boundaries around `code` (see `:h /\<`).

aiotter:
Since the substitution for `%...code...%` is already done at the first substitution, you don't have to concern about `%s` generating additional `%...code...%` string. They won't be substituted twice.

aiotter:
Or are you concerning about the introduction of any other new substitution in `%...something...%` format in the future?

w0rp:
I think the bug fix is correct, we just need to add a test for it. Have a look at some existing tests and add a new one where the linter name includes the word `code`.

aiotter:
@w0rp Added test. I'm new to writing test for vimscript. Please let me know something is wrong.

ipkiss42:
@aiotter Sorry, I misread your proposal. I think it works fine as well.

The only situation where it might still go wrong is for the very unlikely case where the pattern itself contains `code` as part of a word (such as `decode`, or whatever). IMHO it doesn't hurt to add word boundaries in the regular expression, but I leave it up to you.

aiotter:
@ipkiss42 Ah, you're right. Patterns like `%s%: code - %code%` may be interpreted in a wrong way. This is a problem.

Introducing word boundaries is simpler solution, but there seems to be a potential bug in it. We cannot guarantee `%s` create another `%...code...%`. I mean, 'code' between two '%'.

aiotter:
I don't know why Travis CI makes a build error.
`./run-tests test/test_cursor_warnings.vader` in my environment makes no errors, although I have changed only `test_cursor_warnings.vader` since `c525db8`, which passes the entire tests.

I'm doing `./run-tests` to do all the tests, but it is taking too long on my laptop.

aiotter:
I've run `run_tests` and got following stdout. Now it seems stopped working, with CPU usage went quite low, however the process is still alive.
(I'm using Manjaro Linux with neovim and without vim8 installed, but it is not related to the problem as the test is running on docker, is it?)
```
❯ ./run-tests
Starting Vim: neovim-v0.2.0...
Starting Vim: neovim-v0.3.5...
Starting Vim: vim-v8.0.0027...
Starting Vim: vim-v8.1.0519...
Starting Vint...
Starting Custom checks...


Running tests for neovim-v0.2.0


Tried to run tests 5 times

```

ipkiss42:
> Introducing word boundaries is simpler solution, but there seems to be a potential bug in it. We cannot guarantee `%s` create no additional `%...code...%`. I mean, 'code' between two '%'.

I don't think it matters, because the `%s` part comes from the linter and we shouldn't mess with it. I guess this is the reason why it is currently substituted last.

aiotter:
Ah, you're right. I thought `%...code...%` comes the last in current implementation, but actually `%s%` is the last.

w0rp:
Cheers! :beers:

The tests didn't run properly on Travis, so I'll see what's up with that.

PR #3116, opened 2020-04-17
Do not match the 'code' keyword when part of of longer word

ipkiss42:
Fixes #3114

### DIFF
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -260,7 +260,7 @@ function! ale#GetLocItemMessage(item, format_string) abort
     " \=l:variable is used to avoid escaping issues.
     let l:msg = substitute(l:msg, '\V%severity%', '\=l:severity', 'g')
     let l:msg = substitute(l:msg, '\V%linter%', '\=l:linter_name', 'g')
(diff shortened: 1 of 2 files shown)


aiotter:
It looks a good approach. 
Just a reminder that you need to fix the doc as well.
https://github.com/dense-analysis/ale/blob/dc054f30c8356fecbe19e54aab60ff2587a6d71b/doc/ale.txt#L849-L879

ipkiss42:
To be clear, this commit does not introduce `<code>` instead of `code` as keyword (it wouldn't be backwards-compatible). It still uses `code`, but uses the `\<` and `\>` zero-width patterns to enforce word boundaries. Note that the backslashes are not there due to the `\v` in the regular expression.

So I don't think that the doc needs to be changed, but let me know if I missed something.

w0rp:
Thanks for the attempt! I went with #3114 in case someone wants to use some character around `code` that doesn't count as a boundary character. It should fix the issue.

aiotter:
According to the current doc, they can set the formatting variable `%Code_code%` to make it shown as `Code_E001`. In your implementation they cannot.
But it's rare case. You may have no need to specify it on doc.

aiotter:
Whoops I reloaded the browser to know w0rp has already closed the issue XD

PR #3121, opened 2020-04-18
linter/scala/metals: Fix return value of GetProjectRoot()

### LINKED ISSUE #3120: Wrong return type for ale_linters#scala#metals#GetProjectRoot()
## Information

**VIM version**

NVIM v0.4.3
Build type: Release

Operating System: macOS 10.15.3

## What went wrong

The function `ale_linters#scala#metals#GetProjectRoot(buffer)` would return `0` whenever an appropriate project root cannot be found.  The correct return value should be an empty string.

## Reproducing the bug

1. Enable Scala metals for ALE
2. Open a file in vim where no `.bloop`, `build.sbt`, `.metals`, `build.sc` are present.

### :ALEInfo

N/A

zoonfafer:
It was returning 0 when it should be returning an empty string.

The `AssertEqual` in the ale docker image is from an old version so it does not
check the types of the arguments.

This is already fixed in https://github.com/junegunn/vader.vim/commit/427fe19104c15066e4c1d5d385076e8e07a0dee8

Closes #3120

### DIFF
--- a/ale_linters/scala/metals.vim
+++ b/ale_linters/scala/metals.vim
@@ -32,6 +32,8 @@ function! ale_linters#scala#metals#GetProjectRoot(buffer) abort
             \)
         endif
     endfor
+
+    return ''
 endfunction
 
 function! ale_linters#scala#metals#GetCommand(buffer) abort


w0rp:
Cheers! :beers:

PR #3130, opened 2020-04-22
WIP: LSP: Support workspace/configuration

dotdash:
According to
https://github.com/microsoft/language-server-protocol/issues/676 the
didChangeConfiguration method only triggers a workspace/configuration
from the server, which the client needs to handle, so effectively the
current configuration support in ALE is not working at all with modern
LSP servers.

This is a work-in-progress to add support for workspace/configuration,
currently only hacked together far enough that it works for my use with
rust-analyzer. I hope to extend it / clean it up, but if anyone wants to
pick up from here, feel free to.

### DIFF
--- a/autoload/ale/lsp.vim
+++ b/autoload/ale/lsp.vim
@@ -143,6 +143,19 @@ function! ale#lsp#CreateMessageData(message) abort
     return [l:is_notification ? 0 : l:obj.id, l:data]
 endfunction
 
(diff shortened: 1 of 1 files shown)


dotdash:
Uhm, ok, I made this a draft PR for a reason, and editing the comment about tests into my PR message instead of adding a standalone comment does feel pretty bad (also it's rather lucky that I noticed it, because I don't get a notification that way).

Guess I'm done with this PR then...

PR #3134, opened 2020-04-24
Add shellcheck as linter for bats files

### LINKED ISSUE #3133: Add support for running the shellcheck linter on Bats test files
**Linter:** Shellcheck
**URL:** https://www.shellcheck.net/
**Filetype:** Bats
**URL:** https://github.com/bats-core/bats-core

Shellcheck is already an ale linter for shell files but has support for [Bats](https://github.com/bats-core/bats-core) files too. Bats is a testing framework for bash. Please add shellcheck as a linter for the 'bats' filetype.

Ian2020:
Fixes #3133 

I will add tests if anybody else can confirm my approach is the right way to go. Is there a better way to reuse the shellcheck linter from 'sh' for 'bats' files? I have duplicated some code, the handle function in particular which is not nice. Also the only real difference in the GetCommand function is that I hardcode the dialect to be 'bats' for shellcheck (line 25)

The other way to get linting on bats files is to change their filetype to 'sh' however not all bash linters recognise bats syntax. And we lose the ability to do different things with bats files like different rule exclusions and anything else a user may want to do with that filetype. Seems it deserves its own settings.


hsanson:
@Ian2020 you can move any shared code to `autoload/ale/handlers/` folder, for example to a `shellcheck.vim` file.  Then use the functions defined there in each filetype linter definition.

Check the `ktlint` linter for simple example and all `eslint` based linters for more complex one.

Ian2020:
Thanks - I'll rework what I've done in line with that.

Ian2020:
I am still looking at this in my spare time - just working out how to convert linter tests to handler tests now the code has moved.

Ian2020:
All done, tests pass. Can someone review please? Thanks.

Ian2020:
Yeah I couldn't understand why the vars didn't work where they were. I looked at ktlint and eslint but couldn't wrangle the code to work with the existing shellcheck tests. In the end I found the alex linter code and copied its approach.

PR #3137, opened 2020-04-25
Fix eclipselsp command function.

hsanson:
Changes in eclipselsp and java caused the command needed to run the
language server fail to build properly.

Thi PR fixes those issues.


w0rp:
It's very hard for the test suite to test this, so I just merged it.

PR #3144, opened 2020-04-30
don't append a newline to temp file when buffer is noeol and nofixeol is set

jamescdavis:
The `Writefile` util will append a newline to the end of the last line in the buffer, even if it is `noeol` and `nofixeol` is set. This breaks linting for `noeol` (temp file will always have `eol`). This PR fixes the issue by calling `writefile` with the binary flag when the buffer is `noeol` and `nofixeol` is set, thus preserving the `noeol`.


jamescdavis:
@w0rp, @hsanson, stale bot says to bother you :wink:

w0rp:
Nice addition. Cheers! :beers:

Issue #2389, opened 2019-03-24
Add support for xo + typescript


nikersify:
**Name:** xo/typescript
**URL:** https://github.com/xojs/xo

[If configured correctly](https://github.com/xojs/eslint-config-xo-typescript#use-with-xo) `xo` can be used to lint typescript files, in addition to javascript ones.

The implementation should be pretty much the same as in `ale_linters/javascript/xo.vim`, although I've tried blindly copying that file to `ale_linters/typescript/xo.vim` replacing `javascript` with `typescript` but that unfortunately turned out to be too silly to work:

![image](https://user-images.githubusercontent.com/7091471/54886784-0bb04480-4e8c-11e9-9195-e60aad49923b.png)
(output after running `:ALEInfo`)

Thanks in advance!

lo1tuma:
I’ve ran into the same issue today. Enabling xo for typescript would be awesome.

chaucerbao:
This was easy enough. I did the same thing you did, @nikersify; copied the file and replaced references to JavaScript with TypeScript. PR available at #2453.

Running **xo** as a fixer took a few seconds, but that's probably an issue with **xo** itself.

hsanson:
ALE already supports XO linter/fixer for typescript (see #2453).

PR #3148, opened 2020-04-30
fixers/xo: enhance `xo` fixer

charlesbjohnson:
👋 

This is primarily directed at upgrading the `xo` fixer. From a feature perspective, a few things have been done:

1. The fixer uses now supports `let g:ale_javascript_xo_options` for custom flags in JS files
2. The fixer uses now supports `let g:ale_typescript_xo_options` for custom flags in TS files
3. The fixer now supports stdin by virtue of `--stdin --stdin-filename` (fully usable in [`0.30.0`](https://github.com/xojs/xo/releases/tag/v0.30.0), though technically available before then)

The diff is a bit on the larger side, though, in order to consolidate some of the duplication between the existing linters and the fixer through a shared handler file. I've tried to make changes one-step-at-a-time to make review easier.

In terms of testing, I've personally been using these changes for the past several months. The only issue I noticed was a [minor bug with XO itself](https://github.com/xojs/xo/issues/459), which has recently been fixed.

Some issues that this work relates to:
- #2389: I think this one is already done?

### DIFF
--- a/autoload/ale/fixers/xo.vim
+++ b/autoload/ale/fixers/xo.vim
@@ -1,20 +1,9 @@
 " Author: Albert Marquez - https://github.com/a-marquez
 " Description: Fixing files with XO.
 
-call ale#Set('javascript_xo_executable', 'xo')
-call ale#Set('javascript_xo_use_global', get(g:, 'ale_use_global_executables', 0))
-call ale#Set('javascript_xo_options', '')
-
-function! ale#fixers#xo#GetExecutable(buffer) abort
-    return ale#node#FindExecutable(a:buffer, 'javascript_xo', [
-    \   'node_modules/xo/cli.js',
-    \   'node_modules/.bin/xo',
-    \])
-endfunction
-
 function! ale#fixers#xo#Fix(buffer) abort
-    let l:executable = ale#fixers#xo#GetExecutable(a:buffer)
-    let l:options = ale#Var(a:buffer, 'javascript_xo_options')
+    let l:executable = ale#handlers#xo#GetExecutable(a:buffer, 'javascript')
+    let l:options = ale#handlers#xo#GetOptions(a:buffer, 'javascript')
 
     return {
     \   'command': ale#node#Executable(a:buffer, l:executable)
--- a/autoload/ale/fixers/xo.vim
+++ b/autoload/ale/fixers/xo.vim
@@ -2,17 +2,8 @@
 " Description: Fixing files with XO.
 
 function! ale#fixers#xo#Fix(buffer) abort
-    let l:filetype = getbufvar(a:buffer, '&filetype')
-    let l:type = ''
-
-    if l:filetype =~# 'javascript'
-        let l:type = 'javascript'
-    elseif l:filetype =~# 'typescript'
-        let l:type = 'typescript'
-    endif
-
-    let l:executable = ale#handlers#xo#GetExecutable(a:buffer, l:type)
-    let l:options = ale#handlers#xo#GetOptions(a:buffer, l:type)
+    let l:executable = ale#handlers#xo#GetExecutable(a:buffer)
+    let l:options = ale#handlers#xo#GetOptions(a:buffer)
 
     return ale#semver#RunWithVersionCheck(
     \   a:buffer,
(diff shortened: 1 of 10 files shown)


autoload/ale/handlers/xo.vim
```diff
@@ -0,0 +1,35 @@
...
+endfunction
+
+function! ale#handlers#xo#GetOptions(buffer, type) abort
+    return ale#Var(a:buffer, a:type . '_xo_options')
+endfunction
+
+" xo uses eslint and the output format is the same
+function! ale#handlers#xo#HandleJSON(buffer, lines) abort
+    return ale#handlers#eslint#HandleJSON(a:buffer, a:lines)
+endfunction
```
charlesbjohnson, line 29: This layer of indirection isn't really necessary, but my thinking is that if this has to change at some point that it would easier to update it once here rather than in every XO related linter/fixer.

autoload/ale/handlers/xo.vim
```diff
@@ -0,0 +1,35 @@
...
+
+function! ale#handlers#xo#GetProjectRoot(buffer) abort
+    let l:package_path = ale#path#FindNearestFile(a:buffer, 'package.json')
+
+    return empty(l:package_path) ? '' : fnamemodify(l:package_path, ':p:h')
+endfunction
```
charlesbjohnson, line 35: Admittedly I haven't tested this much in a [monorepo](https://classic.yarnpkg.com/blog/2017/08/02/introducing-workspaces/) setup, so I expect that this might not be exactly what is needed but it [might still be fine](https://github.com/xojs/xo#using-a-parents-config).

charlesbjohnson: Changes this to search for `node_modules` the same way `handlers/eslint.vim` does:

https://github.com/dense-analysis/ale/blob/834d6f9c48ce349a4c3d72e98bd0de81aaa84042/autoload/ale/handlers/eslint.vim#L45-L53

charlesbjohnson: Ended up walking this back because it isn't necessary. AFAICT [`cosmiconfig`](https://github.com/davidtheclark/cosmiconfig) (which `xo` is using) is already doing this.

autoload/ale/handlers/xo.vim
hsanson: I also prefer this kind of indirection as long as performance is not an issue. :+1:

autoload/ale/handlers/xo.vim
hsanson: :+1:

autoload/ale/fixers/xo.vim
```diff
@@ -1,23 +1,45 @@
...
+    let l:filetype = getbufvar(a:buffer, '&filetype')
+    let l:type = ''
+
+    if l:filetype =~# 'javascript'
+        let l:type = 'javascript'
+    elseif l:filetype =~# 'typescript'
+        let l:type = 'typescript'
+    endif
+
+    let l:executable = ale#handlers#xo#GetExecutable(a:buffer, l:type)
```
hsanson, line 14: Overall this PR looks really good. Tests are extensive, it uses version check for backward compatibility, and comments explaining why some things are done certain way. 

Just wondering if it is really necessary to pass l:type to GetExecutable(), GetOptions(), etc when this information can be easily extracted from the buffer argument that is also passed in to them.

Instead of passing l:type all over the place I think is better to simply use `getbufvar(a:buffer, '&filetype')` to get this information within those methods directly.

charlesbjohnson: Done.

Calculating `l:type` once and passing was a minor performance optimization. We can't rely on `getbufvar(a:buffer, '&filetype')` alone, though, given that the filetypes we're interested in are:

- `javascript`
- `javascriptreact`
- `typescript`
- `typescriptreact`

But we're limiting options to just `javascript_xo_...` and `typescript_xo_...`.

hsanson (dismissed):
Good to merge. Waiting Travis issue #3548 to get solved to be able to merge.

hsanson:

```
# needed if already added.

git checkout master
git fetch upstream master
git merge upstream/master

git checkout my-branch-name

```

charlesbjohnson:
Done, thanks for your support @hsanson.

hsanson (approved):
Good to go. Thanks.

Issue #2831, opened 2019-10-13
Support compile_flags.txt with clang-based tools


sam-lunt:
Clang-based tools also support a file `compile_flags.txt`, which is a simpler alternative to a full `compile_commands.json` file (see [here](https://clang.llvm.org/docs/JSONCompilationDatabase.html#alternatives)). However, the [`ale#c#FindProjectRoot`](https://github.com/dense-analysis/ale/blob/b8949aaac33e152b195d3adf11ddbe863c0feef1/autoload/ale/c.vim#L190) function does not attempt to locate such a file, since [`ale#c#FindCompileCommands`](https://github.com/dense-analysis/ale/blob/b8949aaac33e152b195d3adf11ddbe863c0feef1/autoload/ale/c.vim#L161) does not look for `compile_flags.txt`.

This seems like an easy feature to add, and I'd be happy to do it. The only question I had was whether this should be made in `FindCompileCommands` or in `FindProjectRoot` (or neither). Changing `FindCompileCommands` would require the smallest diff, but it would also make the name slightly misleading. It also might break some other user of the function.

The main impetus for this is that it doesn't seem like `clangd` will start if ALE cannot find a `compile_commands.json` file to use, even if there is a `compile_flags.txt` file available. (Also, `clangd` can provide useful results even without a compilation database present. Is there a way to force it to be started?)

w0rp:
One way you make sure any language server is started is to configure `ale_lsp_root` for your project to indicate where the project root is.

proxict:
I second this request. It would possibly be enough to just make the [`ale_c_project_filenames` variable](https://github.com/dense-analysis/ale/blob/master/autoload/ale/c.vim#L9) configurable. Right now, the workaround for this issue is to create one of the files from the list (`'.git/HEAD', 'configure', 'Makefile', 'CMakeLists.txt'`) as a dummy file.

proxict:
I just opened #3165 which should address this issue.

w0rp:
I would like to see a PR for parsing `compile_flags.txt`, which should be similar to parsing `command` keys from `compile_commands.json`. We could make our lives easy by reading the lines of output and converting them into the same JSON Dictionaries we get from `compile_commands.json`. We should always prefer `compile_commands.json` over `compile_flags.txt`, because the quoting of arguments will be better if the `arguments` keys are used in output, say from [Bear](https://github.com/rizsotto/Bear).

PR #3165, opened 2020-05-10
Make C/C++ project files configurable

proxict:
New global variable `g:ale_c_project_filenames` allows users to configure which files trigger a directory as a C/C++ project root.

Should address https://github.com/dense-analysis/ale/issues/2831

### DIFF
--- a/autoload/ale/c.vim
+++ b/autoload/ale/c.vim
@@ -5,8 +5,12 @@ call ale#Set('c_parse_makefile', 0)
 call ale#Set('c_parse_compile_commands', 0)
 let s:sep = has('win32') ? '\' : '/'
 
-" Set just so tests can override it.
-let g:__ale_c_project_filenames = ['.git/HEAD', 'configure', 'Makefile', 'CMakeLists.txt']
+let g:ale_c_project_filenames = get(g:, 'ale_c_project_filenames', [
+\   '.git/HEAD',
+\   'configure',
+\   'Makefile',
+\   'CMakeLists.txt',
+\])
 
 function! ale#c#GetBuildDirectory(buffer) abort
     " Don't include build directory for header files, as compile_commands.json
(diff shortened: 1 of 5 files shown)


proxict:
Can somebody review this please?

autoload/ale/c.vim
```diff
@@ -192,7 +196,7 @@ function! ale#c#FindProjectRoot(buffer) abort
 
     " Fall back on detecting the project root based on other filenames.
     if empty(l:root)
-        for l:project_filename in g:__ale_c_project_filenames
+        for l:project_filename in g:ale_c_project_filenames
```
w0rp, line 199: The documentation says you can use `b:ale_c_project_filenames`, so make that true.

```suggestion
        for l:project_filename in ale#Var(a:buffer, 'c_project_filenames')
```

autoload/ale/c.vim
```diff
@@ -5,8 +5,12 @@ call ale#Set('c_parse_makefile', 0)
 call ale#Set('c_parse_compile_commands', 0)
 let s:sep = has('win32') ? '\' : '/'
 
-" Set just so tests can override it.
-let g:__ale_c_project_filenames = ['.git/HEAD', 'configure', 'Makefile', 'CMakeLists.txt']
+let g:ale_c_project_filenames = get(g:, 'ale_c_project_filenames', [
+\   '.git/HEAD',
+\   'configure',
+\   'Makefile',
+\   'CMakeLists.txt',
```
w0rp, line 12:
```suggestion
\   'CMakeLists.txt',
\   'compile_flags.txt',
```

Let's add this here too, and write the same in the documentation for the option, so we do a little bit better project detection by default. You may have to break the default over many lines if it's long. You can use the hidden characters `>` and `<` in documentation to write multi-line code blocks. Search with `/>` and you'll find an example somewhere in the documentation.

w0rp (changes requested):
This is a good idea.

This doesn't address #2831 totally. I would like to see another PR for parsing `compile_flags.txt`, which should be similar to parsing `command` keys from `compile_commands.json`. We could make our lives easy by reading the lines of output and converting them into the same JSON Dictionaries we get from `compile_commands.json`. We should always prefer `compile_commands.json` over `compile_flags.txt`, because the quoting of arguments will be better if the `arguments` keys are used in output, say from [Bear](https://github.com/rizsotto/Bear).

PR #2782, opened 2019-09-20
Fix 2777 - Add IBM openapi validator

hsanson:
Adds IBM openapi validator to the ever growing list of ALE linters. This allows us to edit open api specs in yaml within vim/neovim with confidence.

Closes #2777

PR #3172, opened 2020-05-14
More docs for beginners to use cloudformation linter

0xMH:
I added more info to cloudformation linter docs cause its hard to get it to work for beginners because it won't work out of the box. I made the easiest changes that I could think above. We can make it more robust using  a func to detect cloudformation files based on its contents. Will add it later if this PR get through.

### DIFF
--- a/doc/ale-cloudformation.txt
+++ b/doc/ale-cloudformation.txt
@@ -8,7 +8,22 @@ cfn-python-lint                             *ale-cloudformation-cfn-python-lint*
 cfn-python-lint is a linter for AWS CloudFormation template file.
 
 https://github.com/awslabs/cfn-python-lint
+===============================================================================
+Getting it to work:                            *ale-cloudformation-activate*
 
+To get cloudformation linter to work on only CloudFormation files put 
+
+ au BufRead,BufNewFile *.template.yaml set filetype=yaml.cloudformation
+
+on `ftdetect/cloudformation.vim`
+
+and 
+
+let b:ale_linter_aliases = {'cloudformation': ['cloudformation', 'yaml']}
+
+on `after/ftplugin/yaml.vim`
+
+This will get both cloudformation and yaml linters on work on any file with `template.yaml` ext.
 ===============================================================================
   vim:tw=78:ts=2:sts=2:sw=2:ft=help:norl:
 
--- a/doc/ale-cloudformation.txt
+++ b/doc/ale-cloudformation.txt
@@ -7,23 +7,40 @@ cfn-python-lint                             *ale-cloudformation-cfn-python-lint*
 
 cfn-python-lint is a linter for AWS CloudFormation template file.
 
-https://github.com/awslabs/cfn-python-lint
-===============================================================================
-Getting it to work:                            *ale-cloudformation-activate*
+Website: https://github.com/awslabs/cfn-python-lint
 
-To get cloudformation linter to work on only CloudFormation files put 
+Installation
+-------------------------------------------------------------------------------
+
+
+Install cfn-python-lint using either pip or brew: >
+
+`pip install cfn-lint`. If pip is not available, run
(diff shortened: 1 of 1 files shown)


hsanson:
This is interesting and really useful. I applied the same concept to openapi files that can be written in YAML or JSON (see #2782 ).

Just a note: setting the filetype to `cloudformation.yaml` causes ALE to execute both cloudfront and yaml configured linters without need to set `b:ale_linter_aliases`.

0xMH:
> Just a note: setting the filetype to cloudformation.yaml causes ALE to execute both cloudfront and yaml configured linters without need to set b:ale_linter_aliases.

Didn't know that, Will Change it thanks! 
What do you think of using a func like this? It sets the filetype based on the file contents.
```
 function! DetectCfn(type)
     let likely = 0
     let pointsRequired = 10

     " A mapping of all the important words in a CloudFormation template to the
     " number of points they're worth when detecting a file type. The values
     " were chosen fairly arbitrarily, but the section headers are worth 1
     " point, intrinsic functions are worth 2, and pseudo parameters are worth
     " 4. AWSTemplateFormatVersion is used as a sure sign its a Cfn template,
     " and AWS::\w+::\w+ is given 5 points since you're specifying resources
     " using the CloudFormation name.
     let pointMap = [
         \['AWSTemplateFormatVersion', 100],
         \['\vAWS::\w+::\w+', 5],
         \['AWS::AccountId', 4],
         \['AWS::NotificationARNs', 4],
         \['AWS::NoValue', 4],
         \['AWS::Partition', 4],
         \['AWS::Region', 4],
         \['AWS::StackId', 4],
         \['AWS::StackName', 4],
         \['AWS::URLSuffix', 4],
         \['Fn::Base64', 2],
         \['!Base64', 2],
         \['Fn::Cidr', 2],
         \['!Cidr', 2],
         \['Fn::FindInMap', 2],
         \['!FindInMap', 2],
         \['Fn::GetAZs', 2],
         \['!GetAZs', 2],
         \['Fn::ImportValue', 2],
         \['!ImportValue', 2],
         \['Fn::Join', 2],
         \['!Join', 2],
         \['Fn::Select', 2],
         \['!Select', 2],
         \['Fn::Split', 2],
         \['!Split', 2],
         \['Fn::Sub', 2],
         \['!Sub', 2],
         \['Fn::Transform', 2],
         \['!Transform', 2],
         \['!Ref', 2],
         \['Description', 1],
         \['Metadata', 1],
         \['Parameters', 1],
         \['Mappings', 1],
         \['Conditions', 1],
         \['Transform', 1],
         \['Resources', 1],
         \['Outputs', 1],
         \]
     for lineContents in getline(1, line('$'))
         for strPoints in pointMap
             if lineContents =~ strPoints[0]
                 let likely += strPoints[1]
             endif
             if likely > pointsRequired
                 if a:type =~ "yaml"
                     set filetype=yaml.cloudformation
                 elseif a:type =~ "json"
                     set filetype=json.cloudformation
                 endif
                 return
             endif
         endfor
     endfor
 endfunction

 augroup filetypedetect
     au BufRead,BufNewFile *.yaml,*.yml call DetectCfn('yaml')
     au BufRead,BufNewFile *.json call DetectCfn('json')
     au BufNewFile,BufRead *.template setfiletype yaml.cloudformation
 augroup END
```

hsanson:
I am not familiar with Cloudformation file format so I can only provide generic advice:

 1. Usually you don't want to inspect the whole file to determine the filetype. Specially if you expect cloudformation files to be large or to open with many of them open at the same time.
From what I can see all these files will have `AWSTemplateFormatVersion:` key at the beginning so looking for it within the first 10~20 lines should be enough to determine if the file is cloudfront template or not. Is better to look for one or two strings that are mandatory, that is, if they are not present then linter validation will fail and look for them.

2. To be on the safe side I would also use the detect function on *.template files. There seems to be other files that use that extension already.

PR #3173, opened 2020-05-15
feat(javac): Add java_javac_sourcepath variable

nibocn:
This variable can set multiple source code paths, the source code path is a relative path (relative to the project root directory)


nibocn:
> The classpath variable works with a string instead of an array. For consistency I think would be good to make this new but related variable too a string or even better make both variables `classpath` and `sourcepath` support both strings and arrays.

Okay, I will adjust the `sourcepath` variable to string later

hsanson:
@nibocn Are you planning more changes? if not I can merge this change.

nibocn:
@hsanson That's all, thanks.

PR #3178, opened 2020-05-21
Fixes #3092 - Implement loading `@file` c arguments

### LINKED ISSUE #3092: C/C++: should add support for @file arguments
## Background
GCC and Clang allow you to place command line flags together in a file, and include all of those flags as if they were on the command line by using a `@file` flag.

For example, suppose I had the file `./debug_args` with the content:
```
-DDEBUG
-O0
-g
```
and I ran gcc like this:
```
gcc @debug_args -o main main.c
```
It would be equivalent to running:
```
gcc -DDEBUG -O0 -g -o main main.c
```
## Request

Currently, ALE will filter out `@file` flags. I would like them to be properly handled by ALE.

I came across this recently when compiling a project which used `@file` flags. I used [bear](https://github.com/rizsotto/Bear) to generate a `compile_commands.json` file, but had issues because ALE was filtering out the `@file` flag `gcc` needed.

sudobash1:
This is a fix for the bug #3092 I filed on expanding `@file` c arguments.

Just as background, a `@file` argument works like this: If `gcc` or `clang` encounters a flag which starts with `@` it will treat the remainder of that flag as a file path, and it will use the contents of that file as flags. (It is somewhat like an "include file" for command line arguments).

Initially I implemented this as simply passing the `@file` arguments directly to the compiler, but then I realized that this could cause problems since the arguments contained with the file would not get parsed and filtered. If you are curious this implementation is [here](https://github.com/sudobash1/ale/commit/798eb57b5f6728fb6556bf9051b404b7dfda54ef).

This PR instead has vim read the contents of the `@file` and will add those to the argument list to be filtered.

I have added a Vader test for this.

Thanks for you work!


w0rp (approved):
I say we just give this a shot and see if anyone notices any issues. I can't think of a better way to handle this.

w0rp:
Cheers! :beers:

w0rp:
If we end up with someone having issues with `readfile()` calls being too slow, we have the option later of storing the result of `readfile()` for `@file` arguments in a Dictionary, so we don't read the file from disk all the time.

sudobash1:
I think this is my first open-source PR to get merged. Thanks for your work on this project @w0rp.

PR #3190, opened 2020-05-23
OmniSharp C# linter over LSP

luizribeiro:
This sets up a linter for C# with [OmniSharp](https://github.com/OmniSharp/omnisharp-roslyn) over LSP.

I wrote documentation explaining how to use it and added some tests too.

Been using this successfully with C# projects with OmniSharp 1.32.5 on macOS. See [luizribeiro/mgsb](https://github.com/luizribeiro/mgsb) for an example of a project.

### DIFF
--- a/ale_linters/cs/omnisharp.vim
+++ b/ale_linters/cs/omnisharp.vim
@@ -1,0 +1,29 @@
+" Author: Luiz Ribeiro <luizribeiro@gmail.com>
+" Description: C# support via omnisharp
+
+call ale#Set('cs_omnisharp_mono_executable', 'mono')
+call ale#Set('cs_omnisharp_executable', 'OmniSharp.exe')
+call ale#Set('cs_omnisharp_options', '-lsp')
+
+function! ale_linters#cs#omnisharp#GetProjectRoot(buffer) abort
+    let l:omnisharp_json = ale#path#FindNearestFile(a:buffer, 'omnisharp.json')
+    return !empty(l:omnisharp_json) ? fnamemodify(l:omnisharp_json, ':h') : ''
+endfunction
+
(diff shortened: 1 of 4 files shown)


luizribeiro:
@w0rp, any reason why we wouldn't want to merge this?

ale_linters/cs/omnisharp.vim
```diff
@@ -0,0 +1,31 @@
+" Author: Luiz Ribeiro <luizribeiro@gmail.com>
+" Description: C# support via omnisharp
+
+call ale#Set('cs_omnisharp_mono_executable', 'mono')
+call ale#Set('cs_omnisharp_executable', 'OmniSharp.exe')
```
w0rp, line 5: Try using just `OmniSharp`. Because of `PATHEXT`, you should be able to run it without `.exe` in the filename on Windows, if it's in `PATH`.

ale_linters/cs/omnisharp.vim
```diff
@@ -0,0 +1,31 @@
...
+
+function! ale_linters#cs#omnisharp#GetMonoExecutable(buffer) abort
+    return ale#Var(a:buffer, 'cs_omnisharp_mono_executable')
+endfunction
+
+function! ale_linters#cs#omnisharp#GetCommand(buffer) abort
+    let l:executable = ale#Var(a:buffer, 'cs_omnisharp_executable')
+    let l:options = ale#Var(a:buffer, 'cs_omnisharp_options')
+
+    return '%e ' . l:executable . ' ' . l:options
```
w0rp, line 22: Apply `ale#Escape` to paths or executables in commands.

```suggestion
    return '%e ' . ale#Escape(l:executable) . ' ' . l:options
```

PR #3193, opened 2020-05-29
Use old solc error reporter

karlb:
Solc switched to a new error reporter with
https://github.com/ethereum/solidity/releases/tag/v0.6.0

Use a flag to bring back the old reporting.


karlb:
@w0rp Without this PR, you can't use any recent solc version with ale. Is there a reason against merging this?

PR #3215, opened 2020-06-19
psscriptanalyzer: Add support for configurable options

kevinoid:
For parity with many other linters, and to support my current desire for specifying `-Settings`, add support for configurable command options via the `ale_powershell_psscriptanalyzer_options` variable.

Note: This change is not covered by tests because I could not find any existing tests for `ale_linters#powershell#psscriptanalyzer#GetCommand`.  If tests are a hard requirement, can we discuss removing the temporary script generation and passing arguments directly to make it easier to test?

Thanks,
Kevin


kevinoid:
@w0rp I think this PR is good to merge and not stale.  (Unless you want to consider major changes to psscriptanalyzer to support testing.)

PR #3219, opened 2020-06-27
Add integration for tremor's languages

anupdhml:
Tremor is an event processing system for unstructured data with rich support for structural pattern matching, filtering and transformation.

For more details on tremor, see:
https://www.tremor.rs/

Current tremor languages:
https://docs.tremor.rs/tremor-script/
https://docs.tremor.rs/tremor-query/


anupdhml:
@w0rp this change has been very useful for our vim users and would be nice to have this merged so that we don't have to keep maintaining ale fork and pointing folks towards it to use our integration. 

Also, thanks for writing ale -- I have been using it daily for quite a while now and it works great!

Issue #3143, opened 2020-04-29
ESLint doesn't work as a fixer when ESLint used with plugins and run in a parent directory of the project


jgonera:
## Information

```
NVIM v0.4.3
Build type: Release
```

Operating System: macOS Catalina 10.15.2 (19C57)

## What went wrong
When running ESLint as a fixer ALE is not cd-ing into the file's directory like it does when it runs ESLint as a linter:

```bash
# As a linter
(finished - exit code 1) ['/bin/zsh', '-c', 'cd ''/Users/juliusz/repos/ale-bug/project'' && ''/Users/juliusz/repos/ale-bug/project/node_modules/eslint/bin/eslint.js'' -f json --stdin --stdin-filename ''/Users/juliusz/repos/ale-bug/project/src/index.js'' < ''/var/folders/x6/bcf_tsv923nbj4mxn67vhwbh0000gn/T/nvimbxfwpz/7/index.js''']

# As a fixer
(finished - exit code 2) ['/bin/zsh', '-c', '''/Users/juliusz/repos/ale-bug/project/node_modules/eslint/bin/eslint.js'' --stdin-filename ''/Users/juliusz/repos/ale-bug/project/src/index.js'' --stdin --fix-dry-run --format=json < ''/var/folders/x6/bcf_tsv923nbj4mxn67vhwbh0000gn/T/nvimbxfwpz/5/index.js''']
```

In such case ESLint is run in Vim's `:pwd`, in my case `/Users/juliusz/repos/ale-bug`, while it should run in `/Users/juliusz/repos/ale-bug/project` which is where `.eslintrc.js` and `node_modules` are. This results in ESLint not being able to find the plugins and exiting with status 2. Doing `:cd project` makes ESLint work as a fixer.

If I try running ESLint manually while in `/Users/juliusz/repos/ale-bug` it outputs this (this output doesn't show up in `:ALEInfo` since it's stderr):

```
$ /bin/zsh -c '/Users/juliusz/repos/ale-bug/project/node_modules/eslint/bin/eslint.js --stdin-filename /Users/juliusz/repos/ale-bug/project/src/index.js --stdin --fix-dry-run --format=json < /Users/juliusz/repos/ale-bug/project/src/index.js'

Oops! Something went wrong! :(

ESLint: 6.8.0.

ESLint couldn't find the plugin "eslint-plugin-node".

[...]

PR #3222, opened 2020-07-02
eslint: Use cwd from executable location to fix nested projects

kevinoid:
Since  #2937, ESLint is run from the nearest directory with `node_modules`.  This does not work correctly for nested projects where the eslint and any required config/plugin packages are installed in the outer project.  For example: https://github.com/dense-analysis/ale/issues/3143#issuecomment-652452362

Adopt the behavior of SublimeLinter, which runs from `project_root` determined by the presence of the `eslint` executable in `node_modules/.bin` (or `eslint` in `dependencies`/`devDependencies` of `package.json`, which we can add later as necessary).  See SublimeLinter/SublimeLinter#1649 and [`NodeLinter#find_local_executable`](https://github.com/SublimeLinter/SublimeLinter/blob/056e6f6/lint/base_linter/node_linter.py#L109).

Thanks for considering,
Kevin

cc: @w0rp, @joeldodge79

### DIFF
--- a/autoload/ale/handlers/eslint.vim
+++ b/autoload/ale/handlers/eslint.vim
@@ -1,6 +1,11 @@
 " Author: w0rp <devw0rp@gmail.com>
 " Description: Functions for working with eslint, for checking or fixing files.
 
+let s:executables = [
+\   'node_modules/.bin/eslint_d',
+\   'node_modules/eslint/bin/eslint.js',
+\   'node_modules/.bin/eslint',
+\]
 let s:sep = has('win32') ? '\' : '/'
 
 call ale#Set('javascript_eslint_options', '')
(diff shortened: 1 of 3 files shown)


joeldodge79:
this works perfectly for my setup in https://github.com/dense-analysis/ale/issues/3143#issuecomment-652452362 

thanks a ton for working on this!

w0rp:
Cheers! :beers:

kevinoid:
Thanks for reviewing and merging @w0rp!  :clinking_glasses:

jakebrinkmann:
I am having a problem which is related to this MR. Should I open a new issue, or ask the questions here? 

In summary, running eslint from my source code directory is desired, but this MR introduced changing directory to the location of eslint (thereby using the wrong configuration, instead of my projects eslint configuration)

PR #3227, opened 2020-07-06
Update error-parser for `nix-instantiate` on Nix 2.4+

Ma27:
In the (unreleased) Nix 2.4 the error-messages have been reformatted[1].
This patch aims to retain proper `.nix`-support in `ale`, for both
stable Nix (2.3 and older) and unstable Nix (2.4 and newer).

[1] NixOS/nix#3590

### DIFF
--- a/ale_linters/nix/nix.vim
+++ b/ale_linters/nix/nix.vim
@@ -1,18 +1,51 @@
 " Author: Alistair Bill <@alibabzo>
+" Author: Maximilian Bosch <maximilian@mbosch.me>
 " Description: nix-instantiate linter for nix files
 
+function! ale_linters#nix#nix#Command(buffer, output, meta) abort
+    let l:version = a:output[0][22:]
+
+    if l:version =~# '^\(2.4\|3\).*'
+        return 'nix-instantiate --log-format internal-json --parse -'
+    else
+        return 'nix-instantiate --parse -'
+    endif
+endfunction
+
 function! ale_linters#nix#nix#Handle(buffer, lines) abort
-    let l:pattern = '^\(.\+\): \(.\+\), at .*:\(\d\+\):\(\d\+\)$'
     let l:output = []
 
-    for l:match in ale#util#GetMatches(a:lines, l:pattern)
-        call add(l:output, {
-        \   'lnum': l:match[3] + 0,
-        \   'col': l:match[4] + 0,
-        \   'text': l:match[1] . ': ' . l:match[2],
-        \   'type': l:match[1] =~# '^error' ? 'E' : 'W',
-        \})
-    endfor
+    if empty(a:lines)
+        return l:output
+    endif
+
+    if a:lines[0] =~# '^@nix .*'
+        for l:line in a:lines
+            if l:line =~# '^@nix .*'
+                let l:result = json_decode(strpart(l:line, 4))
+
+                if has_key(l:result, 'column')
+                    call add(l:output, {
+                    \     'type': 'E',
+                    \     'lnum': l:result.line,
+                    \     'col': l:result.column,
+                    \     'text': l:result.raw_msg
+                    \})
+                endif
+            endif
+        endfor
+    else
+        let l:pattern = '^\(.\+\): \(.\+\) at .*:\(\d\+\):\(\d\+\)$'
+
+        for l:match in ale#util#GetMatches(a:lines, l:pattern)
+            call add(l:output, {
+            \   'lnum': l:match[3] + 0,
+            \   'col': l:match[4] + 0,
+            \   'text': l:match[1] . ': ' . substitute(l:match[2], ',$', '', ''),
+            \   'type': l:match[1] =~# '^error' ? 'E' : 'W',
+            \})
+        endfor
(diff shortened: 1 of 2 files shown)


Ma27:
Well, it's still WIP unfortunately as the relevant PR in Nix must be merged.

Ma27:
Note to self: we probably need an improved version supporting both Nix 2.3 and 3.0 as a 3.0 release is not in sight. With that being implemented, this PR should be mergeable.

Ma27:
cc @w0rp @hsanson this should be reviewable now btw :)

Ma27:
cc @hsanson @w0rp anything tbd to get this merged? :)

Ma27:
cc @w0rp would you mind having a look? :)

ale_linters/nix/nix.vim
```diff
@@ -1,18 +1,51 @@
 " Author: Alistair Bill <@alibabzo>
+" Author: Maximilian Bosch <maximilian@mbosch.me>
 " Description: nix-instantiate linter for nix files
 
+if !exists('g:ale_linters_nix_version')
+    let g:ale_linters_nix_extra_cmd = ''
+
+    " strip away `nix-instantiate (Nix) `
+    let g:version = system('nix-instantiate --version')[22:]
```
hsanson, line 9: This `system()` method is blocking so cannot be merged into ALE. I got PR's [rejected](https://github.com/dense-analysis/ale/pull/2121#discussion_r239214535) by @w0rp time ago for the same reason.

The way to do this now is by chaining commands using `ale#command#Run()` function. Read the help of this function and look into `semver.vim` file the function `ale#semver#RunWithVersionCheck()` for an example of how to achieve what you want.

There are are many other linters/fixers that check the tool version using the same chained command method. Grepping ALE's code for WithVersion will let you find them.

hsanson (changes requested):

hsanson:

```
# needed if already added.

git checkout master
git fetch upstream master
git merge upstream/master

git checkout my-branch-name

```

ale_linters/nix/nix.vim
Ma27: Hmm the problem is that this doesn't control which command to execute, but how to interpret the output (in contrast to the other linters AFAIU). Do you have a suggestion how to implement this with `RunWithVersionCheck`?

hsanson: You can use ale#command#Run() to build the command line including or not the ` ' --log-format internal-json '` flag depending on the tool version.

For the Handle() method you first parse using the pattern and check if l:output is empty or not. If not empty return it and if empty try parsing with the json format:

```
function! ale_linters#nix#nix#Handle(buffer, lines) abort
    let l:pattern = '^\(.\+\): \(.\+\), at .*:\(\d\+\):\(\d\+\)$'
    let l:output = []

        call add(l:output, {
        \   'lnum': l:match[3] + 0,
        \   'col': l:match[4] + 0,
        \   'text': l:match[1] . ': ' . l:match[2],
        \   'type': l:match[1] =~# '^error' ? 'E' : 'W',
        \})
    endfor

    if !empty(l:output)
        return l:output
    endif

    for l:line in a:lines
        if l:line =~# '^@nix .*'
            let l:result = json_decode(strpart(l:line, 4))

            if has_key(l:result, 'column')
                call add(l:output, {
                \     'type': 'E',
                \     'lnum': l:result.line,
                \     'col': l:result.column,
                \     'text': l:result.raw_msg
                \})
            endif
        endif
     endfor
   
    return l:output
endfunction
```

Ma27: Fixed, thanks for the hint :)

test/handler/test_nix_handler.vader
```diff
@@ -5,6 +5,31 @@ After:
   call ale#linter#Reset()
 
 Execute(The nix handler should parse nix-instantiate error messages correctly):
+  let g:ale_linters_nix_version = 'new'
```
hsanson, line 8: This variable is no longer needed.

test/handler/test_nix_handler.vader
```diff
@@ -5,6 +5,31 @@ After:
...
+  \
+  \ ],
+  \ ale_linters#nix#nix#Handle(bufnr(''), [
+  \   "@nix {\"line\":6,\"column\":3,\"raw_msg\":\"syntax error, unexpected ']', expecting ';'\"}",
+  \   "@nix {\"line\":3,\"column\":5,\"raw_msg\":\"undefined variable 'foo'\"}",
+  \   "@nix {\"unrelated\":\"message\"}"
+  \ ])
+
+Execute(The nix handler should parse message from old nix-instantiate correctly):
+  let g:ale_linters_nix_version = 'old'
```
hsanson, line 32: This variable is no longer needed.

hsanson (changes requested):

test/handler/test_nix_handler.vader
Ma27: both fixed :)

hsanson (approved):
Looks good, thanks.

Issue #3228, opened 2020-07-07
Add support for astyle


cjappl:
**Name:** astyle 
**URL:** http://astyle.sourceforge.net/

Astyle is a code format fixer for c and c++. It is run in the command line using:

`astyle` 

It also respects project based config files using 

`astyle --project`

I attempted to implement this myself but unfortunately my vimscript skills are very lacking. Here is what could maybe be the bones of an `autoload/ale/fixers/astyle.vim`

```
" Author:
" Description:

call ale#Set('c_astyle_executable', 'astyle')
call ale#Set('c_astyle_options', '')

function! ale#fixers#astyle#Fix(buffer)
    let l:executable = ale#Var(a:buffer, 'c_astyle_executable')
    let l:options = ale#Var(a:buffer, 'c_astyle_options')

    return {
    \   'command': ale#Escape(l:executable)
    \       . ' --project '
    \       . (empty(l:options) ? '' : ' ' . l:options)
    \}
endfunction
```

skorokithakis:
This is being worked on in #3231.

hsanson:
This is already done for C/C++ thanks to @jhlink efforts.

Issue #1695, opened 2018-07-03
ALEFix deletes rubocop comments


nyarly:
## Information

**VIM version**
NVIM v0.2.2
Build type: Release

Operating System: 
NAME=NixOS
ID=nixos
VERSION="18.03.131807.489a14add9a (Impala)"

### :ALEInfo
```
let g:ale_ruby_brakeman_options = ''
let g:ale_ruby_reek_show_context = 0
let g:ale_ruby_reek_show_wiki_link = 0
let g:ale_ruby_rubocop_executable = 'rubocop'
let g:ale_ruby_rubocop_options = ''
let g:ale_ruby_ruby_executable = 'ruby'
 Global Variables:
let g:ale_cache_executable_check_failures = 0
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = 100
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = 50
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {'ruby': 'rubocop'}
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_filetype_changed = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'never'
let g:ale_lint_on_insert_leave = 1
let g:ale_linter_aliases = {}
let g:ale_linters = {'ruby': 'all'}
let g:ale_linters_explicit = 0
let g:ale_list_window_size = 10
let g:ale_list_vertical = 0
let g:ale_loclist_msg_format = '%code: %%s'
let g:ale_max_buffer_history_size = 20
let g:ale_max_signs = -1
let g:ale_maximum_file_size = 0
let g:ale_open_list = 'on_save'
let g:ale_pattern_options = {}
let g:ale_pattern_options_enabled = 0
let g:ale_set_balloons = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
[...]

PR #3231, opened 2020-07-10
Add astyle for C/C++ formatting

jhlink:
Hi all~~ First time commit!! Phew~~ 💦 

I'd like to add the astyle as a C formatter to ale. 

Technically, astyle supports [other formats](http://astyle.sourceforge.net/astyle.html) (Java, C#, ObjC, C++), but for simplicity, I'm adding astyle specifically for C. 

Please let me know if there are any changes I can make for a successful PR merge. 😄 

**--Edit--**
Closes issue #3228 

This PR also adds support for the following:
- C++
- project level options by providing the filename containing astyle options to `ale_c_astyle_options` or `ale_cpp_astyle_options`
  - e.g. If .astylerc is used in a project, add `let g:ale_c_astyle_options = .astylerc` to your vimrc, which passes the file name via [--project cmd line argument](http://astyle.sourceforge.net/astyle.html#_Option_Files). 
- redirection via stdin😅

### DIFF
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -160,6 +160,11 @@ let s:default_registry = {
 \       'suggested_filetypes': ['php'],
 \       'description': 'Fix PHP files with php-cs-fixer.',
 \   },
+\    'astyle': {
+\       'function': 'ale#fixers#astyle#Fix',
+\       'suggested_filetypes': ['c', 'cpp'],
+\       'description': 'Fix C/C++ with astyle.',
+\    },
 \   'clangtidy': {
 \       'function': 'ale#fixers#clangtidy#Fix',
 \       'suggested_filetypes': ['c', 'cpp', 'objc'],
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -163,7 +163,7 @@ let s:default_registry = {
 \    'astyle': {
 \       'function': 'ale#fixers#astyle#Fix',
 \       'suggested_filetypes': ['c', 'cpp'],
-\       'description': 'Fix C with astyle.',
+\       'description': 'Fix C/C++ with astyle.',
 \    },
 \   'clangtidy': {
 \       'function': 'ale#fixers#clangtidy#Fix',
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -162,8 +162,8 @@ let s:default_registry = {
 \   },
 \    'astyle': {
 \       'function': 'ale#fixers#astyle#Fix',
-\       'suggested_filetypes': ['c', 'cpp'],
-\       'description': 'Fix C/C++ with astyle.',
+\       'suggested_filetypes': ['c'],
+\       'description': 'Fix C with astyle.',
 \    },
 \   'clangtidy': {
 \       'function': 'ale#fixers#clangtidy#Fix',
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -162,7 +162,7 @@ let s:default_registry = {
 \   },
 \    'astyle': {
 \       'function': 'ale#fixers#astyle#Fix',
-\       'suggested_filetypes': ['c'],
+\       'suggested_filetypes': ['c', 'cpp'],
 \       'description': 'Fix C with astyle.',
 \    },
 \   'clangtidy': {
--- a/autoload/ale/fixers/astyle.vim
+++ b/autoload/ale/fixers/astyle.vim
@@ -21,10 +21,11 @@ function! ale#fixers#astyle#Fix(buffer) abort
     let l:executable = ale#fixers#astyle#Var(a:buffer, 'executable')
     let l:filename = ale#Escape(bufname(a:buffer))
     let l:options = ale#fixers#astyle#Var(a:buffer, 'options')
-    let l:command = ' --stdin=' . l:filename
+    let l:command = ' --stdin='
 
     return {
-    \   'command': ale#Escape(l:executable) . l:command
+    \   'command': ale#Escape(l:executable)
     \     . (empty(l:options) ? '' : ' --project=' . l:options)
+    \     . l:command
     \}
 endfunction
--- a/autoload/ale/fixers/astyle.vim
+++ b/autoload/ale/fixers/astyle.vim
@@ -1,14 +1,27 @@
 " Author: James Kim <jhlink@users.noreply.github.com>
-" Description: Fix C files with astyle.
+" Description: Fix C/C++ files with astyle.
 
-call ale#Set('c_astyle_executable', 'astyle')
+function! s:set_variables() abort
+    for l:ft in ['c', 'cpp']
+        call ale#Set(l:ft . '_astyle_executable', 'astyle')
+    endfor
+endfunction
+
+call s:set_variables()
+
+function! ale#fixers#astyle#Var(buffer, name) abort
+    let l:ft = getbufvar(str2nr(a:buffer), '&filetype')
+    let l:ft = l:ft =~# 'cpp' ? 'cpp' : 'c'
+
+    return ale#Var(a:buffer, l:ft . '_astyle_' . a:name)
+endfunction
 
 function! ale#fixers#astyle#Fix(buffer) abort
-    let l:executable = ale#Var(a:buffer, 'c_astyle_executable')
+    let l:executable = ale#fixers#astyle#Var(a:buffer, 'executable')
+    let l:command = ' %t'
 
     return {
-    \   'command': ale#Escape(l:executable)
-    \       . ' %t',
+    \   'command': ale#Escape(l:executable) . l:command,
     \   'read_temporary_file': 1,
     \}
 endfunction
--- a/autoload/ale/fixers/astyle.vim
+++ b/autoload/ale/fixers/astyle.vim
@@ -18,15 +18,42 @@ function! ale#fixers#astyle#Var(buffer, name) abort
     return ale#Var(a:buffer, l:ft . '_astyle_' . a:name)
 endfunction
 
+" Try to find a project options file.
+function! ale#fixers#astyle#FindProjectOptions(buffer) abort
+    let l:proj_options = ale#fixers#astyle#Var(a:buffer, 'project_options')
+
+    " If user has set project options variable then use it and skip any searching.
+    " This would allow users to use project files named differently than .astylerc.
+    if !empty(l:proj_options)
+         return l:proj_options
+    endif
+
+    " Try to find nearest .astylerc file.
+    let l:proj_options = fnamemodify(ale#path#FindNearestFile(a:buffer, '.astylerc'), ':t')
+
+    if !empty(l:proj_options)
+        return l:proj_options
+    endif
+
+    " Try to find nearest _astylerc file.
+    let l:proj_options = fnamemodify(ale#path#FindNearestFile(a:buffer, '_astylerc'), ':t')
+
+    if !empty(l:proj_options)
+        return l:proj_options
+    endif
+
+    " If no project options file is found return an empty string.
+    return ''
+endfunction
+
 function! ale#fixers#astyle#Fix(buffer) abort
     let l:executable = ale#fixers#astyle#Var(a:buffer, 'executable')
-    let l:filename = ale#Escape(bufname(a:buffer))
-    let l:options = ale#fixers#astyle#Var(a:buffer, 'project_options')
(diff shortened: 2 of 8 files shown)


skorokithakis:
Sorry, approved by mistake.

jhlink:
No problem! If you see anything that I can do for a successful merge, please let me know! :D

skorokithakis:
Oh I'm just a user, not a maintainer, sorry :(

jhlink:
Hahaha, it's okay! Even so, I'm still open to feedback! :)

skorokithakis:
My feedback is that I love this commit and am waiting for it to be merged because I use astyle and can't wait to have Ale support for it!

hsanson:
@jhlink would it be too difficult for you to add support for C++, C#, and ObjC? seems a waste to only support C. The clangtidy fixer is a good example of how to handle multiple languages with the same fixer.

@jhlink  @skorokithakis  I cannot test this myself since I do not develop C languages these days but I can merge it if it works for you.

skorokithakis:
@hsanson I can test it today and report back. Astyle reports C++ as well, by the way, so it should just be a matter of configuration.

jhlink:
@hsanson , Sure!!
I'll go ahead and add support for C++, C#, and ObjC.

skorokithakis:
Hmm, this doesn't work for me, either on a `.ino` file that I have mapped to C or a `.c` file. It doesn't format the file on save. Ale works for, say, Python files otherwise.

jhlink:
@skorokithakis , I haven't checked whether this works on .ino files, but I've verified it works for c and cpp files. 

Is **astyle** globally accessible?
You can test whether this works by running `astyle _sourcepath_` in terminal.

Although you shouldn't need it, have you created an .astylerc [config](http://astyle.sourceforge.net/astyle.html) at the project/root level?

skorokithakis:
@jhlink it is, yes, and I can run it fine on project files (`astyle src/main.c`), it just doesn't work inside vim. I installed the PR by pulling the changes into my ~/.vim/bundles/ale repo, and can see the changes there, so that should be fine.

jhlink:
@skorokithakis , rockin'
In your .vimrc, did you define `astyle` to work for `c`  in the `ale_fixers` array? 
```
let g:ale_fixers = {
...
      \ 'c': ['astyle'],
      \ }
```

Also, are you using the `g:ale_fix_on_save` flag to fix the c files?

skorokithakis:
Ah, doh, I didn't realize I needed to enable it. It works fine, thanks!

hsanson:
@jhlink Note that the .astylerc config file won't take any effect if the fixer uses temporary files to work. There are many issues similar to this: #1695 #3119 #3046 
Since astyle supports stdin (redirection) I recommend the fixer is changed to use that instead of temporary files.

jhlink:
Thanks for the heads up @hsanson !!  🙏

skorokithakis:
Can this be fixed/finalized? It's a great piece of functionality that I find myself missing.

jhlink:
Hey @skorokithakis !! 
Yeah, sorry about the hold up. 
It's been difficult making time for this, but here are the goals outstanding for the PR. 

- [ ] Add support for the other languages for astyle (Java, C++, ObjC)
- [ ] Apply astyle changes in accordance to what hsanson described. There seems to be an issue with using temporary files to lint/fix a file. 

My plan is to knock this out this weekend. I'll post an update on Sunday. :)

skorokithakis:
Is there anything I can do to help, @jhlink? The first item seems pretty straightforward, though my vimscript knowledge is nonexistent.

jhlink:
Sure! @skorokithakis Right now, I'm adding C++ support then afterwards work on stdin (redirection) instead of temp. files. 

So far, I'm finishing up adding C++ to fixers/astyle.vim based on clangtidy.
Do the following tasks and work on stdin based fixing instead of temp files. 

- [ ] Add them to docs. There are three places to update, I believe. 
- [ ] Write the Vader test case for each. 
- [ ] Modify the registry.vim to include new lang supp. 
- [ ] Modify fixer/astyle.vim to handle new lang supp. 

I've still got Java and ObjC left to do, but if you can work on either of those, that'd be great!!

jhlink:
Hey @hsanson , 
I've added support for C/C++, project level options, and stdin/redirection. 😅 
As it is, the PR is ready for merge if it looks good to you!! 

I haven't had time to include Java, ObjC, or C#, but figured it could be a challenge for another day. For now, I was hoping a PR including astyle with C/C++ support would suffice for now.

What do you think @skorokithakis  ?

skorokithakis:
Sorry, I haven't had time to work on it these days, but I agree. If it were trivial to add language aliases (ie if you could just say "use the C cleaner for C++") I'd say we should do that, but since it seems a fair amount of work to redo everything for another language, I think this can be merged and C++/ObjC can be added later.

jhlink:
@skorokithakis , No problem. :) 
(Don't forget! I already added C++ support!)

Here are the languages that haven't been added and propose to be added later:
- Java
- C#
- ObjC

hsanson:
@jhlink this is looking good, great work. I just have some recommendations:

1. Rename `g:ale_cpp_astyle_options` and `g:ale_c_astyle_options` to `g:ale_cpp_astyle_project_options` and `g:ale_c_astyle_project_options` respectively.  This matches the actual flags and usage of this configuration file and would allow  in the future to add support for default options configuration file (e.g. `g:ale_cpp_astyle_default_options`).  I do not use astyle myself but I do see the benefit of having a defaults (global to all projects) config file and per project config file.

2. Would be better if ALE tried to find the presence of a .astylerc file in the project and set the --project flag accordingly if found automatically. Something like this should work:

```vimscript
" Try to find a project options file.
function! ale#fixers#astyle#FindProjectOptions(buffer) abort
    let l:options = ale#fixers#astyle#Var(a:buffer, 'options')
 
    " If user has set project options variable then use it and skip any searching.
    " This would allow users to use project files named differently than .astylerc.
    if !empty(l:options)
         return l:options
    endif

    " Try to find nearest .astylerc file.
    let l:options = fnamemodify(ale#path#FindNearestFile(a:buffer, '.astylerc'), ':t')

    if !empty(l:options)
        return l:options
    endif

    " Try to find nearest _astylerc file.
    let l:options = fnamemodify(ale#path#FindNearestFile(a:buffer, '_astylerc'), ':t')

    if !empty(l:options)
        return l:options
    endif

    " If no project options file is found return an empty string.
    return ''
endfunction

function! ale#fixers#astyle#Fix(buffer) abort
    let l:executable = ale#fixers#astyle#Var(a:buffer, 'executable')
    let l:filename = ale#Escape(bufname(a:buffer))
    let l:options = ale#fixers#astyle#FindProjectOptions(a:buffer)
    let l:command = ' --stdin='

    return {
    \   'command': ale#Escape(l:executable)
    \     . (empty(l:options) ? '' : ' --project=' . l:options)
    \     . l:command
    \}
endfunction
``` 

Note I have not tested the above code, please use with discretion.

jhlink:
@hsanson , thanks a bunch for the feedback!! I'll start on this later tonight!! 😄

jhlink:
@hsanson, Thanks a bunch for the feedback!! Learned some reaaaal cool things about ALE, Vimscript, and Vader. Pretty slick!!

Alrighty~ 😅  

1. I renamed `options` to `project_options` and cascaded the change to the documentation. 

2. I had no idea ALE had a function like `FindNearestFile`. That makes things waaay easier! I wrote a test case that detects `_astylerc` by default, but I had to add a dummy `_astylerc` file in `test/test_c_projects/makefile_project`. 

I was thinking about adding another test to check for `.astylerc`, but that would require making another dummy file in some test directory. I wasn't sure if this was acceptable since leaving stray configs in various test directories felt dubious. (Open to add it though!)

Test cases aside, I've confirmed that both default types are detected and an empty string is returned when failing to find either. 

Let me know if there's anything I can do to facilitate the merge! :D

hsanson:
@jhlink good to go. Great work!.

skorokithakis:
Fantastic, this is extremely useful. Thank you @jhlink!

cjappl:
Hey all, having a problem testing this out. let me know if I should open another issue.

Here is a snippet of lines from ALEInfo when I try to run :ALEFix 

```
let g:ale_cpp_astyle_executable = 'astyle'
let g:ale_cpp_astyle_project_options = '.astylerc'
...
let g:ale_fixers = {'cmake': ['remove_trailing_lines'], 'cpp': ['remove_trailing_lines', 'astyle'], 'json': ['prettier'], 'javascript': ['prettier'], 'css': ['prettier'], 'python': ['remove_trailing_lines', 'trim_whitespace']}
...
(finished - exit code 1) ['bash', '-c', '''astyle'' --project=.astylerc --stdin= < ''/var/folders/n0/0fkj29b90q3_nlv8jkrxprr8m01t_4/T/nvimX4NjIS/3/test_DelayBlockBasics.cpp''']
```

I can successfully run on the command line using the following command:
`$ bash -c "astyle --project=.astylerc --stdin=''./processorblocks/delayblock/test/test_DelayBlockBasics.cpp"`

but not when I use the following (exit code 1, as above):
```
$ bash -c "astyle --project=.astylerc --stdin= < ''./processorblocks/delayblock/test/test_DelayBlockBasics.cpp"                                                                                                                                                         
Cannot open project option file .astylerc
Artistic Style has terminated
```

Anything you could think of that I'm doing wrong here? The `<` in the second command, and in the one that ALE is running could be a hint? 

Happy to test anything out! Been waiting for having astyle in ALE for so long :)

skorokithakis:
Ah, looks like this assumes that an `.astyle.rc` will exist, even though it doesn't.

cjappl:
I don't think that's the problem. I do have an .astylerc and I have defined it in my .vimrc using
`let g:ale_cpp_astyle_project_options = '.astylerc'`

jhlink:
Hey @cjappl !

The plugin sets these flags automatically so long as you've added `astyle` in the `ale_fixer` dictionary. Unless your project options file is named differently than `.astylerc` or `_astylerc`, you don't have to explicitly set them.

```
let g:ale_cpp_astyle_executable = 'astyle'                 " Not needed
let g:ale_cpp_astyle_project_options = '.astylerc'     " Maybe needed if default names don't float your boat. :) 
```

The only way I've replicated your error is if I place `.astylerc` in the same directory as some C/C++ file. I think it's because astyle assumes that the project options file is at the project root directory and not in the same directory as the source. 

```
The project option file has precedence over the default option file but not the command line options. The project option file should be in the top directory of the project being formatted. The file is identified by a file name only. One of the command line project options must be used to indicate a file is available, or it must be referred to by the environment variable. Artistic Style looks for the file in the current directory or one of its parent directories in the following order.
```  (src)[http://astyle.sourceforge.net/astyle.html#_Option_Files]

Or the fact that it doesn't detect the config file in the same source directory could be a bug. (It's weird, I think this is a plus. Better to put all your source files in a dedicated directory than make a smorgasbord of jungle files.)

You can resolve this in a couple ways. 
1. In your project directory, keep your `.astylerc` at the root directory and place all your cpp files in a `src` dir. 
2. Place a `.astylerc` in your home dir. (e.g. `~/`), and use it as the default for all your cpp files. 

Either way, you should see astyle make formatting changes to your code when executing ALEFix. 

Hope this helps!!

**Comment by cjappl** [comment]
Ah interesting! That clarifies a lot! Thanks so much for taking the time to write all that out, it's much appreciated

So I think I found a bug/improvement.

here's the issue.

My directory structure is something similar to the following:
- rootDir
   -.astylerc
   -src

I normally launch vim from rootDir. This was causing the bug above.

When I instead launch vim from src, it works as intended.

This leads me to believe that you must have .astylerc in a directory that is a parent of the directory you launch vim in.

Is there a way to get it to look in the launch directory + parent directories instead of just parent directories? I would guess my use case isn't uncommon.

It does appear to be working 100% fine as long as astylerc is ABOVE the launch directory for vim. This is so great, been waiting for a long time for this exact thing

**Comment by hsanson** [comment]
Yes, it is a bug. My limited tests did not include .astylerc files.  This can be easily fixed by changing the command to include the relative path of the file being formatted in the --stdin argument:

```
let l:command = ' --stdin=''' . expand('#'.a:buffer) . ''''
```

@jhlink if you can make a PR with this change I can approve and merge it ASAP. If I do it myself I cannot approve nor merge my own PRs.

jhlink:
@hsanson , sure!! I'll hammer it out today!!

georgexsh:
thanks for add astyle for ale, it works great.
but it has an odds behavior which I am uncertain a bug or feature: the astyle fixer appears to use the content on disk to astyle executable, rather than the content in the current buffer, which means if I changed the buffer without saving than run astyle fixer, the change will be reverted.

PR #3237, opened 2020-07-19
Restore old behavior of ALEFix command for Rubocop

NARKOZ:
Since RuboCop 0.60 ALEFix command stopped to fix all found offenses. This change restores the 
previous behavior by allowing rubocop to fix all detected offenses.

Also see: https://docs.rubocop.org/rubocop/usage/auto_correct.html

### DIFF
--- a/autoload/ale/fixers/rubocop.vim
+++ b/autoload/ale/fixers/rubocop.vim
@@ -1,15 +1,18 @@
 call ale#Set('ruby_rubocop_options', '')
+call ale#Set('ruby_rubocop_auto_correct_all', '')
 call ale#Set('ruby_rubocop_executable', 'rubocop')
 
 function! ale#fixers#rubocop#GetCommand(buffer) abort
     let l:executable = ale#Var(a:buffer, 'ruby_rubocop_executable')
     let l:config = ale#path#FindNearestFile(a:buffer, '.rubocop.yml')
     let l:options = ale#Var(a:buffer, 'ruby_rubocop_options')
+    let l:auto_correct_all = ale#Var(a:buffer, 'ruby_rubocop_auto_correct_all')
 
     return ale#ruby#EscapeExecutable(l:executable, 'rubocop')
     \   . (!empty(l:config) ? ' --config ' . ale#Escape(l:config) : '')
     \   . (!empty(l:options) ? ' ' . l:options : '')
-    \   . ' --auto-correct-all --force-exclusion %t'
+    \   . (!empty(l:auto_correct_all) ? ' --auto-correct-all' : ' --auto-correct')
+    \   . ' --force-exclusion %t'
 endfunction
 
 function! ale#fixers#rubocop#Fix(buffer) abort
--- a/autoload/ale/fixers/rubocop.vim
+++ b/autoload/ale/fixers/rubocop.vim
@@ -1,5 +1,5 @@
 call ale#Set('ruby_rubocop_options', '')
-call ale#Set('ruby_rubocop_auto_correct_all', '')
+call ale#Set('ruby_rubocop_auto_correct_all', 0)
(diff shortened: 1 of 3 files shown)


autoload/ale/fixers/rubocop.vim
```diff
@@ -9,7 +9,7 @@ function! ale#fixers#rubocop#GetCommand(buffer) abort
     return ale#ruby#EscapeExecutable(l:executable, 'rubocop')
     \   . (!empty(l:config) ? ' --config ' . ale#Escape(l:config) : '')
     \   . (!empty(l:options) ? ' ' . l:options : '')
-    \   . ' --auto-correct --force-exclusion %t'
+    \   . ' --auto-correct-all --force-exclusion %t'
 endfunction
```
w0rp, line 13: Let's make this optional, and use `--auto-correct` as the default. Reading the documentation, they say that some fixes are safe, and others are unsafe, so let's use only the safe fixes by default.

paul: Is there a way to fix this in my config in the meantime? Setting the option in my vim config thus:

```vim
let g:ale_ruby_rubocop_options = '--auto-correct-all'
```

results in a rubocop command like:

```
 ['/usr/bin/zsh', '-c', '''rubocop'' --config ''.../.rubocop.yml'' --auto-correct-all --auto-correct --force-exclusion ''/tmp/nvim7B7OMS/.../myfile.rb''']
```

It seems the `--auto-correct-all` I provided comes first, so the `--auto-correct` default here "wins". Could we swap the order that `options` occurs to after the defaults in the command line?

```vim
    return ale#ruby#EscapeExecutable(l:executable, 'rubocop')
    \   . (!empty(l:config) ? ' --config ' . ale#Escape(l:config) : '')
    \   . '--auto-correct --force-exclusion'
    \   . '%t'
```

NARKOZ: @w0rp I have made the requested changes. Now safe auto correction is by default. Users can enable non safe autocorrection by setting `ruby_rubocop_auto_correct_all` variable:

```
let g:ale_ruby_rubocop_auto_correct_all = 1
```

autoload/ale/fixers/rubocop.vim
```diff
@@ -1,15 +1,18 @@
 call ale#Set('ruby_rubocop_options', '')
+call ale#Set('ruby_rubocop_auto_correct_all', '')
```
w0rp, line 2: Set this to `0` by default.

autoload/ale/fixers/rubocop.vim
```diff
@@ -1,15 +1,18 @@
...
     let l:executable = ale#Var(a:buffer, 'ruby_rubocop_executable')
     let l:config = ale#path#FindNearestFile(a:buffer, '.rubocop.yml')
     let l:options = ale#Var(a:buffer, 'ruby_rubocop_options')
+    let l:auto_correct_all = ale#Var(a:buffer, 'ruby_rubocop_auto_correct_all')
 
     return ale#ruby#EscapeExecutable(l:executable, 'rubocop')
     \   . (!empty(l:config) ? ' --config ' . ale#Escape(l:config) : '')
     \   . (!empty(l:options) ? ' ' . l:options : '')
-    \   . ' --auto-correct --force-exclusion %t'
+    \   . (!empty(l:auto_correct_all) ? ' --auto-correct-all' : ' --auto-correct')
```
w0rp, line 14:
```suggestion
    \   . (l:auto_correct_all ? ' --auto-correct-all' : ' --auto-correct')
```

Do this instead, and document this new option as a `Number`. Update the tests to check the command we run when it's set to `1`.

NARKOZ:
@w0rp done.

w0rp:
Cheers! :beers:

PR #3196, opened 2020-05-31
Add autoimport and rename support for langservers

jeremija:
This PR contains the following updates:

- `ALERename` command now works with langservers (previously only with tsserver).
- Autoimport functionality now works with langservers (previously only with tsserver).

I've been using both with `gopls` for a few months and it's been working pretty well.

Closes #2976

PR #3240, opened 2020-07-20
Fixing gopls, Improving alerename to support TextDocumentEdit[]

lightyear15:
- Fixing gopls launch options
- improving ale rename to support ``WorkspaceEdit`` as list of ``TextDocumentEdit``

### DIFF
--- a/autoload/ale/rename.vim
+++ b/autoload/ale/rename.vim
@@ -95,41 +95,65 @@ function! ale#rename#HandleLSPResponse(conn_id, response) abort
         endif
 
         let l:workspace_edit = a:response.result
+        let l:changes = []
 
-        if !has_key(l:workspace_edit, 'changes') || empty(l:workspace_edit.changes)
+        if has_key(l:workspace_edit, 'changes') && !empty(l:workspace_edit.changes)
+            for l:file_name in keys(l:workspace_edit.changes)
+                let l:text_edits = l:workspace_edit.changes[l:file_name]
+                let l:text_changes = []
+
+                for l:edit in l:text_edits
+                    let l:range = l:edit.range
+                    let l:new_text = l:edit.newText
+
+                    call add(l:text_changes, {
+                    \ 'start': {
+                    \   'line': l:range.start.line + 1,
+                    \   'offset': l:range.start.character + 1,
+                    \ },
+                    \ 'end': {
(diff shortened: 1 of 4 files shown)


autoload/ale/rename.vim
```diff
@@ -95,41 +95,65 @@ function! ale#rename#HandleLSPResponse(conn_id, response) abort
         endif
 
         let l:workspace_edit = a:response.result
+        let l:changes = []
 
-        if !has_key(l:workspace_edit, 'changes') || empty(l:workspace_edit.changes)
+        if has_key(l:workspace_edit, 'changes') && !empty(l:workspace_edit.changes)
```
w0rp, line 100: See if you can handle the LSP data so we keep most of the lines we had before, but merge one of the two different formats into a common `List` of `Dict`s we can use below.

lightyear15: Sorry, your comment is not 100% clear to me.
You meant, rather than my change which in pseudocode looks something like this
```
changes = []
if LSP answer is type A
   " cycle over changed files and parse changes according to type A format
elseif LSP answer is type B
   " cycle over changed files and parse changes according to type B format
endif
" cycle over changes, and apply
```
you'd like to see something more or less like 
```
file_list = []
changes = []
if LSP answer is type A
   " parse file list according to type A format
elseif LSP answer is type B
   " parse file list according to type B format
endif
for file_name in file_list
   if LSP answer is type A
      " parse changes on file_name according to type A format
   elseif LSP answer is type B
      " parse changes on file_name according to type B format
endfor
" cycle over changes, and apply
```
correct?

lightyear15: I don't think I can
the original LSP response presents a dictionary, where key is file_name and value is a list of changes
the one I added, presents a list of structs which  contain the filename and the list of changes for that file.
I can't think of a way to iterate over those different responses while unifying the way I add the change in the ``changes`` list and make one single call to ``add`` on ``changes``.

lightyear15: I can refactor out the call to ``add`` so to avoid code repetition...

w0rp: I just figured out that this was actually already done in #3196, so I'll merge that and you can remove your changes to this here.

PR #3252, opened 2020-07-29
Support yarn berry (pnp)

myobie:
Support for `eslint` and `tsserver` with latest yarn.

_I didn't change every possible node package's binary path because I wouldn't be able to test them to make sure they work._


toastal:
This can be expanded to several other tools as well like `stylelint` or `elm-format`

myobie:
@toastal yep 🥳  I'd like to get the two largest tools merged and then let others contribute to their tools's files using this as an example.

myobie:
Any chance this can be re-opened and considered?

PR #3282, opened 2020-08-12
fix(ale_linters/phpcs): add support for multiline error messages

zanona:
This PR adds support for `phpcs` when there are errors resulting on messages containing more than one line.

An example would be :

```php
<?php
$foo = 'bar';
echo "
<div>
    <h1>hello world</h1>
    <p>$foo</p>
</div>
";
```

Before this fix, the command below would return an error, but Ale wouldn't catch it.
```
$ phpcs --report=emacs theme/single-member.php
/home/user/file-name.php:14:10: error - All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found '"
'.
```

### DIFF
--- a/ale_linters/php/phpcs.vim
+++ b/ale_linters/php/phpcs.vim
@@ -23,7 +23,7 @@ function! ale_linters#php#phpcs#Handle(buffer, lines) abort
     " Matches against lines like the following:
     "
     " /path/to/some-filename.php:18:3: error - Line indented incorrectly; expected 4 spaces, found 2 (Generic.WhiteSpace.ScopeIndent.IncorrectExact)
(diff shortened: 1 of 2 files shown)


zanona:
Unrelated test error at `/testplugin/test/command_callback/test_bingo_command_callback.vader`

zanona:
Rebasing seems to have fixed tests, as it was unrelated. cc / @w0rp

zanona:
Oh my, this is really embarrassing.

zanona:
@w0rp can you point someone to review this. I'm not quite sure what else I could've done here. Tests passed and all...

w0rp (approved):
This is all good, I've just been too busy to look at ALE pull requests in recent months.

w0rp:
Cheers! :beers:

Issue #1557, opened 2018-05-06
Consider supporting dmypy


w0rp:
The latest version of mypy now ships with its own special daemon for checking for type errors faster. It might be nice if ALE supported this in some way or another, possibly including an additional plugin which uses ALE.

alok:
Has the `0.610` release of mypy improved this enough?

w0rp:
I'll have to look into it.

ckarnell:
It looks like the daemon is now considered stable, as of the 0.750 release of mypy: https://mypy-lang.blogspot.com/2019/11/mypy-0.html

ckarnell:
@w0rp If you gave basic instructions on how to contribute to get this working I might be able to help out.

w0rp:
@ckarnell It's great that the daemon is stable now. Here is how I think we should handle this.

I think `dmypy` is different enough that we'd have to avoid breaking how people have configured `mypy` by adding a new linter file. We can duplicate the `mypy.vim` file, and if the error format is the same, we can use the same `Handle` function for both linters.

We can decide if `dmypy` should be run by searching for the `dmypy.json` file, using the functions we usually use in the codebase. That's probably the directory we should use as the working directory.

This is probably something I might end up implementing, if someone else doesn't, as it's one of those tools I'll probably end up needing for my day job.

w0rp:
Another thing I think worth implementing is checking in the `mypy` linter file if both `dmypy` and `mypy` linters are enabled, with both being enabled by default, and if the `dmypy.json` file exists. If the file exists, we could skip running the `mypy` linter, with a new option to force running `mypy`.

ascandella:
I started looking into this as I recently introduced mypy to a large codebase at work (30 seconds to run mypy).

I started with the approach of two separate linters as suggested (copy mypy.vim -> dmypy.vim) but it was almost exactly the same. Then, I couldn't figure out the best way to detect which linters were enabled (I couldn't find other examples of linters checking `ale#Var(a:buffer, 'linters')` or similar.

In addition, in my usage `dmypy run -- ` is basically a drop in replacement for running `mypy`, except that it's faster and has the potential to create a `.dmypy.json` file if you ask it to start when one's not running.

Given the above, I wanted to see what it would take to consolidate the linters so we didn't have nearly-identical logic in multiple places.

Here is the draft PR I put together (on my personal fork, not ready for upstream yet). It needs tests and documentation, but if others want to take it for a spin, I'm interested in hearing feedback.

[...]

PR #2385, opened 2019-03-22
Mypy: try to find folder containing mypy.ini to use as cwd.

hjwp:
attempts to fix #2383 

no tests i'm afraid (yet? will see what i can do.  i don't know vimscript...)

PR #3283, opened 2020-08-12
Add dmypy(mypy daemon) as a linter

davidatbu:
Two years ago, in #1557 , `dmypy` support was discussed. @ascandella implemented a proof-of-concept on his own fork, but that never made it through to an actual PR. This is another attempt. This time, I tried to follow @w0rp 's suggestions on keeping `dmypy` and `mypy` as separate linters.

## Question
This PR is a draft PR because I am not sure how to go about disabling `mypy` if `dmypy` is installed (and there's a `.dmypy.json` file somewhere). This idea was also @w0rp 's suggestion in #1557 . If someone can point me to what the best way to do that would be, that's be great. 

## Other important points
 - There's a new `python_mypy_run_even_with_dmypy` flag, according to @w0rp  's suggestion(it's not being used right now, look at "Question" above).
This linter uses the same `Handle` function as the `mypy` linter(Which means it uses the flags `python_mypy_ignore_invalid_syntax` and `python_mypy_show_notes`). 
- Additionally, this linter also uses the `python_mypy_options`, _in addition_ to `python_dmypy_options`. The latter is meant for `dmypy`-only options.
 
Once the question above is answered, I'd be happy to write tests, and once we finalize the implementation, the documentation.

### DIFF
--- a/ale_linters/python/mypy.vim
+++ b/ale_linters/python/mypy.vim
@@ -7,6 +7,29 @@ call ale#Set('python_mypy_show_notes', 1)
 call ale#Set('python_mypy_options', '')
 call ale#Set('python_mypy_use_global', get(g:, 'ale_use_global_executables', 0))
 call ale#Set('python_mypy_auto_pipenv', 0)
+call ale#Set('python_mypy_run_even_with_dmypy', 0)
+
+function! ale_linters#python#mypy#upgadeToDmypyIfPossible(buffer) abort
+    if !ale#Var(a:buffer, 'python_mypy_run_even_with_dmypy')
+        return
+    endif
+
+    let l:enabled_linters = ale#linter#GetLintersLoaded()
+
+    " Check if dmypy is enabled
+    if exists('enabled_linters["python"]')
+        for l:linter in l:enabled_linters['python']
+            if l:linter['name'] is? 'dmypy'
+                " Check if there's a .dmypy.json file
+                for l:path in ale#path#Upwards(expand('#' . a:buffer . ':p:h'))
+                    if filereadable(l:path . '/.dmypy.json')
+                        " TODO: Ok, how do we abort running mypy now?
+                    endif
+                endfor
+            endif
+        endfor
+    endif
+endfunction
 
 function! ale_linters#python#mypy#GetExecutable(buffer) abort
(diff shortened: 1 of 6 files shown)


ale_linters/python/mypy.vim
```diff
@@ -7,6 +7,29 @@ call ale#Set('python_mypy_show_notes', 1)
 call ale#Set('python_mypy_options', '')
 call ale#Set('python_mypy_use_global', get(g:, 'ale_use_global_executables', 0))
 call ale#Set('python_mypy_auto_pipenv', 0)
+call ale#Set('python_mypy_run_even_with_dmypy', 0)
+
+function! ale_linters#python#mypy#upgadeToDmypyIfPossible(buffer) abort
```
w0rp, line 12: To make this work, you can return `0` or `1` from this function, and call it in `ale_linters#python#mypy#GetExecutable`. You can return an empty string from that function to skip running the linter.

w0rp, line 12:
```suggestion
function! ale_linters#python#mypy#WillUseDmypyInstead(buffer) abort
```

Name the function this instead. It will make more sense.

davidatbu:
Hi @w0rp , sorry for the silence. So, I have incorporated your comments and I'm the process of adding tests(all of this locally, to be clear). But, there's one (I think big) issue: `dmypy` reloads itself if you pass `--shadow-file` to it. I have created an issue [here on the mypy repo](https://github.com/python/mypy/issues/9309), but haven't gotten a response yet.

This basically defeats the point of supporting dmypy, since, (in my experiennce), reloading the daemon is no faster than just running `mypy`. 

That is of course, unless #2385 lands. In that case, we can perhaps provide an option such as `python_dmypy_lint_file` that lets us avoid `--shadow-file`. 

How do you think we should proceed?

w0rp:
Let's wait for them to fix `--shadow-file`. It probably won't work well without it.

davidatbu:
Thanks @w0rp , thanks for the reply!

Small note: I meant to reference issue #3285 in my post above(I've since fixed my post).

sbv-csis:
Sounds like the --shadow-file fix wasnt quite doable - 
Any hope for:
> That is of course, unless #3285 lands. In that case, we can perhaps provide an option such as python_dmypy_lint_file that lets us avoid --shadow-file

PR #3284, opened 2020-08-13
Update ale_ruby_rubocop_auto_correct_all tag

resong:
Looks like updating the tags for `ale_ruby_rubocop_auto_correct_all` was missed; they're the same tags as `ale_ruby_rubocop_options`.


hsanson (approved):
Good catch and thanks for the fix.

PR #3302, opened 2020-08-21
latexindent: Run fixer from stdin instead of a temporary file

khaveesh:
Latexindent supports reading from STDIN since 2018 as specified [here](https://latexindentpl.readthedocs.io/en/latest/sec-how-to-use.html?highlight=stdin#from-the-command-line). So, I have modified the fixer to read from the file directly through STDIN redirection instead of creating temporary files.

I have also fixed the tests which were reading from temporary files to work on STDIN. All tests are passing in my fork and code formatting should be good to go. Thanks!

### DIFF
--- a/autoload/ale/fixers/latexindent.vim
+++ b/autoload/ale/fixers/latexindent.vim
@@ -10,9 +10,7 @@ function! ale#fixers#latexindent#Fix(buffer) abort
 
     return {
     \   'command': ale#Escape(l:executable)
-    \       . ' -l -w'
+    \       . ' -l'
     \       . (empty(l:options) ? '' : ' ' . l:options)
-    \       . ' %t',
-    \   'read_temporary_file': 1,
     \}
 endfunction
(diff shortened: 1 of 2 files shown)


hsanson (approved):
Works fine, tested with my own latex documents.

Sh3Rm4n:
Hey i just wanted to say, that this change broke my setup. I'm using the following options

```vim
let g:ale_tex_latexindent_options = "-l 'indent.yaml' -m"
```

With `indent.yaml` being the local configuration file. When I do `ALEFix` now, the file will be replaced with the content of the config file itself. 

Renaming `indent.yaml` to `defaultSettings.yaml` and changing the options to:

```vim
let g:ale_tex_latexindent_options = " -m"
```

fixes the issue. So I have this workaround at hand. So a fix is not urgent. Just wanted to leave the info here. I haven't looked into ale yet, to see, how to fix this behavior.

PR #3326, opened 2020-08-28
Allow option for showing cursor echos in popup

jsit:
I'm afraid I'm not well-equipped to write a test for this


w0rp (changes requested):
`TruncatedEcho` isn't the right place to insert new popups. I'm considering something closer to #2925.

I think short messages should go to the echo line, and detailed messages could appear in a popup.

Issue #3320, opened 2020-08-27
ALE doesn't lint some errors gcc spits out


Kimplul:
## Information

**VIM version**
NVIM v0.4.4
Build type: Release

Operating System: Debian 10

## What went wrong
ALE doesn't lint certain errors, such as the one that occurs in this specific situation:
c.cpp:
```
#include <string>

int main(){
    std::string str = TEST;
    return 0;
}
```

Output of :ALEInfo:
```
(executable check - success) g++
(finished - exit code 1) ['/bin/bash', '-c', '''g++'' -S -x c++ -o /dev/null -iquote ''/home/ki
mi/Documents/Projects/minimalale'' ''-DTEST=''\''''h''\'''''' -std=c++14 -Wall - < ''/tmp/nvim8
0dfds/2/c.cpp''']
<<<OUTPUT STARTS>>>
<stdin>: In function ‘int main()’:
<command-line>: error: conversion from ‘char’ to non-scalar type ‘std::string’ {aka ‘std::__cxx
11::basic_string<char>’} requested
<stdin>:4:23: note: in expansion of macro ‘TEST’
<<<OUTPUT ENDS>>>
```

To clear it up a bit, I'm using a compile-time macro to assign a single character to a string. This is an error, as they're not the same type, but because of the somewhat odd circumstances of the error message, ale skips over it.

I found a quick botch. I modified line 12 in ```autoload/ale/handlers/gcc.vim``` to be:
```
let s:pattern = '\v^([a-zA-Z]?:?[^:]+):(\d+)?:?(\d+)?:? ([^:]+): (.+)$'
```

This fix passes all test cases, but again, don't know if it has any weird side-effects that the tests don't pick up on.

## Reproducing the bug
1. Create example file
2. Create compile_commands.json file with a macro that corresponds to a single character, in my case
```
[
    {
        "arguments": [
            "g++",
            "-c",
            "-DTEST='h'",
            "c.cpp"
        ],
        "directory": "/home/kimi/Documents/Projects/minimalale",
        "file": "c.cpp"
    }
]
```

### :ALEInfo
```
 Current Filetype: cpp
Available Linters: ['cc', 'ccls', 'clangcheck', 'clangd', 'clangtidy', 'clazy', 'cppcheck', 'cpplint', 'cquery', 'flawfinder']
   Linter Aliases:
'cc' -> ['gcc', 'clang', 'g++', 'clang++']
[...]

PR #3327, opened 2020-08-29
Improved macro handling in gcc

Kimplul:
Here is the fix for #3320. I went with a solution where if the macro expansion error happens in the current file, set ```lnum``` and ```col``` accordingly. If the error happens in some other file, there's really no easy way to know on which line the error should be placed on in the current file, so just let ```lnum``` be 0.

For example, consider this situation:
```
/* c.cpp */
#include <string>
#include "header1.h"

int main(){
   return 0;
}
```
```
/* header1.h */
#include "header2.h"
```
```
/* header2.h */
std::string str = TEST;
```
```
# compile with
g++ c.cpp -DTEST='w'
```

The error reported by GCC would be
```
<command-line>: error: ‘w’ was not declared in this scope
header2.h:1:19: note: in expansion of macro ‘TEST’
    1 | std::string str = TEST;
      |                   ^~~~
```
To figure out on which line in ```c.cpp``` we should place the error we would need to go through the different headers in ```c.cpp``` until we find ```header2.h```. I think that's a bit overkill, and just letting the ```lnum``` be zero with the ```text``` ```Error found in macro expansion. See :ALEDetail``` is (hopefully) good enough.

### DIFF
--- a/autoload/ale/handlers/gcc.vim
+++ b/autoload/ale/handlers/gcc.vim
@@ -10,7 +10,7 @@ let s:pragma_error = '#pragma once in main file'
 " <stdin>:8:5: warning: conversion lacks type at end of format [-Wformat=]
 " <stdin>:10:27: error: invalid operands to binary - (have ‘int’ and ‘char *’)
 " -:189:7: note: $/${} is unnecessary on arithmetic variables. [SC2004]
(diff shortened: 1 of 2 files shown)


w0rp (approved):
Yep, this is good. We handle just this particular kind of error in this specific way, and it hopefully shouldn't break anything.

w0rp:
Cheers! :beers:

Kimplul:
Cheers! :beer:

PR #3334, opened 2020-09-04
Fix asciidoc languagetool integration

remi6397:
This fixes a typo in `asciidoc/languagetool.vim` that prevented languagetool from being loaded in asciidoc files at all.

Also, big thanks to all the contributors for bringing this awesome project :)


hsanson (approved):
Indeed all other linters are defined for 'asciidoc' instead of 'asciidoctor'.

Issue #2418, opened 2019-04-05
Support neovim floatwindow


glepnir:
Coc.nvim can show the lsp linter message in the float window。can ale support this feature？

gbrlsnchs:
There are three things I think would be nice to show in a floating window:
- Hover information
- Documentation while navigating the items in popup menu
- Linting info

pianocomposer321:
Any update on this? I would really like to see floating windows implemented.

rbhanot4739:
I would also love to have this feature implemented as it makes vieweing linter messages much user friendly than having them displayed through `echo` or `virtualtext`.

hsanson:
@glepnir with #3341 maybe this can be closed now?

derlaft:
>with #3341 maybe this can be closed now?

Works for me:

![image](https://user-images.githubusercontent.com/852606/104747157-78c35700-5750-11eb-9d52-6d7067299c5d.png)

Big thanks to @kevinclark and everyone else involved.

Related: I wonder if there's a way to automatically turn on `:ALEHover` tips without an actual mouse.

savchenko:
> Works for me:

Have you manually merged the #3341 into own Ale branch?.. Don't see it working with either nvim 0.4.4 or 0.5.0.

derlaft:
> Have you manually merged the #3341 into own Ale branch?.. Don't see it working with either nvim 0.4.4 or 0.5.0.

No, it's this PR: https://github.com/dense-analysis/ale/pull/3470

At the time I wrote the comment, 0.5.0 was not released and I was using nvim-git AUR package. I think it should work with ale@master and latest neovim release.

savchenko:
> I think it should work now with ale@master and the latest neovim release.

Unfortunately it does not. Assuming you refer to the v0.5.0 which is the latest _released_ version of nVim.

kevinclark:
@savchenko Works for me on both 0.5.0 and 0.6.0+dev44 against c5c58f5bf82bf (committed yesterday). It was also working fine on 0.6.0+dev44 and master from back in April, last I updated before I pulled and then downgraded neovim to try to repro your issue.

I suspect it's a config issue. If you'd like to open a new issue with your config and @ me on it, I can take a look.

PR #3341, opened 2020-09-08
Add nvim floating window support

jargonzombies:
I'm in need of some guidance. From some floating window handling code from [gitgutter](https://github.com/airblade/vim-gitgutter), I've been able to cobble together neovim floating window support.

I've done the following:
- Added a `g:ale_float_preview` variable that enables/disables forwarding calls to `ale#preview#Show` to a function opening a floating window
- Tried to get to feature parity with everything going on in `ale#preview#Show` within the new function `ale#preview#ShowFloating` (commands, filetype, `stay_here`, etc.)

In combination with `g:ale_hover_to_preview`, this generates floats for LSP hover and lints (with `:ALEDetail`), and probably in a lot of other situations too.

Sticking points:
- Is this a good approach? Should it be an alternative to balloons instead?
- Does it make sense to implement `stay_here` functionality for a floating window? Should we bug out onto a regular preview if `stay_here` is false?
- How to test?
- In #2418, a requested feature is floating windows for "documentation while navigating the items in popup menu", but I have no idea what is meant. If they mean that they want hover for symbols in the completion popup, how could I go about adding that?

This is my first time messing around in vimscript, so please explain things as if I were a five year old, but also reasonably competent with google and `:help`.


lithammer:
This is related to https://github.com/dense-analysis/ale/pull/2925.

kevinclark:
@w0rp I'm interested in this too and would consider working through a patch, but I saw this close without a response. Is this something that would be accepted?

w0rp:
I've reopened this. It really just need some kind of test for it.

kevinclark:
Oh, great! @jargonzombies do you want to write the test and push it over the line, or should I pick it up?

w0rp:
I think I might prefer to see this put in its own file and called from other files. `ale#preview#Show` is used for:

* Hover information
* `:ALEDetail`
* Errors if you configure them the right way
* Finding references
* Symbol search
* Handling an LSP "show message" response

Finding references and symbol search should always use the preview window so the `ftplugin` file can be loaded for making a selection. (`ftplugin/ale-preview-selection.vim`) The floating windows might be good for the rest, though I'm not sure about the "show message" response. It would be nice to be able to configure everything in one go, and also individually.

jargonzombies:
@kevinclark Unfortunately, I'm really busy all the way until mid-January, so if you're willing to push this over the line, I would appreciate that.

kevinclark:
@jargonzombies I'll give it a shot!

Issue #3322, opened 2020-08-27
ALERename applies changes in the wrong order


Andoku:
## Information

**VIM version**

VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Mar 18 2020 18:29:15)
Included patches: 1-1453   

Operating System: Ubuntu 18.04

clangd: 10.0.0

## What went wrong

`:ALERename` command is not working properly with `clangd`. 
It messes up code. 
And if i undo changes by pressing `u` and quit without saving `:q!` changes are still saved in the file somehow.

## Reproducing the bug

1. Move to a symbole
2. Run :ALERename, type the new symbol name and hit
3. Everything gets messed up

### :ALEInfo

```
 Current Filetype: cpp
Available Linters: ['cc', 'ccls', 'clangcheck', 'clangd', 'clangtidy', 'clazy', 'cppcheck', 'cpplint', 'cquery', 'flawfinder']
   Linter Aliases:
'cc' -> ['gcc', 'clang', 'g++', 'clang++']
  Enabled Linters: ['clangd']
 Suggested Fixers: 
  'astyle' - Fix C/C++ with astyle.
  'clang-format' - Fix C/C++ and cuda files with clang-format.
  'clangtidy' - Fix C/C++ and ObjectiveC files with clang-tidy.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'uncrustify' - Fix C, C++, C#, ObjectiveC, ObjectiveC++, D, Java, Pawn, and VALA files with uncrustify.
 Linter Variables:

let g:ale_cpp_clangd_executable = '/home/anton/clangd_10.0.0/bin/clangd'
let g:ale_cpp_clangd_options = ''
let g:ale_cpp_gcc_options = '-std=c++11 -Wall'
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = v:null
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = v:null
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
[...]

PR #3344, opened 2020-09-09
Fix handling of ranges at file end.

dfn-certling:
Something along these lines solves the range ending at the file end in my case

### DIFF
--- a/autoload/ale/code_action.vim
+++ b/autoload/ale/code_action.vim
@@ -135,14 +135,7 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
         endif
 
         let l:lines_before_change = len(l:lines)
-
-        if l:end_line < len(l:lines)
-            let l:end = l:lines[l:end_line :]
-        else
-            let l:end = []
-        endif
-
-        let l:lines = l:start + l:middle + l:end
+        let l:lines = l:start + l:middle + l:lines[l:end_line :]
 
         let l:current_line_offset = len(l:lines) - l:lines_before_change
         let l:line_offset += l:current_line_offset
--- a/autoload/ale/code_action.vim
+++ b/autoload/ale/code_action.vim
@@ -127,10 +127,20 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
         endif
 
         call extend(l:middle, l:insertions[1:])
-        let l:middle[-1] .= l:lines[l:end_line - 1][l:end_column - 1 :]
+        if l:end_line <= len(l:lines)
+            " Only extend the last line if end_line is within the range of
+            " lines.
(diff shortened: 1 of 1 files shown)


autoload/ale/code_action.vim
```diff
@@ -127,10 +127,22 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
...
+        endif
 
         let l:lines_before_change = len(l:lines)
-        let l:lines = l:start + l:middle + l:lines[l:end_line :]
+
+        if l:end_line < len(l:lines)
+            let l:end = l:lines[l:end_line :]
+        else
+            let l:end = []
+        endif
```
w0rp, line 143: You don't need to check the length for a slice. You'll only ever get an error when you can't access an index. `x[1]` might raise an error, but `x[1:]` never will, it will give you `[]`.

w0rp (changes requested):

AlxHnr:
I've run into the same issue and can confirm it. Using your fix however leads to ALE breaking my code after renaming a variable with clangd. Here is what happens if I try to to rename `setValue` to `setValue123`:

Before:

```cpp
void Bar::setValue(int value) {
  this->value = value;
}
```

After:

```cpp
void Bar::setValue123(int value) { s->value = value; 
```

The new code is missing `}` and `this` gets replaced with `s`.

w0rp:
@AlxHnr Can you propose another patch on top of this, and attach the output of `ch_logfile` from Vim so we can add a test case? You have to use Vim for testing it to get a log file, as NeoVim doesn't offer message logging like Vim does.

AlxHnr:
> @AlxHnr Can you propose another patch on top of this, and attach the output of `ch_logfile` from Vim

No.

I assume this PR will already fix _this_ issue. The file scrambling problem is a different one, described in #3322. I've posted a more detailed, reproducible example there.

w0rp:
Okay, I asked for the information I'll need in that issue.

This patch can be merged once there's a test that covers this. Have a look at the test cases, or I'll do it myself at some point. (Could be weeks away from now. It's just whenever I have some spare time.)

dfn-certling:
I'm puzzled. Looking through the tests 

https://github.com/dense-analysis/ale/blob/08295ce17405cb5f6c80d2f726262493bfd21210/test/test_code_action.vader#L179

seems to directly test this. And indeed, in my setup this test already fails without my changes. So either this test already fails or there is something else at work.

w0rp:
I think this has been fixed in a different way according to recent changes. Let me know if you still have any issues.

w0rp (approved):

w0rp:
Apparently this is still needed, so I'll just merge it.

Cheers! :beers:

Issue #2469, opened 2019-04-30
Add support for pandoc


pbnj:
**Name:** pandoc
**URL:** pandoc.org

I have been contemplating the possibility of using `pandoc` as a fixer for some filetypes.

For example:

Given this markdown file:
```
# GitHub-Flavored Markdown File

    ## Some Malformed header

- bullet a
- bullet b
  - bullet c

some really really really really really really really really really really really really long line
```

With ALE configured to use `pandoc --from gfm --to gfm --columns=80` as a markdown fixer, invoking `ALEFix` would generate:

```
# GitHub-Flavored Markdown File

## Some Malformed header

  - bullet a
  - bullet b
      - bullet c

some really really really really really really really
really really really really really long line
```

Would this be a good idea to implement?
If so, I would also like the pandoc flags to be configurable by the user since pandoc supports various input/output filetypes.

petermbenjamin:
I also would not mind working on a PR for this, granted it is a good idea in the first place.

w0rp:
Sounds like a good idea to me.

hsanson:
Pandoc support as fixer already exists in ALE.

PR #3349, opened 2020-09-11
Add Pandoc for markdown formatting

Dophin2009:
Hi there, 

I'm not very experienced with Vim script, but I'd like to add [Pandoc](https://pandoc.org) as a fixer to format markdown files. This potentially closes issue #2469.

I'd imagine that this fixer would be better to have disabled by default, though I'm not sure how that's done.


Dophin2009:
Oops, I just realized there's another PR #3141 that adds Pandoc as a fixer as well.

PR #3354, opened 2020-09-14
Collapse spaces and lines in the completion menu.

rphln:
Some language servers (i.e., Rust's) send line breaks (which Vim renders as `^@`) and indentations on the completion details, which ends up taking a lot of space on the popup menu.

This patch collapses those characters into a single space.


w0rp (approved):

w0rp:
Cheers! :beers:

PR #3357, opened 2020-09-17
feat: add intelephense support for php

creativenull:
This PR will add initial support for [intelephense](https://github.com/bmewburn/intelephense-docs) for PHP.

I've added some tests to go with it. I've never done vim tests with vader before so I took inspiration from the tests implemented from `test/command_callback/test_php_langserver_callbacks.vader` file.

For the config options, I'm not too sure how I can check if `initialization_options` were properly read by intelephense. But I have only one test case for that, if someone knows how to do this I won't mind updating this PR with it.

### DIFF
--- a/ale_linters/php/intelephense.vim
+++ b/ale_linters/php/intelephense.vim
@@ -1,0 +1,32 @@
+" Author: Eric Stern <eric@ericstern.com>,
+"         Arnold Chand <creativenull@outlook.com>
+" Description: Intelephense language server integration for ALE
+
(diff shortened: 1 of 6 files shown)


creativenull:
@w0rp @hsanson  or any other maintainer out there to take a look at this? All checks are passed, I have couple tests added, unless more is needed.

w0rp (approved):

w0rp:
Cheers! :beers:

PR #3358, opened 2020-09-18
Improves fixer performance for large buffers

lukaswozniak:
Hi,

The usecase here is using the default trim_whitespace fixer.

This commit improves performance of ale#util#SetBufferContents . It turns out that setbufline function takes really long time (over 400 seconds for 7k lines file), which effectively made the fixer part of the plugin unusable for me. I found out that if I save the cursor pos, delete contents, append new contents, and restore cursor pos, then the same task is accomplished in less than 3 seconds for 7k lines file.

I didn't make a bug report or anything, because I don't think it is a problem of Ale plugin, but the internals of setbufline function. I also didn't bother to test this on different setups than mine, so it definitely needs further checking.

### DIFF
--- a/autoload/ale/util.vim
+++ b/autoload/ale/util.vim
@@ -486,7 +486,7 @@ function! ale#util#Input(message, value) abort
 endfunction
 
 function! ale#util#HasBuflineApi() abort
-    return exists('*deletebufline') && exists('*setbufline')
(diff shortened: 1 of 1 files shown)


ipod825:
This makes all fixers much fast in general.

w0rp (approved):
Looks good. We can assume `getpos` and `setpos` exist, so I'll remove those checks.

w0rp:
Cheers! :beers:

PR #3361, opened 2020-09-21
ALEInfo now shows enabled fixers

bukzor:
> Where are the tests? Have you added tests? Have you updated the tests? 

Test added to `test/test_ale_info.vader`: `ALEInfo should calculate enabled fixers correctly`

### DIFF
--- a/autoload/ale/debugging.vim
+++ b/autoload/ale/debugging.vim
@@ -197,6 +197,18 @@ function! s:EchoLSPErrorMessages(all_linter_names) abort
     endfor
 endfunction
 
+function! ale#debugging#FormatFixer(fixer) abort
+    let [l:name, l:Func] = a:fixer
+
+    if type(l:name) is v:t_string
+        return l:name
+    elseif type(l:Func) is v:t_func
+        return execute('function l:Func')
+    else
+        throw printf('Unexpected value. %s', a:fixer)
+    endif
+endfunction
+
 function! ale#debugging#Info() abort
     let l:buffer = bufnr('')
     let l:filetype = &filetype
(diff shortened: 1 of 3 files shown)


autoload/ale/debugging.vim
```diff
@@ -225,6 +225,9 @@ function! ale#debugging#Info() abort
     let l:fixers = uniq(sort(l:fixers[0] + l:fixers[1]))
     let l:fixers_string = join(map(copy(l:fixers), '"\n  " . v:val'), '')
 
+    let l:enabled_fixers = deepcopy(ale#fix#Get(l:buffer, '', []))
+    let l:enabled_fixers = map(copy(l:enabled_fixers), 'v:val[''name'']')
```
w0rp, line 229: There's no need to `deepcopy` here.

Where does `'name'` come from? `s:GetCallbacks` returns a List of two-item Lists with `[name, Callback]`. `name` can be `v:null`. You can select not only fixers from the registry, but also arbitrary functions for fixing code, like lambdas. Make sure lambda functions in `ale_fixers` are handled too. They need to be printed in a reasonable way.

w0rp (changes requested):
I think this is worth doing. See my comments here.

autoload/ale/debugging.vim
bukzor: > There's no need to deepcopy here.

This is just copy-pasta from line 205 where the same is done for linters. Should I change both?

> Where does 'name' come from?

Again, just copy-pasta from line 217. It worked for me and passed tests so I didn't think about it too hard.

> s:GetCallbacks returns a List of two-item Lists with [name, Callback]. name can be v:null. You can select not only fixers from the registry, but also arbitrary functions for fixing code, like lambdas. See :help ale-fix. Make sure lambda functions in ale_fixers are handled too.

Is there an example in the test suite of lambda fixers?

> They need to be printed in a reasonable way.

Is that possible? Is there an existing function for it?

PR #3362, opened 2020-09-23
Show tsserver hints/suggestions in Ale.

daliusd:


evmorov:
The first section of the documentation says:

> Populating the |loclist| with warning and errors

tsserver hints/suggestions are infos. Is it a bug that the loclist is populated with them? Personally, I don't like it. VSCode and IDEA don't populate "Problems" with hints either:

image

evmorov:
Is it possible to make a setting for that? coc.vim and vscode has `typescript.suggestionActions.enabled`.

daliusd:
Interestingly info is expected in loclist, e.g. ALENext has `-info` flag. From documentation:
```
  `-error`, `-warning` and `-info` enable jumping to errors, warnings or infos
    respectively, ignoring anything else. They are mutually exclusive and if
    several are provided the priority is the following: error > warning > info.
```

So question is what setting is for? So you can use `-error` and `warning` flags if info does not bother you.

evmorov:
> e.g. ALENext has -info

I'm not using `ALENext`. I have `let g:ale_open_list = 1`. In my workflow if there is something in the loclist then it must be fixed.

> So question is what setting is for?

To turn off tsserver hints.

daliusd:
I'm only user as well :smile:. One idea: `g:ale_lsp_show_message_severity` should be respected by tsserver as well. That would solve this problem. However I'm not sure if I'm right of wrong here. Let's wait for more input from core developers (e.g. @w0rp)

evmorov:
Nevermind. I've started to use `ALENext` and set `g:ale_open_list` to `0`. I'll continue using your branch until this PR is merged.

daliusd:
@w0rp any comments?

w0rp (changes requested):
This is almost good. Just make sure that the items use the type `'I'` for "info", and that we have a setting to ignore suggestions. I think it's worth turning them on by default, so people are aware they exist, and letting people turn suggestions off with one setting. Let's call this setting `ale_lsp_suggestions`, set it to `1` by default, and document that setting it to `0` will disable suggestions for `tsserver` and from LSP servers.

We could add `ale_minimum_severity` at some point for everything, but I think it's worth being able to control specifically suggestions from `tsserver` and LSP servers as a separate setting.

daliusd:
Sorry, I will not be able to fix it completely as I don't see using tsserver with Ale myself in the future.

daliusd:
Update:

Items use type 'I' for suggestions. It is set in `ale/autoload/ale/lsp/response.vim` line 109:
```
        if get(l:diagnostic, 'category') is# 'suggestion'
            let l:loclist_item.type = 'I'
        endif
```

I'm working on setting now.

daliusd:
@w0rp please review again

daliusd:
@w0rp please merge this one as well. I personally find it useful.

daliusd:
@w0rp :wink:

hsanson (approved):
Thanks for the patience. This looks good to me.

daliusd:
🍻

muzuiget:
Finally found this issue, this change confuse me.

Usually I get same result from command `tsc` and `tsserver` in Ale, so I can lint my code without open it in Vim.

After this change, when I run `tsc`, no warnings/errors, but open it in Vim, some warnings/errors exits, eg:

```
6133: 'url' is declared but its value is never read.
```

This message is look like it lint by `tsc --noUnusedParameters` or `unusedParameters=true` in `tsconfig.json`,  it looks like `tsc` and `tsserver` treat `tsconfig.json` in difference behavior. It takes some time to find this issue.

After reading above comments, set `ale_lsp_suggestions=1` seen to solve my problem.

w0rp:
I think you won't be alone in finding that confusing. I'll turn them off by default.

PR #3216, opened 2020-06-24
Fix 3103 - add shellcheck shell directive detection.

hsanson:
Searches for shellcheck shell directive to detect dialects for scripts
that do not have shebang.

Closes #3103

PR #3364, opened 2020-09-24
Add support for shellcheck "shell" directives

IngoMeyer441:
shellcheck allows to use a special shell directive instead of a shebang
line in shell scripts to specify the shell type. This is useful for
sourced scripts which won't be executed by a shell directly and thus
don't need a shebang line.


hsanson:
@IngoMeyer441  There is a PR that adds support for this already: #3216, if you see anything that can be improved on that PR or if you think this PR is better please comment.

IngoMeyer441:
@hsanson Sorry, did not see this! :sweat_smile: Your PR looks good to me and is better than mine (you check more than the first line and the configured shell and you wrote tests :smile:). So, my PR can be closed.

PR #3381, opened 2020-10-03
Use stdin/stdout to communicate with gofmt

andreypopp:
Problem: I use docker based workflow where all devtools including `gofmt` are inside a container. The vim is running on host system though and runs `gofmt` via a wrapper script which does `docker exec gofmt`. Because currently fixer uses temp files `gofmt` cannot see those files produced on the host system.

Solution: use stdin/stdout to communicate with `gofmt`

I might be missing something why it's not using stdin/stdout currently.

cc @aliou @eliath as contributors to this fixer

### DIFF
--- a/autoload/ale/fixers/gofmt.vim
+++ b/autoload/ale/fixers/gofmt.vim
@@ -11,9 +11,6 @@ function! ale#fixers#gofmt#Fix(buffer) abort
 
     return {
     \   'command': l:env . ale#Escape(l:executable)
-    \       . ' -l -w'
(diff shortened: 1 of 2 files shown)


andreypopp:
@w0rp hi, can you take a look 👀?

andreypopp:
Not stale!

hsanson (approved):
Simple and works fine with my limited testing on some small go projects. Probably gofmt did not support stdin/stdout before and that is why ALE did not use it this way.

PR #3384, opened 2020-10-04
feat: Add support for Erlang escript

m-ildefons:
Add support and documentation for Erlang escript.

Conventional erlang linters like `erlc` choke on shebangs in escript
files, which can cause them to not report on other language violations
or report false-positives. `escript` does have a syntax and semantics
check mode of its own. This linter utilizes that to check escript files.


w0rp:
I've reopened this. Add some tests for this.

m-ildefons:
Hi @w0rp ,
thanks for revisiting this. I've put tests analogous to the dialyzer linter in. I hope this suffices.

PR #3386, opened 2020-10-06
Make isort fixer recognize auto_pipenv flag

ivorpeles:
I noticed with `python_auto_pipenv = 1` the `black` fixer would run on any file in a project containing a Pipfile whether vim was run inside the virtual environment or not. I wanted to make `isort`'s fixer behave the same way in conjunction with the `python_auto_pipenv` flag.

What I have below generally works although I do have one concern: the `!executable(l:executable)` check fails when the executable ends up being `pipenv`. I added a hacky workaround to exclude this case but ideally there would be a test fixture that makes `executable(pipenv)` evaluate as true. We would expect this for any user setting `python_auto_pipenv` or `python_isort_auto_pipenv` to 1.

Would appreciate some feedback here, thanks!

### DIFF
--- a/autoload/ale/fixers/isort.vim
+++ b/autoload/ale/fixers/isort.vim
@@ -24,7 +24,7 @@ function! ale#fixers#isort#Fix(buffer) abort
     \   ? ' run isort'
     \   : ''
 
-    if !executable(l:executable)
+    if !executable(l:executable) && l:executable != 'pipenv'
         return 0
     endif
 
--- a/autoload/ale/fixers/isort.vim
+++ b/autoload/ale/fixers/isort.vim
@@ -24,7 +24,7 @@ function! ale#fixers#isort#Fix(buffer) abort
     \   ? ' run isort'
     \   : ''
 
-    if !executable(l:executable) && l:executable !=# 'pipenv'
+    if !executable(l:executable) && l:executable isnot# 'pipenv'
         return 0
     endif
 
--- a/autoload/ale/fixers/isort.vim
+++ b/autoload/ale/fixers/isort.vim
@@ -24,7 +24,7 @@ function! ale#fixers#isort#Fix(buffer) abort
     \   ? ' run isort'
     \   : ''
 
-    if !executable(l:executable) && l:executable != 'pipenv'
+    if !executable(l:executable) && l:executable !=# 'pipenv'
         return 0
     endif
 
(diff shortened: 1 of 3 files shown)


autoload/ale/fixers/isort.vim
```diff
@@ -2,24 +2,35 @@
...
-    \   ['isort'],
-    \)
+    let l:executable = ale#fixers#isort#GetExecutable(a:buffer)
+
+    let l:exec_args = l:executable =~? 'pipenv$'
+    \   ? ' run isort'
+    \   : ''
 
-    if !executable(l:executable)
+    if !executable(l:executable) && l:executable != 'pipenv'
```
crimsonknave, line 27: The current CI failure is complaining about `!=` here: `autoload/ale/fixers/isort.vim:27:50: Use robust operators `!=#` or `!=?` instead of `!=` (see Google VimScript Style Guide (Matching))`

crimsonknave (commented):
I'd love to see this! It appears that the PR adding similar function for `black` simply removed that section: https://github.com/dense-analysis/ale/pull/1988/files#diff-b9a6cf2d19edcb105ed741db26eed8b09f0be5348801b806e9dca68050255756L15-L17.

autoload/ale/fixers/isort.vim
ivorpeles: Looks like it also wanted `isnot` in place of `!=`, more robust when one of the operands isn't a string

ivorpeles:
@crimsonknave Sorry for the massive delay here, took me a while to re-acquaint myself with whatever I was doing here. Tests are fixed and it works for me locally, let me know if anything else needs doing. Thanks!

crimsonknave:
@ivorpeles Thanks for fixing it up! I'm just a user of ALE, so I have no real idea what else if anything you need to do. I was just looking forward to this getting merged and took a look at what the black plugin did to see if I could help get it unstuck.

ivorpeles:
Oh cheers, thanks!

ivorpeles:
@w0rp didn't see anything in the contributing docs about review process, could I get a review on this?

ivorpeles:
@hsanson bumping for review, this has been open + passing tests for a while. Would be nice to get it merged

hsanson (approved):
Not really familiar with pipenv or how this is supposed to affect the fixed but comparing the changes with the black fixer seems to be correct.

PR #3390, opened 2020-10-08
dafny: include correct filename in lint results

benknoble:
Results can come from included files, not just the current buffer.


benknoble:
Not stale. /cc @w0rp again

benknoble:
Still not stale. /cc @w0rp

hsanson (approved):
Simple and seems correct.

hsanson:
@benknoble thanks for the persistence. I avoid reviewing or merging stuff I don't use on a daily basis since I have no way to properly test it but this change is small enough and I believe you use this frequently so it must work as expected.

PR #3402, opened 2020-10-20
Add verible verilog linter

zegervdv:
Adds [verible](https://github.com/google/verible) linting for verilog/SystemVerilog.

I've added an options test and updated the default linters test.


zegervdv:
@w0rp is this ready to merge, or do I need to add more tests?

PR #3403, opened 2020-10-21
Add Ruby sorbet option to enable watchman

will:
Follow up to https://github.com/dense-analysis/ale/pull/2614

The sorbet support already in ale works great, and I think it makes sense to default to `--disable-watchman` as @rintaun did in the original PR. 

A little while ago I had hacked up my copy in my `.vim` dir to remove that though, so files changed outside of vim (such as auto generated rbi files) would be picked up by the lsp server. 

I figured it'd be better to contribute that back up to ale though. I've done very little vim script though, so this was all just done by looking around the project a bit, so I'm not sure if things are wrong. It seems to work for me though, for what it's worth.

### DIFF
--- a/ale_linters/ruby/sorbet.vim
+++ b/ale_linters/ruby/sorbet.vim
@@ -1,14 +1,17 @@
 call ale#Set('ruby_sorbet_executable', 'srb')
 call ale#Set('ruby_sorbet_options', '')
+call ale#Set('ruby_sorbet_enable_watchman', 0)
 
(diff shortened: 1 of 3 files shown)


will:
It's cool that it all works so fast, even though there are many moving pieces. Thanks for making ale quick :)
vimalelsp

will:
posting to keep that sale bot away

hsanson (approved):
Looks good.

will:
Thanks!

will:
1337 closed pr  😎 
image

PR #3407, opened 2020-10-22
Make executable for nix linter configurable

LnL7:
Especially useful when using the development release of nix, since the error format changed there.


Ma27:
@Lnl7 you may want to test https://github.com/dense-analysis/ale/pull/3227 for `nixUnstable` :)

PR #3422, opened 2020-10-31
add windows iconv

cyyever:

### DIFF
--- a/autoload/ale.vim
+++ b/autoload/ale.vim
@@ -260,6 +260,13 @@ function! ale#GetLocItemMessage(item, format_string) abort
     " Windows may insert carriage return line endings (^M), strip these characters.
     let l:msg = substitute(l:msg, '\r', '', 'g')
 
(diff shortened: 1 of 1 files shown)


autoload/ale.vim
```diff
@@ -260,6 +260,13 @@ function! ale#GetLocItemMessage(item, format_string) abort
     " Windows may insert carriage return line endings (^M), strip these characters.
     let l:msg = substitute(l:msg, '\r', '', 'g')
 
+    if has('win32')
+        let l:to = &encoding
+        let l:res = iconv(l:msg, 'default', l:to)
+        if l:res !=# '' && stridx(l:res, '??') == -1
+            let l:msg=l:res
+        endif
+    endif
```
w0rp, line 269: There's no description for this pull request. What does this do, and why do we need it? Could you add a test for this?

w0rp (changes requested):

PR #3427, opened 2020-11-02
Check for Vim's shell type buffer variables before falling back to the file type

lifecrisis:
This PR fixes an interesting error I noticed when working on shell scripts that don't have a `#!` line.

To reproduce:
1. create a "test.bash" file, with no `#!` line
2. open the file and verify that `b:is_bash` is set
3. notice that, when using `shellcheck` the file is linted as a simple `sh` file

This is because we fall back to the file type before checking for buffer-local shell type variables.

See my commit messages for more details.

Thanks!


hsanson (approved):
Good catch.

PR #3436, opened 2020-11-13
Add dprint fixer

myobie:
Useful when working on deno projects.


myobie:
I think this should be kept alive.

myobie:
I still think this should be kept alive.

hsanson (changes requested):
@myobie sorry for taking too long. I do not actively develop in Javascript/Typescript .  Currently the PR has conflicts with master that should be easy to solve.

Also I can see from dprint documentation that it support STDIN input. This is optional but I recommend use STDIN instead of temporary files. This ensures that any local configuration works as expected.

PR #3441, opened 2020-11-17
Add support for local solhint executables

hbarcelos:
This is a follow-up PR for [#3305 ](https://github.com/dense-analysis/ale/pull/3305).

I've rebased my code from the master branch after the release of version 3.x.
The tests which were previously failing are now passing when I run `:Vader` on the test file.


chanhosuh:
ALE isn't looking for a local solhint: `(executable check - failure) solhint`

I assume it's because this PR was never merged...  can we fix this and get it in?  Also it'd be good to have the same search logic for `solc` too.

hbarcelos:
This PR is stale because if was not merged and a new major version was released in the meantime.
There is a (new PR)[https://github.com/dense-analysis/ale/pull/3682] for this though.

Issue #1466, opened 2018-04-02
Add support for LSP/tsserver Code Actions


rlerm:
Right now, there is no support for using :ALEFix for LSP-based linters. The protocol supports this (via codeAction), although it might be somewhat complex.

w0rp:
After giving this some thought, I think the best way to support fixing files via Language Server Protocol is to not apply fixes to files in the pipeline of fixes for `:ALEFix`, but to handle code actions separately. Code actions will likely need to write the files they change to disk in many cases, and will also support actions on ranges. What ALE ought to do is ask the server what actions are available for either a range in a document or the whole document, and then have you selection an option to apply.

ALE should support code actions primarily via some commands, and secondarily via the GVim mouse-controlled context menu. There should be some way to issue code actions without first loading the list of actions.

w0rp:
If there's something to update this issue with, I'll post a comment. What I said before about code actions still applies.

jazmit:
+1. Code actions are the last blocker before I can dispense with VSCode.  Maybe it would make sense to rename this issue to 'Support code actions'?  Are there any known workarounds/hacks to get code actions working?

w0rp:
https://github.com/dense-analysis/ale/commit/ca2026bd13346cb7e1afeecb54dbec3da0bd5151

I was playing around with adding support for code actions with popup menu support, so you can just right click on regions of code to perform action. GVim supports updating the popup menu while it's still open, and you can trigger requesting some actions whenever the menu is opened, so it will be pretty good.

I'll try to get back on this when I can. I plan to get the popup menu support to be pretty good first, and then add console menu support second for terminals. The hardest part is just applying edits to files that may or may not be open in Vim already.

w0rp:
Parts of the code required for implementing code actions were previously added to the codebase to support rename operations and organising imports. See here: #2709

NikosEfthias:
:+1:

NikosEfthias:
just waiting for this feature to go back to ale from vim-lsp I love ale's plug-n-play approach

daliusd:
If would appreciate if somebody would test my `tsserver` code actions pull request. In order to do that you need to:

1. `git clone https://github.com/daliusd/ale.git`

1. `git checkout codefix`

1. Add in your .vimrc (with assumption that you are using Plug): `Plug '~/projects/ale'` (with assumption that you have cloned ale into projects folder).

1. Add mappings for `ALECodeAction`:

```
[...]

PR #3442, opened 2020-11-17
Code actions fix: diagnostics end character change

daliusd:
@leo60228 proposed fix for clangd here:
https://github.com/dense-analysis/ale/issues/1466#issuecomment-729091829

This specific fix is not harming `typescript-language-server` and
`jedi-language-server` code actions (the only LSPs that were tested with
code actions implementation). Therefore if it is helping with `clangd` I
think that's what we want to have.


w0rp:
I have redone this fix based on the changes I just pushed for the GUI right click menu support.

Cheers! :beers:

PR #3461, opened 2020-11-24
feat: add yamlfix fixer

lyz-code:
Add yamlfix fixer support

### DIFF
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -110,6 +110,11 @@ let s:default_registry = {
 \       'suggested_filetypes': [],
 \       'description': 'Remove all trailing whitespace characters at the end of every line.',
 \   },
+\   'yamlfix': {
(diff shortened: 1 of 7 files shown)


autoload/ale/fix/registry.vim
```diff
@@ -110,6 +110,11 @@ let s:default_registry = {
 \       'suggested_filetypes': [],
 \       'description': 'Remove all trailing whitespace characters at the end of every line.',
 \   },
+\   'yamlfix': {
+\       'function': 'ale#fixers#yamlfix#Fix',
+\       'suggested_filetypes': ['python'],
```
hsanson, line 115: Is this correct? I would expect the suggested filetype to be yaml.

lyz-code: You're right @hsanson , fixed in 12eb8d15.

Thanks

hsanson (approved):
Looks good to merge.

Issue #3448, opened 2020-11-22
Unit tests don't run on master via ./run_tests and docker


kevinclark:
On 681a6e371d02 (which is top of master at the moment). Logs [here](https://gist.github.com/kevinclark/bc457f247940dd80c87c93149ddadde4). Though I ran it with `--fast -q` in the log, it fails without those flags as well.

I'm on macOS 10.15.7. Docker version 19.03.13, build 4484c46d9d. Let me know what other versions might matter. I see CI is passing and figured docker should be isolating the environment, but 🤷‍♂️

kevinclark:
Looks like there's a few issues. One is that many of the test scripts run grep/sed/ls/sort using the local system instead of the docker instance and then you get slight differences in implementation that causes failures. Another is that the tag reference check is using a unified diff to look for invalid references, but it just does it based on a '+' line, which doesn't allow for reordering. F.e:

```diff

-ALEComplete
 ALECompletePost
-ALECompletePost-autocmd
+ALEComplete
```

Here's a case where `ALEComplete` shows up at an invalid reference when it's just out of expected order.

I'm working up a PR to resolve these issues since it blocks my work on #3470. It's going to result in some of the doc-sorting enforcement changing order slightly, but it'll be consistent across platforms.

hsanson:
@kevinclark can this be closed now that #3471 is merged?

kevinclark:
Yep!

PR #3471, opened 2020-11-28
Make test/scripts/custom-checks portable

kevinclark:
As reported in #3448, `./run-tests` doesn't work universally. The issue appears to be partly a dependence on system tools (grep and sed in particular) from the user environment which aren't exactly equivalent across systems. This PR moves the execution of those checks into the docker environment, which is updated to use gnu versions of the tools which are more robust than the busybox equivalent.

As a result of the change, the TOC and supported-tools tests, which rely on sort order, where changed slightly. The custom-linter-rules test now, seemingly correctly, fails because there are several lint names that aren't snake case. It's an open question what to do about that as I suspect there are users of those lints that need to be supported. Grandfathered in whitelist?

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -9,8 +9,8 @@ set -u
 # options, or read the output below.
 #
 
-image=w0rp/ale
-current_image_id=f58c7bf8900f
+image=akevinclark/ale
+current_image_id=8e1802ca34b5
 
 # Used in all test scripts for running the selected Docker image.
 DOCKER_RUN_IMAGE="$image"
--- a/test/script/check-tag-alignment
+++ b/test/script/check-tag-alignment
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+exit_code=0
+
+# Documentation tags need to be aligned to the right margin, so look for
+# tags which aren't at the right margin.
+grep ' \*[^*]\+\*$' doc/ -r \
+    | awk '{ sep = index($0, ":"); if (length(substr($0, sep + 1 )) < 79) { print } }' \
+    | grep . && exit_code=1
+
+exit $exit_code
--- a/test/script/custom-checks
+++ b/test/script/custom-checks
@@ -46,7 +46,9 @@ echo '========================================'
 echo 'Differences follow:'
 echo
 
-test/script/check-supported-tools-tables || exit_code=$?
+set -o pipefail
+docker run "${docker_flags[@]}" test/script/check-supported-tools-tables || exit_code=$?
+set +o pipefail
 
 echo '========================================'
 echo 'Look for badly aligned doc tags'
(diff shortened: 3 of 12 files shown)


run-tests
```diff
@@ -9,8 +9,8 @@ set -u
 # options, or read the output below.
 #
 
-image=w0rp/ale
-current_image_id=f58c7bf8900f
+image=akevinclark/ale
+current_image_id=8e1802ca34b5
```
kevinclark, line 13: Note that this change likely needs to be reverted back to @w0rp's docker if this is to be merged. But in the meantime I wanted CI to be able to use the new Dockerfile.

test/script/check-tag-alignment
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+exit_code=0
+
+# Documentation tags need to be aligned to the right margin, so look for
+# tags which aren't at the right margin.
+grep ' \*[^*]\+\*$' doc/ -r \
```
kevinclark, line 7: I think it's debatable whether this is working as intended. Because of the leading space, tags on the far left margin are missed. If I remove it, I get these back:

```
doc/ale-development.txt:*ale-dev*
doc/ale-development.txt:*ale-development*
doc/ale-powershell.txt:*g:ale_powershell_psscriptanalyzer_executable*
doc/ale-powershell.txt:*g:ale_powershell_psscriptanalyzer_module*
doc/ale-powershell.txt:*g:ale_powershell_psscriptanalyzer_exclusions*
doc/ale-supported-languages-and-tools.txt:*ale-supported-list*
doc/ale.txt:*ale*
```

Some of those (ale-dev, ale) are heading tags and seem like they should potentially be allowed. But the [powershell](https://github.com/dense-analysis/ale/blob/master/doc/ale-powershell.txt#L30) examples seem pretty awkward and maybe not in the spirit of the check?

test/script/custom-checks
```diff
@@ -13,7 +13,7 @@ echo 'Custom warnings/errors follow:'
 echo
 
 set -o pipefail
-docker run -a stdout "${docker_flags[@]}" test/script/custom-linting-rules . || exit_code=$?
```
kevinclark, line 16: Only binding stdout here was hiding the warnings, which went to stderr.

kevinclark:
@w0rp ping

kevinclark:
This now grandfathers in the non-snakecased lints. Namely:

  * clojure/clj_kondo.vim
  * elixir/elixir_ls.vim
  * go/golangci_lint.vim
  * swift/swiftformat.vim

I believe this is ready to go.

kevinclark:
@hsanson Looks like you're the person actually merging PRs. Can you take a look at this?

w0rp (approved):
This is all good. I'll merge this in, generate a new Docker image, and swap it out.

The Powershell file where the tags aren't matched by the regex aren't formatted correctly, so I wouldn't worry about it too much. That file just needs to be fixed.

w0rp:
Cheers! :beers:

kevinclark:
Thank you!

PR #3475, opened 2020-11-30
dafny: add a timeLimit option and message-parsing

benknoble:
The option itself was manually tested.

The message-parsing has a new test.

#3390 should probably merge first, which will necessitate some updates to this PR.

### DIFF
--- a/ale_linters/dafny/dafny.vim
+++ b/ale_linters/dafny/dafny.vim
@@ -14,13 +14,28 @@ function! ale_linters#dafny#dafny#Handle(buffer, lines) abort
         \ })
     endfor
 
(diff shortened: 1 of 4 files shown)


benknoble:
@w0rp should be passing

hsanson (approved):
Looks good. Tests in order too and no visible backward compatibility issues.

benknoble:
@hsanson should we update `bufnr` to `filename` now that #3390 is merged, too? I'm actually surprised the tests passed after that

hsanson:
@benknoble I do not use Dafny but from what I have read it seems using filenames would be useful in this case.  It will allow ALE to show errors  on files still not open (no buffer) in vim/nvim.

PR #3478, opened 2020-12-02
Fixes to code actions (cursor moving, tests, EOL/EOF corner cases)

liskin:
This PR adds tests for many corner cases (mostly involving newlines and changes near newlines) around applying changes from LSP, fixes these corner cases, and also adds a workaround for LSPs that just replace the entire document instead of sending individual changes.

More details in the individual commits:

---

commit fa8398970918200d1237988a2fd5598e51571ae0

    Minor cleanup in test/test_code_action.vader
    
    Dropped unnecessary sourcing, which probably originated from
    test_codefix.vader which overrides some functions, but
    test_code_action.vader does not.
    
    Also, Save &fixeol before changing it, and Restore at the end after
    closing buffers, otherwise fixeol isn't fully restored.

commit 354b5b74a96f68cd7f66836678831b1a19443b9d

    code_action: Don't move cursor when change covers entire file
    
    This prevents the cursor jumping to end of file after every
    rename/refactor. This happens with e.g. python-language-server which
    rewrites the entire buffer regardless of backend:
    
    https://github.com/palantir/python-language-server/blob/3d045b8db9b53d688f433fb2f2e5df27cbc1f788/pyls/plugins/jedi_rename.py#L31-L37
    https://github.com/palantir/python-language-server/blob/10cd98c898c89a8c9961b25725a6cd913485e1d0/pyls/plugins/rope_rename.py#L43-L49

commit a4b7069e7f80345731848e21a506966da765dc90

    code_action: Refactor/simplify ApplyChanges
    
    No behaviour change, just simplification that will make further fixes
    a bit easier.

commit 53f8856288bc454dbbe2fcc4804869dd06e5515f

    code_action: Fix EOL at EOF corner cases while performing no changes
    
    Instead of always removing the trailing empty line and hoping it wasn't
    supposed to be there, only remove it when we know ale#util#Writefile
    will add a trailing newline, and additionally tell ale#util#Writefile to
    not add that newline if any change explicitly deleted that newline.
    
    (Unfortunately with a:should_save=0, adding a trailing newline that
    wasn't there results in noeol and empty line at end of buffer, because
    for some reason setbufvar(l:buffer, '&eol', 1) doesn't work in buffers
    that started as noeol. This is likely a bug in vim itself.)
    
    Includes a minor fix that prevents invoking ale#util#SetBufferContents
    without a buffer. Not sure whether that might really happen anywhere.
    
    (test_code_action_python.vader depends on two incorrect behaviours
    cancelling each other, so it's temporarily adjusted to pass)

commit bae3c8bc528ae4f134b680544caa0e9d8764d693

    code_action: Fix column around EOL corner cases
    
    The original code didn't allow changing the last character in a line
    without also replacing the newline, due to an off-by-one error.
    
    Or, from a different point of view, it didn't allow replacing the
    newline without also replacing the preceding character. Therefore it
    comes as no surprise that it didn't correctly handle whether to replace
    or leave the newline at the end of the changed range.
    
    This commit fixes both, and also adjusts two tests that depended on the
    incorrect behaviour. Also, the temporary change to
    test_code_action_python.vader introduced earlier can now be reverted.
    
    Finally, this adds a _lot_ of tiny unit tests for various corner cases.
    These were verified against the reference implementation in
    [vscode-languageserver-node][], and a javascript source to run them is
    added in a followup commit. The added tests take an additional second or
    two. It might be worth reducing them somewhat to only test what was
    broken…?
    
    [vscode-languageserver-node]: https://github.com/microsoft/vscode-languageserver-node/

commit b96472443a102ee63fd21c176cdbdaf47981d5c7

    code_action: Handle positions out of bounds
    
    This makes the code more robust, and perhaps even a little easier to
    understand: less exceptional cases, more bounds clamping.

commit 12b40b7e7a8f8cb46df689d1be6a8a91ce7c40c4

    code_action: Add instructions for verifying corner case tests against vscode

### DIFF
--- a/autoload/ale/code_action.vim
+++ b/autoload/ale/code_action.vim
@@ -96,21 +96,34 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
 
         let l:insertions = split(l:text, '\n', 1)
 
-        " Special case when text must be added after new line
-        if l:column > len(l:lines[l:line - 1]) + 1
-            let l:line += 1
-            let l:column = 1
+        " Fix invalid columns
+        let l:column = l:column > 0 ? l:column : 1
+        let l:end_column = l:end_column > 0 ? l:end_column : 1
+
+        " Clamp start to BOF
+        if l:line < 1
+            let [l:line, l:column] = [1, 1]
         endif
 
-        " Adjust end if we moved start past end
+        " Clamp start to EOF
+        if l:line > len(l:lines) || l:line == len(l:lines) && l:column > len(l:lines[-1]) + 1
+            let [l:line, l:column] = [len(l:lines), len(l:lines[-1]) + 1]
+        " Special case when start is after EOL
+        elseif l:line < len(l:lines) && l:column > len(l:lines[l:line - 1]) + 1
+            let [l:line, l:column] = [l:line + 1, 1]
+        endif
+
+        " Adjust end: clamp if invalid and/or adjust if we moved start
         if l:end_line < l:line || l:end_line == l:line && l:end_column < l:column
             let [l:end_line, l:end_column] = [l:line, l:column]
         endif
 
-        " Special case when end is after new line
-        if l:end_column > len(l:lines[l:end_line - 1]) + 1
-            let l:end_line += 1
-            let l:end_column = 1
+        " Clamp end to EOF
+        if l:end_line > len(l:lines) || l:end_line == len(l:lines) && l:end_column > len(l:lines[-1]) + 1
+            let [l:end_line, l:end_column] = [len(l:lines), len(l:lines[-1]) + 1]
+        " Special case when end is after EOL
+        elseif l:end_line < len(l:lines) && l:end_column > len(l:lines[l:end_line - 1]) + 1
+            let [l:end_line, l:end_column] = [l:end_line + 1, 1]
         endif
 
         " Careful, [:-1] is not an empty list
@@ -119,14 +132,9 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
 
         let l:middle[-1] .= l:insertions[0]
         let l:middle     += l:insertions[1:]
+        let l:middle[-1] .= l:lines[l:end_line - 1][l:end_column - 1 :]
 
-        if l:end_line <= len(l:lines)
-            " Only extend the last line if end_line is within the range of
-            " lines.
-            let l:middle[-1] .= l:lines[l:end_line - 1][l:end_column - 1 :]
-        endif
-
-        let l:end_line_len = l:end_line > len(l:lines) ? 0 : len(l:lines[l:end_line - 1])
+        let l:end_line_len = len(l:lines[l:end_line - 1])
         let l:lines_before_change = len(l:lines)
         let l:lines = l:start + l:middle + l:lines[l:end_line :]
 
--- a/autoload/ale/code_action.vim
+++ b/autoload/ale/code_action.vim
@@ -89,38 +89,20 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
         let l:end_column = l:code_edit.end.offset
         let l:text = l:code_edit.newText
 
-        " Adjust the ends according to previous edits.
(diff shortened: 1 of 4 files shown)


liskin:
@jeremija @daliusd You folks authored most of the code I'm touching, so pinging you if you want to have a look. Thanks!

daliusd:
It is cool to see that you have decided to improve existing code. So far everything looks good. Cheers 🍺

jeremija:
@liskin Thanks for pinging me. Previously, I did notice there were a few corner cases in which the LSP rename messed up my document, but I never had the time to debug it - thanks for taking the time to do it. I'll be sure to try out your changes over the next few days!

liskin:
TODOs (add instructions for verifying the tests against vscode reference impl, and rebase) done. I believe this PR is now really ready for review. @daliusd if you'd like to have a look at the commits I added since Wednesday they're ready.

daliusd:
IMHO, everything is OK.

tmc:
What's next to merge this?

daliusd:
@w0rp @hsanson or someone with merge rights should review and approve this.

davidatbu:
@hsanson , I believe that you're holding off on some PRs because you think they need @w0rp 's attention, but I wasn't sure if you had seen this at all. Thank you for your time!

hsanson:
@davidatbu thanks for the heads up. I do avoid merging changes that affect core parts of ALE. Those need to be reviewed by @w0rp (in my opinion).  For added linters/fixers/lsp tools that I can actually test on my setup I review and merge when time allows.

davidatbu:
I've been trying out this fork for a couple of days now, and there are sporadic occurrences of ALE completely truncating in an external file file when trying to use `ALERename` on a variable defined in that file. I'm trying to come up with a small reproducible example, but I can't figure out what causes it even after trimming down my vimrc to very minimal. I was wondering if anyone else experienced this.

FYI, I'm using `pyright` as an LSP for Python files.

hsanson:

```
# needed if already added.

git checkout master
git fetch upstream master
git merge upstream/master

git checkout my-branch-name

```

liskin:
@hsanson A rebase isn't strictly necessary as GH merges PRs with master before running the checks, so I've only forcepushed to re-trigger the tests. (I don't mind rebasing, I just wanted to double check that I'm not incorrect.)

@davidatbu I do have to admit that I haven't done a lot of renaming since creating this PR (to be entirely honest, most of this PR was done for fun rather than solving a specific problem that I had), so if there were some problems, I wouldn't encounter them anyway. Also, I use pyls which always replaces the entire file contents, so it's unlikely to trigger any interesting bugs.

Anyway, the issue you described should definitely be taken care of. Perhaps the best way to debug this is to set up `ale#code_action#ApplyChanges` to dump the `changes` array to a file so that when this happens, we can look back and see what might be triggering this. Will you be able to do something like that or shall I try to come up with a patch that enables such logging?

w0rp (approved):
It's hard to follow the code, just by the nature of what is being done, but that doesn't matter because the extensive tests you've provided are good. :+1: I'll merge this now, and if anyone has any issues with it, I'll let you know.

w0rp:
Cheers! :beers:

I really appreciate all of the time and care you've put into this.

liskin:
Oh, but I'm a bit sad that you squashed the commits and threw all my comments away. I really did try to make the code changes, if not easy, at least possible to follow, and now that's been dumped. If anyone comes and finds a problem, I'll need to look into a backup of my branch to see my individual commits. :-/

w0rp:
I don't like to have git commit messages that are way, way too long, as it makes it hard to read the `git log`. Anything that long ought to be written in test comments or documented in the code or documentation.

liskin:
If you want git log to be easy to read, use `--pretty=oneline` or `tig` or something. There really is no good reason to just cut all that info. It's good for reviewers, and it's good for anyone wondering why a certain change was made. Comments are good, but they can only explain why the code is what it is now, not how it became to be, which is often important when debugging.

If you really intend to throw all this away, then I'm afraid I'll need to reconsider my future contributions. :-/

PR #3491, opened 2020-12-17
Add salt-lint support

Poulpatine:
This PR adds support for salt-lint which is a linter for salt-stack files (see https://github.com/warpnet/salt-lint and https://github.com/saltstack/salt ).


hsanson (approved):
Looks very good. Thanks for the work.

PR #3417, opened 2020-10-29
Add linter for Inko

YorickPeterse:
This adds a linter for Inko (https://inko-lang.org/). The linter makes use of Inko's own compiler, and a newly introduced --check flag to only check for errors; instead of also compiling source code.

PR #3494, opened 2020-12-18
Add linter for Inko

YorickPeterse:
This adds a linter for Inko (https://inko-lang.org/). The linter makes use of Inko's own compiler, and a newly introduced --check flag to only check for errors; instead of also compiling source code.

This was originally submitted in https://github.com/dense-analysis/ale/pull/3417, but was closed by the stale PR bot. Since I haven't received any comments since asking it to be reopened, I'm resubmitting the PR here.


hsanson (approved):
I do not use Inko so have not tested this. Trusting @YorickPeterse uses this and works as expected.

YorickPeterse:
@hsanson Thanks!

PR #3261, opened 2020-08-02
Enhanced Dhall fixer support

toastal:
Should fix #2820 since it's just asking about `format` and `lint`, but there's room to also do the language server, but one thing at a time. I also included `freeze` as a fixer as well to get sha256 for remote resources which is handy.

There's support for adding options to the main `dhail` executable such as `--ascii` (something `dhall-vim` doesn't support) as well as flags for `freeze` and `--all` if you'd really want that.

- - -

[PR #3310](https://github.com/dense-analysis/ale/pull/3310) somewhat addresses Dhall support (only `dhall format`), but the author forgot to search for the appropriate referencing the linked issue, #2820, or helping out with this already-open PR (which has more features), and was merged in before this PR got any attention :disappointed:. That PR, #3310, doesn't provide any configuration options (I know some people will want the the `--ascii` flag for sure), and doesn't include the other fixer-related tools, `dhall-lint` or `dhall-freeze`. This PR, #3261, should replace it.

However, that PR#3310 did appropriately try to find the file as a Stack binary as well.

PR #3501, opened 2020-12-27
Add support for `--ascii` for dhall format

imran-iq:
I just started using dhall and prefer the ascii format to the unicode version. I've added an option `g:ale_dhall_format_ascii` (not on by default) so that the `--ascii` flag is passed the formatter. I've also added relevant documentation + tests.

(P.S super love the plugin, thanks for creating and sharing!)


imran-iq:
Closing in favour of: https://github.com/dense-analysis/ale/pull/3261 (a more complete implementation)

PR #3470, opened 2020-11-27
Add nvim floating window support (replaces #3314)

kevinclark:
This continues the work begun in #3341 by @jargonzombies and responds to the feedback from @w0rp. In particular, it splits out configuration of the individual floating-window usecases and pulls that code into a separate file.

Known issues:
  * ~~Needs tests. Since it's my first time working in viml, I wanted to make sure the basics were right before digging in here.~~ Done and passing as of 60287b0.
  * ~~When floating preview coupled with g:ale_cursor_detail, the window is redrawn in a tight loop. Presumably in the non-floating version, moving into the preview window (via `stay_here`) keeps this from happening. But `stay_here` doesn't make sense in a floating-window world, so we may need another fix.~~ This now fixed in c39a27c.

As far as I'm concerned this is ready to go. Just waiting on review and a merge.

PR #3505, opened 2020-12-30
Use ale_echo_msg_format with cursor details

kevinclark:
This was raised in [this comment on the floating-preview pr](https://github.com/dense-analysis/ale/pull/3470#issuecomment-752225184) where it became apparent that cursor-details were using the ale_echo_message_format code but the preview window path wasn't.

Seems reasonable to want consistent data coming through the different view paths. If this is worth doing, I'll write up a test before merge.

### DIFF
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -137,7 +137,8 @@ function! s:ShowCursorDetailForItem(loc, options) abort
     let l:stay_here = get(a:options, 'stay_here', 0)
 
     let s:last_detailed_line = line('.')
-    let l:message = get(a:loc, 'detail', a:loc.text)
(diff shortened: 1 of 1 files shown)


kevinclark:
There's a behavioral difference that needs to be worked out before this moves forward. While this change adds echo format support (and prepends the message with the 'code' section of loclist) it also makes the detail use 'text' instead of 'detail' which seems wrong.

So, open behavior question: should there be a separate detail format that allows this kind of distinction? Cutting out the 'detail' part seems wrong, but allowing for more data also seems useful.

/cc @cj who brought up the original issue

kevinclark:
Open question (possibly for a follow up PR): when using ALEDocumentation, should floating windows be an option?

PR #3521, opened 2021-01-06
3498 change standardrb fixer to use stdin

hsanson:
Seems standardrb fails to properly use the --config option when using temporary files but works fine when reading from stdin. This commit changes the fixer so it uses stdin instead of temporary files.


grosscr:
I will say though (now that I'm using it for more than just a test), it appears that it now jumps you up to the top of file, instead of leaving the view of the file where it was after applying the changes. This is a bit annoying :/

hsanson:
@w0rp seems travis CI hanged somehow... there are many PRs in "Waiting for status to be reported" state.

hsanson:
This was tested by others and works for them. I will take that as a review and merge this.

Issue #3504, opened 2020-12-30
ALE fix causes jump


basharh:
## Information

**VIM version**

VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Dec  3 2020 01:24:40)
macOS version

macOS Big Sur version 11.0.1 

## What went wrong

- running with `g:ale_fix_on_save=1` and saving a file with errors causes VIM/ALE to jump in the file after those errors are fixed

## Reproducing the bug

1. Run with `g:ale_fix_on_save=1`
2. introduce an error that a linter(eg prettier) can fix(eg, missing trailing semicolon).
3. navigate to the line that contains the error and press `zt` to make sure it's at the first line
4. save and notice that line jumps to the middle. Behavior is inconsistent.

### :ALEInfo

 Current Filetype: typescriptreact
Available Linters: ['eslint', 'standard', 'tslint', 'tsserver', 'typecheck', 'xo']
  Enabled Linters: ['eslint', 'tsserver']
  Ignored Linters: []
 Suggested Fixers: 
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = v:null
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = v:null
let g:ale_disable_lsp = 0
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 1
let g:ale_fixers = {'typescript': ['eslint', 'prettier'], 'json': ['prettier'], 'typescriptreact': ['eslint', 'prettier'], 'java': ['google_java_format'], 'javascript': ['eslint']}
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = -1
let g:ale_lint_on_enter = 1
let g:ale_lint_on_filetype_changed = 1
let g:ale_lint_on_insert_leave = 1
let g:ale_lint_on_save = 1
[...]

PR #3535, opened 2021-01-12
Revert "Improves fixer performance for large buffers"

hsanson:
Reverts dense-analysis/ale#3358 that caused several issues with fixers dense-analysis/ale#3504. The problems are a result of the way the PR updates the buffer that also result in changes in the jump list and undo list. This causes unwanted cursor jumps and messes the undo and folding.

The original PR fixes some performance issues on large files but the issues it causes may not be worth the speed up benefit. Testing a 68Kb kotlin file with the ktlint fixer works in less than a second without dense-analysis/ale#3358 and I get not unwanted jumps and undo works as expected.

### DIFF
--- a/autoload/ale/util.vim
+++ b/autoload/ale/util.vim
@@ -486,7 +486,7 @@ function! ale#util#Input(message, value) abort
 endfunction
 
 function! ale#util#HasBuflineApi() abort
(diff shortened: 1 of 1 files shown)


w0rp (approved):
Yeah, I'm happy to undo this if it cause other issues.

PR #3541, opened 2021-01-15
Fix 3167 - add custom fixer example to help docs.

### LINKED ISSUE #3167: Add example for a custom command fixer
I wrote a program that accepts code in stdin and writes the fixed code in stdout. However, I'm not sure how to add it as a fixer to Ale. Could a simple example be added to the README on how to do this, since most custom tools read from stdin and write to stdout anyway?

I think it would be useful for many users, as it's a fairly common scenario.

hsanson:
Adds a custom fixer example to the docs.

Closes #3167


w0rp (approved):

w0rp:
Cheers! :beers:

PR #3549, opened 2021-01-19
Fix 354 - Migrate CI from travis to Github Actions

### LINKED ISSUE #3548: Migrate ALE from Travis CI to Github Actions
Travis-CI.com no longer provides unlimited credits to open source projects:

 - https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works
 
And ALE already ran out of credits, therefore there are many PR's stuck waiting for checks from travis that will never complete. One way to solve this is to migrate from Travis-CI to Github actions:

 - https://fbernutz.github.io/posts/github-actions

Since ALE used Docker to run tests we can use these actions for the migration:

 - https://github.com/marketplace/actions/docker-run-action
 - https://github.com/marketplace/actions/build-and-push-docker-images

hsanson:
Migrate CI from travis to Github Actions. Note this is the first time I have ever implemented Github Actions so this requires testing. Any @dense-analysis/maintainers  with experience in this is welcome to review and comment.

FYI @w0rp

Closes #3548


MindTooth:
One option is to use a matrix if these are individual checks.  Meaning they can run in parallel instead.

hsanson:
@MindTooth thanks for the tip. Wondering if that works fine when there are multiple steps but one of them does not depend on the matrix variable. For example would this work?

```
jobs:
  test_ale:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        test:
          - "--vim-80-only"
          - "--vim-81-only"
          - "--neovim-only"
          - "--linters-only"
    steps:
      - uses: docker/build-push-action@v2
        with:
          repository: w0rp/ale
          context: .
          file: ./Dockerfile
          push: false
          tags: w0rp/ale:latest
      - uses: addnab/docker-run-action@v1
        with:
          image: w0rp/ale:latest
          run: ./run-tests -v ${{ matrix.test }}
```

Still learning about Github actions and really have no idea how to test this.

.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+          - '--neovim-only'
+          - '--linters-only'
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build test docker image
+        uses: docker/build-push-action@v2
+        with:
+          repository: w0rp/ale
+          context: .
+          file: ./Dockerfile
```
MindTooth, line 25: Seems this is redundant when context is set.

Ref.: https://github.com/docker/build-push-action

.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
+name: CI
+on:
+  push:
+    branches: [ master ]
```
MindTooth, line 4: Not adding https://github.com/dense-analysis/ale/blob/7e3d2930d8defbcb233b0bc21822cb8e8377bec2/.travis.yml#L9 from the original travis file?

MindTooth:
Looks great.  Added som comments.

MindTooth:
The test image could be pushed to the container registry here on GitHub.  Need a workflo to build one when the Dockerfile changes.  I’ll see if I can take a look at it.  Seeing that the Dockefile almost never changes, the building of the image could extracted from the regular workflow.

.github/workflows/main.yml
hsanson: I removed the docker build step as it takes 5+ minutes that may exhaust the free minutes Github provides monthly.

.github/workflows/main.yml
hsanson: This was not working for some reason.  Not sure what is the exact syntax in Github actions to trigger on tags instead of branches.

MindTooth: Yeah.  Just noticed that it already exists an image on Docker Hub.

.github/workflows/main.yml
MindTooth: Yeah.  Just noticed that it already exists an image on Docker Hub.

.github/workflows/main.yml
```diff
@@ -0,0 +1,21 @@
...
+      matrix:
+        vim-version:
+          - '--vim-80-only'
+          - '--vim-81-only'
+          - '--neovim-only'
+          - '--linters-only'
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run tests
+        run: ./run-tests -v ${{ matrix.test }}
```
MindTooth, line 21: Believe it should be `matrix.vim-version` instead.  Need to use the name give inside the strategy block.

hsanson: Fixed, thanks.

.github/workflows/main.yml
MindTooth: Some example here: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-branches-and-tags

Maybe just use:

```
on:
  push:
    tags:
      - /^v\d+\.\d+\.(x|\d+)$/ 
      ```

Regex syntax: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

w0rp:
The only comment I'd make is what @MindTooth  already said about caching the Docker images, but let's get this in now so tests will run again. I'm still likely to get an unlimited plan from Travis CI, so I will keep that as a fallback option if we also run out of time for GitHub Actions.

w0rp:
Cheers @hsanson! :beers: Thanks for sorting out these kinds of issues while I have been doing much less myself.

MindTooth:
https://hub.docker.com/r/w0rp/ale

This seems to be built already and seeing that the `Dockerfile` almost never changes, I'm not sure that the project needs to take any additional actions.  I believe the rate limit will not be reached based on the current activity of the project.

w0rp:
We'll see how it goes. I don't think we're likely to hit usage limits, and if we do, we could switch back to Travis CI in whole or in part if they give us more time.

PR #1847, opened 2018-08-24
1846 support solargraph lsp

hsanson:
Adds support for solargraph language server.

Closes #1846

PR #1725, opened 2018-07-15
Add kotlin language server support

MTDL9:
Added support for the following unofficial Kotlin language server:
https://github.com/fwcd/KotlinLanguageServer

I used `languageserver` as the ALE internal name, let me know you prefer another name since this is an unofficial project and not a Language Server released by JetBrains themselves. An (unlikely) official Language Server release by JetBrains would probabily supersede this unofficial project

Referenced by issue #1701

PR #3556, opened 2021-01-22
Add rnix-lsp for Nix diagnostics and completion

jD91mZM2:


hsanson (changes requested):
Please also include respective documentation changes and add some tests. Check these PRs as examples:

   - https://github.com/dense-analysis/ale/pull/1725/files
   - https://github.com/dense-analysis/ale/pull/1847/files

jD91mZM2:
There, is that better? :)

test/nix/test.nix
```diff
@@ -0,0 +1,3 @@
+{
```
hsanson, line 1: This file can be empty as we do not actually run anything in the test. Is just to check if the command to run it is correct.

hsanson (approved):

test/nix/test.nix
jD91mZM2: Thanks, updated :)

hsanson (approved):
Good to go. Thanks or the work.

Issue #3560, opened 2021-01-24
Update tests to run on vim 8.2 and nvim 4.4


hsanson:
Both vim 8.2 and nvim 4.4 have been out for a while. Would be good to update the test scripts to run on them.

Issue #3411, opened 2020-10-24
Failed tests against Vim 8.2


mymedia2:
## Information

**VIM version**

    VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Oct 24 2020 11:11:04)
    Included patches: 1-1895
    Compiled by vim-testbed

Operating System: Alpine Linux (in Docker)

## What went wrong

    (1/3) [EXECUTE] (X) [] should be equal to [{'lnum': 2, 'bufnr': 3, 'vcol': 0, 'pattern': '', 'valid': 1, 'nr': -1, 'type': 'E', 'col': 3, 'text': 'foo bar'}]
    (5/7) [EXECUTE] (X) Vim(let):E491: json decode error at 'x'
      > command line..function <SNR>23_vader[5]..vader#run[45]..<SNR>25_run[41]..<SNR>25_execute[2]..vader#window#execute[11]..script /tmp/vAFhjPN/772[1]..function ale#util#FuzzyJSONDecode, line 8
    (6/7) [EXECUTE] (X) Vim(let):E491: json decode error at 'x'
      > command line..function <SNR>23_vader[5]..vader#run[45]..<SNR>25_run[41]..<SNR>25_execute[2]..vader#window#execute[11]..script /tmp/vAFhjPN/773[1]..function ale#util#FuzzyJSONDecode, line 8
    (1/1) [EXECUTE] (X) [{'lnum': 1, 'bufnr': 3, 'col': 10, 'linter_name': 'testlinter', 'vcol': 0, 'nr': -1, 'type': 'W', 'text': 'Infix operators must be spaced. [Warning/space-infix-ops]', 'sign_id': 1000002}, {'lnum': 2, 'bufnr': 3, 'col': 10, 'linter_name': 'testlinter', 'vcol': 0, 'nr': -1, 'type': 'E', 'text': 'Missing semicolon. [Error/semi]', 'sign_id': 1000003}] should be equal to [{'lnum': 1, 'bufnr': 3, 'vcol': 0, 'linter_name': 'testlinter', 'nr': -1, 'sign_id': 1000001, 'type': 'W', 'col': 10, 'text': 'Infix operators must be spaced. [Warning/space-infix-ops]'}, {'lnum': 2, 'bufnr': 3, 'vcol': 0, 'linter_name': 'testlinter', 'nr': -1, 'sign_id': 1000002, 'type': 'E', 'col': 10, 'text': 'Missing semicolon. [Error/semi]'}]
[...]

PR #3561, opened 2021-01-25
3560 add vim 8.2 and nvim 0.4 to ci tests

hsanson:
- Add vim 8.2 and nvim 0.4.4 to the CI script.
- Fix tests to work on vim 8.2 that has breaking changes.
- Update Docker image to use latest vim-vader.

Closes: #3560 and #3411

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,8 @@ RUN apk --update add $PACKAGES && \
 
 RUN pip install vim-vint==0.3.15
 
-RUN git clone https://github.com/junegunn/vader.vim vader
+RUN git clone https://github.com/junegunn/vader.vim vader && \
+    cd vader && git checkout c6243dd81c98350df4dec608fa972df98fa2a3af
 
 ARG GIT_VERSION
 LABEL Version=${GIT_VERSION}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,8 @@
 FROM tweekmonster/vim-testbed:latest
 
 RUN install_vim -tag v8.0.0027 -build \
-                -tag v8.1.0519 -build \
                 -tag v8.2.2401 -build \
                 -tag neovim:v0.2.0 -build \
-                -tag neovim:v0.3.5 -build \
                 -tag neovim:v0.4.4 -build
 
 ENV PACKAGES="\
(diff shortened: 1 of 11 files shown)


hsanson:
@w0rp before this can be merged a new Docker image with the changes here must be build and pushed to Docker hub.

Dockerfile
```diff
@@ -18,5 +20,8 @@ RUN apk --update add $PACKAGES && \
 
 RUN pip install vim-vint==0.3.15
 
-RUN git clone https://github.com/junegunn/vader.vim vader && \
-    cd vader && git checkout c6243dd81c98350df4dec608fa972df98fa2a3af
+RUN git clone https://github.com/junegunn/vader.vim vader
```
w0rp, line 23: I recommend still targeting a specific commit or tag so it doesn't change later and break running some tests. Use the same commit ID in `.appveyor.yml` so the Windows tests run the same version.

Dockerfile
```diff
@@ -2,8 +2,10 @@ FROM tweekmonster/vim-testbed:latest
 
 RUN install_vim -tag v8.0.0027 -build \
                 -tag v8.1.0519 -build \
+                -tag v8.2.2401 -build \
                 -tag neovim:v0.2.0 -build \
-                -tag neovim:v0.3.5 -build
+                -tag neovim:v0.3.5 -build \
+                -tag neovim:v0.4.4 -build
```
w0rp, line 8: I think we can cut out the versions in the middle for Vim and NeoVim. Testing with the the oldest and newest supported versions should be enough to catch issues with functions being missing and so on.

w0rp (changes requested):
I replied to your email about Docker access. I can add you to the dense-analysis organisation there, and then you should be able to set that up.

Dockerfile
hsanson: Got it, removed Vim 8.1 and Neovim 0.3.5

Dockerfile
hsanson: Done.  I am also trying to migrate appveyor to Github actions. At that point I may as well update vader on both sides.

hsanson:
@w0rp I think this is ready to merge now. Some pointers:

 - The required checks must be updated in the settings afterwards.
 - The CI workflow will build and push the ALE docker image under these conditions:
    - The Dockerfile is modified.
    - The PR that modifies this file is not from a fork. This is a security limitation of Github Actions.

w0rp (approved):

w0rp:
I've merged this and updated the branch protection rules to require the new checks. Great stuff.

Cheers! :beers:

PR #3568, opened 2021-01-31
Add jq linter for JSON diagnostics

jD91mZM2:
It's present on more machines than jsonlint :)

### DIFF
--- a/ale_linters/json/jq.vim
+++ b/ale_linters/json/jq.vim
@@ -1,9 +1,7 @@
 " Author: jD91mZM2 <me@krake.one>
 
-call ale#Set('json_jq_executable', 'jq')
-
 function! ale_linters#json#jq#GetCommand(buffer) abort
-    let l:executable = ale#Var(a:buffer, 'json_jq_executable')
+    let l:executable = ale#fixers#jq#GetExecutable(a:buffer)
 
     return ale#Escape(l:executable)
 endfunction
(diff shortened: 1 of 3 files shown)


hsanson (changes requested):
Please add documentation and command callback tests. 

On another note I wouldn't recommend using jq as a linter. From what I see on the Internet the parser is very lax and can let pass many invalid strings as valid JSON. Used in conjunction with other linters may be useful I guess.

jD91mZM2:
Hmm... what about `python -m json.tool`? It's also widely installed and should work everywhere

hsanson:
Note that I am not against adding jq as a validation tool.  Just ensure proper tests and documentation are included in the PR and I can merge it.  I believe, but have not tested, that `python -m json.tool` will have same limitations as `jq`. These are stream parser/filter tools so they will fail to raise issues such as duplicate keys that require full view of the JSON document in order to find.

ale_linters/json/jq.vim
```diff
@@ -0,0 +1,34 @@
+" Author: jD91mZM2 <me@krake.one>
+
+call ale#Set('json_jq_executable', 'jq')
```
jD91mZM2, line 3: @hsanson Note, this variable is shadowing a variable from the formatting counterpart (jq is used for formatting already, where this option is set). I was going to point this out earlier, but forgot - in any case I need help with what the proper way of dealing with this scenario is.

(This is also why I haven't documented this option - there's already docs for an option with the same name)

hsanson: Oh, in this case I would simply call the fixer GetCommand() method within this one or use it directly in the linter definition:

```
function! ale_linters#json#jq#GetCommand(buffer) abort
    return ale#fixers#jq#GetExecutable(a:buffer)
endfunction
```

If  don't think anyone would need to set different binaries for the linter and the fixer. Even if this is the case it can be done using bufffer variables (e.g. b:ale_json_jq_executable).

hsanson (approved):
Good, thanks.

Issue #2839, opened 2019-10-16
Expose gopls initialization options (e.g. completeUnimported)


ratorx:
According to [gopls docs](https://github.com/golang/tools/blob/master/gopls/doc/settings.md), there are additional settings that can be passed to the server by the client. From the source code, it seems that these need to be passed as initialization options.

Please could ale expose this? Specifically I'm interested in `completeUnimported`.

w0rp:
Someone just needs to update the linter file to accept a Dictionary for those options. The first example I found in the codebase is the `ccls` linter.

hsanson:
This is now supported thanks to #3571.

PR #3571, opened 2021-02-03
gopls: add go_gopls_init_options

### LINKED ISSUE #3563: Support for passing custom analysis options to gopls
I'm trying to configure ALE to pass custom analyser options to gopls, similar to vim-go:

https://github.com/fatih/vim-go/blob/master/doc/vim-go.txt#L1821-L1831

For example, to enable or disable individual analysers listed here:

https://github.com/golang/tools/blob/master/gopls/doc/analyzers.md

I can see ALE exposes `g:ale_go_gopls_options` but I don't believe the gopls CLI interface allows the above configuration to be passed. Did I miss something?

rfwatson:
Closes #3563 and #2839

### DIFF
--- a/ale_linters/go/gopls.vim
+++ b/ale_linters/go/gopls.vim
@@ -4,6 +4,7 @@
 
 call ale#Set('go_gopls_executable', 'gopls')
 call ale#Set('go_gopls_options', '--mode stdio')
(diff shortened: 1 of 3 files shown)


hsanson (approved):
Great contribution. Thanks.

Issue #3433, opened 2020-11-11
Option to select first completion provided by ALE (without inserting)


MeTheFlea:
Is there a way to auto select the first option in the completion menu whenever `:ALEComplete` is used? I have `set completeopt=menu,menuone,preview,noinsert` in my `.vimrc` but every time the menu opens I have no entry selected.

I saw commit 25b572b fixed an issue with automatically inserting the completion text by forcing the selection away from the first match but after I reverted this commit locally I got exactly the behaviour I wanted (without knowing the full consequences of such a change). 

Is there a supported way of getting this same behaviour?

w0rp:
I think this is worth doing when you request completion result manually. I added that fix to prevent a selection from being made as you type with automatic completion. We can send a different key sequence to open the completion menu when you manually request completion results, from the one we send when they are shown automatically.

rouge8:
@MeTheFlea i think #3573 might fix this for you?

w0rp:
@rouge8 @MeTheFlea I've pushed a commit now which only sets `noselect` in the options if you've configured Vim that way.

rouge8:
Thanks! 🎉

PR #3573, opened 2021-02-03
Support selecting the first completion option / Remove 'noselect' from completeopts

rouge8:
This patch removes `noselect` from the default `completeopts`. This should no longer be necessary thanks to 25b572b3bf3cc0229ed836eb3fa80c4c74f2c247, because the `<C-p>` added to the map will deselect the first entry.

Combined with something like this in your `~/.vimrc`:

```vim
function! RemapAleCompletion()
  iunmap <Plug>(ale_show_completion_menu)
  inoremap <silent> <Plug>(ale_show_completion_menu) <C-x><C-o>
endfunction

augroup remap_ale_completion
  autocmd!
  autocmd VimEnter * call RemapAleCompletion()
augroup END
```

This can address #3433.


MeTheFlea:
> ```viml
> function! RemapAleCompletion()
>   iunmap <Plug>(ale_show_completion_menu)
>   inoremap <silent> <Plug>(ale_show_completion_menu) <C-x><C-o>
> endfunction
> 
> augroup remap_ale_completion
>   autocmd!
>   autocmd VimEnter * call RemapAleCompletion()
> augroup END
> ```

I'm a bit of a vimscript novice but this snippet doesn't seem to work, since `<Plug>(ale_show_completion_menu)` doesn't seem to be set by the time `VimEnter * call RemapAleCompletion()` is executed. It seems like `<Plug>(ale_show_completion_menu)` is only set after the first time `ALEComplete` is called, and I'm not sure if there's an `autocmd` moment where I can automatically override it once it gets set. However, calling `RemapAleCompletion()` manually after doing `:ALEComplete` once does seem to provide the behaviour I originally wanted in #3433.

rouge8:
> I'm a bit of a vimscript novice but this snippet doesn't seem to work, since `<Plug>(ale_show_completion_menu)` doesn't seem to be set by the time `VimEnter * call RemapAleCompletion()` is executed. It seems like `<Plug>(ale_show_completion_menu)` is only set after the first time `ALEComplete` is called, and I'm not sure if there's an `autocmd` moment where I can automatically override it once it gets set. However, calling `RemapAleCompletion()` manually after doing `:ALEComplete` once does seem to provide the behaviour I originally wanted in #3433 for the rest of the vim session.

I think I can reproduce what you're seeing when I *don't* use `let g:ale_completion_enabled = 1`. In that case, I believe this PR plus this function fixes #3433:

```vim
function! RemapAleCompletion()
  " Load autoload/ale/completion.vim and the <Plug>(ale_show_completion_menu) mapping
  ALEComplete
  " or:
  " call ale#completion#GetCompletions('ale-manual')

  iunmap <Plug>(ale_show_completion_menu)
  inoremap <silent> <Plug>(ale_show_completion_menu) <C-x><C-o>
endfunction

augroup remap_ale_completion
  autocmd!
  autocmd VimEnter * call RemapAleCompletion()
augroup END
```

w0rp (changes requested):
I think instead of this change, we need a separate `Plug` mapping for manual completion (`ale-manual`) so we don't send `C-p` to back out of selecting the first item. This can be done by adding an argument to `s:OpenCompletionMenu` to change which `Plug` key it sends. We can't select the first item from automatic completion as it will interrupt typing, but we can when you open the completion menu through a keybind.

rouge8:
> I think instead of this change, we need a separate `Plug` mapping for manual completion (`ale-manual`) so we don't send `C-p` to back out of selecting the first item. This can be done by adding an argument to `s:OpenCompletionMenu` to change which `Plug` key it sends. We can't select the first item from automatic completion as it will interrupt typing, but we can when you open the completion menu through a keybind.

I actually use this with automatic completion. The `noinsert` ensures it doesn't insert the selected item unless I request it with `C-y`.

w0rp:
I'm gonna replace this with an alternate patch which respects the value of `completeopt`, so you'll get `noselect` if you have your options configured that way, otherwise you won't.

PR #3574, opened 2021-02-03
Bugfix: ParseCFlagsFromMakeOutput was never being called

hugomg:
Previously, the make output was never being parsed, even when the
`c_parse_makefile` option was set. I think this is because of a bug in GetCFlags, which would only try parsing the makefile if the compile_commands.json was ALSO present.

I'm not sure if I need to add tests for this, because it's a small change. If it is then I can come back later to add those.

### DIFF
--- a/autoload/ale/c.vim
+++ b/autoload/ale/c.vim
@@ -491,7 +491,7 @@ function! ale#c#GetCFlags(buffer, output) abort
         endif
     endif
 
-    if s:CanParseMakefile(a:buffer) && !empty(a:output) && !empty(l:cflags)
+    if empty(l:cflags) && s:CanParseMakefile(a:buffer) && !empty(a:output)
         let l:cflags = ale#c#ParseCFlagsFromMakeOutput(a:buffer, a:output)
     endif
 


hsanson (commented):
I am not familiar with this part of ALE but the fact that there are many tests for Makefile flags parsing (see test_c_flag_parsing.vader) and they were passing before this fix makes me suspect something is wrong.  I would recommend to check the tests and see why they worked before and after this fix.

hugomg:
On the matter of `test_c_flag_parsing`, my impression is that the tests that we have are unit tests to see if the `ParseCFlagsFromMakeOutput` function does the right thing. However, the bug that I found was that `ParseCFlagsFromMakeOutput` wasn't being called in the first place...

It's possible that there really are no tests that could have caught this. I might be missing something but I can't find any tests that call `ale_linters#c#cc#GetCommand` or `ale#c#GetCFlags`.

By the way, according to the git log the commit that introduced the contentious if statement appears to be this one: affeed7a

hsanson:
I would recommend this PR be reviewed by @w0rp.

w0rp (approved):
Yeah, this is just a mistake I made. I wouldn't worry about unit testing every single thing in the universe, I'll just merge this.

w0rp:
Cheers! :beers:

PR #3575, opened 2021-02-03
Have c_parse_makefile look for GNUmakefile as well

hugomg:
Makefiles using GNU Make features might be called "GNUmakefile" instead
of "Makefile". This commit teaches the `c_parse_makefile` feature to
look for a GNUmakefile file if a Makefile is not present.

I'm putting this in a separate PR from #3574, because it's an unrelated feature request.
Again, I'm not sure if this needs adding tests as  well. If it does then I can try to come back later to do that.

### DIFF
--- a/autoload/ale/c.vim
+++ b/autoload/ale/c.vim
@@ -506,6 +506,10 @@ function! ale#c#GetMakeCommand(buffer) abort
     if s:CanParseMakefile(a:buffer)
         let l:path = ale#path#FindNearestFile(a:buffer, 'Makefile')
 
+        if empty(l:path)
+            let l:path = ale#path#FindNearestFile(a:buffer, 'GNUmakefile')
+        endif
+
         if !empty(l:path)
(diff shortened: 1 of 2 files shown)


hsanson (changes requested):
Adding tests would be good but I am not too familiar with C tests.  At least I think replicating the tests in `test_c_flag_parsing.vader` for the `ale#c#GetMakeCommand()`  method should be added. Using a different sample project than `makefile_project` that uses the GNUMakefile instead of a Makefile.

hugomg:
Adding another test for `ale#c#GetMakeCommand` makes total sense. I'll try doing that later.

hugomg:
A quick question... Is there a way to run the test suite without using Docker? The documentation mentions using docker for the test suite but Docker is a bit tricky to install on Fedora, which is what I'm using now.

hsanson (approved):
Tests seems in order. Thanks.

PR #3582, opened 2021-02-08
mypy: Pass user options before any others

dhleong:
This enables us to use a custom `python` exe as the "mypy" executable
and pass `-m mypy` in `mypy_options`


w0rp (approved):

w0rp:
Cheers! :beers:

PR #3588, opened 2021-02-13
Update elixir-ls link in supported tools

axelson:
https://github.com/elixir-lsp/elixir-ls/ is now the canonical repo: 

> It's now being maintained by proactive volunteers from the Elixir community over at elixir-lsp/elixir-ls. Updates will continue to be published from that repo to the original VS Code extension, so no need to switch plugins if you're using VS Code.

from: https://github.com/JakeBecker/elixir-ls#this-project-has-moved


hsanson (approved):
Thanks for the update.

PR #3593, opened 2021-02-17
add support for prettier-ruby to prettier fixer

deecewan:
it _does_ need an additional plugin before it'll run ([prettier/plugin-ruby](https://github.com/prettier/plugin-ruby)), but when it has the plugin, it works as expected.


hsanson (approved):
Thanks for the contribution.

PR #3598, opened 2021-02-20
Add support for clangd with CUDA

oToToT:
I think clangd actually supports CUDA, so I simply copy some code from clangd support with C++ to get support for CUDA.  
I didn't write any tests since I think the C/C++ version tests works fine.


oToToT:
Hi, @w0rp are you willing to check this? Or maybe I should do something more?  
I think this would help some CUDA developers who use clangd.

w0rp (approved):
I'll add a test file for this.

w0rp:
Cheers! :beers:

PR #3601, opened 2021-02-20
Fix ansible-lint linter definition.

serapas:
Use ansible-lint's feature auto-detection instead of temporary file.
For auto-detection to work, ansible project has to also be a git repository.

Don't use yaml rules. These are checked by yamllint.

Refactor pattern to work with ansible-lint >=5.0 version.

Clean-up obsolete test cases.


ale_linters/ansible/ansible_lint.vim
```diff
@@ -18,28 +18,16 @@ function! ale_linters#ansible#ansible_lint#Handle(buffer, lines) abort
         endif
     endfor
 
-    " Matches patterns line the following:
-    "
-    " test.yml:35: [EANSIBLE0002] Trailing whitespace
```
w0rp, line 23: Keep an example here in the comments showing what the pattern is trying to match.

ale_linters/ansible/ansible_lint.vim
```diff
@@ -18,28 +18,16 @@ function! ale_linters#ansible#ansible_lint#Handle(buffer, lines) abort
...
     let l:output = []
 
     for l:match in ale#util#GetMatches(a:lines, l:pattern)
-        let l:code = l:match[4]
-
-        if l:code is# 'EANSIBLE0002'
-        \&& !ale#Var(a:buffer, 'warn_about_trailing_whitespace')
-            " Skip warnings for trailing whitespace if the option is off.
-            continue
-        endif
```
w0rp, line 34: Keep support for ignoring warnings about trailing whitespace. The ALE option is there for users who have already configured Vim to automatically remove trailing whitespace, so they can edit code without seeing problems they'll fix automatically anyway.

ale_linters/ansible/ansible_lint.vim
```diff
@@ -18,28 +18,16 @@ function! ale_linters#ansible#ansible_lint#Handle(buffer, lines) abort
...
-
         if ale#path#IsBufferPath(a:buffer, l:match[1])
             call add(l:output, {
             \   'lnum': l:match[2] + 0,
             \   'col': l:match[3] + 0,
             \   'text': l:match[5],
-            \   'code': l:code,
-            \   'type': l:code[:0] is# 'E' ? 'E' : 'W',
+            \   'code': l:match[4],
```
w0rp, line 30: We should check what the code is and be sure what we're setting. The only supported codes by Vim are `'E', 'W'`, and `'I'`,

test/handler/test_ansible_lint_handler.vader
```diff
@@ -1,58 +0,0 @@
...
-  call ale#test#SetFilename('main.yml')
-
-  let b:ale_warn_about_trailing_whitespace = 1
-
-After:
-  unlet! b:ale_warn_about_trailing_whitespace
-
-  call ale#linter#Reset()
-
-Execute(The ansible-lint handler should handle basic errors):
```
w0rp, line 12: Replace these tests with new ones that test the parsing of errors. We ought to be able to still parse these messages.

ale_linters/ansible/ansible_lint.vim
```diff
@@ -51,6 +39,6 @@ call ale#linter#Define('ansible', {
 \   'name': 'ansible_lint',
 \   'aliases': ['ansible', 'ansible-lint'],
 \   'executable': function('ale_linters#ansible#ansible_lint#GetExecutable'),
-\   'command': '%e -p %t',
+\   'command': '%e -p -x yaml',
```
w0rp, line 42: Find out what the earliest version of `ansible-lint` is that supports the `-x` option. We need to make sure it's old enough that we won't break linting for users who are stuck on older versions. About 1-2 years ago is usually old enough.

w0rp (changes requested):

ale_linters/ansible/ansible_lint.vim
serapas: _ansible-lint_ version 4.1.0, which is around 2 years old, already supports -x option.

Autodetection is only supported from version 4.3.0. And just from version 5.0.0, which is around half a month old, it's actually working as expected.
 
Working as expected means that if any _.yaml_ file in a _git_ repo of ansible project is opened and correctly identified as _ansible_ filetype, the linter will correctly identify the associated playbook, report all findings for all the involved tasks, roles, etc.. PR ale _ansible-lint_ definition code will filter only the ones appropriate for the opened file. Which is the expected behavior as I understand.

All previous versions had to get the playbook's absolute path as an argument. The findings won't be shown if another ansible file would be opened, e.g. some tasks file.

So my suggested solution works only from version 5.0.0. But it works as expected of any other linter - linting errors are returned for any opened ansible file in a project. And not only for playbook.

The current implementation doesn't work at all - it tries to lint temporary file, so all referenced files in a playbook are not being linted and reported. If one opens a task file, nothing will be reported at all, because it's not a playbook file.

ale_linters/ansible/ansible_lint.vim
serapas: With option -x yaml no reporting of trailing whitespace will be done. It's responsibility of yamllint linter. Which is a separate definition.

ale_linters/ansible/ansible_lint.vim
serapas: _ansible-lint_ (version >=5.0.0) can report following severity codes: “VERY_HIGH”, “HIGH”, “MEDIUM”, “LOW”, “VERY_LOW”, “INFO”. Since they can't be mapped 1-to-1 to vim codes, I would prefer just to mark every finding as an `E` and keep it simple, than to develop a biased mapping.

ale_linters/ansible/ansible_lint.vim
hsanson: Examples are below.

ale_linters/ansible/ansible_lint.vim
hsanson: Agree with this. Having both yamllint and ansible-lint reporting same thing is unnecessary.

ale_linters/ansible/ansible_lint.vim
hsanson: I would prefer to differentiate at least between high severity and low severity issues. Specially when dealing with hundreds of issues from roles that I do not maintain but need to use.

What about mapping like this:

VERY_HIGH, HIGH ->  E
MEDIUM, LOW, VERY_LOW -> W
INFO -> I

ale_linters/ansible/ansible_lint.vim
hsanson: Can confirm the -x option exists since 4.0.0 that is from 2018. I think is safe to assume the option is available for most users by now.

test/handler/test_ansible_lint_handler.vader
hsanson: As @w0rp says these tests should not be modified. Instead new ones are to be added and the Handle() method must work for both the old and new error messages.

hsanson (approved):
Thanks for the contribution.

PR #3602, opened 2021-02-20
Add support for `erlfmt` fixer

AntoineGagne:
As the title says, this PR adds support for [`erlfmt`](https://github.com/WhatsApp/erlfmt).

**Note:** The supported tools list seemed to be missing `dialyzer` for Erlang so I added it to the list at the same time as `erlfmt`.


hsanson (approved):
Looks good, thanks.

Issue #2790, opened 2019-09-26
ALE doesn't start bash-language-server or texlab


rharish101:
## Information

**VIM version**
```
VIM - Vi IMproved 8.1 (2018 May 18, compiled Sep  3 2019 19:23:34)
Included patches: 1-1968
```

Operating System: Arch Linux, with zsh as the default shell

## What went wrong

I have installed bash-language-server and texlab through the Arch Linux package. However, ALE does not start them when opening a `.sh` file or a `.tex` file (as seen through ALE's command history, and evidenced by the lack of completion), even when `:ALEInfo` reports that it has enabled `language_server` as a linter for the `sh` filetype and `texlab` for the `tex` filetype.

## Reproducing the bug

1. Install bash-language-server and texlab
2. Open a `.sh` file or `.tex` file
3. ALE does not show completion

### :ALEInfo

For bash:
```

 Current Filetype: sh
Available Linters: ['language_server', 'shell', 'shellcheck']
  Enabled Linters: ['language_server', 'shell', 'shellcheck']
 Suggested Fixers: 
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'shfmt' - Fix sh files with shfmt.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

let g:ale_sh_language_server_executable = 'bash-language-server'
let g:ale_sh_language_server_use_global = 0
let g:ale_sh_shell_default_shell = 'zsh'
let g:ale_sh_shellcheck_change_directory = 1
let g:ale_sh_shellcheck_dialect = 'auto'
let g:ale_sh_shellcheck_exclusions = ''
let g:ale_sh_shellcheck_executable = 'shellcheck'
let g:ale_sh_shellcheck_options = ''
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = 100
let g:ale_completion_enabled = 1
let g:ale_completion_max_suggestions = 50
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 1
[...]

Issue #2699, opened 2019-08-13
Texlab linter not run


AntoinePrv:
## Information

**VIM version** 
NVIM v0.3.7
Build type: Release

**Operating System**
macOS Mojave 10.14.6

## What went wrong

Thank you for ALE, it's a great project that I use daily!

The Texlab linter does not seem to run. Specifically, looking at `:ALEInfo`, all Latex linters are enabled, but TexLab is the only one not present in the command history.

I think my ALE setup is working alright since I use it without running into any issues with Python.

## Reproducing the bug

1. Install Texlab binary (in `~/.local/bin` in my case). Texlab is properly found in `PATH` and can be executed in the terminal.
2. Open a `.tex` file in vim
3. Start writing unproper Latex, no completion or linting appear, even after `:ALELint`
4. Run `:ALEInfo`, `texlab` appears in both available and enabled linters. However, in the Command history, all linter are run (tentatively as they are not installed), except for `texlab`

### :ALEInfo
```
 
 Current Filetype: tex
Available Linters: ['alex', 'chktex', 'lacheck', 'proselint', 'redpen', 'texlab', 'textlint', 'vale', 'writegood']
   Linter Aliases:
'writegood' -> ['write-good']
  Enabled Linters: ['alex', 'chktex', 'lacheck', 'proselint', 'redpen', 'texlab', 'textlint', 'vale', 'writegood']
 Suggested Fixers: 
  'latexindent' - Indent code within environments, commands, after headings and within special code blocks.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'textlint' - Fix text files with textlint --fix
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

let g:ale_tex_chktex_executable = 'chktex'
let g:ale_tex_chktex_options = '-I'
let g:ale_tex_lacheck_executable = 'lacheck'
let g:ale_tex_texlab_executable = 'texlab'
let g:ale_tex_texlab_options = ''
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = v:null
let g:ale_completion_delay = 100
let g:ale_completion_enabled = 1
let g:ale_completion_max_suggestions = 50
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
[...]

PR #3610, opened 2021-03-02
Fix texlab#GetProjectRoot

ourigen:
This fixes the texlab#GetProjectRoot function based on the discussion in #2501 of using a .git folder as an indicator of a TeX project root. This should address the issue of the texlab LSP server not starting at all, as raised by #2790 and #2699

### DIFF
--- a/ale_linters/tex/texlab.vim
+++ b/ale_linters/tex/texlab.vim
@@ -7,7 +7,7 @@ call ale#Set('tex_texlab_options', '')
 function! ale_linters#tex#texlab#GetProjectRoot(buffer) abort
     let l:project_root = ale#path#FindNearestDirectory(a:buffer, '.git')
 
-    return !empty(l:project_root) ? fnamemodify(l:project_root, ':h') : fnamemodify(a:buffer, ':h')
+    return !empty(l:project_root) ? fnamemodify(l:project_root, ':h') : ''
 endfunction
 
 function! ale_linters#tex#texlab#GetCommand(buffer) abort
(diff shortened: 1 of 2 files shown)


ale_linters/tex/texlab.vim
```diff
@@ -5,7 +5,9 @@ call ale#Set('tex_texlab_executable', 'texlab')
 call ale#Set('tex_texlab_options', '')
 
 function! ale_linters#tex#texlab#GetProjectRoot(buffer) abort
-    return ''
+    let l:project_root = ale#path#FindNearestDirectory(a:buffer, '.git')
+
+    return !empty(l:project_root) ? fnamemodify(l:project_root, ':h') : fnamemodify(a:buffer, ':h')
```
hsanson, line 10: Returning the directory of the current buffer when no .git directory is found may cause ALE to spawn an instance of the language server for every tex file open. See the comment about this: https://github.com/dense-analysis/ale/pull/2501#issuecomment-491993295

Here would be better to return an empty string if no root is found.

ourigen: Good point. I personally didn't mind an LSP server starting for every tex file, but I agree it would not be preferred as the default. I will fix that now, and I'll also try to fix the tests after I understand them more.

hsanson:
please fix the failing test so I can merge this:

```
  Starting Vader: /testplugin/test/command_callback/test_texlab_command_callbacks.vader
    (1/5) [EXECUTE] The language string should be correct
    (2/5) [EXECUTE] The default executable path should be correct
    (3/5) [EXECUTE] The project root should be detected correctly
    (3/5) [EXECUTE] (X) '/testplugin/.git' should be equal to ''
    (4/5) [EXECUTE] The executable should be configurable
    (5/5) [EXECUTE] The options should be configurable
  Success/Total: 4/5
```

ourigen:
@hsanson Tests should be fixed now. Thank you for the review!

hsanson (approved):
Looking good. Thanks.

Issue #3611, opened 2021-03-03
dmd doesn't use dub


vabenil:
## Information

**VIM version**

NVIM v0.4.4
Build type: Release

Arch Linux

## What went wrong

The dmd linter is not using my dub.json file to get import paths. Searching through the source code I found the error to be in this commit 9fe7b1fe6a23fb55e6d782293585d58193123f59, to be more specific in ale_linters/d/dmd.vim line 6.

## Reproducing the bug

Open vim in any project with a dub.json, you will notice ale complaining about not being able to find your libraries.

### :ALEInfo

 Current Filetype: d
Available Linters: ['dls', 'dmd']
  Enabled Linters: ['dls', 'dmd']
  Ignored Linters: []
 Suggested Fixers: 
  'dfmt' - Fix D files with dfmt.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'uncrustify' - Fix C, C++, C#, ObjectiveC, ObjectiveC++, D, Java, Pawn, and VALA files with uncrustify.
 Linter Variables:

let g:ale_d_dls_executable = 'dls'
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = 100
let g:ale_completion_enabled = 1
let g:ale_completion_max_suggestions = 50
let g:ale_disable_lsp = 0
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {'glsl': ['glslangValidator'], 'c': ['clangtidy'], 'd': ['dfmt'], '*': ['remove_trailing_lines', 'trim_whitespace'], 'cpp': ['clangdidy']}
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_filetype_changed = 1
let g:ale_lint_on_insert_leave = 1
let g:ale_lint_on_save = 1
[...]

PR #3612, opened 2021-03-03
Solve #3611

vabenil:
Fix dmd not using dub

I did not write any tests, I'm sorry.


hsanson (approved):
This is a bug fix for already existing linter.  Seems to work as expected.

PR #3617, opened 2021-03-06
Add ocamllsp for reason and esy

andreypopp:
This commit does the following:

- We add support for ocamllsp reason syntax.

  We reuse ocamllsp's ocaml definition here.

- We add support for esy (in addition to opam) for both ocaml and reason.

  The opam workflow stays default for ocaml but for reason I've added esy as the default one.

- We modify executable callback to check for existence of ocamllsp within the opam switch or esy sandbox using "which ocamllsp" invocation.

  This way it doesn't require that the opam switch is active when the vim is started. Same holds for esy projects - there's no need to do "esy vim" and just "vim" will be able to discover "ocamllsp".

### DIFF
--- a/ale_linters/reason/ocamllsp.vim
+++ b/ale_linters/reason/ocamllsp.vim
@@ -1,10 +1,10 @@
 " Author: Risto Stevcev <me@risto.codes>
 " Description: The official language server for OCaml
 
-call ale#Set('ocaml_ocamllsp_use_opam', 1)
-call ale#Set('ocaml_ocamllsp_use_esy', 0)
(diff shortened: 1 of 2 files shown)


andreypopp:
cc @Risto-Stevcev do you think doing `opam config exec -- which ocamllsp` makes sense for opam? As I understand this will make it so opam local switch could be inactive in the shell where vim is started or am I missing something?

Risto-Stevcev:
@andreypopp Ah my bad, I thought that command made sure that it uses the ocamllsp version in the local switch. I wanted to make sure that it honors whatever version is being used in the local switch by default

PR #3625, opened 2021-03-12
julia: respect $JULIA_PROJECT env var, set project_root explicitly

andreypopp:
First we modify `ale#Julia#FindProjectRoot` to respect `$JULIA_PROJECT` environment variable. This is what julia [does][1].

Second, instead of using `--project=@.` when launching languageserver process we set `--project` to the value obtained from `ale#Julia#FindProjectRoot`. This is because using `--project=@.` might pick another directory than the one specified by `$JULIA_PROJECT` environment variable, again [see][1] for docs on `--project=@.`.

[1]: https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_PROJECT

### DIFF
--- a/ale_linters/julia/languageserver.vim
+++ b/ale_linters/julia/languageserver.vim
@@ -5,10 +5,16 @@
 call ale#Set('julia_executable', 'julia')
 
 function! ale_linters#julia#languageserver#GetCommand(buffer) abort
+    let l:project_root = ale#julia#FindProjectRoot(a:buffer)
+
+    if l:project_root is# ''
+        let l:project_root = '@.'
+    endif
+
     let l:julia_executable = ale#Var(a:buffer, 'julia_executable')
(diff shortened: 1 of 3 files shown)


andreypopp:
cc @bstellato for feedback (as the original author of ALE linter for Julia's languageserver): hi, do you think it makes sense?

I've found out that when I have an environment like this:
```
.                           # Julia environment root (this is the root I want)
├── Project.toml
└── dev
    └── PostgresCatalog.jl  # Package linked for development (root ALE picks up if I launch vim from this dir)
        └── Project.toml
```

I have to be careful always to launch `vim` from the root so it picks up the correct `Project.toml`. Now if I set `$PROJECT_ROOT` in my env (I do this with `direnv`) and with this patch the situation is resolved and ALE always launches languageserver with the correct root.

PR #3639, opened 2021-03-18
Add more parameters to the DMD linting command

### LINKED ISSUE #3637: Support for D versions + other flags with the DMD linter
Some code rely on some versions passed to DMD. For example:
```d
module bindbc.lua.v53.bindstatic;

version(BindBC_Static) version = BindLua_Static;
version(BindLua_Static) {
    version(LUA_53) version = LUA_53_STATIC;
}

version(LUA_53_STATIC):
// lua-5.3 static bindings follow
```
To use this code I define versions `LUA_53` and `BindBC_Static` in my `dub.json` source file. 
The ALE linter is not using it and it creates false lint errors when I use lua symbols in my code.

The versions can be found from `dub describe --data=versions`. Versions is not the only thing that can create false lint errors.
There are also debug versions, imported files, imported string files.

My suggestion would be to replace the current `dub describe --import-paths` with `dub describe --data=import-paths --data=string-import-paths --data=import-files --data=versions --data=debug-versions` and to parse the output of that command to retrieve all the options to give to DMD for linting.

rtbo:
fixes #3637

### DIFF
--- a/ale_linters/d/dmd.vim
+++ b/ale_linters/d/dmd.vim
@@ -11,7 +11,12 @@ function! s:GetDUBCommand(buffer) abort
         " directory where we found the dub config, and then run `dub describe`
         " from that directory.
         if !empty(l:config)
-            return [fnamemodify(l:config, ':h'), 'dub describe --import-paths']
+            return [fnamemodify(l:config, ':h'), 'dub describe --data-list
+            \ --data=import-paths
+            \ --data=string-import-paths
+            \ --data=versions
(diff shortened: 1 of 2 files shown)


w0rp (changes requested):
Looks okay to me. Add some tests for this. You can provide example lines of output from the `dub` command and check the resulting command. See `:help ale-dev`.

hsanson (approved):
Thanks for adding the tests.

PR #3640, opened 2021-03-18
Switch to using buildifier's -path option

jparise:
Buildifier offers a -path command line option:

> Buildifier's reformatting depends in part on the path to the file
> relative to the workspace directory. Normally buildifier deduces
> that path from the file names given, but the path can be given
> explicitly with the -path argument. This is especially useful when
> reformatting standard input, or in scripts that reformat a temporary
> copy of a file.

This lets us drop our `-type`-guessing logic.

For the moment, we rely on the buffer filename being "relative to the
workspace directory", which it generally is, but in a future change, we
should introduce logic that will locate the `WORKSPACE` file and adjusts
the filename relative to that directory. This could be complicated based
on the working directory in which `buildifier` is executed, so a little
more research is necessary to implement that logic correctly.


hsanson (approved):
Thanks, this simplifies the fixer a lot.

Issue #3599, opened 2021-02-20
Linting Entire Projects


w0rp:
ALE has been focused on linting a single file at a type as you type for a long time. The functionality of ALE for doing this has been working so well for my personal work that I haven't felt the need to add much of anything to ALE for about half a year: it does what I want and I don't want much else. I've been thinking of something I would like to see in ALE, and that's an issue that came up a long time ago: Linting entire projects. (The earliest issue I could find is #669, and the most recent is #3524.)

It's possible to add functionality to ALE for linting an entire project, provided the following restrictions are applied.

1. For a linter to to check an entire project it must either be:
    1. A linter with a new `project_command` option defined for it, optionally `project_root`.
    1. An LSP linter whose server supports commands to check an entire project.
1. The `cwd` for the command for running against the project will be set to the project root. A new `ale_root` option will deprecate and replace the `ale_lsp_root` option, and function in exactly the same manner, but for linters that check entire projects.
1. Results for projects will always populate the quickfix list, and in order to use a list for problems local to a file you will have to use the `loclist` instead. You cannot populate local errors in the quickfix list at the same time that you're populating the quifckfix list with project-wide errors.
1. Errors in projects will not show visual cues for them, they will just populate the quickfix list and possibly ALE's own lists so you can jump through errors. Linters should separately run on files as you type to point out problems that can be seen from looking at only one file. This can be amended in future, but only after just the most basic functionality is working first.
1. Problems parsed from `callback` functions will be discarded when running over a whole project if they are lacking the `bufnr` or `filename` keys.
1. Checking an entire project will only be invoked via an `ALELintProject` command, never automatically.

As a pre-requisite for this work, #2281 will have to be implemented first, so the working directory for commands can be separated from the commands, such that the working directory can be swapped out when running against an entire project.

The `project_command` option implies that some linters will have `_project_options` settings, in addition to `_options`, so separate arguments can be set when checking entire projects. I know I personally use two different ESLint configuration files for checking individual files and projects, because the full type-aware configuration is too slow.

The support table will have to be updated to indicate which linters support checking an entire project. They won't all support it overnight. ESLint will probably be the first linter that gets this kind of support just because I happen to use it.

w0rp:
I've added the `ale_root` setting to deprecate `ale_lsp_root`. The old settings will quietly continue to be loaded until some time in the future when support for them can be removed.

WhyNotHugo:
[...]

PR #3647, opened 2021-03-21
improve DMD handler

rtbo:
- ignore errors from other files
 - catch 'Deprecation' as warning
 - add tests

(+ an easier to read pattern IMO)

### DIFF
--- a/ale_linters/d/dmd.vim
+++ b/ale_linters/d/dmd.vim
@@ -54,16 +54,19 @@ function! ale_linters#d#dmd#Handle(buffer, lines) abort
     " /tmp/tmp.G1L5xIizvB.d(8,8): Error: module weak_reference is in file 'dstruct/weak_reference.d' which cannot be read
     let l:pattern = '\v^(\f+)\((\d+)(,(\d+))?\): (\w+): (.+)$'
     let l:output = []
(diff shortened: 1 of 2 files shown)


ale_linters/d/dmd.vim
```diff
@@ -52,16 +52,18 @@ function! ale_linters#d#dmd#Handle(buffer, lines) abort
...
     let l:output = []
 
     for l:match in ale#util#GetMatches(a:lines, l:pattern)
-        call add(l:output, {
-        \   'lnum': l:match[1],
-        \   'col': l:match[2],
-        \   'type': l:match[3] is# 'Warning' ? 'W' : 'E',
-        \   'text': l:match[4],
-        \})
+        if ale#path#IsBufferPath(a:buffer, l:match[1])
```
hsanson, line 59: If you can get the absolute path of the file where the issue is reported then is better to add the *filename* key instead of filtering out errors for other files.  This would allow this linter to work project wide in the future: #3599

See [terraform.vim](https://github.com/dense-analysis/ale/blob/master/ale_linters/terraform/terraform.vim#L37) file for example.

hsanson (approved):
Looks good to me, thanks.

hsanson (approved):
Thanks for the fix.

PR #3648, opened 2021-03-22
Fix 3624 - parse buffer signs with getplaced() function

### LINKED ISSUE #3524: Linting the entire project and not just a single file
Hi all,

I would like to propose linting the entire project and not just the current buffer.
I am using Rust and rust-anylzer.

Thanks all.

### LINKED ISSUE #3624: ReadSigns: ale#sign#ReadSigns causes redirect error if run during execute()
## Information

VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Feb 21 2021 21:02:22)
Included patches: 1-2539

Operating System: Linux

## What went wrong

[`ale#sign#ReadSigns`](https://github.com/dense-analysis/ale/blob/c21d6afd2fb799013e3894d393bf976d9da31e65/autoload/ale/sign.vim#L172) uses redirect, which can fail with [E930](https://vimhelp.org/eval.txt.html#E930):

`E930: Cannot use :redir inside execute()`

If another script uses `execute()` where ALE is triggered this result in uncaught error.

An example trace:

```vim
Error detected while processing function foo[1]
..BufDelete Autocommands for "*"..function ale#engine#Cleanup[14]
..ale#engine#RunLinters[28]
..ale#engine#SetResults[6]
..ale#sign#SetSigns[8]
..ale#sign#FindCurrentSigns[1]
..ale#sign#ReadSigns
```

## Reproducing the bug

Trace in this *reproduce* differs from the example above, which is way to big to include, but should give same effect:

**Note**: `call` is needed here. Would more often be in a function, where call, let or the like is used.

```vim
$ vim foo.vim
:command! -nargs=1 TriggerAleRedir call execute('bdelete ' . <f-args>)
:e bar
:TriggerAleRedir bar
```

## Error output:

```
Error detected while processing BufWinEnter Autocommands for "*"
..function ale#events#LintOnEnter[5]
..ale#Queue[33]
..<SNR>91_Lint[26]
..ale#engine#RunLinters[28]
..ale#engine#SetResults[6]
..ale#sign#SetSigns[8]
..ale#sign#FindCurrentSigns[1]
..ale#sign#ReadSigns:
line    1:
E930: Cannot use :redir inside execute()
```

The above example also likely then produce E714.

```
Error detected while processing BufWinEnter Autocommands for "*"
..function ale#events#LintOnEnter[5]
..ale#Queue[33]
..<SNR>91_Lint[26]
..ale#engine#RunLinters[28]
..ale#engine#SetResults[6]
..ale#sign#SetSigns:
line    8:
E714: List required
```

## Fix

Not sure what the rationale is behind using redirect and parsing of that output, but the `sign_getplaced()` looks like a good alternative.

```vim
" Some error checking should perhaps be done here
" As long as a:buffer exists, it should be OK.
let sig = sign_getplaced( a:buffer, { group: s:supports_sign_groups ? 'ale' : '' })[0].signs

if len(sig) && s:supports_sign_groups
    let sig = filter(sig, 'v:val.priority == g:ale_sign_priority')
endif
```

Which should return something like:

```vim 
[
    {
        lnum: 123,
        id: 2
        name: 'foo',
        priority: 10
        group: 'ale'
    },
    {
        lnum: 456,
        id: 3
        name: 'bar',
        priority: 10
        group: 'ale'
    },
]
```

[`filter()` was added in Vim 7.](https://vimhelp.org/version7.txt.html#version7.txt)

hsanson:
Use getplaced() function when available to retrieve buffer signs. 

Closes #3624

### DIFF
--- a/autoload/ale/sign.vim
+++ b/autoload/ale/sign.vim
@@ -203,6 +203,27 @@ function! ale#sign#ParsePattern() abort
     return l:pattern
 endfunction
 
+" Given a buffer number, return a List of placed signs [line, id, group]
+function! ale#sign#ParseSignsWithGetPlaced(buffer) abort
+    let l:signs = sign_getplaced(a:buffer, { 'group': s:supports_sign_groups ? 'ale' : '' })[0].signs
+    let l:result = []
+    let l:is_dummy_sign_set = 0
+
+    for l:sign in l:signs
+        if l:sign['name'] is# 'ALEDummySign'
+            let l:is_dummy_sign_set = 1
+        else
+            call add(l:result, [
+            \   str2nr(l:sign['lnum']),
+            \   str2nr(l:sign['id']),
+            \   l:sign['name'],
+            \])
+        endif
+    endfor
+
+    return [l:is_dummy_sign_set, l:result]
+endfunction
+
 " Given a list of lines for sign output, return a List of [line, id, group]
 function! ale#sign#ParseSigns(line_list) abort
     let l:pattern =ale#sign#ParsePattern()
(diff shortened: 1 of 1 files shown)


hsanson:
This is affecting many, review is slow, and some people has reported this MR fixes the issue for them so merging.

w0rp:
Thanks for merging this. I am still alive. I've just been busy.

PR #3652, opened 2021-03-22
Add support for `ptop` fixer

BarrOff:
As the title says, this PR adds support for [ptop](https://www.freepascal.org/tools/ptop.var), a Pascal fixer and part of the [freepascal](https://www.freepascal.org/) project.

### DIFF
--- a/autoload/ale/fixers/ptop.vim
+++ b/autoload/ale/fixers/ptop.vim
@@ -11,7 +11,7 @@ function! ale#fixers#ptop#Fix(buffer) abort
     return {
     \   'command': ale#Escape(l:executable)
     \       . (empty(l:options) ? '' : ' ' . l:options)
-    \       . ' %t %t',
+    \       . ' %s %t',
     \   'read_temporary_file': 1,
     \}
 endfunction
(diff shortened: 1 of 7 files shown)


autoload/ale/fixers/ptop.vim
```diff
@@ -0,0 +1,17 @@
...
+call ale#Set('pascal_ptop_options', '')
+
+function! ale#fixers#ptop#Fix(buffer) abort
+    let l:executable = ale#Var(a:buffer, 'pascal_ptop_executable')
+    let l:options = ale#Var(a:buffer, 'pascal_ptop_options')
+
+    return {
+    \   'command': ale#Escape(l:executable)
+    \       . (empty(l:options) ? '' : ' ' . l:options)
+    \       . ' %t %t',
```
hsanson, line 14: ALE replaces all instances of %t with the same temporary file. This will result in ptop reading from and writing to the exact same file. This may result in corrupted output on some filesystems.

I think replacing this to ' %s %t' has the desired effect and would be safer.

hsanson (approved):
Thanks, looks good.

PR #3672, opened 2021-04-02
Fix `ale_sign_column_always` for Neovim 0.5

Svetlitski:
Neovim 0.5 (not yet released) changes the behavior of the sign column such that (quoting from the relevant commit):
> If all signs of the same group have no width, the signcolumn will not be rendered for that group.

See the following for more details: 
1. [The commit that introduced this change](https://github.com/neovim/neovim/commit/88ae03bcdb8992fd91a3efdb61dbd7e2aa395eff)
2. [This GitHub issue](https://github.com/neovim/neovim/issues/13635) where I (unsuccesfully) tried to argue that this was more of a bug than a feature.

What this means for ALE is that unfortunately `ale_sign_column_always` stopped working – the sign column would collapse when there were no warnings/errors, just like the default behavior. This pull request fixes that.

In addition I made a small, unrelated fix to `run-tests`, which wasn't working on my machine because I only have the `md5` tool and not `md5sum`. The fix ensures that `md5` is used if `md5sum` does not exist, to make the script even more portable.

### DIFF
--- a/autoload/ale/sign.vim
+++ b/autoload/ale/sign.vim
@@ -354,10 +354,17 @@ function! ale#sign#GetSignCommands(buffer, was_sign_set, sign_map) abort
     let l:command_list = []
     let l:is_dummy_sign_set = a:was_sign_set
 
+    " Neovim 0.5 changed the way the sign column behaves.
+    " See this commit for more details:
+    " https://github.com/neovim/neovim/commit/88ae03bcdb8992fd91a3efdb61dbd7e2aa395eff
+    if has('nvim-0.5') && g:ale_sign_column_always &&
+    \  ! getbufvar(a:buffer, 'ale_nvim_05_sign_fixed', v:false)
+        setlocal signcolumn=yes:1
+        let b:ale_nvim_05_sign_fixed=v:true
     " Set the dummy sign if we need to.
     " The dummy sign is needed to keep the sign column open while we add
     " and remove signs.
-    if !l:is_dummy_sign_set && (!empty(a:sign_map) || g:ale_sign_column_always)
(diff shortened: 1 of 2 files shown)


lukas-reineke:
This is not a good solution. Just overwriting `signcolumn` with `yes:1` would break everyone's settings.

Like I wrote in the nvim PR already, all you have to do is give the dummy sign a non 0 width. Settings the text to a space should be all you need.
You can try this like so

No signcolumn
```vim
:set signcolumn=auto
:sign define NoTextSign text=
:exe ":sign place 2 line=1 name=NoTextSign file=" . expand("%:p")
```

signcloumn
```vim
:set signcolumn=auto
:sign define TextSign text=\ " traling space here
:exe ":sign place 3 line=1 name=TextSign file=" . expand("%:p")
```

amree:
@lukas-reineke Your fix works, but it will throw error if I open neovim without any file. I'm getting `E158: Invalid buffer name:` on this line:

```
exe ":sign place 3 line=1 name=TextSign file=" . expand("%:p")
```

Any idea to improve this?

lukas-reineke:
The code I posted was an example to demonstrate how text length in signs work now. That code will not fix anything if you put it in your config.

To fix the issue, set the text of the dummy sign to a space here:
https://github.com/dense-analysis/ale/blob/ad27e921d71c5264382512dae7fce5292491e05f/autoload/ale/sign.vim#L88

I already gave you the code to fix it.
Nothing has changed from the comment I made in the Neovim repo
https://github.com/neovim/neovim/issues/13635#issuecomment-752380312

Replacing that one line is all you need to do.

Issue #2333, opened 2019-03-05
Hadolint: Show rule ID


varac:
When hadolint gets called by  ALE, it shows the rule ID before the message:

```
(executable check - success) docker                                                                                                                                                                                 
(started) ['/bin/zsh', '-c', 'docker run --rm -i hadolint/hadolint < ''/tmp/nvimR7OOcs/1/Dockerfile''']                                                                                                             
(finished - exit code 1) ['/bin/zsh', '-c', 'docker run --rm -i hadolint/hadolint < ''/tmp/nvimR7OOcs/4/Dockerfile''']                                                                                              
<<<OUTPUT STARTS>>>                                                                                                                                                                                                 
/dev/stdin:7 DL3018 Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`                                                                                                       
<<<OUTPUT ENDS>>> 
```

However, in the text buffer the rule ID is omitted. Please show it there, it's extremely useful to look it up in the hadolint docs.

SuperSandro2000:
I think it is mandatory to fully understood the rules and get the right fixes.

Issue #958, opened 2017-09-26
Parse and report syntax errors for hadolint


varac:
I'm using hadolint and ale to lint my dockerfiles.
When the syntax of the keywords are ok, it shows me hints what to improve - fine!
But when the Dockerfile contains a line starting with a non recognized keyword or junk text, ale won't show any error or hints unfortunatly.

```
 Current Filetype: dockerfile
Available Linters: ['hadolint']
  Enabled Linters: ['hadolint']
 Linter Variables:

let g:ale_dockerfile_hadolint_docker_image = 'lukasmartinelli/hadolint'
let g:ale_dockerfile_hadolint_use_docker = 'never'
 Global Variables:

let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%s'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'always'
let g:ale_linter_aliases = {}
let g:ale_linters = {}
let g:ale_open_list = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_column_always = 0
let g:ale_sign_error = '>>'
let g:ale_sign_offset = 1000000
let g:ale_sign_warning = '--'
let g:ale_statusline_format = ['%d error(s)', '%d warning(s)', 'OK']
let g:ale_warn_about_trailing_whitespace = 1
  Command History:

(executable check - success) hadolint
(started) ['/usr/bin/zsh', '-c', 'hadolint - < ''/tmp/nvimRsbhPT/2/Dockerfile''']
(finished - exit code 0) ['/usr/bin/zsh', '-c', 'hadolint - < ''/tmp/nvimRsbhPT/3/Dockerfile''']

<<<OUTPUT STARTS>>>
"/dev/stdin" (line 4, column 1):
unexpected 'd'
[...]

PR #3678, opened 2021-04-04
hadolint: multiple fixes and adapt expected format

m-ildefons:
- Disable color output, so severity detection works again
- Match on the correct filename
- Show hadolint rule number in vim gutter in addition to `ALEDetails`
- Capture and show error in case of syntax errors
- Add tests for error capture
- Adapt existing tests

39d22ad changed the file name, when reading from stdin from `/dev/stdin`
to `-`. This broke integration with ALE.
c00fc5c colorized the output to tty by default and added the flag
`--no-color` to disable it again. This broke detecting issue severity.

fixes: #2333
fixes: #958

### DIFF
--- a/ale_linters/dockerfile/hadolint.vim
+++ b/ale_linters/dockerfile/hadolint.vim
@@ -7,7 +7,7 @@ call ale#Set('dockerfile_hadolint_docker_image', 'hadolint/hadolint')
 function! ale_linters#dockerfile#hadolint#Handle(buffer, lines) abort
     " Matches patterns line the following:
     "
-    " /dev/stdin:19 DL3001 Pipe chain should start with a raw value.
+    " -:19 DL3001 warning: Pipe chain should start with a raw value.
     " /dev/stdin:19:3 unexpected thing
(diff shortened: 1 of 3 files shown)


m-ildefons:
@infokiller beat me to fixing the color output and input file name in  #3680. But this commit still fixes the other issues like showing the code in the vim gutter and test coverage.

infokiller:
@m-ildefons sorry I missed your PR. As I mentioned in the other one, there is now also json output which will make parsing much easier, but I'm not sure if it should be used since it can break ALE with older versions of hadolint that don't support it.

m-ildefons:
No problem :smile: . Considering that switching to the json output basically involves writing the plugin and its tests from scratch, I don't think its worth the effort right now. This may change if there ever needs to be a fix done to this again.

ddnomad (approved):
Just testing this one locally. It works fine. 

Any idea when this will be merged? Really helpful :)

hsanson (approved):
Looks good to me and seems to be working for some.

PR #3680, opened 2021-04-05
hadolint: fix color output and stdin shown as "-"

infokiller:
I started working on using the hadolint json format instead, but I'm blocked on
an upstream issue: https://github.com/hadolint/hadolint/issues/593


infokiller:
The issue was fixed quickly, should I send the other PR to use the json output instead of this one?

hsanson (approved):
Looks ok. Thanks.

hsanson:
@infokiller JSON output support seems very recent. Another PR to support it when available would be good as long as it does not break current tool support.

PR #3682, opened 2021-04-06
Updated `solhint` linter to be able to use a local installation

hbarcelos:
`solhint` is usually managed through `npm`/`yarn`, therefore it should behave like other linters in the node.js ecosystem.

The changes allow to use the solhint executable installed locally, which is important since globally installed `solhint` does not work with local `.solhintrc` files.


hsanson (approved):
Thanks, looks good.

Issue #3584, opened 2021-02-09
Floating preview closes visual modes


axeldetlofsson:
## Information

**VIM version**
NVIM v0.5.0-dev+1025-g0fad952a2
Build type: RelWithDebInfo

Operating System: 
Linux Atch 5.10.14 Kernel

## What went wrong
With ale_floating_preview enabled visual I have trouble with visual modes when cursor is at a error/warning. Entering visual mode wont close open floating preview, causing my next cursor move to close open preview but also exit visual mode. I dont know if this a bug or request for a setting like 'let g:ale_close_preview_when_entering_visual = 1' 

## Reproducing the bug

1. Configure ale with theese settings:
let g:ale_floating_preview = 1
let g:ale_hover_to_floating_preview = 1
let g:ale_detail_to_floating_preview = 1
let g:ale_use_global_executables = 1
let g:ale_sign_error = 'x'
let g:ale_sign_warning = '?'
let g:float_preview#docked = 0
let g:ale_cursor_detail = 1
let g:ale_hover_cursor = 1
let g:ale_close_preview_on_insert = 1

2. Go to a line with a error / warning causing floating preview to show.
3. Enter visual mode.
4. Move down a line with j.
5. BUG: you are no longer in visual mode.

### :ALEInfo

 Current Filetype: javascript.jsx
Available Linters: ['eslint', 'fecs', 'flow', 'flow-language-server', 'jscs', 'jshint', 'standard', 'tsserver', 'xo']
  Enabled Linters: ['eslint', 'tsserver']
  Ignored Linters: []
 Suggested Fixers: 
  'eslint' - Apply eslint --fix to a file.
  'fecs' - Apply fecs format to a file.
  'importjs' - automatic imports for javascript
  'prettier' - Apply prettier to a file.
  'prettier_eslint', 'prettier-eslint' - Apply prettier-eslint to a file.
  'prettier_standard', 'prettier-standard' - Apply prettier-standard to a file.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'standard' - Fix JavaScript files using standard --fix
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'xo' - Fix JavaScript/TypeScript files using xo --fix.
 Linter Variables:

let g:ale_javascript_eslint_executable = 'eslint'
let g:ale_javascript_eslint_options = ''
let g:ale_javascript_eslint_suppress_eslintignore = 0
[...]

PR #3689, opened 2021-04-12
Don't exit visual mode on floating preview close

kevinclark:
As reported in #3584, the floating preview doesn't interact properly with visual modes. Because the window creation is a new command, it exits visual mode in the process. This PR checks to see if visual mode was in use before and then restores it after the fact if so.

Unfortunately I wasn't able to add tests for this fix. Support for visual mode handling doesn't *appear* to exist in vader. Since there's no change in buffer input, that doesn't seem to be a viable strategy either. Existing preview tests do pass (though they're stubbing out a lot of the actual functionality).

I've been living on this branch for a couple of weeks with success and another respondent on the issue verifies the fix as well.

### DIFF
--- a/autoload/ale/floating_preview.vim
+++ b/autoload/ale/floating_preview.vim
@@ -113,6 +113,9 @@ function! s:Create(options) abort
 endfunction
 
 function! s:Close() abort
+    let l:mode = mode()
+    let l:restore_visual = l:mode is# 'v' || l:mode is# 'V' || l:mode is# "\<C-V>"
+
     if !exists('w:preview')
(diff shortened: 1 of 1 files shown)


hsanson (approved):
This fixes and issue affecting many and the fix has been reported to work by those affected. The change seems correct to me too.

kevinclark:
@hsanson Thank you!

w0rp:
Thanks for this one. It's one of those things that you pretty much can't test with Vader, so don't worry about it.

Cheers! :beers:

PR #3717, opened 2021-05-09
Don't use a temporary file for tflint

dimbleby:
Standard terraform module structure is to split files into `main.tf`, `variables.tf`, `output.tf`.

Current `tflint` linting creates a temporary file.  But taking eg the `main.tf` in isolation prevents `tflint` from doing its job, `ALEInfo` shows output like this:
```
{"issues":[],"errors":[{"message":"Failed to check `azurerm_public_ip_invalid_sku` rule: Failed to eval an expression i
n /tmp/nvimcFkTcA/2/main.tf:49; Reference to undeclared input variable: An input variable with the name \"vm\" has not
been declared. This variable can be declared with a variable \"vm\" {} block."}]}
```
... where the `vm` was indeed declared, in the `variables.tf` file, where it should be.

Simply not using a temporary file resolves this.

### DIFF
--- a/ale_linters/terraform/tflint.vim
+++ b/ale_linters/terraform/tflint.vim
@@ -99,6 +99,7 @@ endfunction
 call ale#linter#Define('terraform', {
 \   'name': 'tflint',
 \   'executable': {b -> ale#Var(b, 'terraform_tflint_executable')},
+\   'cwd': '%s:h',
 \   'command': function('ale_linters#terraform#tflint#GetCommand'),
 \   'callback': 'ale_linters#terraform#tflint#Handle',
 \})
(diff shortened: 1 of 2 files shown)


dimbleby:
Using a temporary file was deliberately introduced, at 22e8050639cc81a35c49479839a4c3e6ea9d9cc.

That was some time ago and I could believe that either ALE or tflint has changed since then, but if you can remember the reason for that change it would be interesting...

dimbleby:
Ah, but then the linter become useless in a different way: tflint doesn't recurse into subdirectories.  So you need to be in the working directory of the file being linted.

Therefore I'ev also set `cwd` for this linter.

hsanson (approved):
Change seems correct and basic testing seems to work for me.

Issue #2624, opened 2019-06-27
cmake-format doesn't find cmake-format.yaml/json/py


notskm:
## Information

**VIM version**
VIM - Vi IMproved 8.1 (2018 May 18, compiled Jun 25 2019 22:02:57)
MS-Windows 64-bit console version

NVIM v0.4.0-1134-g615fc6826

Operating System: Windows 10 1903

## What went wrong

I have `cmake-format` set up as a fixer and ALE configured to fix on save. When run through ALE, `cmake-format` is using its default settings to format my `CMakeLists.txt` files. It should be automatically detecting and using `cmake-format.yaml/json/py` in the current working directory.

If I manually run `cmake-format CMakeLists.txt -i`, it finds and uses my `cmake-format.yaml` file.

## Reproducing the bug

1. Set up `cmake-format` as a fixer
2. Create a cmake-format.yaml file with non-default settings in the current working directory
3. Run ALEFix in a CMakeLists.txt file

### :ALEInfo

 Current Filetype: cmake
Available Linters: ['cmakelint']
  Enabled Linters: ['cmakelint']
 Suggested Fixers: 
  'cmakeformat' - Fix CMake files with cmake-format.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

let g:ale_cmake_cmakelint_executable = 'cmakelint'
let g:ale_cmake_cmakelint_options = ''
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = v:null
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = v:null
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 1
let g:ale_fixers = {'cmake': ['cmakeformat'], 'cpp': ['clang-format'], 'python': ['autopep8']}
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
[...]

PR #3725, opened 2021-05-15
Use stdin/out for cmake-format

ts826848:
cmake-format added support for reading from/outputting to stdin/out as
of v0.3.6, released 2018-04-10 (commit 2e2aff2) [0].

Reading from stdin is preferable over reading from a temporary file
because when given a concrete file cmake-format will look for its config
file (.cmake-format.py or similar) in the parent directories of the
provided file. If the temporary file is off in a tmpdir somewhere (e.g.,
/tmp on *nix), cmake-format will almost certainly not come across the
user's intended format configuration file, making it appear that
cmake-format is ignoring the config file.

If cmake-format reads from stdin, though, it'll look for its config file
in its current working directory and its parent directories, in a
similar manner to clang-format. This has a much higher chance of running
across the intended config file.

[0]: https://github.com/cheshirekow/cmake_format/releases/tag/v0.3.6

### DIFF
--- a/autoload/ale/fixers/cmakeformat.vim
+++ b/autoload/ale/fixers/cmakeformat.vim
@@ -10,9 +10,7 @@ function! ale#fixers#cmakeformat#Fix(buffer) abort
 
     return {
     \   'command': ale#Escape(l:executable)
-    \       . ' -i '
     \       . (empty(l:options) ? '' : ' ' . l:options)
-    \       . ' %t',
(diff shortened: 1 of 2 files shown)


ts826848:
I believe this solves #2624. It Works On My Machine™️, at least.

hsanson (approved):
Thanks, since stdin/stdout are supported since 2018 I think is ok to set this as default.

PR #3727, opened 2021-05-17
Allow lintr to consider .Renviron settings. Fix #3391

### LINKED ISSUE #3391: Vanilla mode ignores .Renviron
# Background 
Calling `Rscript` with `--vanilla` switch on:
https://github.com/dense-analysis/ale/blob/b91c6c2edd20794ad5637b561ed4c678647a76e5/ale_linters/r/lintr.vim#L24-L27
forces R to ignore `.Renviron`, as stated `Rscript --help`

>   **--vanilla**           Combine --no-save, --no-restore, --no-site-file --no-init-file and --no-environ

# Example
```sh
Rscript --vanilla -e "library(lintr)"
Error in library(lintr) : there is no package called ‘lintr’
Rscript --no-save --no-restore --no-site-file --no-init-file -e "library(lintr)"
```

# Question
What are the reasons for calling `Rscript` using `--vanilla` switch? Wouldn't be better to replace `--vanilla` option with `--no-save --no-restore --no-site-file --no-init-file`. Furthermore, for fairly popular practice of defining libraries via `~/.Rprofile` it could be good to let user modify switches with which they wish for the `Rscript` command to be called?

ourigen:
Using `--vanilla` switch prevents R from reading .Renviron settings, which is problematic for users who define their R libraries in locations separate from the default as it prevents `library(lintr)` from being called. Fixes #3391, which explains the issue more.

### DIFF
--- a/ale_linters/r/lintr.vim
+++ b/ale_linters/r/lintr.vim
@@ -1,5 +1,6 @@
 " Author: Michel Lang <michellang@gmail.com>, w0rp <devw0rp@gmail.com>,
-"         Fenner Macrae <fmacrae.dev@gmail.com>
+"         Fenner Macrae <fmacrae.dev@gmail.com>,
+"         ourigen <ourigen [at] pm.me>
 " Description: This file adds support for checking R code with lintr.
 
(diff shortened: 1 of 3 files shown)


hsanson (approved):
Looks good, thanks.

PR #3728, opened 2021-05-20
Add fixer to remove whitespace at end of lines, except for double whitespace

johnmathews:
Add a generic fixture called `trim_whitespace_except_double` which calls `ale#fixers#generic#TrimWhitespaceExceptDoubleWhitespace`

This fixer removes whitespace at the end of a line but leaves a double whitespace. (It will remove a single whitespace, it will remove 3 or more whitespaces. It will not remove 2 whitespaces).

This is useful in markdown, where a double whitespace is used to create a line break.

I wrote a test `test_trim_whitespace_except_double.vader` and ran `./run-tests` and the tests seemed  to run without problems. The test is based on `test_trim_whitespace.vader`, I simply changed the first argument of the result so that it was unchanged compared to the code being fixed (to not remove the double whitespace after `first_argument`).

Any comments are appreciated, I'm still learning how to make pull requests.

### DIFF
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -141,6 +141,11 @@ let s:default_registry = {
 \       'suggested_filetypes': [],
 \       'description': 'Remove all trailing whitespace characters at the end of every line.',
 \   },
+\   'trim_whitespace_except_double': {
+\       'function': 'ale#fixers#generic#TrimWhitespaceExceptDoubleWhitespace',
+\       'suggested_filetypes': [],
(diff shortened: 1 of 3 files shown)


johnmathews:
Is there a better way to suggest new fixers? Was there something wrong with this PR?

hsanson (commented):
I don't think this is good as a generic fixer since it is specific of Markdown and can conflict with others configuration.  ALE by default will run all fixers/linters it finds.

I rather add markdownlint as a fixer that can handle double trailing spaces and is configurable:

  - https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md009---trailing-spaces

PR #3732, opened 2021-05-22
ocaml: prepare for  filetype separation, drop ols support

undu:
The ocaml filetype is currently used for several, different file
formats. This causes problems as not all tools support all formats.
New filetypes are introduced to support this separation, this needs some
changes in ale that are fortunately backwards-compatible.

These change add ocamlinterface file support for ocp-indent, merlin,
ocamlformat and ocaml-lsp. For ocaml-lsp I took the liberty to
add all recognised language ids, even if they are not supported.

ols has been removed as the project has been abandoned since 2019,
users are better served by other tools.

The PR for splitting the filetypes is at https://github.com/ocaml/vim-ocaml/pull/61

### DIFF
--- a/ale_linters/ocamlinterface/merlin.vim
+++ b/ale_linters/ocamlinterface/merlin.vim
@@ -1,0 +1,17 @@
+" Author: Andrey Popp -- @andreypopp
+" Description: Report errors in OCaml code with Merlin
+
+if !exists('g:merlin')
+    finish
+endif
+
+function! ale_linters#ocamlinterface#merlin#Handle(buffer, lines) abort
+    return merlin#ErrorLocList()
+endfunction
+
+call ale#linter#Define('ocamlinterface', {
+\   'name': 'merlin',
+\   'executable': 'ocamlmerlin',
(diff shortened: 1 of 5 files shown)


undu:
appveyor failure is unrelated to the change:

```
git clone -q --depth=10 https://github.com/dense-analysis/ale.git C:\testplugin
fatal: unable to access 'https://github.com/dense-analysis/ale.git/': Could not resolve host: github.com
Command exited with code 128
```

hsanson (changes requested):
Changes seem ok except for removing ols that is a potential breaking change. Currently there is no defined procedure to remove supported tools from ALE and I would refrain from doing so unless is first announced (somehow) to give users time to phase the tool out or request it not to be removed.

I recommend moving the changes to remove ols support to a separate PR with an associated issue to inform and start discussion on if is ok to remove it from ALE.

undu:
That makes sense, I'll remove the changes to remove ols.

hsanson (approved):
Thanks for the changes. Looks good to me.

PR #3744, opened 2021-05-27
Let R_languageserver consider .Renviron

ourigen:
Addresses a similar problem to #3727 but for the R language server. 

Using `--vanilla` switch prevents R from reading `.Renviron` settings, which for users who define their R libraries in locations different from the default prevents `library(languageserver)` from being called


hsanson (approved):
Looks good, thanks.

PR #3754, opened 2021-06-03
add bsfmt brightscript fixer

andersondanilo:
Add bsfmt (brightscript) fixer


hsanson (changes requested):
Looks good except for the fact that the PR includes extra 3989 files that I am sure are not needed.

PR #3757, opened 2021-06-06
Add support for rpmlint 2.0.0

carlsmedstad:
The -o/--option flag was removed in version 2.0.0 of rpmlint. Providing
this causes rpmlint to fail and provide no output. Only provide this
flag to rpmlint if the version is less than 2.0.0.


hsanson (approved):
Thanks for the fix.

PR #3759, opened 2021-06-07
black fixer: --pyi option was appended without a space

bukzor:
Also a new test: The black callback should not concatenate options


bukzor:
Is the testing pipeline stuck?

hsanson (approved):
Thanks for the fix.

PR #3769, opened 2021-06-17
Add support for bicep language server

JohanBjoerklund:
Add support for bicep language server.


hsanson (changes requested):
Please add an entry in the supported-tools.md and ale-supported-languages-and-tools.txt files. Note these files must be properly sorted by name or the tests will fail, recommended that you run the test suite before submitting to review.

PR #3775, opened 2021-06-22
fix: cflags parser: no absolute path for '-include'

mjeanson:
Both '-include' and '-imacros' take a file as an argument that will then
be searched in the include path like a regular '#include "..."'
statement in a source file. As such, they should not have their path
converted to an absolute path.


hsanson (approved):
Thanks for the fix.

PR #3777, opened 2021-06-23
racket/raco: add filename to loc list

benknoble:
This allows the location list from one buffer to point to an issue in
another; previously, the error message would be shown but with no way to
jump to it.


hsanson (approved):
Thanks, looks good.

PR #3780, opened 2021-06-24
Change checkstyle's output sub_type to style

Frederick888:
Checkstyle's output should have `style` as `sub_type`, otherwise it causes
issues such as wrong sign column icons.

`./run-tests` all passed.


hsanson (approved):
Thanks for the fix.

PR #3782, opened 2021-06-26
ALERename: Fix an issue with unlisted buffers

### LINKED ISSUE #3781: ALERename: unlisted buffers are modified incorrectly
## Information

**VIM version**

NVIM v0.4.4
Build type: Release

Operating System: Linux

## What went wrong

After calling ALERename on certain variable,s, some files would be incorrectly replaced. The whole file would be truncted, and the only changes would be the new variable name.

## Reproducing the bug

I was able to track it down to the following configuration:

```vimscript
let g:ale_linters = {
\  'go': ['gopls', 'golangci-lint'],
\}
```

I then opened a `myfile.go`. In the same folder I had a test file `myfile_test.go`. After renaming a symbol in `myfile.go`, which was also referenced from `myfile_test.go`, the whole `myfile_test.go` would be truncated.

I tracked it down to`bufnr('/path/to/myfile_test.go')` returning number for a file that I never opened, even though the buffer wasn't there. I later noticed that this buffer was actually unlisted (visible only via `:ls!`). So there must be something in the `golangci-lint` linter that opens this file in the background.

### :ALEInfo

I don't think it's necessary since I'll shortly submit an MR that fixes it.

jeremija:
After calling ALERename on certain symbols, some files would be incorrectly replaced. The whole file would be truncated, and the only contents would be the new variable name.

I was able to consistently reproduce it using the following configuration:

    let g:ale_linters = {
    \  'go': ['gopls', 'golangci-lint'],
    \}

Whenever I rename a symbol in `myfile.go` which is also referenced from an unopened file `myfile_test.go`, the whole `myfile_test.go` is truncated.

I tracked it down to `bufnr('/path/to/myfile_test.go')` returning number for a buffer that I never opened.

I later noticed that this buffer was unlisted (visible only via `:ls!`). So there must be something in the golangci-lint linter that opens this file in the background.

This fix adds a check for hidden buffers before applying code actions.

Closes #3781

### DIFF
--- a/autoload/ale/code_action.vim
+++ b/autoload/ale/code_action.vim
@@ -67,9 +67,13 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
     let l:current_buffer = bufnr('')
     " The buffer is used to determine the fileformat, if available.
     let l:buffer = bufnr(a:filename)
-    let l:is_current_buffer = l:buffer > 0 && l:buffer == l:current_buffer
+    " Buffer might be unlisted after calling :bd on it, we want to behave
+    " as if these buffers don't exist because they cannot be read from or
+    " modified anymore..
+    let l:is_opened = l:buffer > 0 && buflisted(l:buffer)
+    let l:is_current_buffer = l:is_opened && l:buffer == l:current_buffer
 
-    if l:buffer > 0
+    if l:is_opened
         let l:lines = getbufline(l:buffer, 1, '$')
 
         " Add empty line if there's trailing newline, like readfile() does.
@@ -155,7 +159,7 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
         endif
     endfor
 
-    if l:buffer > 0
+    if l:is_opened
         " Make sure ale#util#{Writefile,SetBufferContents} add trailing
         " newline if and only if it should be added.
         if l:lines[-1] is# '' && getbufvar(l:buffer, '&eol')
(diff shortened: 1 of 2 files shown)


jeremija:
cc: @w0rp @liskin

daliusd (approved):
Looks good to me. That’s all I can do however.

liskin:
@davidatbu Is this the problem you encountered in https://github.com/dense-analysis/ale/pull/3478#issuecomment-761332395?

liskin (commented):
Good catch!

Should the `if a:should_save || l:buffer < 0` line become `if a:should_save || !l:is_opened`?

jeremija:
Hey all, thanks for taking a look at this so quickly! After I made this change yesterday, I realized we can further improve the rename operation, since there are some cases when renaming simply fails after doing it twice.

So I'm going to close this MR in favor of #3783 - I'd appreciate if you could have a look at the change there :bowing_man:

davidatbu:
@liskin, this was exactly the problem i described in that comment you mentioned! Hope it gets fixed soon.

jeremija:
Thanks @davidatbu, I'd appreciate if you could have a look at MR #3783 to see if you can find any bugs?

I think it should all be resolved now, but it's still possible there are some corner cases I haven't tested.

davidatbu:
Thanks for your contributions. I'm currently toying with Neovim's built in
LSP, so i don't have the bandwidth to try this out, but if I ever revert
back to ALE, i will be sure to let you know.

Issue #3642, opened 2021-03-19
ALERename fails on double renaming


bogdan:
## Information

**VIM version**

NVIM v0.4.3
Build type: Release

Operating System:  10.15.5

## What went wrong

ALERename fails on double renaming with typescript and tsserver

## Reproducing the bug

Suppose you have  2 files:

``` typescript
// ./one.ts
export default class One {
}
// ./two.ts
import One from './a';

class Two extends One {
}
```

1. Open `one.ts`.
2. Put the cursor on `One` class name.
3. Call `:ALERename`
4. Rename to `One1`
5. Call `:ALERename`
6. Rename to `One11`
7. Open `two.ts` file.

Expected `Two` to extend `One11`  but it is extending `One1` instead.

### :ALEInfo

https://gist.github.com/bogdan/cd13d82cd808b844f99c1695aeb3974a

w0rp:
Potentially tricky to fix, but definitely worth fixing.

jeremija:
@bogdan I think I was able to come up with a proof of concept that fixes this issue: the solution is to send the `textDocument/didChange` message after modifying the files. You can have a look at the changes here:

https://github.com/jeremija/ale/commit/db6b7d1c0acf6cb1af19b6e278deb16a7e984830

I just don't have the time to fix all the broken tests and submit a complete MR at the moment, but if anyone want's to do it - feel free!

hsanson:
Fixed with #4049

PR #3783, opened 2021-06-27
Dispatch textDocument/didChange after rename

### LINKED ISSUE #3343: Send didChange/didSave events after rename
Rename operations modify files other than the current buffer, and language servers (and `tsserver`) need to know that these files are changed. We should make ALE tell language servers that files are changed or saved when they change as the result of a code action or rename operation. This will keep type information up to date when you rename something across many files from outside of the file where a name originates.

jeremija:
Previously whenever we renamed a symbol that was referenced from other files we'd just edit those files in the background, and the LSP wouldn't know about these changes. If we tried to rename the same symbol again, the renaming would fail. In some scenarios, the operation would just be wrong. Here is an attempt to fix this issue.

I also noticed another bug when using Go with `gopls` LSP and the `gofmt` fixer. Whenever the file was saved, the `gofmt` would run and reformat the file. But it seems there was some kind of a race condition so I disabled saving for now, and all of the modified files will be unsaved, so the user should call `:wa` to save them. I personally like this even better because I can inspect exactly what changes happened, and I instantly see them in the other opened buffers, which was previously not the case.

Finally, this change fixes a bug when trying to replace the contents of a hidden buffer.

Fixes #3343, #3642, #3781.

Video example:

Rename example

### DIFF
--- a/test/test_code_action_corner_cases.vader
+++ b/test/test_code_action_corner_cases.vader
@@ -28,6 +28,11 @@
 "   $ node test_code_action_corner_cases.js <test_code_action_corner_cases.vader
 
 Before:
+  runtime autoload/ale/lsp.vim
+
+  function! ale#lsp#NotifyForChanges(conn_id, buffer) abort
+  endfunction
+
   Save &fixeol
   set nofixeol
 
@@ -43,7 +48,9 @@ Before:
     if a:mode isnot 'file'
       execute 'edit ' . l:file
     endif
-    call ale#code_action#ApplyChanges(l:file, a:changes, a:mode isnot 'buffer')
+    call ale#code_action#ApplyChanges(l:file, a:changes, {
+    \ 'should_save': a:mode isnot 'buffer',
+    \})
     if a:mode is 'buffer'
       execute 'write ' . l:file
     endif
@@ -92,7 +99,7 @@ Execute(Preserve (no)eol at eof):
 
   " there doesn't seem to be a way to tell if a buffer is empty or contains one
   " empty line :-(
-  AssertEqual "", TestChanges("", [], 'file')
+  AssertEqual "\n", TestChanges("", [], 'file')
 
 Execute(Respect fixeol):
   set fixeol
--- a/test/test_code_action_corner_cases.vader
+++ b/test/test_code_action_corner_cases.vader
@@ -28,11 +28,6 @@
 "   $ node test_code_action_corner_cases.js <test_code_action_corner_cases.vader
 
 Before:
-  runtime autoload/ale/lsp.vim
-
-  function! ale#lsp#NotifyForChanges(conn_id, buffer) abort
-  endfunction
-
   Save &fixeol
   set nofixeol
 
@@ -96,10 +91,6 @@ Execute(Preserve (no)eol at eof):
     AssertEqual "eol\n",    TestChanges("eol\n",    [], g:mode)
     AssertEqual "eols\n\n", TestChanges("eols\n\n", [], g:mode)
   endfor
-
-  " there doesn't seem to be a way to tell if a buffer is empty or contains one
-  " empty line :-(
-  AssertEqual "\n", TestChanges("", [], 'file')
 
 Execute(Respect fixeol):
   set fixeol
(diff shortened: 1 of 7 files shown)


test/test_code_action_corner_cases.vader
```diff
@@ -92,7 +99,7 @@ Execute(Preserve (no)eol at eof):
 
   " there doesn't seem to be a way to tell if a buffer is empty or contains one
   " empty line :-(
-  AssertEqual "", TestChanges("", [], 'file')
+  AssertEqual "\n", TestChanges("", [], 'file')
```
liskin, line 102: This makes the above comment meaningless.

The point of that comment was that this test can't be done for the `save` and `buffer` modes, as these go through a buffer and change the file contents even though they should not. Now that all edits go through a buffer, not even the `file` mode works properly. At this point we might as well test all three modes and add a comment explaining that the behaviour is wrong and the only point of the test will be to catch changes in behaviour, not to test that it behaves correctly.

test/test_code_action_corner_cases.vader
```diff
@@ -28,6 +28,11 @@
 "   $ node test_code_action_corner_cases.js <test_code_action_corner_cases.vader
 
 Before:
+  runtime autoload/ale/lsp.vim
+
+  function! ale#lsp#NotifyForChanges(conn_id, buffer) abort
```
liskin, line 33: Tests that override ale functions should `runtime` the overridden files again in `After:`

liskin:
I don't really know enough about vim scripting to be able to tell whether doing `:edit` and `:w!` is a good idea, so I only commented on the parts that I do understand. But overall it looks like it might just work. :-)

test/test_code_action_corner_cases.vader
liskin: Also, since all modes go through a buffer, it may not be necessary to test everything three times. The only differences now are whether the buffer is switched and whether it's saved (the test plumbing would save it anyway if it was meant to not be saved), none of which is relevant to these corner case tests any more.

jeremija: Thanks, I was going to ask you if it's okay to remove this line!

test/test_code_action_corner_cases.vader
jeremija: Yeah that's way better than what I did, will update it!

jeremija: Done!

test/test_code_action_corner_cases.vader
jeremija: Done!

daliusd:
So it is closed. No need to review anymore. Anyway @liskin covered everything perfectly.

jeremija:
Oops if I closed this, it must've been by accident. Reopening now, sorry about this!

daliusd:
Overall I usually forget about `ALERename` and when I do I follow up it with `:wa`. So overall I think this is good fix.

bogdan:
@jeremija I tried your branch with tsserver and tried to rename `A` to `A1` from `b.ts` and here is what I get:

image

liskin:
I kind of forgot to review the followups to my previous review, so if this ever gets reopened and the issue reported by @bogdan is dealt with, I'll be happy to publish some fixups to better address my concerns about the tests (I vaguely remember I wanted to do more than the current version of this PR does).

Oh and unfortunately I still haven't managed to actually test this fix because I don't use any language/LSP combination where I'd need code actions across multiple files. :-/

jeremija:
Hey @liskin, @bogdan sorry for going MIA.

@bogdan I'm failing to reproduce your error, and I've been using this fork since I opened this MR full time.

Any chance you can post your `ALEInfo` and vim/nvim version?

I'd love to have this merged because I really think it improves the usability.

hsanson:
@jeremija cannot reopen the MR, Github tells me is because the branch has been re-created or force-pushed.

jeremija:
@hsanson shall I open a new one instead?

liskin:
> Hey @liskin, @bogdan sorry for going MIA.

Happens to all of us :-)

(I've had a taskwarrior task to review this PR since 2021-07-14 which I kept replanning to "tomorrow" until 2021-08-29, when I decided to stop lying to myself and just unscheduled it :-))

jeremija:
Ah sorry about that @liskin 🙂

PR #3789, opened 2021-06-29
Add stylua fixer for lua

rliebz:
Based this off the docs and a couple other existing fixers. Should be relatively straightforward, but definitely let me know if there's anything I can improve on here


hsanson (approved):
Seems correct, thanks for the contribution.

PR #3792, opened 2021-06-29
Deduplicate items in location list

gpanders:
Simple PR to deduplicate items added to the location list.

---

I recently added a hook to the Nvim LSP configuration to pipe all of its diagnostics through ALE. This works really well, except that there are often duplicate issues found by the LSP and ALE linters. This is a small change that simply deduplicates those entries.

I did not add a new test as the existing `test_quickfix_deduplication` test already covers deduplication. Let me know if you still want an additional test though.

N.B. I originally wrote this to be configurable with a new option `g:ale_deduplicate_loclist`, but in the spirit of avoiding adding more configuration levers than necessary I simply made it the default. If you think this should be configurable, let me know and I'll update the PR with a config option.

### DIFF
--- a/autoload/ale/list.vim
+++ b/autoload/ale/list.vim
@@ -44,6 +44,15 @@ function! s:ShouldOpen(buffer) abort
     return l:val is 1 || (l:val is# 'on_save' && l:saved)
 endfunction
 
+function! s:Deduplicate(list) abort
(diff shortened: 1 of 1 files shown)


hsanson (approved):
Looks good to me.

Issue #3492, opened 2020-12-17
g:ale_filename_mappings doesn't work for elixir-ls


giddie:
## Information

**VIM version**
NVIM v0.4.4
Build type: Release

Operating System: 
ArchLinux
## What went wrong

With the following config:
```vim
let g:ale_filename_mappings = {
\  'elixir-ls': [
\    ['/home/dannpg/Work/project', '/opt/app']
\  ]
\}
```

I see the following output from elixir-ls:
```
{"jsonrpc":"2.0","method":"window/logMessage","params":{"message":"\n10:29:34.893 [error] GenServer ElixirLS.LanguageServer.Server terminating\n** (File.Error) could not set current working directory to \"/home/dannpg/Work/project\": no such file or directory\n    (elixir 1.10.3) lib/file.ex:1523: File.cd!/1\n    (language_server 0.6.2) lib/language_server/serve ...
```

If I set:

```vim
let g:ale_lsp_root = {'elixir-ls': '/opt/app'}
```

Then `elixir-ls` operates OK, but it reports the docker paths (/opt/app), so Vim is unable to jump to the correct file.

### :ALEInfo
```
 Current Filetype: elixir
Available Linters: ['credo', 'dialyxir', 'dogma', 'elixir-ls', 'mix']
  Enabled Linters: ['credo', 'elixir-ls']
  Ignored Linters: []
 Suggested Fixers: 
  'mix_format' - Apply mix format to a file.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

let g:ale_elixir_elixir_ls_config = {'elixirLS': {'dialyzerEnabled': v:false}}
let g:ale_elixir_elixir_ls_release = '.vim-config/elixir-ls-wrappers'
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = v:null
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = v:null
let g:ale_disable_lsp = 0
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
[...]

PR #3805, opened 2021-07-07
Perform inverse filename mapping on GoToDefinition

giddie:
Rough fix for https://github.com/dense-analysis/ale/issues/3492. I'm afraid I don't know much vimscript and am short on time. I haven't been able to figure out how to pass the linter name through sensibly, so it's hard-coded, which works OK as a hack for my purposes.

It looks to me like this problem is likely to affect *all* LSP linters, so I'm surprised noone else has encountered this issue so far.


giddie:
Can we re-open this please? Is something needed to get this merged?

PR #3808, opened 2021-07-08
racket: support racket-langserver lsp

benknoble:
Not sure what tests are reasonable for an LSP linter. This worked for me after `raco pkg install racket-langserver`.

Notably, before I did that, the LSP didn't work (obviously), but I could see no error messages; only "executable check—success" for `racket`.

### DIFF
--- a/ale_linters/racket/langserver.vim
+++ b/ale_linters/racket/langserver.vim
@@ -1,0 +1,7 @@
+call ale#linter#Define('racket', #{
+\   name: 'racket-langserver',
+\   lsp: 'stdio',
+\   executable: 'racket',
+\   command: '%e -l racket-langserver',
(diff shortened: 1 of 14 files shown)


hsanson:
Check ruby solargraph and eclipselsp test for example on how to write tests for a language server. In short:

1. Create directory and files that simulate a racket project in test-files. From what I see a racket project must have a init.rkt file, also I see it looks for the top most one so to test this you can create a directory tree with init.rkt files and test that the linter detects the correct project dir with AssertLSPProject.
2. Use AssertLSPLanguage, AssertLSPConfig, AssertLSPOptions, and AssertLinter to check that things are properly set.

benknoble:
@hsanson done

hsanson (approved):
Thanks for the changes.

Issue #3412, opened 2020-10-26
[support/improvement] clj-kondo needs to change directory before running


NickeZ:
Hello,

clj-kondo doesn't correctly find the project file if it is executed in the wrong working directory. [the original author is not open to fix it](https://github.com/borkdude/clj-kondo/issues/1051). So my question is: is it possible to configure ALE to change directory to the directory of the file before running?

Thanks!

nospam2998:
I've got no insight at all, but notice that #3812 appears to say it fixes this issue. Would you agree @NickeZ or would this this benefit from some other kind of change in ale?

NickeZ:
Don't know, I don't use clojure anymore. Feel free to close.

nospam2998:
Thanks for your answer @NickeZ. Lets leave the issue open for now. Anyone stumbling upon this who actually use closure/clj-kondo is encouraged to please clarify the current situation and whether some action is desired before this can be considered to be resolved.

PR #3812, opened 2021-07-10
Add option to clojure clj-kondo linter

bootleq:
Allow define `ale_clojure_clj_kondo_options` variable to change command options.

This would help #3412, by setting `--config path/to/config.edn` can make `clj-kondo` work correctly.

Small change, so only mininal linter test added here.


bootleq:
Updated: I've done a rebase because found the test fail was due to the branch I forked out, now the `run-tests` were all green on my local environment.

hsanson (approved):
Thanks, looks good.

Issue #3806, opened 2021-07-08
isort needs to have the filename passed in as an arg


jeffwillette:
Using the isort fixer is convenient in python except for when editing `__init__.py` files which often need their imports in a specifici order which gets undone when I save the file and the fixers run. 

- ALE used to never give a problem (because I had added custom CLI args `let g:ale_python_isort_options = '--skip __init__.py --filter-files`)
- It seems like something changed along the way and now the file cannot be ignored via the above CLI args because the command which gets passed to isort reads from stdin `isort --skip __init__.py --filter-files - < /tmp/nvim0aAePZ/10745/__init__.py`
- isort issue reference https://github.com/PyCQA/isort/issues/1771#event-4993433333
- The solution is to make the command passed to isort have the filename passed in manually, the correct command should be...

```
isort --skip __init__.py --filter-files --filename /tmp/nvim0aAePZ/10745/__init__.py - < /tmp/nvim0aAePZ/10745/__init__.py
``` 

so the command called by ALE directly should have the `--filename` option added so that users custom args given to isort can work off the filename since the stdin stream doesn't give any file information. I don't have time to do this now, but I will leave this issue here and get to it when I have some free time if no one else takes it up.

jeffwillette:
Actually, I just had a look at the fixer itself and I am not sure how to go about fixing this since the stdin stream filename isn't anywhere I can see. Does anyone have an idea of how to accomplish this?

https://github.com/dense-analysis/ale/blob/f0887d3e6178482255f11aa378124aef3699245f/autoload/ale/fixers/isort.vim#L32

hsanson:
@jeffwillette you can use %s placeholder and it will be replaced with the absolute filename by ALE. Something like below should do what you need:

```
    return {
    \   'cwd': '%s:h',
    \   'command': ale#Escape(l:executable) . l:exec_args
    \   . ale#Pad('--filename %s')
    \   . (!empty(l:options) ? ' ' . l:options : '') . ' -',
    \}
```

Feel free to make a PR if the change fixes this issue and does not break anything else.

hsanson:
Should be fixed with #3815

PR #3815, opened 2021-07-14
added filename to stdin stream of isort

jeffwillette:
modified tests to run with new format


jeffwillette:
Solves #3806

hsanson (approved):
Thanks for the fix.

Issue #3703, opened 2021-04-21
Vim-8 Popup Windows


ldelossa:
Hello,

I am curious if there are any plans to support vim 8 popup window in addition to nvim's floating windows. Vim-lsp supports this: https://github.com/prabirshrestha/vim-lsp/blob/master/doc/vim-lsp.txt#L298 (not to make a comparison, just to say its feasible).

benknoble:
I would love this. _Theoretically_, with appropriate mods to autoload/ale/floating_preview.vim, it shouldn't be too hard.

hsanson:
Implemented in #3817

PR #3817, opened 2021-07-14
Vim popup

benknoble:
This PR adds vim-popups as an alternative to neovim's floating windows for things like ALEHover.

Not sure about testing this, but _it works_! At least for ALEHover and cursor_detail.

Related: #3703


hsanson (approved):
The changes look good and I tested it does not break NeoVim. I do not have a recent version of Vim that support popupwin so I am trusting this works and will merge as is a requested feature by many.

PR #3823, opened 2021-07-16
Update erlang format for OTP24

Kuroneer:
Erlang's erlc error format includes the column in OTP24.
See https://blog.erlang.org/My-OTP-24-Highlights/#column-number-in-warnings-and-errors


hsanson (approved):
Good, adds support for new format without breaking previous one and tests both. Thanks for the contribution.

Kuroneer:
Cheers! :beers: :)

PR #3825, opened 2021-07-17
Add support for `dart analyze`

ghsang:
It seems like `dartanalyzer` has been deprecated in favor of `dart analyze` (https://dart.dev/tools/dart-analyze).


hsanson (approved):
Thanks for the update.

PR #3826, opened 2021-07-18
Correct typo in python config filenames

foobarbyte:
In a previous PR I added asdf-vm's config file to the python project config files checked for. Unfortunately there was a typo in the file name .tool_versions should be .tool-versions. This PR corrects the typo in the autoload file and in the corresponding documentation.


hsanson (approved):
Thanks for the quick fix.

PR #3827, opened 2021-07-19
Add importMap option to deno Initialization Options

creativenull:
This PR introduces the addition of `importMap` key used in the Deno LSP to be able to map URLs into module names. [Import maps manual can be found here](https://deno.land/manual@v1.12.0/linking_to_external_code/import_maps) and the json file found in their tests use `importMap` as part of the `InitializationOptions` [which can be found here](https://github.com/denoland/deno/blob/8ed83cba7e2eeabb3af1fd65579dacd782f50644/cli/tests/lsp/initialize_params.json#L16).

By default, `import_map.json` file will be searched in the project root directory and be used, if found, and gives the user the choice to use another filename to identify the `importMap` file.

~~In addition to that, I've also added deno as a linter option for `javascript`, let me know if you'd like that as a separate PR instead.~~

For tests, I've added two different import map files (default and a user defined file) in the `test/test-files/typescript` directory.

Edit: I've removed deno as a linter for javascript for now, making a separate PR for it. This will just focus on adding `importMap` option to deno lsp initialization options.


hsanson (approved):
Thanks. Looks good.

PR #2963, opened 2020-01-10
Python support poetry

unc0:
Add poetry automatic detect support to python linters and black fixer.
https://github.com/dense-analysis/ale/issues/1656#issuecomment-414760077

PR #3834, opened 2021-07-24
Python support poetry

danielroseman:
Updated version of #2963 that fixes the conflicts.


hsanson (approved):
Thanks for the big contribution.

PR #3855, opened 2021-08-04
Use md5(1) when available

jparise:
`md5sum` isn't available by default on macOS. Instead, it ships the
BSD-style `md5(1)` command, which does the same thing but with different
arguments.

With this change, `run-tests` works out-of-the-box on macOS.


hsanson (approved):
Saw this on a larger PR before. Thanks for making it self-contained and easier to review/merge.

PR #3867, opened 2021-08-09
purs-tidy fix

toastal:
Sorry, I had made an error in my last merge request around how the stdin stuff works. I just tested locally and this works now.


hsanson (approved):
Thanks for finding out and quickly fixing it.

PR #3868, opened 2021-08-09
Dhall fixes: use stdin, doc errors

toastal:
Retested locally on Nix and is working for me


hsanson (approved):
Great, thanks for the fixes.

PR #3873, opened 2021-08-14
Close #3872 - Add eslint-plugin-jsonc as a linter for JSON, JSONC and JSON5

### LINKED ISSUE #3872: Add support for eslint-plugin-jsonc plugin
**Name:** eslint with eslint-plugin-jsonc plugin
**URL:** https://ota-meshi.github.io/eslint-plugin-jsonc/ 

Linter and fixer for JSON, JSONC and JSON5 files that uses `eslint`. Since `eslint` is already supported for other filetypes, this should be trivial. I'll make a very basic pull request with the minimum required configuration and let me know if it makes sense supporting it so I can improve it. 🐸

Cheers

jpesce:
Use the same lint configuration as eslint for javascript in JSON, JSONC and JSON5 files. Requires `eslint-plugin-jsonc` in `.eslintrc`.

This is the first time working with the codebase, so let me know if there's an easier way to reuse the eslint configuration without having to copy everything. :)

If you think the PR makes sense, I'll try to write the tests.

Closes #3872


ale_linters/json/eslint.vim
```diff
@@ -0,0 +1,16 @@
...
+"
+" Requires eslint-plugin-jsonc or a similar plugin to work
+"
+" Uses the same funtcions as ale_linters/javascript/eslint.vim by w0rp
+" <devw0rp@gmail.com>
+
+call ale#linter#Define('json', {
+\   'name': 'eslint',
+\   'output_stream': 'both',
+\   'executable': function('ale#handlers#eslint#GetExecutable'),
```
hsanson, line 12: Would be better to create a different GetExecutable function that somehow checks if the plugin for JSON linting is available. If not available return an empty string so the linter is not run for these filetypes if the plugin is not installed.

To run commands (e.g. for checking if the plugin is present) without freezing vim/neovim make sure to use the ale#comand#Run() function that works in background. Check any linter that uses this function (e.g. eclipselsp.vim) to determine the version of the tools used  for examples.

hsanson:
Looks good. You are already reusing well the code for eslint and just defining the linter for additional languages. The only problem I see is that this would enable the eslint linter for JSON, JSONC, and JSON5 files even if the plugin is not present.

Ensure to update the list of supported tools in  supported-tools.md, ale-supported-languages-and-tools.txt, and add entries to ale.txt for the newly supported languages. See ale-graphql-eslint for example.

ale_linters/json/eslint.vim
kevinoid: Hi @hsanson, thanks for reviewing this PR!  I'm not the author, but am excited to see this added.

> Would be better to create a different GetExecutable function that somehow checks if the plugin for JSON linting is available.

Although I agree with this idea, I think it's a bit unreasonable to require it as part of this PR, since neither `graphql/eslint` nor `typescript/eslint` implement such checks as far as I can see.  Also, doing the check in `GetExecutable()` would cause eslint to be invoked twice on every call, unless some (tricky) caching logic were added, correct?

jpesce:
> Ensure to update the list of supported tools in supported-tools.md, ale-supported-languages-and-tools.txt, and add entries to ale.txt for the newly supported languages. See ale-graphql-eslint for example.

Updated 🐸

ale_linters/json/eslint.vim
jpesce: Hmmm, to me it makes sense what @kevinoid said. 

Anyway, I was trying to figure out how to check if a plugin is installed or not, but I haven't found a solid approach:

Using `npm list` or something similar requires the user to be actually using `npm` (we would also need to check with `yarn`). Plus, even if the plugin is installed, it may not be properly configured in `.eslintrc` (not using `jsonc-eslint-parser` as the parser for JSON* files for example).

`eslint --print-info ficticiousfile.json` might work, but I found out there are multiple eslint plugins for JSON (for example ([eslint-plugin-json](https://www.npmjs.com/package/eslint-plugin-json) and [eslint-plugin-jsonc](https://www.npmjs.com/package/eslint-plugin-jsonc) both work with JSON).

I think the ALE commands (e.g. `:ALEInfo`) gives the user enough info to figure out what's going on and decide for themselves.

Let us know what you think @hsanson. 🐸

hsanson: > Although I agree with this idea, I think it's a bit unreasonable to require it as part of this PR, since neither graphql/eslint nor typescript/eslint implement such checks as far as I can see.

Indeed, this is just a recommendation, not a requirement. I do not use eslint myself so I have no strong opinions on the matter.  If this works for those that do use the tool I am fine with it.

> Anyway, I was trying to figure out how to check if a plugin is installed or not, but I haven't found a solid approach

Thanks for taking the time to investigate. If it is too complicated to do then better leave it as it is. Simple is best.
We can improve on this in the future if it becomes an issue .

hsanson:
@jpesce sorry I forgot to mention that the list of supported linters must be sorted alphabetically or the tests would fail. This is the correct sorting:

```
* JSON
  * [eslint](http://eslint.org/)
  * [fixjson](https://github.com/rhysd/fixjson)
  * [jq](https://stedolan.github.io/jq/)
  * [jsonlint](https://github.com/zaach/jsonlint)
  * [prettier](https://github.com/prettier/prettier)
  * [spectral](https://github.com/stoplightio/spectral)
* JSON5
  * [eslint](http://eslint.org/)
* JSONC
  * [eslint](http://eslint.org/)
```

1. JSON5 section must be before JSONC
2. eslint in the JSON section must be the first entry.

jpesce:
> @jpesce sorry I forgot to mention that the list of supported linters must be sorted alphabetically or the tests would fail.

Makes sense, thanks! Updated them 🐸

jpesce:
Sorry, there was a misalignment in `ale.txt`

Fixed it in https://github.com/dense-analysis/ale/pull/3873/commits/d0156e6fe6bfcc4a99f6bd24ad849dbd559a504e

hsanson:
Test still failing... I think because the order within ale.txt must also match the order in supported tools.

If you have docker daemon would be good to run the tests locally before pushing the chages:

> ./run-tests -v --linters-only

jpesce:
Sorry about that, I installed docker and ran the tests and now it's passing.

One remark though — the test expects the following order in `ale.txt`:
1. json
2. jsonc
3. json5

but it wierdly expects json5 to be before jsonc in `ale-supported-languages-and-tools.txt`, like so:
1. json
2. json5
3. jsonc

<details>
<summary>Here's the test output</summary>

```
./run-tests -v --linters-only
Docker run image denseanalysis/ale:180c2d67c6a055af98b836ae9d28e3db ready
Starting Vint...
Starting Custom checks...

Running Vint to lint our code
Vint warnings/errors follow:


Running custom linting rules


Checking for duplicate tags
Duplicate tags follow:


Checking for invalid tag references
Invalid tag references tags follow:

diff supported-tools.md and doc/ale-supported-languages-and-tools.txt tables
Differences follow:

Look for badly aligned doc tags
Badly aligned tags follow:

Look for table of contents issues

Check for bad ToC sorting:

Check for mismatched ToC and headings:

Check Python code

.....
----------------------------------------------------------------------
Ran 5 tests in 0.002s

OK

All tests passed!
```

</details>


Let me know how you prefere to continue 🐸
Thanks!

hsanson (approved):
Thanks for the contribution and patience.

PR #3878, opened 2021-08-24
Introduce an ALELSPStarted autocommand

### LINKED ISSUE #3861: Fire a User event when an LSP starts
Inspired by https://github.com/prabirshrestha/vim-lsp, it would be nice to have the ability to make maps for ALE LSP commands only when the LSP is active. Then I would work as follows:

- use normal `gd` and other vim commands, possibly configured by ftplugins, as an approximation when no LSP is available, ALE is off, etc.
- when LSP is available, have the commands delegate to ALE for a more exact experience

Additionally, I would be able to `setlocal omnifunc=ale#completion#omnifunc` for LSP-enabled buffers, rather than my current `setlocal completefunc` so that I don't lose a good `omnifunc` when I don't have an LSP.

## Technical details

Since the autocmd User event doesn't permit patterns, we have to be careful to fire it for the right buffers. I've left a link that explores that code below.

Possibly useful for implementation: https://github.com/prabirshrestha/vim-lsp/blob/fb0a72306e1fc42c1ce8909aa1e420ddfccf6fbf/autoload/lsp.vim#L345

jparise:
This User autocommand is trigged immediately after an LSP process is
successfully initialized. This provides a way to perform any additional
initialization work, such as setting up buffer-level mappings.

Fixes #3861

### DIFF
--- a/autoload/ale/lsp_linter.vim
+++ b/autoload/ale/lsp_linter.vim
@@ -271,6 +271,30 @@ function! ale#lsp_linter#OnInit(linter, details, Callback) abort
         call ale#lsp#NotifyForChanges(l:conn_id, l:buffer)
     endif
 
+    " Tell the relevant buffer that the LSP has started via an autocmd.
+    if l:buffer > 0
+        if l:buffer == bufnr('')
+            silent doautocmd <nomodeline> User ALELSPStarted
+        else
+            execute 'augroup ALELSPStartedGroup' . l:buffer
+                autocmd!
+
+                execute printf(
+                \   'autocmd BufEnter <buffer=%d>'
+                \       . ' doautocmd <nomodeline> User ALELSPStarted',
+                \   l:buffer
+                \)
+
+                " Replicate ++once behavior for backwards compatibility.
+                execute printf(
+                \   'autocmd BufEnter <buffer=%d>'
+                \       . ' autocmd! ALELSPStartedGroup%d',
+                \   l:buffer, l:buffer
+                \)
+            augroup END
+        endif
+    endif
+
     call a:Callback(a:linter, a:details)
 endfunction
 
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -4307,6 +4307,13 @@ ALEJobStarted                                           *ALEJobStarted-autocmd*
   |ale#engine#IsCheckingBuffer()| over |ALELintPre-autocmd|, which is actually
   triggered before any linters are executed.
 
+ALELSPStarted                                           *ALELSPStarted-autocmd*
+                                                                *ALELSPStarted*
+
+  This |User| autocommand is trigged immediately after an LSP connection is
+  successfully initialized. This provides a way to perform any additional
+  initialization work, such as setting up buffer-level mappings.
+
 
 ALEWantResults                                         *ALEWantResults-autocmd*
                                                                *ALEWantResults*
--- a/test/lsp/test_lsp_startup.vader
+++ b/test/lsp/test_lsp_startup.vader
@@ -14,6 +14,7 @@ Before:
   let g:job_map = {}
   let g:emulate_job_failure = 0
   let g:next_job_id = 1
+  let g:lsp_started = 0
 
   let g:socket_map = {}
   let g:emulate_socket_failure = 0
@@ -102,17 +103,17 @@ Before:
     call ale#lsp#HandleMessage(l:conn_id, l:data)
   endfunction
 
-  function! Start() abort
+  function! Start(buffer) abort
     let l:linter = values(ale#linter#GetLintersLoaded())[0][0]
 
     return ale#lsp_linter#StartLSP(
-    \ bufnr(''),
+    \ a:buffer,
     \ l:linter,
     \ {linter, details -> add(g:calls, [linter.name, details])},
     \)
   endfunction
 
-  function! AssertInitSuccess(linter_name, conn_prefix, language, root, command) abort
+  function! AssertInitSuccess(linter_name, conn_prefix, language, root, command, buffer) abort
     let l:messages = PopMessages()
 
     if a:linter_name is# 'tsserver'
@@ -121,7 +122,7 @@ Before:
       \   {
       \     'seq': v:null,
       \     'arguments': {
-      \       'file': expand('%:p'),
+      \       'file': expand('#' . a:buffer . ':p'),
       \     },
       \     'type': 'request',
       \     'command': 'open',
@@ -252,7 +253,7 @@ Before:
       \     'jsonrpc': '2.0',
       \     'params': {
       \       'textDocument': {
-      \         'uri': ale#path#ToURI(expand('%:p')),
+      \         'uri': ale#path#ToURI(expand('#' . a:buffer . ':p')),
       \         'version': ale#lsp#message#GetNextVersionID() - 1,
       \         'languageId': a:language,
       \         'text': "\n",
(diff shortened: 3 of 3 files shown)


autoload/ale/lsp_linter.vim
```diff
@@ -271,6 +271,20 @@ function! ale#lsp_linter#OnInit(linter, details, Callback) abort
         call ale#lsp#NotifyForChanges(l:conn_id, l:buffer)
     endif
 
+    " Tell the relevant buffer that the LSP has started.
+    if l:buffer > 0 && l:buffer == bufnr('')
```
jparise, line 275: This `l:buffer > 0` check might not be relevant / useful here.

autoload/ale/lsp_linter.vim
```diff
@@ -271,6 +271,20 @@ function! ale#lsp_linter#OnInit(linter, details, Callback) abort
...
+        silent doautocmd <nomodeline> User ALELSPStarted
+    else
+        execute 'augroup ALELSPStartedGroup' . l:buffer
+            autocmd!
+
+            execute printf(
+            \   'autocmd BufEnter <buffer=%d> User ALELSPStarted',
```
benknoble, line 282: This looks wrong. It doesn't use `doautocommand`.

It also doesn't get removed once fired. 

See https://github.com/prabirshrestha/vim-lsp/blob/fb0a72306e1fc42c1ce8909aa1e420ddfccf6fbf/autoload/lsp.vim#L345 for one way to approach it. 

The test likely doesn't see it because it doesn't switch buffers after asking the LSP to start but before this code-path is hit.

autoload/ale/lsp_linter.vim
benknoble: Probably neither branch should run if the buffer number is invalid.

autoload/ale/lsp_linter.vim
jparise: Yeah, I totally lost the plot on this branch when I started reading about `++once` and didn't come back to finish it up.

jparise:
I'd like to write a test for the "`ALELSPStarted` in another buffer" case, but I think that's going to require a bit more work. I couldn't find any existing test patterns that work across buffers like that so I'll be back once I can get something working.

jparise:
I needed to evolve the test helper functions in `test/lsp/test_lsp_startup.vader` a bit so make them buffer-aware, but now we have good test coverage for the new functionality.

I would have liked to use `++once` (which is already used in `autoload/ale/floating_preview.vim`), but that code path appears to be implicitly protected by some version/feature detection that coincides with `++once`'s availability. It isn't too burdensome to add a backwards compatible expression similar to what vim-lsp is doing, though.

autoload/ale/lsp_linter.vim
```diff
@@ -271,6 +271,29 @@ function! ale#lsp_linter#OnInit(linter, details, Callback) abort
...
+
+                execute printf(
+                \   'autocmd BufEnter <buffer=%d>'
+                \       . ' doautocmd <nomodeline> User ALELSPStarted',
+                \   l:buffer
+                \)
+
+                " Replicate ++once behavior for backwards compatibility.
+                execute printf(
+                \   'autocmd! ALELSPStartedGroup%d BufEnter <buffer>',
```
benknoble, line 290: This should be `autocmd! ALELSPStartedGroup BufEnter <buffer=%d>`, and see other comment.

autoload/ale/lsp_linter.vim
```diff
@@ -271,6 +271,29 @@ function! ale#lsp_linter#OnInit(linter, details, Callback) abort
...
+    if l:buffer > 0
+        if l:buffer == bufnr('')
+            silent doautocmd <nomodeline> User ALELSPStarted
+        else
+            execute 'augroup ALELSPStartedGroup' . l:buffer
+                autocmd!
+
+                execute printf(
+                \   'autocmd BufEnter <buffer=%d>'
+                \       . ' doautocmd <nomodeline> User ALELSPStarted',
```
benknoble, line 284: You want to run two commands on `BufEnter`:

1. `doautocmd`…
2. Remove the autocommand

Otherwise you haven't actually replicated `++once` correctly. The placement of the other autocmd means it's deleted before it even runs (though the syntax was wrong).

If we get rid of all the execute/printfs, the template we want is one of

```vim
augroup ALELSPStartedGroup
  autocmd!
  autocmd BufEnter <buffer=N> doautocmd <nomodeline> User ALELSPStarted | autocmd! ALELSPStartedGroup <buffer=N>
augroupEnd
```

_or_

```vim
augroup ALELSPStartedGroup
  autocmd!
  autocmd BufEnter <buffer=N> doautocmd <nomodeline> User ALELSPStarted
  autocmd BufEnter <buffer=N> autocmd! ALELSPStartedGroup <buffer=N>
augroup END
```

Either one is fine. Then just execute/printf to substitute `N` with the `l:buffer`.

benknoble:
In case it's not clear, I'm super stoked for this and very glad you're putting the time in on the feature @jparise ! I just want to make sure it works :) and since we haven't heard from any maintainers yet, you and I have the only eyes on the code. Kudos on getting the tests working, by the way.

benknoble:
> I would have liked to use `++once` (which is already used in `autoload/ale/floating_preview.vim`), but that code path appears to be implicitly protected by some version/feature detection that coincides with `++once`'s availability.

I think that's `has('popupwin')` in vim; they're technically an 8.2 release feature, so `++once` is already there by then.

autoload/ale/lsp_linter.vim
jparise: I wasn't able to get that exact suggestion to work. I would run into errors like this:

```
    (12/12) [EXECUTE] (X) Vim(autocmd):E216: No such event: <buffer>
      > BufEnter Autocommands for "<buffer=3>"
```

I fixed the syntax and retained the buffer-specific autocommand group, and that does seem to act as intended.

jparise:
> > I would have liked to use `++once` (which is already used in `autoload/ale/floating_preview.vim`), but that code path appears to be implicitly protected by some version/feature detection that coincides with `++once`'s availability.
> 
> I think that's `has('popupwin')` in vim; they're technically an 8.2 release feature, so `++once` is already there by then.

There's a small temptation to "cheat" and make this new autocommand feature only available if `++once` is available. 😉

autoload/ale/lsp_linter.vim
benknoble: You know, I might have missed that the group was created with the buffer number attached, in which case your approach is correct. My bad.

autoload/ale/lsp_linter.vim
benknoble: (See another comment of mine about the group name: it actually has the buffer number attached!)

hsanson:
@jparise @benknoble I usually leave core changes like this to @w0rp and restrict myself to check new/updated linters but this particular feature is very useful so I am ok merging it if you  can confirm is working for you.

In my testing seems to work fine and I can use it to simplify some of my configuration to set mappings when LSP linters are active.

hsanson (approved):
Looks good and useful.

benknoble:
I haven't tested it out yet, but I should have time this long weekend. Don't feel like you have to wait for me, but a third set of eyes never hurts :)

benknoble:
Adding a buffer-local map worked fine; I didn't test calling a function, but I'm not too worried. I did test a bare `:echom` and never saw the output, though (even in `:messages`).

PR #3887, opened 2021-09-02
Cppcheck backwards compat 1.34

tsjordan-eng:
Add support for cppcheck versions that don't support `{column}` or `{inconclusive:text}` in `--template`. This merge parses the output with and without the support and defaults to column 1.


hsanson (approved):
Thanks for the fix.

PR #3888, opened 2021-09-03
Use echon over echom for cursor echo

amadeus:
I feel like dynamic messages like this really should not clutter the `:messages` view, since they are very ephemeral in nature.

This would be my first time contributing to Ale, I did look over the development stuff but if I missed something please let me know!

### DIFF
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -27,13 +27,11 @@ function! ale#cursor#TruncatedEcho(original_message) abort
         silent! setlocal shortmess+=T
 
         try
-            let l:winwidth = winwidth(0)
-
             " echon will not display the message if it exceeds the width of
             " the window
-            if l:winwidth < strdisplaywidth(l:message)
+            if &columns < strdisplaywidth(l:message)
                 " Truncate message longer than window width with trailing '...'
-                let l:message = l:message[:l:winwidth - 5] . '...'
+                let l:message = l:message[:&columns - 5] . '...'
             endif
 
             echon l:message
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -28,12 +28,14 @@ function! ale#cursor#TruncatedEcho(original_message) abort
 
         try
             let l:winwidth = winwidth(0)
+
             " echon will not display the message if it exceeds the width of
             " the window
             if l:winwidth < strdisplaywidth(l:message)
                 " Truncate message longer than window width with trailing '...'
                 let l:message = l:message[:l:winwidth - 5] . '...'
             endif
+
             echon l:message
         catch /^Vim\%((\a\+)\)\=:E523/
             " Fallback into manual truncate (#1987)
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -27,6 +27,13 @@ function! ale#cursor#TruncatedEcho(original_message) abort
         silent! setlocal shortmess+=T
 
         try
+            let l:winwidth = winwidth(0)
+            " echon will not display the message if it exceeds the width of
+            " the window
+            if l:winwidth < strdisplaywidth(l:message)
+                " Truncate message longer than window width with trailing '...'
+                let l:message = l:message[:l:winwidth - 5] . '...'
+            endif
             echon l:message
         catch /^Vim\%((\a\+)\)\=:E523/
             " Fallback into manual truncate (#1987)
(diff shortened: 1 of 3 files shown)


amadeus:
After some testing of running on this branch, I did notice some issues with the initial implementation.

The basic finding was that any message longer than the window width would just render nothing. I had to adopt the logic from the catch portion to fix it, however, interestingly enough I had to increment the 4 to a 5 in the number of chars it would splice off. We might also want to increment the number pulled off in the catch as well, since I think 4 still triggers a weird newline situation.

But to make this all a bit more complicated, there's almost no use for the catch anymore I don't think, since it essentially tries the same logic but logs to echom instead...

Anyways, what I can say, I really appreciate the quality of life of `:messages` not getting overrun with type logs, but can understand how this PR is now a bit more complicated from an acceptance perspective.

amadeus:
also it looks like vint is failing on the use of echon, saying I should use execute echo, which in this case is not applicable, because echon doesn't work with execute (from what I've been able to test).  Will probably need some sort of ignore for this

hsanson:
That is a custom linting rule (see :h ale-coding-standards). I am not familiar with the details of why the rule exists so I am unable to tell if this can be approved or not.  This change will need to be reviewed and approved by @w0rp.

amadeus:
Well, I fixed the lint warnings (you can't use `echon` with `execute`), but that looks like it just cascades into other test failures because it looks like the existing tests are looking for the echo'd information to be written somewhere and validated. 

Additionally, I just made a small improvement to the output to take into account the full window width, not the split width, which results in proper truncation (before it was truncating based on split widths, which isn't useful).

I've been using this feature in a branch for months and I've found it to be quite a quality of life improvement, and would love it to somehow be merged, just not sure what to do about all the tests that will fail as a result -- not sure if you guys want to ever accept this or not?

w0rp (approved):

w0rp:
I'll force merging this and fix tests if they fail later, as we just updated the build to test with different NeoVim versions.

Cheers! :beers:

w0rp:
I fixed the tests, restored the echo line check, and made it so you can ignore problems form the custom linting rules by adding `" no-custom-checks` above lines you want to ignore.

amadeus:
omg, thanks so much! really appreciate the merge!

ajukraine:
I've just upgraded ALE and this change is not working well in my setup. I use MacOS and tried on different terminals (kitty, Alacritty, Terminal.app).

The problem I have is that whenever my cursor goes to a line with Ale warning/error, then I get a message shown in the bottom that also asks me to "press enter" in order to move back normal mode. Quite annoying as I don't want my editor stop on every linting...

dsifford:
Chiming in here to second @ajukraine's sentiment. This has made my editing experience much worse on (at least) macos.

dsifford:
setting `set noshowcmd` works as a temp workaround possibly.

amadeus:
I haven't been able to reproduce your issues at all. Are there any more details you can provide around how to replicate these issues? I have been running a version of this for a few months and never seen a press enter prompt.

dsifford:
@amadeus Here's my [`init.nvim`](https://github.com/dsifford/.dotfiles/blob/master/config/nvim/init.vim), if that helps narrow it down.

Nothing really more to add that @ajukraine didn't already say.

ajukraine:
Mine config setup is https://github.com/ajukraine/my.nvim, with ALE specific settings settings here https://github.com/ajukraine/my.nvim/blob/main/lua/aj/ale.lua.

hsanson:
I am also unable to reproduce this issue. Tested with Konsole and Kitty on a python hello world file using pylint, flake8, and jedils.  I use neovim 0.7 on Linux.

This is the python file:

```
"""
Hello sample
"""
import numpy as np


def say3(msg):
    """
    This is documentation
    """
    print("Hello %s" % msg)
    str = "This is a very super long line with many many many characters spanning over 100 chars long to test flake configuration"
    str2 = "This is a very super long line with many spanning over 79 chars to test"

    print("Hello %s %s %s" % (str, str2, msg))

    a = np.array([])


say3("World!")
```

This is the vimrc:

```

set runtimepath=$VIMRUNTIME
set runtimepath^=$HOME/Projects/vim/ale
let g:ale_linters = {'python': ['jedils', 'pylint', 'flake8']}
```

This is the command to start the editor:

```
nvim -u vimrc hello.py
```


I recommend creating a separate issue to keep track of this. Adding comments to a closed PR can be missed by other maintainers.

ajukraine:
@hsanson thank you for taking care of the issue and all us affected

Issue #3886, opened 2021-09-01
Error detected decoding JSON from Neovim LSP


hsanson:
## What went wrong

Reported in https://github.com/dense-analysis/ale/issues/3373#issuecomment-895494932 and https://github.com/dense-analysis/ale/issues/3373#issuecomment-910196544. Seems that some language servers (NVIM LSP?) send a list instead of the expected string or list.

Seems to happen with rust-analyzer and PHP LSP.

```
Error detected while processing function ale#completion#HandleUserData[2]..ale#util#FuzzyJSONDecode:
line    5:
E714: List required
```

## Reproducing the bug

???

hsanson:
If someone can provide a configuration that reliably reproduces the issue would greatly help debugging it.  I am unable to get this to happen with any of the LSP I use daily (javalsp, solargraph, pyright).

Mte90:
I am getting with this https://github.com/mte90/dotfiles
I am using also compe (I will switch from that soon) and I don't know if this is part of the issue.

Also I don't see a way to disable ALE to be executed with LSP at all as the globals seems ignored.

Mte90:
I am getting the same error also with coq_nvim.
There is something I can do to help debugging this?

In my case I am not interested in ALE support in completion.

hsanson:
You can try debugging the issue:

  - https://inlehmansterms.net/2014/10/31/debugging-vim/
  
1. Start vim in debug mode "vim -D" using minimal vimrc and file that causes the issue.
2. In the debugger add a breakpoint at the problematic function (e.g. `breakadd func ale#completion#HandleUserData`)
3. Continue using vim as normal (e.g. execute `cont` in debugger prompt).
4. When the function is called and the debugger stops at the breakpoint inspect the value of the argument (e.g. `echo a:completed_item`).
5. If the output is not JSON then we know it will fail latter. Provide the value in this issue to see if we can figure out what it is and where it comes from.

hsanson:
Note that this function is called not only for auto-completion. It is also used to handle LSP CodeActions. That is why this is invoked even if ALE auto-completion is disabled.

Mte90:
Get stopped in that way to that line:
![immagine](https://user-images.githubusercontent.com/403283/132236082-4abaad77-9067-4423-8ee2-8ba5753261fc.png)
If i continue I am getting the usual error.

Added an echo of that variable and is not empty at all:

[...]

PR #3890, opened 2021-09-06
Fix FuzzyJSON when is already a JSON

Mte90:
A json is a dict so json_decode generate an error because expects a list.

Fix https://github.com/dense-analysis/ale/issues/3886

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -1001,9 +1001,13 @@ endfunction
 
 function! ale#completion#HandleUserData(completed_item) abort
     let l:user_data_json = get(a:completed_item, 'user_data', '')
-    let l:user_data = !empty(l:user_data_json)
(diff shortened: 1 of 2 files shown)


hsanson (changes requested):
Thanks for the contribution but I think this is not good way to fix the problem.

The FuzzyJSONDecode() method is clear on what it expects as input (e.g. string or list of strings) so if the caller is sending something different than what is expected then the caller needs to be fixed.

Instead of modifying the FuzzyJSONDecode() method better modify the ale#completion#HandleUserData() method to check the user data and if a dictionary skip the json decoding step.

Mte90:
Done :-)

hsanson:
Seems to still have issues (tests failing).  I am closing this MR in favor of this other one #3894 that addresses the issue and further improves the conditional checks.

Mte90:
Cool the important that a fix landed, I have not so much experience with VimL

PR #3893, opened 2021-09-08
Improve the thriftcheck handler pattern

jparise:
More recent versions of thriftcheck use a more compliant GCC-style
output format which includes a space before the "severity" group.
This matches similar tools, like shellcheck.

This change adjusts the handler's pattern to parse this format in a
backwards-compatible way (even though backwards compatibility isn't
critical long-term as thriftcheck itself is close to its 1.0 release).


hsanson (approved):
Thanks, looks good.

PR #3895, opened 2021-09-08
[python/pyre.vim] Fix pyre persistent behavior

oliverralbertini:
It's necessary to provide a `-l` option to pyre with the closest parent
directory containing a `.pyre_configuration.local` file, or simply
change directory (cwd) to the root of the pyre project. Thanks to Ken
Verbosky for the code that fixes this.

Error seen when not using such a solution:

```
1031.473923 on 6: Dropping message 'ƛ Background task unexpectedly quited: Invalid configuration: Cannot find any source files to analyze. Either `source_directories` or `targets` must be specified.
```

Issue with this approach is that if you are editing files under
different projects, the `pyre persistent` process is not re-created for
each file. We have to do `:ALEStopAlllsps` in order for the process to
start with the new working directory.

### DIFF
--- a/ale_linters/python/pyre.vim
+++ b/ale_linters/python/pyre.vim
@@ -22,12 +22,15 @@ endfunction
 
 function! ale_linters#python#pyre#GetCommand(buffer) abort
     let l:executable = ale_linters#python#pyre#GetExecutable(a:buffer)
-
-    let l:exec_args = l:executable =~? 'pipenv\|poetry$'
-    \   ? ' run pyre persistent'
-    \   : ' persistent'
+    let l:exec_args = (l:executable =~? 'pipenv\|poetry$' ? ' run pyre' : '') . ' persistent'
 
     return ale#Escape(l:executable) . l:exec_args
+endfunction
+
+function! ale_linters#python#pyre#GetCwd(buffer) abort
+    let l:local_config = ale#path#FindNearestFile(a:buffer, '.pyre_configuration.local')
+
+    return fnamemodify(l:local_config, ':h')
 endfunction
 
 call ale#linter#Define('python', {
@@ -37,4 +40,5 @@ call ale#linter#Define('python', {
 \   'command': function('ale_linters#python#pyre#GetCommand'),
 \   'project_root': function('ale#python#FindProjectRoot'),
 \   'completion_filter': 'ale#completion#python#CompletionItemFilter',
+\   'cwd': function('ale_linters#python#pyre#GetCwd'),
 \})


oliverralbertini:
I'm seeing a couple of failures, but I'm not sure they're related to my change:

```
[00:01:44]     (5/6) [EXECUTE] (X) {'line_length': 17, 'conn_id': 347, 'source': 'ale-import', 'column': 7, 'request_id': 1, 'line': 2, 'additional_edits_only': 1, 'completion_filter': function('ale#completion#python#CompletionItemFilter'), 'prefix': 'missingword'} should be equal to {'request_id': 1, 'conn_id': 347, 'source': 'ale-import', 'column': 7, 'line_length': 17, 'line': 2, 'additional_edits_only': 1, 'completion_filter': 'ale#completion#python#CompletionItemFilter', 'prefix': 'missingword'}
[00:01:44]     (6/6) [  GIVEN] Some example Python code
[00:01:44]     (6/6) [EXECUTE] ALEImport should tell the user when no completions were found from a language server
[00:01:44]     (6/6) [EXECUTE] (X) {'line_length': 17, 'conn_id': 347, 'source': 'ale-import', 'column': 7, 'request_id': 1, 'line': 2, 'additional_edits_only': 1, 'completion_filter': function('ale#completion#python#CompletionItemFilter'), 'prefix': 'missingword'} should be equal to {'request_id': 1, 'conn_id': 347, 'source': 'ale-import', 'column': 7, 'line_length': 17, 'line': 2, 'additional_edits_only': 1, 'completion_filter': 'ale#completion#python#CompletionItemFilter', 'prefix': 'missingword'}
```

hsanson:
@oliverralbertini you changed the `completion_filter` value in the linter definition from `ale#completion#python#CompletionItemFilter` to `function('ale#completion#python#CompletionItemFilter')` and that is why the test is failing.
 
 Revert the change or update the tests with the new value, whichever makes the linter work properly.

oliverralbertini:
> @oliverralbertini you changed the `completion_filter` value in the linter definition from `ale#completion#python#CompletionItemFilter` to `function('ale#completion#python#CompletionItemFilter')` and that is why the test is failing.

Thanks @hsanson, great catch. I will just revert it, seems from the docs that it can be either a string of funcref:

```
  `completion_filter`     A |String| or |Funcref| for a callback function
                         accepting a buffer number and a completion item.

                         The completion item will be a |Dictionary| following
                         the Language Server Protocol `CompletionItem`
                         interface as described in the specification,
                         available online here:
                         https://microsoft.github.io/language-server-protocol
```

hsanson (approved):
Thanks, looks good.

PR #3900, opened 2021-09-14
[ale-python-root] add `.pyre_configuration.local` to list of files

oliverralbertini:
Without this, we have one `pyre` process running across different pyre
projects. With this change, files in different projects can be linted
with pyre at the same time.


hsanson (approved):
Thanks for the contribution. Looks good to me.

PR #3901, opened 2021-09-14
ALECodeAction fix: check linter name when searching for nearest error.

daliusd:
It is easier to explain this fix with an example:

* tsserver and LSPs ask for error information when you want to fix
  error. tsserver `ts@getCodeFixes` command needs tsserver error code.

* now let's imagine that user has eslint and tsserver in use. Sometimes
  both can report same error in different way.

* Now there is no guarantee which error will come first and if eslint
  error comes first then tsserver will not return code fixes as we are
  passing wrong error code to it.

This fix will return proper error code based on linter.


hsanson (approved):
Looks like a sensible check to do. Thanks for the fix.

daliusd:
> Looks like a sensible check to do. Thanks for the fix.

For some time I thought that I had misconfigured typescript until I figured out that I have left problem in this code. I think something similar can be done for LSP part as well.

daliusd:
:beers:

PR #3902, opened 2021-09-15
[YAML] CircleCI: skip checking for new tool version

goetzc:
This way the tool runs a bit faster and we don't create unneeded network requests. Don't know if there are other network requests still occurring.

### DIFF
--- a/ale_linters/yaml/circleci.vim
+++ b/ale_linters/yaml/circleci.vim
@@ -28,7 +28,7 @@ endfunction
 call ale#linter#Define('yaml', {
 \   'name': 'circleci',
 \   'executable': {b -> expand('#' . b . ':p') =~? '\.circleci' ? 'circleci' : ''},
-\   'command': 'circleci config validate - < %s',
+\   'command': 'circleci --skip-update-check config validate - < %s',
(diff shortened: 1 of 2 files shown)


hsanson (commented):
Thanks, Can we assume that flag is available independently of circleci version? Just wondering if this change would break ALE for people stuck with older versions of circleci.

goetzc:
That is a good question.

Looks like the check was [introduced](https://github.com/CircleCI-Public/circleci-cli/commit/002bec6978ab1da3985edbb808522131e7502e8b) in version [v0.1.4005](https://github.com/CircleCI-Public/circleci-cli/releases/tag/v0.1.4005) on 13 Nov 2018, like three years ago.

And later they [introduced](https://github.com/CircleCI-Public/circleci-cli/commit/bb7188b801a97a8e8c9ad21d8225f3faa1903e1d) another option to disable the command using an environment variable, `CIRCLECI_CLI_SKIP_UPDATE_CHECK` in version [v0.1.8764](https://github.com/CircleCI-Public/circleci-cli/releases/tag/v0.1.8764) on 16 Jul 2020, last year.

hsanson (approved):
Thanks, change looks good.

PR #3912, opened 2021-09-20
linters: Add Linux Kernel checkpatch for C and DTS

scosu:
checkpatch.pl is a script that is supposed to be executed before your
patch is sent to the mailinglist. It checks your code and complains
about several things.

This patch uses that script for both C and DTS files.

Signed-off-by: Markus Schneider-Pargmann <msp@baylibre.com>


hsanson (changes requested):
Please add tests (see :h  ale-development-linter-tests), documentation for the new configuration options, and update the supported-tools.md and ale-supported-languages-and-tools.txt with the new tool.

PR #3918, opened 2021-09-25
Fix tidy fixer

yoann9344:
Correct the command by adding a '-' at the end
Add a global variable to separate tidy linter and fixer options
Clean the way how the options are collected
Update documentation for tidy

### DIFF
--- a/autoload/ale/fixers/tidy.vim
+++ b/autoload/ale/fixers/tidy.vim
@@ -2,6 +2,7 @@
 " Description: Fixing HTML files with tidy.
 
 call ale#Set('html_tidy_executable', 'tidy')
+call ale#Set('html_tidy_fixer_options', '')
 call ale#Set('html_tidy_use_global', get(g:, 'ale_use_global_executables', 0))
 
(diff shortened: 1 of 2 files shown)


hsanson (changes requested):
Looks good but the tests also need to be updated and a new test for the html_tidy_fixer_options needs to be added.

yoann9344:
I'm not familiar at all with vimscript tests. I saw it uses Vador. So it could take a while for me to be up-to-date !

hsanson:
@yoann9344 in the `test_tidy_fixer_callback.vader` test modify the command output to match the modifications you did:

```
Execute(The tidy callback should return the correct default command):
  AssertEqual
  \ {
  \   'command': ale#Escape('../test-files/tidy/tidy')
  \     . ' -q --tidy-mark no --show-errors 0 --show-warnings 0'
  \ },
  \ ale#fixers#tidy#Fix(bufnr(''))
```

Adding the `-` at the end of the command should let the test pass:

```
  \     . ' -q --tidy-mark no --show-errors 0 --show-warnings 0 -'
```

PR #3922, opened 2021-09-30
Automatically find black config file

yoann9344:
Black uses only one config file (_pyproject.toml_). So I add an option to search automatically for this config file through `ale#nearest` (for the current project).

This option is `g:ale_python_black_use_config`, by default it does nothing more than before this patch.

I created 3 modes :
- Use only the user options (`g:ale_python_black_options`) ; option 0 default
- Use both config and options ; option 1
- Use only the config file ; option 2

And if the config is not found it will always use  the user options, even with the option 2.

### DIFF
--- a/autoload/ale/fixers/black.vim
+++ b/autoload/ale/fixers/black.vim
@@ -4,6 +4,7 @@
 call ale#Set('python_black_executable', 'black')
 call ale#Set('python_black_use_global', get(g:, 'ale_use_global_executables', 0))
 call ale#Set('python_black_options', '')
+call ale#Set('python_black_use_config', 0)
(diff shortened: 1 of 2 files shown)


hsanson (commented):
I think the current logic can be confusing. As other fixers/linters the config file should be used if found and the options used if defined. 

If python_black_options is defined use it
If config file is found use it
If both python_black_options is defined and config file is found then use both

We can keep the python_black_use_config as a boolean that is true by default. Users can set it to false (0) if they want to stop the fixer from using the local configuration file even if present.

yoann9344:
Ok I will change this

PR #3924, opened 2021-10-01
feat: add deno lsp for javascript

creativenull:
This PR adds `deno` as an LSP server for javascript projects and improve the docs for deno. Will update the PR if tests start failing, or if there is anything else missing 😅


hsanson (approved):
Looks good to me. Thanks.

PR #3925, opened 2021-10-02
Implement textDocument/didSave includeText optional argument

### LINKED ISSUE #3923: Incomplete `textDocument/didSave` implementation
## Information

**VIM version**
NVIM v0.6.0-dev+141-gaa07d06bf
Build type: RelWithDebInfo

Operating System: Arch Linux

## What went wrong

Connecting ALE as an LSP client to [Godot Game Engine](https://github.com/godotengine/godot) would occasionally crash the engine. After the bug was fixed on the engine, it was discovered that the `textDocument/didSave` implementation within ALE is missing several fields in the message.

## Reproducing the bug

ALE has an incomplete LSP implementation of `textDocument/didSave`.

See:
https://github.com/dense-analysis/ale/blob/c9c89a1853c01f3451edd7105814d123023ddbd8/autoload/ale/lsp/message.vim#L80-L86

Which is missing the file contents, and possibly other fields according to the official [LSP specification](https://microsoft.github.io/language-server-protocol/specification#textDocument_didSave).

See https://github.com/godotengine/godot/pull/53261#issuecomment-931423393 for more information.

### :ALEInfo
N/A

Rubonnek:
Not sure if I have to add tests to this PR.

Closes #3923.


autoload/ale/lsp.vim
```diff
@@ -260,8 +261,12 @@ function! s:UpdateCapabilities(conn, capabilities) abort
...
     endif
 
-    if type(get(a:capabilities, 'workspaceSymbolProvider')) is v:t_dict
-        let a:conn.capabilities.symbol_search = 1
+    if has_key(a:capabilities, 'textDocumentSync')
+        if has_key(a:capabilities.textDocumentSync, 'save')
+            if get(a:capabilities.textDocumentSync.save, 'includeText') is v:true
+                let a:conn.capabilities.includeText = 1
+            endif
+        endif
```
Rubonnek, line 269: It seems this chunk could use a `try` block like this:

```vimscript
    try
        if get(a:capabilities.textDocumentSync.save, 'includeText') is v:true
            let a:conn.capabilities.includeText = 1
        endif
    endtry
```

But I'm not really familiar with vimscript.

autoload/ale/lsp.vim
```diff
@@ -260,8 +261,12 @@ function! s:UpdateCapabilities(conn, capabilities) abort
         let a:conn.capabilities.symbol_search = 1
     endif
 
-    if type(get(a:capabilities, 'workspaceSymbolProvider')) is v:t_dict
-        let a:conn.capabilities.symbol_search = 1
```
vimpostor, line 264: Why are you deleting this? Is the `get(a:capabilities, 'workspaceSymbolProvider') is v:true` check above really enough?

autoload/ale/lsp.vim
vimpostor: Why do you want a `try` here?

If you are worried about `'includeText'` not existing, you could add a third argument to `get()` that returns a default value in that case. However, since `get()` by default returns `0` if the key is not found, and `0` is **not** `v:true`, this isn't even necessary here.

Rubonnek: I'm worried about `textDocumentSync` and `save` not existing due to a buggy LSP server implementation, which is why I thought the `try` block could help.

autoload/ale/lsp.vim
Rubonnek: Didn't mean to delete this. Change reverted.

autoload/ale/lsp.vim
vimpostor: > I'm worried about `textDocumentSync` and `save` not existing

This isn't possible, since you check `has_key(a:capabilities, 'textDocumentSync')` beforehand.

Rubonnek: I think I didn't explain myself correctly. What I meant to say initially was that it seemed that this chunk could use a `try` block and that those `has_key` calls could be dropped for readability. The `try` block is just for the possibility of `textDocumentSync` and `save` not existing.

Essentially turning this:
```
    if has_key(a:capabilities, 'textDocumentSync')
        if has_key(a:capabilities.textDocumentSync, 'save')
            if get(a:capabilities.textDocumentSync.save, 'includeText') is v:true
                let a:conn.capabilities.includeText = 1
            endif
        endif
    endif
```
Into this:
```
    try
        if get(a:capabilities.textDocumentSync.save, 'includeText') is v:true
            let a:conn.capabilities.includeText = 1 endif endtry
        endif
    endtry
```

hsanson: Not expert in LSP but from what I read in the documentation the "textDocumentSync.save" property can be either a boolean or SaveOptions type. So it must be tested first as a boolean and then as a dict. The same treatment can be seen on the other capabilities where they are tested for v:true first and then for v:t_dict.

Something like this should do and is more readable than using try/catch to handle program logic:

```
if type(get(a:capabilities, 'textDocumentSync')) is v:t_dict
  let l:save = get(a:capabilities.textDocumentSync, 'save')

  if l:save is v:true
    let a:conn.capabilities.includeText = 1
  endif

  if type(l:save) is v:t_dict
    let a:conn.capabilities.includeText = get(l:save, 'includeText', v:false)
  endif
endif
```

Another advantes of this implementation is that in future other attributes in the textDocumentSync dictionary can be easily added.

Note I have not tested this.

Rubonnek: Thanks for the suggestion. I've implemented this as well.

Rubonnek:
I don't get why AppVeyor is failling.

hsanson:
@Rubonnek you changed the signature of the `ale#lsp#message#DidSave()` function so this test [test_lsp_client_messages.vader#L80](https://github.com/dense-analysis/ale/blob/master/test/lsp/test_lsp_client_messages.vader#L80) needs to be updated to use the new signature.

Said that it may be difficult to pass the conn_id to that function in the test. What I recommend is that you modify the function signature to look like this:

```
function ale#lsp#message#DidSave(buffer, includeText) {
  ....
}
```

Where the `includeText` argument is a boolean that is true or false depending of if the capability is provided by the server. You can determine the capability before calling the DidSave() function and pass the boolean instead of the conn_id.

This signature is more consistent with all other message functions already implemented by ALE and allows for easier testing of the function with and without the capability:

```
Execute(ale#lsp#message#DidSave() should return correct message):
  AssertEqual
  \ [
  \   1,
  \   'textDocument/didSave',
  \   {
  \     'textDocument': {
  \         'uri': ale#path#ToURI(g:dir . '/foo/bar.ts'),
  \     },
  \   }
  \ ],
  \ ale#lsp#message#DidSave(bufnr(''), v:false)

Execute(ale#lsp#message#DidSave() should return correct message with includeText capability):
  AssertEqual
  \ [
  \   1,
  \   'textDocument/didSave',
  \   {
  \     'textDocument': {
  \         'uri': ale#path#ToURI(g:dir . '/foo/bar.ts'),
  \         'version': version = ale#lsp#message#GetNextVersionID(),
  \     },
  \     'text': join(getbufline(a:buffer, 1, '$'), '\n') . '\n'
  \   }
  \ ],
  \ ale#lsp#message#DidSave(bufnr(''), v:true)
```


Bonus Points:
  - Use single quotes instead of double quotes.
  - You may want to create a function that does this ` join(getbufline(a:buffer, 1, '$'), '\n') . '\n'` and use that function in the implementation and tests. A good candidate would be `ale#util#GetBufferContents(buffer)` in utils.vim to complement the `ale#util#SetBufferContents()` function.

Rubonnek:
@hsanson Thanks a lot for the feedback!

There's still one issue I'm not sure how to fix, and I'm not sure how my change is related to it.

For some strange reason textDocument/didSave URI changed in the `test_did_save_event.vader` test.

```
[[1, 'textDocument/didChange', {'contentChanges': [{'text': '<contents>
'}], 'textDocument': {'uri': 'file:///C:/testplugin/test/completion/dummy.txt', 'version': 17}}], [1, 'textDocument/didSave', {'textDocument': {'uri': '/testplugin/test/linter/dummy.txt'}}]]

[[1, 'textDocument/didChange', {'contentChanges': [{'text': '<contents>
'}], 'textDocument': {'uri': 'file:///C:/testplugin/test/completion/dummy.txt', 'version': 17}}], [1, 'textDocument/didSave', {'textDocument': {'uri': 'file:///C:/testplugin/test/completion/dummy.txt'}}]]
```

Issue #3911, opened 2021-09-18
Add support for erb-lint


evenreven:
**Name:** erb-lint
**URL:** https://github.com/Shopify/erb-lint

erb-lint is a Rubocop-based linter and fixer developed at Shopify for embedded ruby (.html.erb files). It used to be hard to integrate with editors, but this [recently changed](https://github.com/Shopify/erb-lint/releases/tag/v0.1.0), and there is now a [VS Code integration](https://marketplace.visualstudio.com/items?itemName=manuelpuyol.erb-linter) that works quite well (both as linter and autofixer). I've only used Ruumba (the other Rubocop-based erb linter) for a little while, but from what I can tell the main advantage of erb-lint is that it comes with real HTML parsing and linting as well, i.e. it doesn't just lint the Ruby snippets. (Technically this is done by a different gem called `better-html`, but it's bundled with erb-lint and you don't have to run it separately.)  

I think erb-lint is currently the best linter for this format, and since I use (neo)vim and not VS Code, I'd love to see it supported by ALE! (Sadly I don't know enough to help with a PR.)

PR #3931, opened 2021-10-07
Add support for erblint

ghost:
Add support for [erblint](https://github.com/Shopify/erb-lint) as asked in issue #3911


hsanson:
@evenreven unfortunatelly this PR only adds erblint as a linter. Supporting it as fixer should not be difficult. Check the  ale/autoload/ale/fixers/rubocop.vim  file for a simple example.

PR #3935, opened 2021-10-10
erblint as fixer

ghost:
Add erblint as fixer #3911


evenreven:
Lovely! Good work! (I'm not a reviewer, but FWIW I did test it now by pointing vim-plug to your fork, and it works here on my end. Only tried a simple autoformat operation (single to double quotes), but it works and feels faster than the VS Code extension too. :+1:

PR #3949, opened 2021-10-20
Do not show linebreaks in virtual text warnings

vimpostor:
This replaces all line breaks with whitespace. Before this patch, these line breaks would show up as `'^@'` or `'^M'` - depending on OS, which is kinda ugly. It is better to replace them with whitespace. This also removes trailing whitespace with `trim()`.


hsanson:
I though I merged this, thanks for the notice.

PR #3960, opened 2021-10-31
Initialize colors correctly if virtual text is loaded first

vimpostor:
See commit message for more details.

Do we need a test for this?


vimpostor:
I have added a test now, that fails without this patch and passes after applying this patch.

hsanson (approved):
Sorry for the delay, thanks for the fix.

PR #3961, opened 2021-10-31
Make run-tests compatible with Podman

vimpostor:
This fixes `Error: cannot specify an image and a filter(s)` if one was using Podman instead of a normal Docker installation and of course still works with normal Docker.


hsanson (approved):
Thanks, looks good.

PR #3967, opened 2021-11-06
Add Statix for Linting

houstdav000:
Took a couple of hours fighting with regex, but I managed to implement statix check for linting, left things viable to set up fixing later, and made a few documentation cleanups.

PR #3969, opened 2021-11-08
Implement statix Linter and Fixer

houstdav000:
Write an implementation for statix to use as both a linter and a fixer, using `statix check` and `statix fix` respectively. 

I previously made a PR for this as #3967, and closed it when I realized there were breaking changes with the v0.4.0 release. This implementation works with v0.4.0, and adds fixer support as well.

Made some minor ale-nix.txt formatting fixes as well left over from my previous implementation of nixfmt and from nixpkgs-fmt.


hsanson (changes requested):
Would be great if you can add a test for the linter handler with common output generated by the linter.

houstdav000:
Alright, added a handler test in 0cea76d, and ensured the tests pass locally.

hsanson (approved):
Great, thanks for the contribution.

PR #3972, opened 2021-11-09
Add info about using C/C++ toolchains in build containers

wosrediinanatour:
Using VIM-ALE with CCLS while building in a build container may be tricky. I added short paragraph with a link to CCLS' Wiki.


wosrediinanatour:
@w0rp Could you take a look?

PR #3974, opened 2021-11-11
Improve ALEFix performance for neovim

vimpostor:
`nvim_buf_set_lines()` does not have the same problems with Treesitter, that `setbufline()` does have.

### DIFF
--- a/autoload/ale/fix.vim
+++ b/autoload/ale/fix.vim
@@ -32,7 +32,7 @@ function! ale#fix#ApplyQueuedFixes(buffer) abort
                 endif
             endif
         endif
-    catch /E21/
+    catch /E21\|E5555/
         " If we cannot modify the buffer now, try again later.
         let g:ale_fix_buffer_data[a:buffer] = l:data
 
--- a/autoload/ale/util.vim
+++ b/autoload/ale/util.vim
@@ -522,7 +522,12 @@ function! ale#util#SetBufferContents(buffer, lines) abort
 
     " Use a Vim API for setting lines in other buffers, if available.
     if l:has_bufline_api
-        call setbufline(a:buffer, 1, l:new_lines)
+        if has('nvim')
+            call nvim_buf_set_lines(a:buffer, 0, l:first_line_to_remove, 0, l:new_lines)
+        else
+            call setbufline(a:buffer, 1, l:new_lines)
+        endif
+
         call deletebufline(a:buffer, l:first_line_to_remove, '$')
     " Fall back on setting lines the old way, for the current buffer.
     else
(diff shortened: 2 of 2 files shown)


hsanson:
Using nvim_buf_set_lines() causes the signs column to dissappear momentarily resulting in a very distracting shift movement of all text in the buffer to the left and then back to the right. This does not happen with sefbufline().  Probably an issue on Nvim side?  Apart from this it seems to work fine.

vimpostor:
I can check that problem out inside the neovim code, but it will take a while.

vimpostor:
> Using nvim_buf_set_lines() causes the signs column to dissappear momentarily resulting in a very distracting shift

I found the real problem behind this. Apparently `nvim_buf_set_lines()` always removes all signs and marks from the lines it is replacing, which can easily be tested with:

```vim
nvim -Nu <(cat<<EOF
func Do(t)
	call nvim_buf_set_lines(bufnr(), 0, 0, 0, ['0', '1', '2'])
	sign define piet text=>> texthl=Search
	exe printf('sign place 2 line=2 name=piet buffer=%d', bufnr())
	redraw
	sleep 1

	" 1) Vim API: Keeps signs (which is what we want to avoid flicker)
	" call setbufline(bufnr(), 1, ['0', '1', 'hi'])
	" 2) New Neovim API: Deletes signs
	" call nvim_buf_set_text(bufnr(), 0, 0, 3, 0, ['0', '1', 'hi'])
	" 3) Alternative Neovim API: Also deletes signs
	call nvim_buf_set_lines(bufnr(), 0, 3, 0, ['0', '1', 'hi'])
endfunc
au VimEnter * call timer_start(1,  'Do')
EOF
)

```

In fact neovim recently faced a very similar problem in their own LSP formatting sync: https://github.com/neovim/neovim/issues/14307

To solve this, we can use the same fix that they [merged](https://github.com/neovim/neovim/pull/14630): Just save and restore all signs.

I have updated the PR branch with this fix, would you mind reopening this PR?

w0rp:
@vimpostor GitHub won't let me reopen this PR, but you can open a new PR and @hsanson could test it again.

vimpostor:
@w0rp I have reset the PR branch to the state f9e1c6e3 when it was closed. According to Github you should be able to reopen it now, then afterwards I can reset to ce013ec2c162dfa176475b0fa123f72e879b82b8 again.

w0rp:
@vimpostor Nice, I've done that now.

vimpostor:
> @vimpostor Nice, I've done that now.

I've pushed the new changes and rebased on top of master again.

w0rp (approved):

w0rp:
Cheers! :beers: I've merged this now, and the commit contains your detailed explanation of the problem and how to fix it.

PR #3976, opened 2021-11-12
Add sub_type to Elvis output

dmitrivereshchagin:
Elvis is Erlang style reviewer, thus all loclist objects should have `sub_type` set to `style`.


hsanson (approved):
Nice, thanks.

PR #3979, opened 2021-11-14
Find composer.json when searching for psalm

diskshima:
This PR adds a lookup for a directory with `composer.json` in it when trying to find the `psalm` executable under `vendor/bin`.
Previously it looked for the `.git/` directory but some projects may not necessarily have their `vendor/bin` directory placed at their project root.

### DIFF
--- a/test/linter/test_psalm.vader
+++ b/test/linter/test_psalm.vader
@@ -36,3 +36,8 @@ Execute(User provided options should be used):
   AssertLinter 'psalm',
   \ ale#Escape('psalm')
   \ . ' --language-server --my-user-provided-option my-value'
+
+Execute(The project path should be correct for composer.json file):
+  call ale#test#SetFilename('../test-files/php/with-composer/test.php')
+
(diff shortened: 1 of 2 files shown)


test/linter/test_psalm.vader
```diff
@@ -36,3 +36,8 @@ Execute(User provided options should be used):
   AssertLinter 'psalm',
   \ ale#Escape('psalm')
   \ . ' --language-server --my-user-provided-option my-value'
+
+Execute(The project path should be correct for composer.json file):
+  call ale#test#SetFilename('../test-files/php/with-composer/test.php')
+
+  AssertLSPProject ale#path#Simplify(g:dir . '/../test-files/php/with-composer')
```
diskshima, line 43: I'm using `AssertLSPProject` here but not sure if this is the best way to test this case.
I'll be more than happy to be corrected ;)

hsanson (approved):

PR #3980, opened 2021-11-14
Add cppcheck handler match on misra msg

dgeorge83616:
- [x] Read the Contributing guide linked above first.
- [x] Read the documentation that comes with ALE with `:help ale-dev`.
- [x] Have fun!
- [x] ... write tests.

The incumbent regex did not match the misra code values because the code format includes `-` and `.`. The new regex does not validate the misra code format; rather, it just accepts the chars. For example, `misra-2.` would be accepted.


hsanson (approved):
Thanks for the fix and the corresponding test.

PR #3982, opened 2021-11-15
Rewrite Alex Integration to Use stdin

### LINKED ISSUE #3971: Rewrite alex Integration to Use --stdin
Since originally having been added, it looks like alex has gained stream support via a `--stdin` flag. Rewriting the handler to use this flag would allow potentially faster linting and less disk resource usage.

I may be able to do this at some point, but I just need to have enough time.

houstdav000:
Since having been added, the `alex` tool has added support for linting
on stdin. Rewrite this integration to reduce the number of tools
requiring disk-write access.

Been sitting on this for a second; waited for `alex` to be added to nixpkgs.

Closes #3971

### DIFF
--- a/autoload/ale/handlers/alex.vim
+++ b/autoload/ale/handlers/alex.vim
@@ -11,8 +11,9 @@ endfunction
 
 function! ale#handlers#alex#CreateCommandCallback(flags) abort
     return {b -> ale#node#Executable(b, ale#handlers#alex#GetExecutable(b))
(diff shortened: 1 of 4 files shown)


hsanson (approved):
Thanks.

The --stdin flag seems to have been added on version [6.0.0](https://github.com/get-alex/alex/releases/tag/6.0.0) released in 2018 so I think is safe to assume this won't break for people using older versions of alex.

PR #3983, opened 2021-11-15
Cppcheck buffered file only

dgeorge83616:
Cppcheck can take a long time to check a large project and has --file-filter option to check only the files in the project that match the filter. The PR adds logic to check only the file in the buffer.

Cppcheck isn't designed to work on header files without being included in the c/cc/cpp file. This PR also addresses that issue by skipping `--project` option and adding `--suppress` for discovered false positive, `unusedStructMember`.


dgeorge83616:
Oops. The c test case failes... will fix.

dgeorge83616:
I'm looking for a way to get the full file path within the test code to remedy the windows platform test. Hint, anyone?

dgeorge83616:
Need to handle `.h` files because `cppcheck` stops if it cannot find the file within `compile_commands.json`.

hsanson (approved):
Looks good. Thanks.

PR #3987, opened 2021-11-19
Fix --file-filter option on cppcheck command

dgeorge83616:
This time, the tests pass and works in actual use, too. I had committed a version that passed the tests, but the test patterns didn't match reality. Sorry for not correctly re-verifying actual usage on that commit.


hsanson (approved):
Thanks for the quick fix.

PR #3981, opened 2021-11-15
Add cspell Linter

houstdav000:
Add the `cspell` linter, with the languages it supports.

Took a bit of reading to figure out some test problems, but I think I got it working pretty well.

Supported dictionaries can be found/added at <https://github.com/streetsidesoftware/cspell-dicts>

Closes #3828 and #3970

PR #3990, opened 2021-11-19
Add OCLint (https://oclint.org/) for C and C++ source files

IngoMeyer441:
This PR adds support for OCLint (https://oclint.org/).

### DIFF
--- a/ale_linters/c/oclint.vim
+++ b/ale_linters/c/oclint.vim
@@ -1,0 +1,57 @@
+" Author: Ingo Meyer <https://github.com/IngoMeyer441>
+" Description: oclint linter for c files, based on `clangtidy.vim`
+
+call ale#Set('c_oclint_executable', 'oclint')
+call ale#Set('c_oclint_options', '')
+call ale#Set('c_oclint_compileflags', '')
+call ale#Set('c_build_dir', '')
+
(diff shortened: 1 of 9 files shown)


IngoMeyer441:
Does anyone know why the tests fail, with [this error message](https://github.com/dense-analysis/ale/runs/4262273467?check_suite_focus=true#step:3:3021):
```
(1/7) [EXECUTE] (X) Vim(return):E716: Key not present in Dictionary: ale_c_build_dir
      > function ale#assert#Linter[10]..ale#linter#GetCommand[3]..<lambda>1423[1]..ale#c#RunMakeCommand[4]..ale_linters#c#oclint#GetCommand[2]..ale#c#GetBuildDirectory[1]..ale#Var, line 4
```
I based the tests on the `clangtidy` tests and I don't see a difference which could cause this error message...

hsanson (changes requested):
You could have avoided duplicating code for c and cpp by defining common functions inside handlers/oclint.vim and then call them from c/oclint.vim and cpp/oclint.vim.  Check the cspell PR for a recent example of how to define a linter tool that works with multiple languages:

 - https://github.com/dense-analysis/ale/pull/3981/files
 
 Also please add tests for the handler. The PR above can serve as example on this too.

Issue #3991, opened 2021-11-19
Volar LSP support for Vue


creativenull:
**Name:** volar
**URL:** https://github.com/johnsoncodehk/volar

Volar is a new development tooling support for Vue, especially for Vue 3. For ALE this will comes in the form of a LSP server. Will be making a PR shortly.

PR #3992, opened 2021-11-19
Add volar support for vue

creativenull:
Includes the support of volar as an LSP server, addresses issue #3991


test/test-files/volar/src/App.vue
```diff
@@ -0,0 +1,8 @@
+<script setup>
```
hsanson, line 1: The contents of this file are not needed to run the tests so this can be empty.  No need to to make ALE bigger than what it is already.

test/test-files/volar/src/main.js
```diff
@@ -0,0 +1,4 @@
+import { createApp } from 'vue'
```
hsanson, line 1: This file does not seem needed for the tests so can be removed.

test/test-files/volar/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
```
hsanson, line 1: This file also does not seem necessary.

test/test-files/volar/vite.config.js
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'vite'
```
hsanson, line 1: This file also does not seem necessary.

hsanson (changes requested):

hsanson (approved):
Thanks for the changes.

KKKiraSong:
Thanks for adding volar support for vue3.
But I have some diffculties.
My vue config for ale is:
image
Then I install volar: npm i -g @volar/vue-language-server. And the version is 1.0.24.
But when I open a vue file, it doesn't work.
image
I define 'activeTabs', but I use 'activeTab' in template tag. ALE cannot work.
I have installed ts in my project.
I cannot find the reason. Plz help me.
Thank u.

hsanson:
This kind of support questions are better submitted to the discussions... commenting in a closed issue makes it difficult to spot for people that may be able to help:  https://github.com/dense-analysis/ale/discussions

Also please include input of ALEInfo as part of the question. It may provide details on what may be the problem.

KKKiraSong:
> This kind of support questions are better submitted to the discussions... commenting in a closed issue makes it difficult to spot for people that may be able to help: https://github.com/dense-analysis/ale/discussions
> 
> Also please include input of ALEInfo as part of the question. It may provide details on what may be the problem.

ok. Thanks

PR #3995, opened 2021-11-29
Pass full path to pydocstyle

brendanator:
`pydocstyle` was working only for files at the root of the project as it was using `%s:t`. This PR passes the full path using `%s`


hsanson (approved):
Thanks, looks good.

PR #3998, opened 2021-12-01
WIP - add neovim 0.6 to run-tests

hsanson:
For some unknown reason Vader tests fail when run in Neovim 0.6 but work fine on all previous versions. Specifically if fails when using wildcards in the list of tests.

For context the following command fails to run on 0.6 but works fine when using 0.5 and 0.4.4:

```
docker run -a stderr -e VADER_OUTPUT_FILE=/dev/stderr --rm -v "$PWD:/testplugin" -v "$PWD/test:/home" \
   -w /testplugin denseanalysis/ale:4162227ba361156550caebb5608c91a1 /vim-build/bin/neovim-v0.6.0 \
  -u test/vimrc "+Vader! test/*/*.vader"
```

While this commands runs fine on all versions, including 0.6.  Check the wildcards in the list of vader tests to run:

```
docker run -a stderr -e VADER_OUTPUT_FILE=/dev/stderr --rm -v "$PWD:/testplugin" -v "$PWD/test:/home" \
  -w /testplugin denseanalysis/ale:4162227ba361156550caebb5608c91a1 /vim-build/bin/neovim-v0.6.0 \
  -u test/vimrc "+Vader! test/*.vader"
```

Maybe somthing changed in 0.6 that breaks Vader? This needs more investigation.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,9 +28,8 @@ jobs:
         vim-version:
           - '--vim-80-only'
           - '--vim-82-only'
-          - '--neovim-02-only'
-          - '--neovim-04-only'
           - '--neovim-05-only'
+          - '--neovim-06-only'
           - '--linters-only'
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
         vim-version:
           - '--vim-80-only'
           - '--vim-82-only'
-          - '--neovim-05-only'
+          - '--neovim-02-only'
           - '--neovim-06-only'
           - '--linters-only'
     steps:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
-FROM testbed/vim:23
+FROM testbed/vim:20
 
 RUN install_vim -tag v8.0.0027 -build \
                 -tag v8.2.2401 -build \
-                -tag neovim:v0.5.0 -build \
+                -tag neovim:v0.2.0 -build \
                 -tag neovim:v0.6.1 -build
 
 ENV PACKAGES="\
@@ -24,4 +24,4 @@ RUN git clone https://github.com/junegunn/vader.vim vader && \
 
 ARG GIT_VERSION
 LABEL Version=${GIT_VERSION}
-LABEL Name=w0rp/ale
+LABEL Name=denseanalysis/ale
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,9 @@
-FROM testbed/vim:20
+FROM testbed/vim:23
 
 RUN install_vim -tag v8.0.0027 -build \
                 -tag v8.2.2401 -build \
-                -tag neovim:v0.2.0 -build \
-                -tag neovim:v0.4.4 -build \
                 -tag neovim:v0.5.0 -build \
-                -tag neovim:v0.6.0 -build
+                -tag neovim:v0.6.1 -build
 
 ENV PACKAGES="\
     bash \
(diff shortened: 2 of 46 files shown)


hsanson:
After some more investigation seems the problem is related with the number of tests passed as argument. When running the tests command with these tests if works fine on all versions, including 0.6:

 -  "+Vader! test/*.vader"
 -  "+Vader! test/util/*.vader"
 -  "+Vader! test/linter/a*.vader"    # Note only tests starting with "a" are used

When adding a list with many tests (e.g. 200+) then we get a "read error" message:

 -  "+Vader! test/linter/*.vader"

If we add the "--headless" argument to neovim it seem to be able to load and run all tests but then we get many other errors that need to be investigated one by one.

First it seems that now end_col and end_lnum are always returned causing some handler tests to fail:

```
  Starting Vader: /testplugin/test/lsp/test_engine_lsp_response_handling.vader
    (1/7) [EXECUTE] tsserver syntax error responses should be handled correctly
    (1/7) [EXECUTE] (X) [{'lnum': 1, 'bufnr': 3, 'end_lnum': 2, 'pattern': '', 'valid': 1, 'vcol': 0, 'nr': 1005, 'type': 'E', 'end_col': 14, 'col': 14, 'text': '1005: '','' expected.'}] should be equal to [{'lnum': 1, 'bufnr': 3, 'col': 14, 'pattern': '', 'valid': 1, 'vcol': 0, 'nr': 1005, 'type': 'E', 'text': '1005: '','' expected.'}]
    (2/7) [EXECUTE] tsserver semantic error responses should be handled correctly
    (2/7) [EXECUTE] (X) [{'lnum': 1, 'bufnr': 3, 'end_lnum': 2, 'pattern': '', 'valid': 1, 'vcol': 0, 'nr': 1005, 'type': 'E', 'end_col': 14, 'col': 14, 'text': '1005: Some semantic error'}] should be equal to [{'lnum': 1, 'bufnr': 3, 'col': 14, 'pattern': '', 'valid': 1, 'vcol': 0, 'nr': 1005, 'type': 'E', 'text': '1005: Some semantic error'}]
    (3/7) [EXECUTE] tsserver errors should mark tsserver no longer active
    (4/7) [EXECUTE] LSP diagnostics responses should be handled correctly
    (4/7) [EXECUTE] (X) [{'lnum': 1, 'bufnr': 3, 'end_lnum': 1, 'pattern': '', 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'W', 'end_col': 0, 'col': 1, 'text': 'Missing JRE 1-8'}] should be equal to [{'lnum': 1, 'bufnr': 3, 'col': 1, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'W', 'pattern': '', 'text': 'Missing JRE 1-8'}]
    (5/7) [EXECUTE] LSP diagnostics responses on project root should not populate loclist
    (6/7) [EXECUTE] LSP errors should mark linters no longer active
    (7/7) [EXECUTE] LSP errors should be logged in the history
  Success/Total: 4/7

``` 

Second some tests that change options or executables via options also fail. The interesting part is that the test does pass if run in isolation but fails when the whole test suite is executed:

```
    (10/10) [EXECUTE] (X) ['/testplugin/test/test-files/vim/node_modules/.bin/vim-language-server', '''/testplugin/test/test-files/vim/node_modules/.bin/vim-language-server'' --stdio'] should be equal to ['foobar', '''foobar'' --stdio']
  Success/Total: 9/10
 ```

hsanson:
@w0rp some help would be greatly appreciated.  I am running out of ideas.

w0rp:
I'll have a look. While we're at it I think it would be good to eliminate some versions in-between, so we test a low and high version, and skip the versions in the middle.

w0rp:
We should keep the `0.2.x` tests for now, and we can test just `0.2.x` and 0.6.x`

w0rp:
I'm in the process of fixing this. I had to switch back to the old vim-testbed version because for some reason the one one doesn't build on my machine, which is using stable KDE Neon. There are some weird little issues where the way some linter tests are torn down break following tests that I'm fixing.

PR #4006, opened 2021-12-07
add support for checkov for linting terraform files

thyme-87:
Display rule violations from [checkov](https://github.com/bridgecrewio/checkov) via `ale`. Motivated by [checkov2vim](https://github.com/mhristof/checkov2vim) which I did not manage to set up correctly on my machine.

I provided the most basic tests (derived from the tests for the `terraform` linter and `ansible_lint`) and I do hope that this is sufficient.

Please let me know if I missed anything.

### DIFF
--- a/ale_linters/terraform/checkov.vim
+++ b/ale_linters/terraform/checkov.vim
@@ -1,0 +1,52 @@
+" Author: Thyme-87 <thyme-87@posteo.me>
+" Description: use checkov for providing warnings via ale
+
+call ale#Set('terraform_checkov_executable', 'checkov')
+call ale#Set('terraform_checkov_options', '-o json --quiet')
+
+function! ale_linters#terraform#checkov#GetExecutable(buffer) abort
+    return ale#Var(a:buffer, 'terraform_checkov_executable')
+endfunction
+
+function! ale_linters#terraform#checkov#GetCommand(buffer) abort
+    return '%e ' . '-f %t ' . ale#Var(a:buffer, 'terraform_checkov_options')
+endfunction
+
+function! ale_linters#terraform#checkov#Handle(buffer, lines) abort
+    let l:output = []
+
+    try
+        let l:results = get(ale#util#FuzzyJSONDecode(a:lines, {}), 'results', [])
+    catch
+        return []
+    endtry
+
+    "if no problems are found by checkov, it will provide some statistics
+    "as the JSON format is different then, l:results will then be an empty list
+    "handle this by returning an empty list in this case
+    if empty(l:results)
+        return []
+    endif
+
+    for l:violation in l:results['failed_checks']
+        call add(l:output, {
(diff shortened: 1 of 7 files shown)


ale_linters/terraform/checkov.vim
```diff
@@ -0,0 +1,52 @@
...
+endfunction
+
+function! ale_linters#terraform#checkov#GetCommand(buffer) abort
+    return '%e ' . '-f %t ' . ale#Var(a:buffer, 'terraform_checkov_options')
+endfunction
+
+function! ale_linters#terraform#checkov#Handle(buffer, lines) abort
+    let l:output = []
+
+    try
```
hsanson, line 18: Something like this would be simpler:

```
let l:results = get(get(ale#util#FuzzyJSONDecode(a:lines, {}), 'results', []), 'failed_checks', [])

for l:violation in l:results
    call add(......
```

Notes:
  - Try/Catch block is unnecessary.  JSONDecode already has one inside and get() never throws.
  - The get() is called twice to retrive the 'results' and then the 'failed_checks' in sequence. In both cases an empty array is returned if the key is not present.
  - No need to check l:results for empty() since the loop latter will bail out if the array is empty.

ale_linters/terraform/checkov.vim
```diff
@@ -0,0 +1,52 @@
+" Author: Thyme-87 <thyme-87@posteo.me>
+" Description: use checkov for providing warnings via ale
+
+call ale#Set('terraform_checkov_executable', 'checkov')
+call ale#Set('terraform_checkov_options', '-o json --quiet')
```
hsanson, line 5: Handler depends on the output being json. If a user defines this option in his/her config and do not specify these options it will break the linter.  The "-o json --quiet" options should be added to the command in the GetCommand() method always.

hsanson:
Also add an entry to ale-terraform.txt with the new linter and configuration options.

ale_linters/terraform/checkov.vim
thyme-87: Okay, thanks for the explanation - is there no way, to use one get() and let it look into the json? I really don't like how it looks, to put a get() inside a get() - but I see the advantage and am okay with this.

thyme-87:
@hsanson Thank you for the feedback and the explanations! I updated the branch accordingly and hope, it is now ready for being merged.

ale_linters/terraform/checkov.vim
hsanson: Unfortunately this is the best we can do with vimscript.

hsanson:
@thyme-87 sorry, forgot to mention you have to add checkov entry to the list of tools in ale.txt file. Make sure to add it in the same order as in the ale-terraform.txt file.

hsanson (approved):
Thanks for the contribution.

PR #4010, opened 2021-12-12
Fix crash for Reek 6.0.4

phongnh:
When `let g:ale_ruby_reek_show_wiki_link = 1`, Reek linter is crashed
 because `wiki_link` attribute does not exist in result.

 It seems that `wiki_link` is now replaced with `documentation_link` in
 recent version of Reek

Please help to review this PR, thanks.


hsanson (approved):
Nice fix. Solves the issue while keeping backward compatibility.

PR #4012, opened 2021-12-13
ALEFileRename command added.

daliusd:
This command renames file and uses tsserver `getEditsForFileRename` to
fix import paths in Typescript files.

### DIFF
--- a/autoload/ale/filerename.vim
+++ b/autoload/ale/filerename.vim
@@ -97,7 +97,7 @@ function! ale#filerename#Execute() abort
     let l:lsp_linters = []
 
     for l:linter in ale#linter#Get(&filetype)
-        if l:linter.lsp == 'tsserver'
+        if l:linter.lsp is# 'tsserver'
             call add(l:lsp_linters, l:linter)
         endif
     endfor
--- a/autoload/ale/filerename.vim
+++ b/autoload/ale/filerename.vim
@@ -97,13 +97,13 @@ function! ale#filerename#Execute() abort
     let l:lsp_linters = []
 
     for l:linter in ale#linter#Get(&filetype)
-        if !empty(l:linter.lsp)
+        if l:linter.lsp == 'tsserver'
             call add(l:lsp_linters, l:linter)
         endif
     endfor
 
     if empty(l:lsp_linters)
-        call s:message('No active LSPs')
+        call s:message('No active tsserver LSPs')
 
         return
     endif
@@ -111,6 +111,12 @@ function! ale#filerename#Execute() abort
     let l:buffer = bufnr('')
     let l:old_name = expand('#' . l:buffer . ':p')
     let l:new_name = ale#util#Input('New file name: ', l:old_name, 'file')
+
+    if l:old_name is# l:new_name
+        call s:message('New file name matches old file name')
+
+        return
+    endif
 
     if empty(l:new_name)
         call s:message('New name cannot be empty!')
--- a/autoload/ale/filerename.vim
+++ b/autoload/ale/filerename.vim
@@ -1,0 +1,127 @@
+" Author: Dalius Dobravolskas <dalius.dobravolskas@gmail.com>
+" Description: Rename file support for tsserver
+
+let s:filerename_map = {}
+
+" Used to get the rename map in tests.
+function! ale#filerename#GetMap() abort
+    return deepcopy(s:filerename_map)
+endfunction
+
+" Used to set the rename map in tests.
+function! ale#filerename#SetMap(map) abort
+    let s:filerename_map = a:map
+endfunction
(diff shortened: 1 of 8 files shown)


daliusd:
@hsanson please review, I'm not sure who else could check this.

daliusd:
I think it is good now. Tested with pyright - no problem. It seems all the problems were related to ale#util#Input change.

autoload/ale/filerename.vim
```diff
@@ -0,0 +1,127 @@
...
+    endif
+
+    silent! noautocmd execute 'saveas ' . l:new_name
+    call delete(l:old_name)
+endfunction
+
+function! s:OnReady(options, linter, lsp_details) abort
+    let l:id = a:lsp_details.connection_id
+
+    if !ale#lsp#HasCapability(l:id, 'filerename')
```
hsanson, line 69: @daliusd add the filerename capability to the default capabilities in the ale#lsp#Register() method. If not I get an exception thrown when checking for that capability here when and LSP that is not tsserver is active.

```
New file name: /home/ryujin/Projects/HelloWorld/python/hello-plain/hello.pyError detected while processing function ale#filerename#Execute[26]..<SNR>202_ExecuteFileRename[4]..ale#lsp_linter#StartLSP[26]..<SNR>190_StartIfExecutable[15]..<SNR>190_StartWithCommand[13]..<SNR>190_StartLSP[49]..ale#lsp#On
Init[8]..<lambda>24[1]..ale#lsp_linter#OnInit[45]..<SNR>202_OnReady[3]..ale#lsp#HasCapability:
line    8:
E605: Exception not caught: Invalid capability filerename
Error detected while processing function ale#filerename#Execute[26]..<SNR>202_ExecuteFileRename[4]..ale#lsp_linter#StartLSP[26]..<SNR>190_StartIfExecutable[15]..<SNR>190_StartWithCommand[13]..<SNR>190_StartLSP[49]..ale#lsp#OnInit[8]..<lambda>24[1]..ale#lsp_linter#OnInit[45]..<SNR>202_OnReady:
line    3:
E171: Missing :endif
Error detected while processing function ale#filerename#Execute[26]..<SNR>202_ExecuteFileRename[4]..ale#lsp_linter#StartLSP[26]..<SNR>190_StartIfExecutable[15]..<SNR>190_StartWithCommand[13]..<SNR>190_StartLSP[49]..ale#lsp#OnInit:
line    8:
E171: Missing :endif
Error detected while processing function ale#filerename#Execute:
line   26:
E170: Missing :endfor
```

autoload/ale/filerename.vim
```diff
@@ -0,0 +1,127 @@
...
+
+    let l:Callback = function('s:OnReady', [a:options])
+    call ale#lsp_linter#StartLSP(l:buffer, a:linter, l:Callback)
+endfunction
+
+function! ale#filerename#Execute() abort
+    let l:lsp_linters = []
+
+    for l:linter in ale#linter#Get(&filetype)
+        if !empty(l:linter.lsp)
```
hsanson, line 100: Since this functionality is only available with tsserver maybe is better to explicitly check for it:

```
if l:linter.lsp == 'tsserver'
    call add(l:lsp_linters, l:linter)
endif
```

If not the ALEFileRename command seems to execute fine when other LSPs are active but they do nothing.  Also I am unsure what happens if there are more than one LSP active that does support filerename, if that is even possible..

autoload/ale/filerename.vim
```diff
@@ -0,0 +1,127 @@
...
+            call add(l:lsp_linters, l:linter)
+        endif
+    endfor
+
+    if empty(l:lsp_linters)
+        call s:message('No active LSPs')
```
hsanson, line 106: If we add the explicit check for tsserver above then this message would make more sense if explictily mentions tsserver:

```
call s:message('No active tsserver LSPs')
```

hsanson (changes requested):

autoload/ale/filerename.vim
daliusd: Added. I have not noticed this error before, tested with `pyright` - must be good now.

autoload/ale/filerename.vim
daliusd: I think as of now that's not possible. I think there is no other LSP that can handle file renames except `tsserver`.

daliusd: Explicit check added.

autoload/ale/filerename.vim
daliusd: Good idea.  Done.

hsanson (approved):
Thanks for the contribution and changes.

daliusd:
🍻

PR #4021, opened 2021-12-22
ignore error, when SetUpDefaultColumnWithoutErrorsHighlight fails

manfred-kaiser:
Ignores an errors, when using other plugins, which also uses the sign column.


hsanson (commented):
Catching and swallowing all exceptions is not a proper fix. Can you elaborate exactly what problem you experience to see if we can figure out a better way for ALE to not conflict with other plugins?

PR #4026, opened 2022-01-04
Fix for arch

volucris1:
Property "cmakelint" not working with arch, use "cmake-lint" instead


hsanson (commented):
From what I see the official binary name is cmakelint (see https://github.com/cmake-lint/cmake-lint) so this change will break things for anyone not using arch.

PR #4027, opened 2022-01-04
Fix 4025 - Allow to configure texlab LSP settings

### LINKED ISSUE #4025: TexLab Linter (maybe in general): Allow modifying `lsp_config`
TexLab is configured solely by use of LSP configuration, set via `lsp_config`. ALE does not provide any way of setting that (and does not currently set it at all). Allowing to set that might also be useful for other linters.

hsanson:
This PR adds a new configuration `g:ale_tex_texlab_config` that allows to pass initialization settings to texlab LSP.

Closes #4025

### DIFF
--- a/doc/ale-tex.txt
+++ b/doc/ale-tex.txt
@@ -77,7 +77,7 @@ g:ale_tex_texlab_options                             *g:ale_tex_texlab_options*
 
 g:ale_tex_texlab_config                               *g:ale_tex_texlab_config*
                                                       *b:ale_tex_texlab_config*
-  Type: |String|
+  Type: |Dictionary|
   Default: `{}`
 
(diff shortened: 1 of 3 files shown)


doc/ale-tex.txt
```diff
@@ -73,8 +73,23 @@ g:ale_tex_texlab_options                             *g:ale_tex_texlab_options*
   Type: |String|
   Default: `''`
 
-  This variable can be changed to modify flags given to texlab.
+  This variable can be changed to modify flags given to texlab command.
 
+g:ale_tex_texlab_config                               *g:ale_tex_texlab_config*
+                                                      *b:ale_tex_texlab_config*
+  Type: |String|
```
cpiber, line 80: Type should be `|Dictionary|`

PR #4030, opened 2022-01-10
Add eclipselsp jdt:// support for textDocument/definition

yoshi1123:
This adds jdt:// support in eclipselsp. I referred to #[https://github.com/neoclide/coc.nvim/issues/112](https://github.com/neoclide/coc.nvim/issues/112).

The tests are at `test/test_path_uri_jdt.vader`.

Let me know if there are changes needed.

### DIFF
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -94,6 +94,10 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
                     break
                 endif
 
+                if !empty(getbufvar(bufnr(''), 'ale_lsp_root', ''))
+                    break
+                endif
+
                 " Display java file from jar library as part of current project.
                 set filetype=java
 
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -109,7 +109,7 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
                 \   bufnr(''),
                 \   'eclipselsp',
                 \   [0, 'java/classFileContents', {'uri': l:filename}],
-                \       function('s:handle_class_file_contents'))
+                \       function('s:HandleClassFileContents'))
 
                 while s:cb_status isnot# 'complete'
                     sleep 10m
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -109,7 +109,7 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
                 call ale#lsp_linter#SendRequest(
                 \   bufnr(''),
                 \   'eclipselsp',
-                \   [0, 'java/classFileContents', {'uri': l:filename}],
+                \   [0, 'java/classFileContents', {'uri': ale#path#ToURI(l:filename)}],
                 \       function('s:OpenJdtLink', [a:conn_id, l:filename, l:line, l:column, l:options]))
             else
                 call ale#util#Open(l:filename, l:line, l:column, l:options)
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -116,8 +116,11 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
                 endwhile
 
                 let l:contents = s:result['result']
+
                 call setline(1, split(l:contents, '\n'))
-                call ale#util#Open(l:filename, l:line, l:column, l:options)
+                call cursor(l:line, l:column)
+                normal! zz
+
                 setlocal buftype=nofile nomod readonly
             endif
 
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -80,12 +80,12 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
 
             call ale#definition#UpdateTagStack()
 
-            let l:filename = ale#path#FromURI(l:uri)
-            let l:uri_handler = ale#util#GetURIHandler(l:filename)
-            if l:uri_handler isnot# v:null
-                call l:uri_handler.OpenURILink(l:filename, l:line, l:column, l:options, a:conn_id)
-            else
+            let l:uri_handler = ale#uri#GetURIHandler(l:uri)
+            if l:uri_handler is# v:null
+                let l:filename = ale#path#FromFileURI(l:uri)
                 call ale#util#Open(l:filename, l:line, l:column, l:options)
+            else
+                call l:uri_handler.OpenURILink(l:uri, l:line, l:column, l:options, a:conn_id)
             endif
 
             break
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -68,50 +68,36 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
         for l:item in l:result
             if has_key(l:item, 'targetUri')
                 " LocationLink items use targetUri
-                let l:filename = ale#path#FromURI(l:item.targetUri)
+                let l:uri = l:item.targetUri
                 let l:line = l:item.targetRange.start.line + 1
                 let l:column = l:item.targetRange.start.character + 1
             else
                 " LocationLink items use uri
-                let l:filename = ale#path#FromURI(l:item.uri)
+                let l:uri = l:item.uri
                 let l:line = l:item.range.start.line + 1
                 let l:column = l:item.range.start.character + 1
             endif
 
             call ale#definition#UpdateTagStack()
 
-            if l:filename[:5] is? 'jdt://'
-                let l:found = v:false
-                for l:linter in ale#linter#Get('java')
-                    if l:linter.name is# 'eclipselsp'
-                        let l:found = v:true
+            let l:root = a:conn_id[stridx(a:conn_id, ':')+1:]
+            let found_uri_handler = v:false
+            for l:linter in ale#linter#Get(&filetype)
+                if !empty(l:linter.lsp)
+                    if exists('l:linter.uri_handlers') && !empty(l:linter.uri_handlers)
+                        for l:scheme in keys(l:linter.uri_handlers)
+                            if l:uri =~# '^'.l:scheme.'://'
+                                call l:linter.uri_handlers[scheme](l:root, l:line, l:column, l:options, l:uri)
+                                let l:found_uri_handler = v:true
+                                break
+                            endif
+                        endfor
                     endif
-                endfor
-                if ! l:found
-                    break
                 endif
+            endfor
 
-                function! s:OpenJdtLink(conn_id, uri, line, column, options, result) abort
-                    " Display java file from jar library as part of current project.
-                    let l:contents = a:result['result']
-                    call ale#util#Open(a:uri, a:line, a:column, a:options)
-                    if !empty(getbufvar(bufnr(''), 'ale_lsp_root', ''))
-                        return
-                    endif
-                    let b:ale_lsp_root = a:conn_id[stridx(a:conn_id, ':')+1:]
-                    set filetype=java
-                    call setline(1, split(l:contents, '\n'))
-                    call cursor(a:line, a:column)
-                    normal! zz
-                    setlocal buftype=nofile nomod readonly
-                endfunction
-
-                call ale#lsp_linter#SendRequest(
-                \   bufnr(''),
-                \   'eclipselsp',
-                \   [0, 'java/classFileContents', {'uri': ale#path#ToURI(l:filename)}],
-                \       function('s:OpenJdtLink', [a:conn_id, l:filename, l:line, l:column, l:options]))
-            else
+            if !found_uri_handler
+                let l:filename = ale#path#FromURI(l:uri)
                 call ale#util#Open(l:filename, l:line, l:column, l:options)
             endif
 
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -80,24 +80,11 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
 
             call ale#definition#UpdateTagStack()
 
-            let l:root = a:conn_id[stridx(a:conn_id, ':')+1:]
-            let found_uri_handler = v:false
-            for l:linter in ale#linter#Get(&filetype)
-                if !empty(l:linter.lsp)
-                    if exists('l:linter.uri_handlers') && !empty(l:linter.uri_handlers)
-                        for l:scheme in keys(l:linter.uri_handlers)
-                            if l:uri =~# '^'.l:scheme.'://'
-                                call l:linter.uri_handlers[scheme](l:root, l:line, l:column, l:options, l:uri)
-                                let l:found_uri_handler = v:true
-                                break
-                            endif
-                        endfor
-                    endif
-                endif
-            endfor
-
-            if !found_uri_handler
-                let l:filename = ale#path#FromURI(l:uri)
+            let l:filename = ale#path#FromURI(l:uri)
+            let l:uri_handler = ale#util#GetURIHandler(l:filename)
+            if l:uri_handler isnot# v:null
+                call l:uri_handler.OpenURILink(l:filename, l:line, l:column, l:options, a:conn_id)
+            else
                 call ale#util#Open(l:filename, l:line, l:column, l:options)
             endif
 
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -80,6 +80,47 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
 
             call ale#definition#UpdateTagStack()
             call ale#util#Open(l:filename, l:line, l:column, l:options)
+
+            if l:filename[:5] is? 'jdt://'
+                let l:found = v:false
+
+                for l:linter in ale#linter#Get('java')
+                    if l:linter.name is# 'eclipselsp'
+                        let l:found = v:true
+                    endif
+                endfor
+
+                if ! l:found
+                    break
+                endif
+
+                " Display java file from jar library as part of current project.
+                set filetype=java
+
+                let b:ale_lsp_root = a:conn_id[stridx(a:conn_id, ':')+1:]
+                let s:cb_status = 'pending'
+
+                function! s:HandleClassFileContents(result) abort
+                    let s:cb_status = 'complete'
+                    let s:result = a:result
+                endfunction
+
+                call ale#lsp_linter#SendRequest(
+                \   bufnr(''),
+                \   'eclipselsp',
+                \   [0, 'java/classFileContents', {'uri': l:filename}],
+                \       function('s:handle_class_file_contents'))
+
+                while s:cb_status isnot# 'complete'
+                    sleep 10m
+                endwhile
+
+                let l:contents = s:result['result']
+                call setline(1, split(l:contents, '\n'))
+                call ale#util#Open(l:filename, l:line, l:column, l:options)
+                setlocal buftype=nofile nomod readonly
+            endif
+
             break
         endfor
     endif
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -79,53 +79,40 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
             endif
 
             call ale#definition#UpdateTagStack()
-            call ale#util#Open(l:filename, l:line, l:column, l:options)
 
             if l:filename[:5] is? 'jdt://'
                 let l:found = v:false
-
                 for l:linter in ale#linter#Get('java')
                     if l:linter.name is# 'eclipselsp'
                         let l:found = v:true
                     endif
                 endfor
-
                 if ! l:found
                     break
                 endif
 
-                if !empty(getbufvar(bufnr(''), 'ale_lsp_root', ''))
-                    break
-                endif
-
-                " Display java file from jar library as part of current project.
-                set filetype=java
-
-                let b:ale_lsp_root = a:conn_id[stridx(a:conn_id, ':')+1:]
-                let s:cb_status = 'pending'
-
-                function! s:HandleClassFileContents(result) abort
-                    let s:cb_status = 'complete'
-                    let s:result = a:result
+                function! s:OpenJdtLink(conn_id, uri, line, column, options, result) abort
+                    " Display java file from jar library as part of current project.
+                    let l:contents = a:result['result']
+                    call ale#util#Open(a:uri, a:line, a:column, a:options)
+                    if !empty(getbufvar(bufnr(''), 'ale_lsp_root', ''))
+                        return
+                    endif
+                    let b:ale_lsp_root = a:conn_id[stridx(a:conn_id, ':')+1:]
+                    set filetype=java
+                    call setline(1, split(l:contents, '\n'))
+                    call cursor(a:line, a:column)
+                    normal! zz
+                    setlocal buftype=nofile nomod readonly
                 endfunction
 
                 call ale#lsp_linter#SendRequest(
                 \   bufnr(''),
                 \   'eclipselsp',
                 \   [0, 'java/classFileContents', {'uri': l:filename}],
-                \       function('s:HandleClassFileContents'))
-
-                while s:cb_status isnot# 'complete'
-                    sleep 10m
-                endwhile
-
-                let l:contents = s:result['result']
-
-                call setline(1, split(l:contents, '\n'))
-                call cursor(l:line, l:column)
-                normal! zz
-
-                setlocal buftype=nofile nomod readonly
+                \       function('s:OpenJdtLink', [a:conn_id, l:filename, l:line, l:column, l:options]))
+            else
+                call ale#util#Open(l:filename, l:line, l:column, l:options)
             endif
 
             break
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -81,6 +81,7 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
             call ale#definition#UpdateTagStack()
 
             let l:uri_handler = ale#uri#GetURIHandler(l:uri)
+
             if l:uri_handler is# v:null
                 let l:filename = ale#path#FromFileURI(l:uri)
                 call ale#util#Open(l:filename, l:line, l:column, l:options)
--- a/autoload/ale/lsp_linter.vim
+++ b/autoload/ale/lsp_linter.vim
@@ -8,12 +8,24 @@ if !has_key(s:, 'lsp_linter_map')
     let s:lsp_linter_map = {}
 endif
 
-function! ale#lsp_linter#GetLinterMap() abort
+" A Dictionary to track one-shot handlers for custom LSP requests
+let s:custom_handlers_map = get(s:, 'custom_handlers_map', {})
+
+" Clear LSP linter data for the linting engine.
+function! ale#lsp_linter#ClearLSPData() abort
+    let s:lsp_linter_map = {}
+    let s:custom_handlers_map = {}
+endfunction
+
+" Only for internal use.
+function! ale#lsp_linter#GetLSPLinterMap() abort
     return s:lsp_linter_map
 endfunction
 
-" A Dictionary to track one-shot handlers for custom LSP requests
-let s:custom_handlers_map = get(s:, 'custom_handlers_map', {})
+" Just for tests.
+function! ale#lsp_linter#SetLSPLinterMap(replacement_map) abort
+    let s:lsp_linter_map = a:replacement_map
+endfunction
 
 " Check if diagnostics for a particular linter should be ignored.
 function! s:ShouldIgnore(buffer, linter_name) abort
@@ -479,17 +491,6 @@ endfunction
 
 function! ale#lsp_linter#CheckWithLSP(buffer, linter) abort
     return ale#lsp_linter#StartLSP(a:buffer, a:linter, function('s:CheckWithLSP'))
-endfunction
-
-" Clear LSP linter data for the linting engine.
-function! ale#lsp_linter#ClearLSPData() abort
-    let s:lsp_linter_map = {}
-    let s:custom_handlers_map = {}
-endfunction
-
-" Just for tests.
-function! ale#lsp_linter#SetLSPLinterMap(replacement_map) abort
-    let s:lsp_linter_map = a:replacement_map
 endfunction
 
 function! s:HandleLSPResponseToCustomRequests(conn_id, response) abort
--- a/autoload/ale/lsp_linter.vim
+++ b/autoload/ale/lsp_linter.vim
@@ -8,7 +8,7 @@ if !has_key(s:, 'lsp_linter_map')
     let s:lsp_linter_map = {}
 endif
 
-function! ale#lsp_linter#GetLinterMap()
+function! ale#lsp_linter#GetLinterMap() abort
     return s:lsp_linter_map
 endfunction
 
--- a/autoload/ale/lsp_linter.vim
+++ b/autoload/ale/lsp_linter.vim
@@ -49,7 +49,7 @@ endfunction
 
 function! s:HandleLSPDiagnostics(conn_id, response) abort
     let l:linter_name = s:lsp_linter_map[a:conn_id]
-    let l:filename = ale#path#FromURI(a:response.params.uri)
+    let l:filename = ale#util#ToResource(a:response.params.uri)
     let l:escaped_name = escape(
     \   fnameescape(l:filename),
     \   has('win32') ? '^' : '^,}]'
--- a/autoload/ale/lsp_linter.vim
+++ b/autoload/ale/lsp_linter.vim
@@ -8,6 +8,10 @@ if !has_key(s:, 'lsp_linter_map')
     let s:lsp_linter_map = {}
 endif
 
+function! ale#lsp_linter#GetLinterMap()
+    return s:lsp_linter_map
+endfunction
+
 " A Dictionary to track one-shot handlers for custom LSP requests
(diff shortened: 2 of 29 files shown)


yoshi1123:
Works in vim. In neovim, have to fix.

autoload/ale/definition.vim
```diff
@@ -80,6 +80,47 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
...
+                    let s:result = a:result
+                endfunction
+
+                call ale#lsp_linter#SendRequest(
+                \   bufnr(''),
+                \   'eclipselsp',
+                \   [0, 'java/classFileContents', {'uri': l:filename}],
+                \       function('s:handle_class_file_contents'))
+
+                while s:cb_status isnot# 'complete'
```
hsanson, line 118: I am not convinced blocking until the LSP response arrives is something we want to do.  Better have a callback like OnReady() that can take care of completing the task (e.g. opening a readonly buffer with the source on the desired position) after the call is completed.

hsanson (changes requested):

hsanson:
I like this feature, is useful for my work but i think as it is implemented in the PR cannot be accepted.  Blocking is against ALE filosophy and most of the code can be refactored to allow inclusion of other schemese like "kls://" used by KotlinLanguageServer.

Also I am unable to get this working on my own setup. The classFileContents call results in the language server throwing an exception about "jdt" provider not being installed. Not sure if I need to add some dependency or lib to get this working.

```
 12.879508 RECV on 6(out): '{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":1,"message":"Jan 12, 2022, 12:19:27 AM Error getting content via disassemblerContentProvider\nProvider \"jdt\" not installed\njava.nio.file.FileSystemNotFoundException: Provider \"jdt\" not installed\n\tat java.base/java.nio.file.Path.of(Path.java:212)\n\tat java.base/java.nio.file.Paths.get(Paths.java:98)\n\tat org.eclipse.jdt.ls.core.internal.DisassemblerContentProvider.getContent(DisassemblerContentProvider.java:41)\n\tat org.eclipse.jdt.ls.core.internal.managers.ContentProviderManager.getContent(ContentProviderManager.java:110)\n\tat org.eclipse.jdt.ls.core.internal.managers.ContentProviderManager.getContent(ContentProviderManager.java:81)\n\tat org.eclipse.jdt.ls.core.internal.handlers.JDTLanguageServer.lambda$26(JDTLanguageServer.java:815)\n\tat org.eclipse.jdt.ls.core.internal.BaseJDTLanguageServer.lambda$0(BaseJDTLanguageServer.java:75)\n\tat java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:642)\n\tat java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:479)\n\tat java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:295)\n\tat java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1016)\n\tat java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1665)\n\tat java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1598)\n\tat java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)\n"}}Content-Length: 36
```

yoshi1123:
Yes, refactoring is a good idea.
About the error, do you have the latest eclipse jdtls?

yoshi1123:
Okay. I made a commit to make it not block. I will work on a commit for refactoring, and see how that turns out.

hsanson:
> Yes, refactoring is a good idea. About the error, do you have the latest eclipse jdtls?

Yes, I am using latest version compiled from source.

hsanson:
> Works in vim. In neovim, needs fix.

What fix is required for neovim? I may be able to help here since I mainly use neovim.

yoshi1123:
It seems that it works with neovim after I made it non-blocking. The problem was that, even though in neovim there was a textDocument/definition response, the callback wasn't being called. Without the blocking loop, it seems to be working in neovim.

As for the "provider jdt not installed" error. What filesystem are you using?

yoshi1123:
Okay try now. If it works, forget those questions.

yoshi1123:
Okay, I refactored it.

Now, Define() has a new key 'uri_handlers'.  It looks like this:
```
call ale#linter#Define('java', {
\   'name': 'eclipselsp',
\   'lsp': 'stdio',
\   'executable': function('ale_linters#java#eclipselsp#Executable'),
\   'command': function('ale_linters#java#eclipselsp#RunWithVersionCheck'),
\   'language': 'java',
\   'project_root': function('ale#java#FindProjectRoot'),
\   'uri_handlers': {
\       'jdt': function('ale_linters#java#eclipselsp#HandleJdt')
\   },
\   'initialization_options': {
\     'extendedClientCapabilities': {
\       'classFileContentsSupport': v:true
\     }
\   }
\})
```
So ALL code for uri links like jdt:// are now in their own file (e.g., in `ale_linters/java/eclipselsp.vim`). The only code added outside of a specific language server source file is a section in the function `ale#definition#HandleLSPResponse` that looks for uri handlers in the linters for the file that requested the textDocument/definition, and calls those callbacks.  This way, any LSP server can add its own code in ale_linters/LANG/LS.vim to handle special URIs, without having to touch any other source files.

yoshi1123:
I changed it to be really simple. It just uses an BufNewFile and BufReadPre autocmd. I figured that in a textDocument/definition to a jdt:// link, you don't need to use the same eclipselsp process as the project doing the request. The line and column work also. So all that is added is a simple autocmd in ale_linters/java/eclipselsp.vim.

yoshi1123:
Okay, its fixed.

The PR is now non-blocking and it allows for other URI schemes.

---

There are three URI handlers in a LSP linter. For example,
```
call ale#linter#Define('java', {
\   'name': 'eclipselsp',
\   'lsp': 'stdio',
\   'executable': function('ale_linters#java#eclipselsp#Executable'),
\   'command': function('ale_linters#java#eclipselsp#RunWithVersionCheck'),
\   'language': 'java',
\   'project_root': function('ale#java#FindProjectRoot'),
\   'uri_handlers': {
\       'jdt': {
\           'OpenURILink': function('ale_linters#java#eclipselsp#OpenJDTLink'),
\           'PathFromURI': function('ale_linters#java#eclipselsp#JDTToPath'),
\           'PathToURI': function('ale_linters#java#eclipselsp#PathToJDT')
\       }
\   },
\   'initialization_options': {
\     'extendedClientCapabilities': {
\       'classFileContentsSupport': v:true
\     }
\   }
\})
```

A handler 'PathFromURI' is called in function `ale#path#FromURI`, returning a value. A handler 'PathToURI' in `ale#path#ToURI`. 'OpenURILink' is called in function `ale#definition#HandleLSPResponse` if found, otherwise the regular `ale#util#Open` is called.

hsanson:
@yoshi1123 thanks for the contribution... currently little busy with work now so will take some time before I can review and test this.

hsanson:
After some careful review I could not get my head over this PR.  I think is better to set a more human readable filename to the buffers (e.g. joda-time-2.9.0.jar/org/joda/time/LocalDate.java) instead of just escaping them when converting from URI to Path and viceversa.  The idea of the autocommand seems good but unless I am missing something is not needed. Modifying the references.vim is possible to load jdt files the same way definition.vim does.  Tiying the URI and JDT open methods to a specific LSP (e.g. eclipselsp) does not sound sensible. Other LSP implementations could also implement the jdt:// scheme for example.

I have implemented a different PR #4030 that does what this one does in a simpler (to my view) way. If this PR works for you I would prefer to merge that one.  This PR supports opening JDT URIs in readonly/unmodifiable buffers with some easy to understand filenames when using ALEGoToDefinition and ALEFindReferences.

yoshi1123:
Alright, I'll try to explain what I am trying to do.

I am trying to make `autoload/ale/definition` have no knowledge of any special URI, and have the user or the linter specify how to read a special URI like jdt:// ones. I did this by adding a new option to linter definitions called 'uri_handlers':

```
call ale#linter#Define('java', {
\   'name': 'eclipselsp',
\   'lsp': 'stdio',
\   'executable': function('ale_linters#java#eclipselsp#Executable'),
\   'command': function('ale_linters#java#eclipselsp#RunWithVersionCheck'),
\   'language': 'java',
\   'project_root': function('ale#java#FindProjectRoot'),
\   'uri_handlers': {
\       'jdt': {
\           'OpenURILink': function('ale_linters#java#eclipselsp#OpenJDTLink'),
\           'PathFromURI': function('ale_linters#java#eclipselsp#JDTToPath'),
\           'PathToURI': function('ale_linters#java#eclipselsp#PathToJDT')
\       }
\   },
\   'initialization_options': {
\     'extendedClientCapabilities': {
\       'classFileContentsSupport': v:true
\     }
\   }
\})
```
The linter, or the user, can then specify any 'uri_handlers' value they want for a specific linter. Thus all handling of a certain URI, such as JDT, would be specified in the linter's Define(). One could put these handler functions for JDT URIs in `autoload/ale/util.vim`, to make them more public so any linter can simply just add this to the Define()'s definition:

```
\       'jdt': {
\           'OpenURILink': function('ale#util#OpenJDTLink'),
\           'PathFromURI': function('ale#util#JDTToPath'),
\           'PathToURI': function('ale#util#PathToJDT')
\       }
```
to allow for this implementation of handling jdt:// URIs. Or they could create their own implementation and change the handler functions (e.g., `ale#util#ADifferentOpenJDTLinkFunction`).

With the 'jdt' handlers I made in the Define(), there are two ways to open a jdt:// URI:

1) Use AleGoToDefinition, which will retrieve the java file content in a jar _before_ creating the new buffer for the contents. This is done with the function `ale_linters#java#eclipselsp#OpenJDTLink`. "Open" as in open a _new_ buffer.
2) Manually type in a jdt:// link, (e.g., `:e jdt://contents/file.jar?...`, or with `gf` in a hover preview), which upon opening or reading a filename that starts with 'jdt://', will set the buffer content to the content specified by the jdt:// URI. This is done with the function `ale_linters#java#eclipselsp#ReadJDTLink`. "Read" as in read content into _current_ buffer.

hsanson:
@yoshi1123 thanks for the explanation. I understand better why the autocomand is useful (allows opening JDT files in many other places like on hover, quickfix, etc.).

My comment regarding tying the handlers to a LSP is still an issue:

> Tiying the URI and JDT open methods to a specific LSP (e.g. eclipselsp) does not sound sensible.

For example instead of defining these in the LSP via Define() we can have a separate global function like `s:default_ uri_handlers` inside path.vim:

```
let s:default_uri_handlers = {
\       'jdt': {
\           'OpenURILink': function('ale#util#OpenJDTLink'),
\           'PathFromURI': function('ale#util#JDTToPath'),
\           'PathToURI': function('ale#util#PathToJDT')
\       } 
```

Then we can iterate this dictionary to see if a handler exists for the URI schema and use it instead of looking for the active LSP etc.  We can then remove all check and lookups for the eclipselsp.   If other LSP (e.g. java-language-server) starts returning JDT URIs then this would hopefully work for those too.


Also the tests are failing the the PR is marked as draft.  If you solve the tests I can merge this even if you preffer not to do the changes I mentione above. The functionality as it is is good and very useful.

yoshi1123:
I guess I had the URI handlers as part of the LSP linter in case two LSP linters handle a URI scheme differently. I guess that is a not a realistic scenario. I don't think there is reason to anticipate two jdt:// URI schemes, or two kls:// URI schemes. Do you agree?

hsanson:
@yoshi1123 I see your thinking. But I think the changes of two different jdt:// or kls:// URI schemes are slim.  In such case I would expect the server to provide a different Scheme.

On the other hand being able to define scheme handlers independently of the LSP is similar to how other editros (e.g. VSCode)  does.

yoshi1123:
Alright, take a look.

I put the autocmd in `autoload/ale/events.vim`.

`s:uri_handlers` dict is in `autoload/ale/util.vim` and accessed with function `ale#util#GetURIHandler`.

yoshi1123:
Though, JDT URI methods are still tied to a specific LSP. The lsp requests are specifically for "eclipselsp", as the request 'java/classFileContents' is a eclipselsp extension, and not in the Language Server Protocol:

```
    call ale#lsp_linter#SendRequest(
                \   bufnr(''),
                \   'eclipselsp',
                \   [0, 'java/classFileContents', {
                \       'uri': ale#path#ToURI(l:filename)
                \   }],
                \   function('s:ReadClassFileContents', [l:filename]))
```

hsanson (approved):
Thanks, works beautifully...   could do with more tests but currently I have no idea how to write tests for this so will consider for latter.

yoshi1123:
I fixed all the vint warnings. Didn't mean to dismiss your review.

autoload/ale/util.vim
```diff
@@ -543,3 +543,142 @@ endfunction
...
+function! ale#util#GetURIHandler(uri) abort
+    for l:scheme in keys(s:uri_handlers)
+        if a:uri =~# '^'.l:scheme.'://'
+            return s:uri_handlers[scheme]
+        endif
+    endfor
+    return v:null
+endfunction
+
+function! ale#util#JDTToPath(uri) abort
```
w0rp, line 564: Put all of the JDT logic in a separate file like `path/jdt.vim` and then we'll only load that code when needed. This will save on time needed to start up ALE for people who don't write Java code.

autoload/ale/lsp_linter.vim
```diff
@@ -8,6 +8,10 @@ if !has_key(s:, 'lsp_linter_map')
     let s:lsp_linter_map = {}
 endif
 
+function! ale#lsp_linter#GetLinterMap() abort
+    return s:lsp_linter_map
+endfunction
```
w0rp, line 13: Reorganise this file like so, renaming this new function to `GetLSPLinterMap`:

```vim
" Associates LSP connection IDs with linter names.
if !has_key(s:, 'lsp_linter_map')
    let s:lsp_linter_map = {}
endif

" A Dictionary to track one-shot handlers for custom LSP requests
let s:custom_handlers_map = get(s:, 'custom_handlers_map', {})

" Clear LSP linter data for the linting engine.
function! ale#lsp_linter#ClearLSPData() abort
    let s:lsp_linter_map = {}
    let s:custom_handlers_map = {}
endfunction

" Only for internal use.
function! ale#lsp_linter#GetLSPLinterMap() abort
    return s:lsp_linter_map
endfunction

" Just for tests.
function! ale#lsp_linter#SetLSPLinterMap(replacement_map) abort
    let s:lsp_linter_map = a:replacement_map
endfunction
```

This will make the code a bit easier to read. The two existing functions are around 400 lines below this one.

autoload/ale/path.vim
```diff
@@ -232,6 +238,12 @@ function! ale#path#ToURI(path) abort
 endfunction
 
 function! ale#path#FromURI(uri) abort
+    let l:uri_handler = ale#util#GetURIHandler(a:uri)
```
w0rp, line 241: Instead of checking for URI handlers in the function to convert a URI to a path you could check for a URI handler before turning a URI into a filename. This function is only supposed to convert `file://` into a path.

w0rp (changes requested):
Looks pretty good. I left some comments on how to re-organise the code to avoid mixing behaviour into existing functions that should be simpler and how to avoid loading Vim code for people who won't be using it.

autoload/ale/util.vim
```diff
@@ -543,3 +543,142 @@ endfunction
...
+
+let s:uri_handlers = {
+\   'jdt': {
+\       'OpenURILink': function('ale#util#OpenJDTLink'),
+\       'PathFromURI': function('ale#util#JDTToPath'),
+\       'PathToURI': function('ale#util#PathToJDT')
+\   }
+\}
+
+function! ale#util#GetURIHandler(uri) abort
```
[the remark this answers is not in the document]
yoshi1123: See commit f397c2e.

autoload/ale/path.vim
```diff
@@ -219,6 +219,12 @@ endfunction
 " relatives paths will not be prefixed with the protocol.
 " For Windows paths, the `:` in C:\ etc. will not be percent-encoded.
 function! ale#path#ToURI(path) abort
+    let l:uri_handler = ale#util#GetURIHandler(a:path)
```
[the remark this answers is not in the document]
yoshi1123: See commit f397c2e.

autoload/ale/util.vim
yoshi1123: See commit f397c2e.

autoload/ale/path.vim
yoshi1123: See commit f397c2e.

This is something I was looking to address.

So there are two things in this commit.

1)
I realized that JDTToURI and URIToJDT are not needed. I thought that a vim buffer name had to conform to the OS's filename character set. I do not think this is the case. Thus there is no conversions needed to and from the buffer name.  textDocument/definition can retrieve a jdt:// URI, then it can be opened as a buffer (e.g., `:e jdt://...`). Then this jdt:// buffer name can be used directly as the URI for another textDocument/definition.

2)
To fix conflation of "path" and "URI", I renamed all calls to `ale#path#FromURI(uri)` to `ale#util#ToResource(uri)`, which is a wrapper like this:
```
function! ale#util#ToResource(uri) abort
    let l:uri_handler = ale#uri#GetURIHandler(a:uri)
    if l:uri_handler is# v:null
        " resource is a filesystem path
        let l:resource = ale#path#FromFileURI(a:uri)
    else
        " resource is a URI
        let l:resource = a:uri
    endif
    return l:resource
endfunction
```
The same is done with `ale#path#ToURI(path)` (wrapped in `ale#util#ToURI(resource)`). (Also `ale#util#ToURI` is renamed to `ale#util#ToFileURI`, and similarly, `ale#util#FromURI` is renamed to `ale#util#FromFileURI`. I renamed it temporarily but thought it might be more clear. I can change it back if you like.)

Basically, since now jdt:// URIs are used along with filesystem paths, the "path" is no longer inaccurate and so I generalized it to "resource". A resource can either be (1) a path or (2) a URI.

For example, this function,
```
function! ale#lsp#message#Definition(buffer, line, column) abort
    return [0, 'textDocument/definition', {
    \   'textDocument': {
    \       'uri': ale#path#ToURI(expand('#' . a:buffer . ':p')),
    \   },
    \   'position': {'line': a:line - 1, 'character': a:column - 1},
    \}]
endfunction
```
now looks like this:
```
function! ale#lsp#message#Definition(buffer, line, column) abort
    return [0, 'textDocument/definition', {
    \   'textDocument': {
    \       'uri': ale#util#ToURI(expand('#' . a:buffer . ':p')),
    \   },
    \   'position': {'line': a:line - 1, 'character': a:column - 1},
    \}]
endfunction
```
Its now `ale#util#ToURI(resource)` (the wrapper) instead of `ale#path#ToURI(path)`.

I also reorganized code. I put the `GetURLHandler` function in the file `autoload/ale/url.vim`. I also put jdt.vim in the directory `autoload/ale/url/`.

Let me know what you think of this possible solution.

hsanson: I think this generalization is useful as it would allow to handle other schemas without too much hassle.  I tested this on my machine and works very well.

Just one detail with the auto commands. There seems to be a problem with the load order of the files. In a particular case the autocommand defined in events.vim throws an error sayin  ale#util#ReadJDTLink() in unknown function.

To replicate the error:

 - Open a Java file.
 - Run :ALEFindReferences on a class name.
 - Select any of the jdt:// URIs in the references list.
 - See error happen.

On the other hand if we first call :ALEGoToDefinition this works and loads the jdt.vim file. After this calling ALEFindReferences works as expected sin the ale#util#ReadJDTLink() function is now defined.

yoshi1123: Thanks hsanson. 75e041d should fix that.

autoload/ale/definition.vim
```diff
@@ -68,18 +68,26 @@ function! ale#definition#HandleLSPResponse(conn_id, response) abort
...
+                let l:uri = l:item.uri
                 let l:line = l:item.range.start.line + 1
                 let l:column = l:item.range.start.character + 1
             endif
 
             call ale#definition#UpdateTagStack()
-            call ale#util#Open(l:filename, l:line, l:column, l:options)
+
+            let l:uri_handler = ale#uri#GetURIHandler(l:uri)
+            if l:uri_handler is# v:null
```
w0rp, line 84:
```suggestion
            let l:uri_handler = ale#uri#GetURIHandler(l:uri)

            if l:uri_handler is# v:null
```

The custom linter will complain about this. and doing this will fix it.

autoload/ale/uri.vim
```diff
@@ -25,3 +25,18 @@ function! ale#uri#Decode(value) abort
...
+\   }
+\}
+
+function! ale#uri#GetURIHandler(uri) abort
+    for l:scheme in keys(s:uri_handlers)
+        if a:uri =~# '^'.l:scheme.'://'
+            return s:uri_handlers[scheme]
+        endif
+    endfor
+    return v:null
```
w0rp, line 41:
```suggestion
    endfor

    return v:null
```

The custom linter will insist on this.

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,101 @@
...
+" Description: Functions for working with jdt:// URIs.
+
+function! s:OpenJDTLink(root, uri, line, column, options, result) abort
+    if has_key(a:result, 'error')
+        echoerr a:result.error.message
+        return
+    endif
+
+    let l:contents = a:result['result']
+    if type(l:contents) ==# type(v:null)
```
w0rp, line 11:
```suggestion
    let l:contents = a:result['result']

    if type(l:contents) ==# type(v:null)
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,101 @@
...
+
+" Load new buffer with jdt:// contents and jump to line and column.
+function! ale#uri#jdt#OpenJDTLink(encoded_uri, line, column, options, conn_id) abort
+    let l:found_eclipselsp = v:false
+    for l:linter in ale#linter#Get('java')
+        if l:linter.name is# 'eclipselsp'
+            let l:found_eclipselsp = v:true
+        endif
+    endfor
+    if !l:found_eclipselsp
```
w0rp, line 42:
```suggestion
    let l:found_eclipselsp = v:false

    for l:linter in ale#linter#Get('java')
        if l:linter.name is# 'eclipselsp'
            let l:found_eclipselsp = v:true
        endif
    endfor

    if !l:found_eclipselsp
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,101 @@
...
+
+    let l:root = a:conn_id[stridx(a:conn_id, ':')+1:]
+    let l:uri = a:encoded_uri
+    call ale#lsp_linter#SendRequest(
+                \   bufnr(''),
+                \   'eclipselsp',
+                \   [0, 'java/classFileContents', {
+                \       'uri': ale#util#ToURI(l:uri)
+                \   }],
+                \   function('s:OpenJDTLink', [l:root, l:uri, a:line, a:column, a:options]))
```
w0rp, line 54:
```suggestion
    call ale#lsp_linter#SendRequest(
    \   bufnr(''),
    \   'eclipselsp',
    \   [0, 'java/classFileContents', {'uri': ale#util#ToURI(l:uri)}],
    \   function('s:OpenJDTLink', [l:root, l:uri, a:line, a:column, a:options])
    \)
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,101 @@
...
+endfunction
+
+function! s:ReadClassFileContents(uri, result) abort
+    if has_key(a:result, 'error')
+        echoerr a:result.error.message
+        return
```
w0rp, line 60:
```suggestion
        echoerr a:result.error.message

        return
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,101 @@
...
+    endif
+
+    let l:linter_map = ale#lsp_linter#GetLSPLinterMap()
+    for l:conn_id in keys(l:linter_map)
+        if l:linter_map[l:conn_id] ==# 'eclipselsp'
+            let l:root = l:conn_id[stridx(l:conn_id, ':')+1:]
+        endif
+    endfor
+    if l:root is# v:null
```
w0rp, line 86:
```suggestion
    let l:linter_map = ale#lsp_linter#GetLSPLinterMap()

    for l:conn_id in keys(l:linter_map)
        if l:linter_map[l:conn_id] ==# 'eclipselsp'
            let l:root = l:conn_id[stridx(l:conn_id, ':')+1:]
        endif
    endfor

    if l:root is# v:null
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,101 @@
...
+    let b:ale_lsp_root = l:root
+    set filetype=java
+
+    call ale#lsp_linter#SendRequest(
+                \   bufnr(''),
+                \   'eclipselsp',
+                \   [0, 'java/classFileContents', {
+                \       'uri': ale#util#ToURI(l:uri)
+                \   }],
+                \   function('s:ReadClassFileContents', [l:uri]))
```
w0rp, line 100:
```suggestion
    call ale#lsp_linter#SendRequest(
    \   bufnr(''),
    \   'eclipselsp',
    \   [0, 'java/classFileContents', {'uri': ale#util#ToURI(l:uri)}],
    \   function('s:ReadClassFileContents', [l:uri])
    \)
```

autoload/ale/util.vim
```diff
@@ -543,3 +543,27 @@ endfunction
...
+function! ale#util#ToURI(resource) abort
+    let l:uri_handler = ale#uri#GetURIHandler(a:resource)
+    if l:uri_handler is# v:null
+        " resource is a filesystem path
+        let l:uri = ale#path#ToFileURI(a:resource)
+    else
+        " resource is a URI
+        let l:uri = a:resource
+    endif
+    return l:uri
```
w0rp, line 556:
```suggestion
    let l:uri_handler = ale#uri#GetURIHandler(a:resource)

    if l:uri_handler is# v:null
        " resource is a filesystem path
        let l:uri = ale#path#ToFileURI(a:resource)
    else
        " resource is a URI
        let l:uri = a:resource
    endif

    return l:uri
```

autoload/ale/util.vim
```diff
@@ -543,3 +543,27 @@ endfunction
...
+function! ale#util#ToResource(uri) abort
+    let l:uri_handler = ale#uri#GetURIHandler(a:uri)
+    if l:uri_handler is# v:null
+        " resource is a filesystem path
+        let l:resource = ale#path#FromFileURI(a:uri)
+    else
+        " resource is a URI
+        let l:resource = a:uri
+    endif
+    return l:resource
```
w0rp, line 568:
```suggestion
    let l:uri_handler = ale#uri#GetURIHandler(a:uri)

    if l:uri_handler is# v:null
        " resource is a filesystem path
        let l:resource = ale#path#FromFileURI(a:uri)
    else
        " resource is a URI
        let l:resource = a:uri
    endif

    return l:resource
```

w0rp (approved):
This is a lot better. I noted a bunch of places where the whitespace and formatting need changing to fit with the style of the codebase, and I'll commit those, fix the conflict, and get this in.

autoload/ale/references.vim
```diff
@@ -66,13 +66,13 @@ endfunction
 function! ale#references#FormatLSPResponseItem(response_item, options) abort
     if get(a:options, 'open_in') is# 'quickfix'
         return {
-        \ 'filename': ale#path#FromURI(a:response_item.uri),
+        \ 'filename': ale#util#ToResource(l:response_item.uri),
```
w0rp, line 69:
```suggestion
        \ 'filename': ale#util#ToResource(a:response_item.uri),
```

autoload/ale/references.vim
```diff
@@ -66,13 +66,13 @@ endfunction
...
         \ 'lnum': a:response_item.range.start.line + 1,
         \ 'col': a:response_item.range.start.character + 1,
         \}
     else
         return {
-        \ 'filename': ale#path#FromURI(a:response_item.uri),
+        \ 'filename': ale#util#ToResource(l:response_item.uri),
```
w0rp, line 75:
```suggestion
        \ 'filename': ale#util#ToResource(a:response_item.uri),
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,105 @@
...
+
+function! s:OpenJDTLink(root, uri, line, column, options, result) abort
+    if has_key(a:result, 'error')
+        echoerr a:result.error.message
+        return
+    endif
+
+    let l:contents = a:result['result']
+
+    if type(l:contents) ==# type(v:null)
```
w0rp, line 12:
```suggestion
    if type(l:contents) is# type(v:null)
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,105 @@
...
+function! s:ReadClassFileContents(uri, result) abort
+    if has_key(a:result, 'error')
+        echoerr a:result.error.message
+
+        return
+    endif
+
+    let l:contents = a:result['result']
+
+    if type(l:contents) ==# type(v:null)
```
w0rp, line 68:
```suggestion
    if type(l:contents) is# type(v:null)
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,105 @@
...
+" Read jdt:// contents, as part of current project, into current buffer.
+function! ale#uri#jdt#ReadJDTLink(encoded_uri) abort
+    if !empty(getbufvar(bufnr(''), 'ale_lsp_root', ''))
+        return
+    endif
+
+    let l:linter_map = ale#lsp_linter#GetLSPLinterMap()
+
+    for l:conn_id in keys(l:linter_map)
+        if l:linter_map[l:conn_id] ==# 'eclipselsp'
```
w0rp, line 86:
```suggestion
        if l:linter_map[l:conn_id] is# 'eclipselsp'
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,105 @@
+" Author: yoshi1123 <yoshi1@tutanota.com>
+" Description: Functions for working with jdt:// URIs.
+
+function! s:OpenJDTLink(root, uri, line, column, options, result) abort
+    if has_key(a:result, 'error')
+        echoerr a:result.error.message
```
w0rp, line 6:
```suggestion
        execute 'echoerr a:result.error.message'
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,105 @@
...
+        return
+    endif
+
+    let l:contents = a:result['result']
+
+    if type(l:contents) ==# type(v:null)
+        echoerr 'File content not found'
```
w0rp, line 13:
```suggestion
        execute 'echoerr ''File content not found'''
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,105 @@
...
+    \   bufnr(''),
+    \   'eclipselsp',
+    \   [0, 'java/classFileContents', {'uri': ale#util#ToURI(l:uri)}],
+    \   function('s:OpenJDTLink', [l:root, l:uri, a:line, a:column, a:options])
+    \)
+endfunction
+
+function! s:ReadClassFileContents(uri, result) abort
+    if has_key(a:result, 'error')
+        echoerr a:result.error.message
```
w0rp, line 61:
```suggestion
        execute 'echoerr a:result.error.message'
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,105 @@
...
+
+        return
+    endif
+
+    let l:contents = a:result['result']
+
+    if type(l:contents) ==# type(v:null)
+        echoerr 'File content not found'
```
w0rp, line 69:
```suggestion
        execute 'echoerr ''File content not found'''
```

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,105 @@
+" Author: yoshi1123 <yoshi1@tutanota.com>
+" Description: Functions for working with jdt:// URIs.
+
+function! s:OpenJDTLink(root, uri, line, column, options, result) abort
+    if has_key(a:result, 'error')
+        echoerr a:result.error.message
+        return
```
w0rp, line 7:
```suggestion

        return
```

w0rp (dismissed):
Some things I'll be fixing...

autoload/ale/uri/jdt.vim
```diff
@@ -0,0 +1,106 @@
...
+    if has_key(a:result, 'error')
+        execute 'echoerr a:result.error.message'
+
+        return
+    endif
+
+    let l:contents = a:result['result']
+
+    if type(l:contents) is# type(v:null)
+        execute 'echoerr ''File content not found'''
```
w0rp, line 14:
```suggestion
        execute 'echoerr ''File content not found''
```

w0rp:
A whole bunch of other tests fail that still reference the old function names. I'll just check out the branch and fix them.

w0rp:
Apologies for the GitHub spam. I'll get this in, in a moment. I do wish GitHub had a feature to let me commit all suggestions in one go so you only get one message.

w0rp (approved):

w0rp:
Cheers! :beers:

I've fixed the failing tests and merged this in now. I just removed your old test file referencing functions you wrote that no longer exist.

PR #4033, opened 2022-01-14
Allows to use quickfix for references.

### LINKED ISSUE #1759: quickfix list support for ALEFindReferences and commands for navigating through references

daliusd:
E.g. following mapping could be used to find references for item under
cursor and put result into quickfix list:

```
nnoremap <leader>af :ALEFindReferences -quickfix<CR>
```

Fixes #1759

### DIFF
--- a/autoload/ale/references.vim
+++ b/autoload/ale/references.vim
@@ -16,6 +16,23 @@ function! ale#references#ClearLSPData() abort
     let s:references_map = {}
 endfunction
 
+function! ale#references#FormatTSResponseItem(response_item, options) abort
+    if get(a:options, 'open_in') is# 'quickfix'
+        return {
+        \ 'filename': a:response_item.file,
+        \ 'lnum': a:response_item.start.line,
+        \ 'col': a:response_item.start.offset,
+        \}
+    else
+        return {
+        \ 'filename': a:response_item.file,
+        \ 'line': a:response_item.start.line,
+        \ 'column': a:response_item.start.offset,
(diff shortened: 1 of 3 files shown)


daliusd:
@hsanson please review.

hsanson (changes requested):
Works fine with my limited testing usin Solargraph LSP. Thanks.

I just would like to have the documention also updated to include the new option `:h ALEFindReferences`. Also mention that the -relative option has no effect if -quickfix is used.

Finally not sure but maybe would be good to automatically open the quickfix if it is not open?

daliusd:
> Works fine with my limited testing usin Solargraph LSP. Thanks.

Cool. I have tested with `tsserver` and `pyright`.
 
> I just would like to have the documention also updated to include the new option `:h ALEFindReferences`. Also mention that the -relative option has no effect if -quickfix is used.

Documentation updated.

> Finally not sure but maybe would be good to automatically open the quickfix if it is not open?

Vim `grep` command does not open quickfix list by default. I personally never open quickfix list. Here is suggestion how to open quickfix list window for `grep` command that must work for `ALEFindReferences` if there is a need: https://www.reddit.com/r/vim/comments/bmh977/comment/emwjm7h/?utm_source=share&utm_medium=web2x&context=3

hsanson (approved):
Great, thanks for the changes.

daliusd:
🍻

PR #4036, opened 2022-01-16
Add CMake linter cmake-lint

### LINKED ISSUE #3396: cmake-format also provides cmake-lint
cmake-format support was merged in #2244 as a fixer. ALE also has support for [cmake-lint](https://github.com/richq/cmake-lint) which is currently unmaintained. However, cmake-format also defines cmake-lint:

> The cmake-format project provides Quality Assurance (QA) tools for cmake:
> 
> cmake-annotate can generate pretty HTML from your listfiles
> cmake-format can format your listfiles nicely so that they don't look like crap.
> cmake-lint can check your listfiles for problems
>ctest-to can parse a ctest output tree and translate it into a more structured format (either JSON or XML).

It can be added as a linter, which would currently be the only maintained CMake linter.

carlsmedstad:
Add support for the CMake linter cmake-lint provided by https://github.com/cheshirekow/cmake_format.

A bit unsure about the naming due to the conflict with the other linter for CMake.

Resolves #3396.

### DIFF
--- a/ale_linters/cmake/cmake_lint.vim
+++ b/ale_linters/cmake/cmake_lint.vim
@@ -1,0 +1,41 @@
+" Author: Carl Smedstad <carl.smedstad at protonmail dot com>
+" Description: cmake-lint for cmake files
+
+let g:ale_cmake_cmake_lint_executable =
+\   get(g:, 'ale_cmake_cmake_lint_executable', 'cmake-lint')
+
+let g:ale_cmake_cmake_lint_options =
+\   get(g:, 'ale_cmake_cmake_lint_options', '')
+
+function! ale_linters#cmake#cmake_lint#Executable(buffer) abort
+    return ale#Var(a:buffer, 'cmake_cmake_lint_executable')
+endfunction
+
+function! ale_linters#cmake#cmake_lint#Command(buffer) abort
+    return ale_linters#cmake#cmake_lint#Executable(a:buffer)
+    \   . ' ' . ale#Var(a:buffer, 'cmake_cmake_lint_options') . ' %t'
+endfunction
+
+function! ale_linters#cmake#cmake_lint#Handle(buffer, lines) abort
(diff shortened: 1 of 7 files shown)


offa:
Since the current linter is unmaintained, wouldn't it make sense to replace it with the new one?

carlsmedstad:
The old linter has been forked and now lives in [cmake-lint/cmake-lint](https://github.com/cmake-lint/cmake-lint). Development is far from active, but I asked if the project was still maintained and the answer was [yes](https://github.com/cmake-lint/cmake-lint/issues/8#issuecomment-1008008173).

So I don't think it's fair to replace it really.

offa:
Ah, thank you. I only knew the original repo.

hsanson (changes requested):
Thanks, looks good. Just missing tests for the linter itself. Can you add one similar to `test/linter/test_ruby.vader`?  It tests if the correct default command is returned and if the options are properly set if configured.

carlsmedstad:
Added the test and also wrapped the executable in `ale#Escape()` as I noticed that was the case for most other linters.

hsanson (approved):
Thanks for the contribution.

hsanson:
@carlsmedstad @offa does this new linter replaces the current one available in ALE:

  - https://github.com/richq/cmake-lint

Wondering if we should consider removing it. I see from  that repo that there has not been any activity in the last 5 years. But that alone is not indiciation that the linter is not being used.

carlsmedstad:
See my comment above for my opinion about that:

> The old linter has been forked and now lives in [cmake-lint/cmake-lint](https://github.com/cmake-lint/cmake-lint). Development is far from active, but I asked if the project was still maintained and the answer was https://github.com/cmake-lint/cmake-lint/issues/8#issuecomment-1008008173.
>
> So I don't think it's fair to replace it really.

I think we should update the URL to point to the new repository at least (https://github.com/cmake-lint/cmake-lint).

PR #4038, opened 2022-01-18
Allow to configure haskell-language-server LSP config

liskin:
Somewhat shameless copypaste from 7b8e711877449478796306be9124c1fdd690b987, so yeah it does include tests. :-)


hsanson (approved):
Thanks, simple and effective.

PR #4039, opened 2022-01-18
doc: Clarify the types of process_with

idbrii:
I had trouble writing a process_with. After figuring it out, I updated the docs to what would have helped me understand the arguments.


hsanson (approved):
Thanks for clarifying the documentation.

PR #4043, opened 2022-01-21
Add oelint-adv support

offa:
Adds support for [oelint-adv](https://github.com/priv-kweihmann/oelint-adv), a Yocto / OpenEmbedded BitBake recipe linter.


hsanson (approved):
Thanks, looks good.

PR #4045, opened 2022-01-27
Fix unhandled nix versions for linter 'nix'

zoonfafer:
Previously, it would not generate any lint messages for nix 2.5.

Moreover, it would cause this error whenever the nix command is invoked, when paired with a custom `g:ale_command_wrapper`:

    Error detected while processing function <SNR>92_NeoVimCallback[29]..<lambda>27[
    1]..<SNR>90_ExitCallback[28]..ale_linters#nix#nix#Command:
    line    1:
    E684: list index out of range: 0


hsanson (approved):
Thanks for the fix.

PR #4049, opened 2022-02-02
Dispatch textDocument/didChange after rename (2)

### LINKED ISSUE #3343: Send didChange/didSave events after rename

jeremija:
Opening a new PR because #3783 was closed and we cannot reopen it.

Fixes #3343, #3642, #3781.

Video example:

Rename example

### DIFF
--- a/autoload/ale/code_action.vim
+++ b/autoload/ale/code_action.vim
@@ -16,13 +16,12 @@ endfunction
 function! ale#code_action#HandleCodeAction(code_action, options) abort
     let l:current_buffer = bufnr('')
     let l:changes = a:code_action.changes
-    let l:should_save = get(a:options, 'should_save')
 
     for l:file_code_edit in l:changes
         call ale#code_action#ApplyChanges(
         \   l:file_code_edit.fileName,
         \   l:file_code_edit.textChanges,
-        \   l:should_save,
+        \   a:options,
         \)
     endfor
 endfunction
@@ -63,28 +62,28 @@ function! s:ChangeCmp(left, right) abort
     return 0
 endfunction
 
-function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
-    let l:current_buffer = bufnr('')
+function! ale#code_action#ApplyChanges(filename, changes, options) abort
+    let l:should_save = get(a:options, 'should_save')
+    let l:conn_id = get(a:options, 'conn_id')
+
+    let l:orig_buffer = bufnr('')
+
     " The buffer is used to determine the fileformat, if available.
     let l:buffer = bufnr(a:filename)
-    let l:is_current_buffer = l:buffer > 0 && l:buffer == l:current_buffer
 
-    if l:buffer > 0
-        let l:lines = getbufline(l:buffer, 1, '$')
-
-        " Add empty line if there's trailing newline, like readfile() does.
-        if getbufvar(l:buffer, '&eol')
-            let l:lines += ['']
-        endif
-    else
-        let l:lines = readfile(a:filename, 'b')
+    if l:buffer != l:orig_buffer
+        call ale#util#Execute('edit ' . a:filename)
+        let l:buffer = bufnr('')
     endif
 
-    if l:is_current_buffer
-        let l:pos = getpos('.')[1:2]
-    else
-        let l:pos = [1, 1]
+    let l:lines = getbufline(l:buffer, 1, '$')
+
+    " Add empty line if there's trailing newline, like readfile() does.
+    if getbufvar(l:buffer, '&eol')
+        let l:lines += ['']
     endif
+
+    let l:pos = getpos('.')[1:2]
 
     " Changes have to be sorted so we apply them from bottom-to-top
     for l:code_edit in reverse(sort(copy(a:changes), function('s:ChangeCmp')))
@@ -155,46 +154,25 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
         endif
     endfor
 
-    if l:buffer > 0
-        " Make sure ale#util#{Writefile,SetBufferContents} add trailing
-        " newline if and only if it should be added.
-        if l:lines[-1] is# '' && getbufvar(l:buffer, '&eol')
-            call remove(l:lines, -1)
-        else
-            call setbufvar(l:buffer, '&eol', 0)
-        endif
-    elseif exists('+fixeol') && &fixeol && l:lines[-1] is# ''
-        " Not in buffer, ale#util#Writefile can't check &eol and always adds
-        " newline if &fixeol: remove to prevent double trailing newline.
+    " Make sure to add a trailing newline if and only if it should be added.
+    if l:lines[-1] is# '' && getbufvar(l:buffer, '&eol')
         call remove(l:lines, -1)
-    endif
-
-    if a:should_save || l:buffer < 0
-        call ale#util#Writefile(l:buffer, l:lines, a:filename)
     else
-        call ale#util#SetBufferContents(l:buffer, l:lines)
+        call setbufvar(l:buffer, '&eol', 0)
     endif
 
-    if l:is_current_buffer
-        if a:should_save
-            call ale#util#Execute(':e!')
-        endif
+    call ale#util#SetBufferContents(l:buffer, l:lines)
 
-        call setpos('.', [0, l:pos[0], l:pos[1], 0])
+    call ale#lsp#NotifyForChanges(l:conn_id, l:buffer)
+
+    if l:should_save
+        call ale#util#Execute(':w!')
     endif
 
-    if a:should_save && l:buffer > 0 && !l:is_current_buffer
-        " Set up a one-time use event that will delete itself to reload the
-        " buffer next time it's entered to view the changes made to it.
-        execute 'augroup ALECodeActionReloadGroup' . l:buffer
-            autocmd!
+    call setpos('.', [0, l:pos[0], l:pos[1], 0])
 
-            execute printf(
-            \   'autocmd BufEnter <buffer=%d>'
-            \       . ' call ale#code_action#ReloadBuffer()',
-            \   l:buffer
-            \)
-        augroup END
+    if l:orig_buffer != l:buffer
+        call ale#util#Execute('buf ' . string(l:orig_buffer))
     endif
 endfunction
 
--- a/autoload/ale/code_action.vim
+++ b/autoload/ale/code_action.vim
@@ -72,7 +72,7 @@ function! ale#code_action#ApplyChanges(filename, changes, options) abort
     let l:buffer = bufnr(a:filename)
 
     if l:buffer != l:orig_buffer
-        call ale#util#Execute('edit ' . a:filename)
+        call ale#util#Execute('silent edit ' . a:filename)
         let l:buffer = bufnr('')
     endif
 
@@ -166,13 +166,13 @@ function! ale#code_action#ApplyChanges(filename, changes, options) abort
     call ale#lsp#NotifyForChanges(l:conn_id, l:buffer)
 
     if l:should_save
-        call ale#util#Execute(':w!')
+        call ale#util#Execute('silent w!')
     endif
 
     call setpos('.', [0, l:pos[0], l:pos[1], 0])
 
     if l:orig_buffer != l:buffer && bufexists(l:orig_buffer)
-        call ale#util#Execute('buf ' . string(l:orig_buffer))
+        call ale#util#Execute('silent buf ' . string(l:orig_buffer))
     endif
 endfunction
 
--- a/autoload/ale/code_action.vim
+++ b/autoload/ale/code_action.vim
@@ -171,7 +171,7 @@ function! ale#code_action#ApplyChanges(filename, changes, options) abort
 
     call setpos('.', [0, l:pos[0], l:pos[1], 0])
 
-    if l:orig_buffer != l:buffer
+    if l:orig_buffer != l:buffer && bufexists(l:orig_buffer)
         call ale#util#Execute('buf ' . string(l:orig_buffer))
     endif
 endfunction
--- a/autoload/ale/organize_imports.vim
+++ b/autoload/ale/organize_imports.vim
@@ -17,7 +17,9 @@ function! ale#organize_imports#HandleTSServerResponse(conn_id, response) abort
     \       'description': 'Organize Imports',
     \       'changes': l:file_code_edits,
     \   },
-    \   {}
+    \   {
+    \       'conn_id': a:conn_id,
+    \   },
     \)
 endfunction
 
--- a/autoload/ale/organize_imports.vim
+++ b/autoload/ale/organize_imports.vim
@@ -1,6 +1,8 @@
 " Author: Jerko Steiner <jerko.steiner@gmail.com>
 " Description: Organize imports support for tsserver
-"
+
+let g:ale_organize_imports_save = get(g:, 'ale_organize_imports_save', 1)
+
 function! ale#organize_imports#HandleTSServerResponse(conn_id, response) abort
     if get(a:response, 'command', '') isnot# 'organizeImports'
         return
@@ -19,6 +21,7 @@ function! ale#organize_imports#HandleTSServerResponse(conn_id, response) abort
     \   },
     \   {
     \       'conn_id': a:conn_id,
+    \       'should_save': g:ale_organize_imports_save,
     \   },
     \)
 endfunction
--- a/autoload/ale/organize_imports.vim
+++ b/autoload/ale/organize_imports.vim
@@ -1,8 +1,6 @@
 " Author: Jerko Steiner <jerko.steiner@gmail.com>
 " Description: Organize imports support for tsserver
 
-let g:ale_organize_imports_save = get(g:, 'ale_organize_imports_save', 1)
-
 function! ale#organize_imports#HandleTSServerResponse(conn_id, response) abort
     if get(a:response, 'command', '') isnot# 'organizeImports'
         return
@@ -21,7 +19,7 @@ function! ale#organize_imports#HandleTSServerResponse(conn_id, response) abort
     \   },
     \   {
     \       'conn_id': a:conn_id,
-    \       'should_save': g:ale_organize_imports_save,
+    \       'should_save': !&hidden,
     \   },
     \)
 endfunction
--- a/autoload/ale/rename.vim
+++ b/autoload/ale/rename.vim
@@ -84,7 +84,7 @@ function! ale#rename#HandleTSServerResponse(conn_id, response) abort
     \       'changes': l:changes,
     \   },
     \   {
-    \       'should_save': 1,
+    \       'conn_id': a:conn_id,
     \   },
     \)
 endfunction
@@ -116,7 +116,7 @@ function! ale#rename#HandleLSPResponse(conn_id, response) abort
         \       'changes': l:changes,
         \   },
         \   {
-        \       'should_save': 1,
+        \       'conn_id': a:conn_id,
         \   },
         \)
     endif
--- a/autoload/ale/rename.vim
+++ b/autoload/ale/rename.vim
@@ -17,6 +17,7 @@ function! ale#rename#ClearLSPData() abort
     let s:rename_map = {}
 endfunction
 
+let g:ale_rename_save = get(g:, 'ale_rename_save', 1)
 let g:ale_rename_tsserver_find_in_comments = get(g:, 'ale_rename_tsserver_find_in_comments')
 let g:ale_rename_tsserver_find_in_strings = get(g:, 'ale_rename_tsserver_find_in_strings')
 
@@ -85,6 +86,7 @@ function! ale#rename#HandleTSServerResponse(conn_id, response) abort
     \   },
     \   {
     \       'conn_id': a:conn_id,
+    \       'should_save': g:ale_rename_save,
     \   },
     \)
 endfunction
@@ -117,6 +119,7 @@ function! ale#rename#HandleLSPResponse(conn_id, response) abort
         \   },
         \   {
         \       'conn_id': a:conn_id,
+        \       'should_save': g:ale_rename_save,
         \   },
         \)
     endif
--- a/autoload/ale/rename.vim
+++ b/autoload/ale/rename.vim
@@ -17,7 +17,6 @@ function! ale#rename#ClearLSPData() abort
     let s:rename_map = {}
 endfunction
 
-let g:ale_rename_save = get(g:, 'ale_rename_save', 1)
 let g:ale_rename_tsserver_find_in_comments = get(g:, 'ale_rename_tsserver_find_in_comments')
 let g:ale_rename_tsserver_find_in_strings = get(g:, 'ale_rename_tsserver_find_in_strings')
 
@@ -86,7 +85,7 @@ function! ale#rename#HandleTSServerResponse(conn_id, response) abort
     \   },
     \   {
     \       'conn_id': a:conn_id,
-    \       'should_save': g:ale_rename_save,
+    \       'should_save': !&hidden,
     \   },
     \)
 endfunction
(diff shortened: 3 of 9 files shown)


test/test_code_action.vader
```diff
@@ -261,22 +328,37 @@ Execute(Cursor will not move when it is before text change):
...
+  \})
   AssertEqual [2, 2], getpos('.')[1:2]
 
 # ====C====
 Execute(Cursor column will move to the change end when cursor between start/end):
   let g:test.changes = g:test.create_change(2, 3, 2, 8, 'value2')
 
   for r in range(3, 8)
+    " make tests at the end work
+    let g:notified_changes = []
```
liskin, line 353: It might be better to assert the contents of `g:notified_changes` explicitly/deliberately in some tests (where we test for the notification) rather than to pollute most tests with workarounds for an assertion in the test teardown, I think.

liskin (approved):
I can't recall what changes I wanted to make to the tests, and reading the code again now it looks all right. So I just added some final nitpicks that I spotted now and didn't notice before, but overall it's probably good.

(Oh and I just noticed that python-lsp-server can do cross-file renames, so I could probably switch to this branch and try it. Can't promise today, though.)

bogdan:
The issue still persists for me:

image


Exact steps to reproduce:

1. Clone this repo: https://github.com/unstoppabledomains/resolution
2. Rollout local development: https://github.com/unstoppabledomains/resolution#development
2. Try to rename this method from `call` to `call1` https://github.com/unstoppabledomains/resolution/blob/7e0e9181a3f3d565a029696df393184f9ff15e0f/src/contracts/EthereumContract.ts#L18

My ALEInfo:

```

 Current Filetype: typescript
Available Linters: ['cspell', 'deno', 'eslint', 'standard', 'tslint', 'tsserver', 'typecheck', 'xo']
  Enabled Linters: ['eslint', 'tsserver']
  Ignored Linters: []
 Suggested Fixers: 
  'deno' - Fix TypeScript using deno fmt.
  'dprint' - Pluggable and configurable code formatting platform
  'eslint' - Apply eslint --fix to a file.
  'prettier' - Apply prettier to a file.
  'tslint' - Fix typescript files with tslint --fix.
  'xo' - Fix JavaScript/TypeScript files using xo --fix.
 Linter Variables:

let g:ale_typescript_tsserver_config_path = ''
let g:ale_typescript_tsserver_executable = 'tsserver'
let g:ale_typescript_tsserver_use_global = 0
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = 100
let g:ale_completion_max_suggestions = 50
let g:ale_disable_lsp = 0
let g:ale_echo_cursor = 1
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let b:ale_fixers = ['eslint', 'prettier']
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_filetype_changed = 1
let g:ale_lint_on_insert_leave = 1
let g:ale_lint_on_save = 1
let g:ale_linter_aliases = {}
let g:ale_linters = {}
let b:ale_linters = ['tsserver', 'eslint']
let g:ale_linters_explicit = 0
let g:ale_linters_ignore = {}
let g:ale_list_vertical = 0
let g:ale_list_window_size = 10
let g:ale_loclist_msg_format = '%code: %%s'
let g:ale_max_buffer_history_size = 20
let g:ale_max_signs = -1
let g:ale_maximum_file_size = v:null
let g:ale_open_list = 0
let g:ale_pattern_options = v:null
let g:ale_pattern_options_enabled = v:null
let g:ale_root = {}
let g:ale_set_balloons = 0
let g:ale_set_highlights = 1
let g:ale_set_loclist = 1
let g:ale_set_quickfix = 0
let g:ale_set_signs = 1
let g:ale_sign_error = '>>'
let g:ale_sign_info = '--'
let g:ale_sign_style_error = '>>'
let g:ale_sign_style_warning = '--'
let g:ale_sign_warning = '--'
let g:ale_sign_highlight_linenrs = 0
let g:ale_statusline_format = v:null
let g:ale_type_map = {}
let g:ale_use_global_executables = v:null
let g:ale_virtualtext_cursor = 0
  Command History:

(executable check - success) /Users/bogdan/makabu/ud/resolution/node_modules/eslint/bin/eslint.js
(finished - exit code 0) ['/bin/zsh', '-c', 'cd ''/Users/bogdan/makabu/ud/resolution'' && ''/Users/bogdan/makabu/ud/resolution/node_modules/eslint/bin/eslint.js'' --parser-options=''{project: null}'' --rule=''{''@typescript-eslint/no-floating-promises'': ''off'', ''@typescript-eslint/no-misused-promises'': ''off''}'' -f json --stdin --stdin-filename ''/Users/bogdan/makabu/ud/resolution/src/contracts/EthereumContract.ts'' < ''/var/folders/gd/qqwx5zqx1wx1pxq5s72hzkdh0000gn/T/nvim35GuT1/1/EthereumContract.ts''']

<<<OUTPUT STARTS>>>
[{"filePath":"/Users/bogdan/makabu/ud/resolution/src/contracts/EthereumContract.ts","messages":[],"errorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0,"usedDeprecatedRules":[]}]
<<<OUTPUT ENDS>>>

(executable check - success) /Users/bogdan/makabu/ud/resolution/node_modules/.bin/tsserver
(started) ['/bin/zsh', '-c', '''/Users/bogdan/makabu/ud/resolution/node_modules/.bin/tsserver''']
```

jeremija:
Thanks @bogdan! Would you be able to paste the ALE commit you have checked out just make we're testing on the same codebase?

jeremija:
Got it. Do you have any other plugins installed that might interfere with this functionality?

I haven't gotten the chance to check out your repo yet, but I've definitely tested `tsserver` in other projects. I'll see if I can find the time tonight.

bogdan:
> Got it. Do you have any other plugins installed that might interfere with this functionality?

I tried to disable all plugins except ALE and the issue persisted.

jeremija:
Thanks, just one more question: which version of Neovim are you running?

bogdan:
```
$ nvim --version
NVIM v0.4.3
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/local/Homebrew/Library/Homebrew/shims/mac/super/clang -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/tmp/neovim-20191107-85913-1wbgqz6/neovim-0.4.3/build/config -I/tmp/neovim-20191107-85913-1wbgqz6/neovim-0.4.3/src -I/usr/local/include -I/tmp/neovim-20191107-85913-1wbgqz6/neovim-0.4.3/deps-build/include -I/usr/local/opt/gettext/include -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include -I/tmp/neovim-20191107-85913-1wbgqz6/neovim-0.4.3/build/src/nvim/auto -I/tmp/neovim-20191107-85913-1wbgqz6/neovim-0.4.3/build/include
Compiled by brew@Catalina.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/0.4.3/share/nvim"

Run :checkhealth for more info
```

jeremija:
@bogdan there must be something different in our configurations. I've cloned your repo and it behaves the same as any other TS project:

https://user-images.githubusercontent.com/1489493/152242921-0509956b-5edc-46e5-b615-ecfcad53746a.mp4

I've also tried disabling all of my configs and plugins `ale` with my branch. Running nvim 0.6.1 on Arch Linux.

hsanson:
@bogdan from my limited testing this PR does not break anything on my environment and would like to merge it.  Is the issue you are seeing caused by this PR? that is, without it the issue does not occur?  I cannot merge it if it breaks stuff for others.

test/test_code_action.vader
jeremija: You're right, added more and better assertions and removed these resets.

test/test_code_action_corner_cases.vader
```diff
@@ -35,19 +35,16 @@ Before:
   set fileformats=unix
 
   " two files, one accessed through a buffer, the other using write/readfile only
```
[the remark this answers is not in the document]
jeremija: Right, removed!

jeremija:
@bogdan could you try disabling all fixers like `let b:ale_fixers = ['eslint', 'prettier']`?

bogdan:
> @bogdan[](https://github.com/jeremija) from my limited testing this PR does not break anything on my environment and would like to merge it. Is the issue you are seeing caused by this PR? that is, without it the issue does not occur? I cannot merge it if it breaks stuff for others.

Well in my case, it made `ALERename` functionality broken in more places that it was before. It is hard to evaluate if anyone else is affected, but I am ok with that if anyone else is not affected.

> @bogdan could you try disabling all fixers from let b:ale_fixers = ['eslint', 'prettier']?

Didn't help.

While debugging all that I was under the expression that the plugin tries to apply changes from all files to a currently open file: here is how `EthereumContract.ts` files looks after an error inside `ALERename` command: https://gist.github.com/a29189b4dec2f3acce5778368104c46c

Any idea what may cause that?

I've put some debug info to make sure it is on the same page.

Inside `ale#rename#HandleTSServerResponse`:

```
a:response =>  {'performanceData': {'updateGraphDurationMs': 30}, 'seq': 0, 'request_seq': 1, 'type': 'response', 'success': v:true, 'body': {'info': {'canRename': v:true, 'disp
layName': 'call', 'triggerSpan': {'end': {'offset': 13, 'line': 18}, 'start': {'offset': 9, 'line': 18}}, 'kindModifiers': '', 'fullDisplayName': '"/Users/bogdan/makabu/ud/resol
ution/src/contracts/EthereumContract".EthereumContract.call', 'kind': 'method'}, 'locs': [{'file': '/Users/bogdan/makabu/ud/resolution/src/contracts/EthereumContract.ts', 'locs'
: [{'contextStart': {'offset': 3, 'line': 18}, 'contextEnd': {'offset': 4, 'line': 29}, 'end': {'offset': 13, 'line': 18}, 'start': {'offset': 9, 'line': 18}}]}, {'file': '/Users/bogdan/makabu/ud/resolution/src/UnsInternal.ts', 'locs': [{'end': {'offset': 52, 'line': 64}, 'start': {'offset': 48, 'line': 64}}, {'end': {'offset': 54, 'line': 71}, 'start'
: {'offset': 50, 'line': 71}}, {'end': {'offset': 53, 'line': 86}, 'start': {'offset': 49, 'line': 86}}, {'end': {'offset': 69, 'line': 100}, 'start': {'offset': 65, 'line': 100
}}]}]}, 'command': 'rename'}
```

Inside `ale#code_action#HandleCodeAction`:

```
l:changes =>  [{'fileName': '/Users/bogdan/makabu/ud/resolution/src/contracts/EthereumContract.ts', 'textChanges': [{'end': {'offset': 13, 'line': 18}, 'newText': 'call1', 'star
t': {'offset': 9, 'line': 18}}]}, {'fileName': '/Users/bogdan/makabu/ud/resolution/src/UnsInternal.ts', 'textChanges': [{'end': {'offset': 52, 'line': 64}, 'newText': 'call1', '
start': {'offset': 48, 'line': 64}}, {'end': {'offset': 54, 'line': 71}, 'newText': 'call1', 'start': {'offset': 50, 'line': 71}}, {'end': {'offset': 53, 'line': 86}, 'newText':
 'call1', 'start': {'offset': 49, 'line': 86}}, {'end': {'offset': 69, 'line': 100}, 'newText': 'call1', 'start': {'offset': 65, 'line': 100}}]}]
```

Inside `ale#code_action#ApplyChanges`:

```
/Users/bogdan/makabu/ud/resolution/src/contracts/EthereumContract.ts [{'end': {'offset': 13, 'line': 18}, 'newText': 'call1', 'start': {'offset': 9, 'line': 18}}] {'conn_id': '/Users/bogdan/makabu/ud/resolution/node_modules/.bin/tsserver:/Users/bogdan/makabu/ud/resolution'}

/Users/bogdan/makabu/ud/resolution/src/UnsInternal.ts [{'end': {'offset': 52, 'line': 64}, 'newText': 'call1', 'start': {'offset': 48, 'line': 64}}, {'end': {'offset': 54, 'line
': 71}, 'newText': 'call1', 'start': {'offset': 50, 'line': 71}}, {'end': {'offset': 53, 'line': 86}, 'newText': 'call1', 'start': {'offset': 49, 'line': 86}}, {'end': {'offset'
: 69, 'line': 100}, 'newText': 'call1', 'start': {'offset': 65, 'line': 100}}] {'conn_id': '/Users/bogdan/makabu/ud/resolution/node_modules/.bin/tsserver:/Users/bogdan/makabu/ud
/resolution'} #
```

jeremija:
@bogdan thanks for digging into it! Can you try adding `set hidden` to your `.vimrc`?

bogdan:
> @bogdan thanks for digging into it! Can you try making sure set hidden is in your .vimrc?

I have `set nohidden` in `vimrc` and the bug goes away if I `set hidden` instead :+1:. But I don't really remember why I put `set nohidden` to `.vimrc` long ago and I think the plugin should work both ways.

jeremija:
Maybe we should force save when `set nohidden` is set. IDK if there are any other consequences of doing that.

jeremija:
It makes perfect sense that we can't edit files in the background when `set nohidden` is set:

```
'hidden' 'hid' 'nohidden' 'nohid'
  'hidden' 'hid'          boolean (default on)
                          global
          When off a buffer is unloaded (including loss of undo information)
          when it is abandoned.  When on a buffer becomes hidden when it is                abandoned.  A buffer displayed in another window does not become
          hidden, of course.
          Commands that move through the buffer list sometimes hide a buffer
          although the 'hidden' option is off: when the buffer is modified,
          'autowrite' is off or writing is not possible, and the '!' flag was
          used.  See also windows.                                                         To hide a specific buffer use the 'bufhidden' option.
          'hidden' is set for one command with ":hide {command}" :hide.
```

liskin:
Makes me wonder if perhaps going back to https://github.com/dense-analysis/ale/pull/3782 (with the additional `didChange` dispatch, probably) might be worth considering… :-)

> IDK if there are any other consequences of doing that.

If we stick with your approach here, @bogdan with `set nohidden` will lose undo history of the current file whenever a cross-file rename is done. (If I understand it correctly.) With the existing code + #3782, there's no switching/unloading of buffers, so history isn't lost.

On the other hand your proposed code is way simpler…

bogdan:
The feature that you may undo a change in multiple files is quite superior. I am not sure if any IDE supports that. I would expect the change to multiple files (especially if some files also got renamed) to be automatically saved and irreversible so in case I need to rollback I call `ALERename` again. Do you have any insight on how it is done IDEs like webstorm?

Is it possible to set `should_save` to true  on in case of `set nohidden` when `tsserver` response is passed to `codeaction` here? https://github.com/dense-analysis/ale/pull/4049/files#diff-d89b532c02fb769c9d6b1613b82b7e7c6c1149e61a1f8a87aee0126eeb5b3508R21

jeremija:
I just remembered that I had do disable saving of files to get around a bug when modifying open files in the background: whenever such a file was modified and saved ALE would run fixers if they were enabled and then the files would sometimes be modified, but the LSP wouldn't get notified about this change, which made it really buggy to perform the rename operation twice because the LSP response would contain old positions.

I wrote about it in the commit message of the first commit:

> I also noticed another bug when using Go with `gopls` LSP and the `gofmt` fixer. Whenever the file was saved, the `gofmt` would run and reformat the file. But it seems there was some kind of a race condition so I disabled saving for now, and all of the modified files will be unsaved, so the user should call `:wa` to save them. I personally like this even better because I can inspect exactly what changes happened, and I instantly see them in the other opened buffers, which was previously not the case.

bogdan:
> whenever such a file was modified and saved ALE would run fixers if they were enabled

Is it possible to disable fixers temporarily after when rename operation is performed instead?

jeremija:
I'm not sure at this point. But this PR has been working perfectly for me with `set hidden` for months. The only thing I have to remember is to occasionally do is `:wa`, usually before and after renaming.

hsanson:
@jeremija I really want to merge this PR as it does fixes the three issues mentioned in the description and from my testing using ruby solargraph it works very well.  But considering that "nohidden" is the default in vim/neovim and that if this option is set then the renaming fails I thinks is hard to justify the merge.

After some thinking I believe the best option here is to tell ALE to save the files by setting "should_save" to true when calling the code action handler.  I left comments in the PR showing were to make the changes.  I tested this and works find with hidden and nohidden set and also is the way  filerename works so I think this is the proper way to handle this.

autoload/ale/organize_imports.vim
```diff
@@ -17,7 +17,9 @@ function! ale#organize_imports#HandleTSServerResponse(conn_id, response) abort
     \       'description': 'Organize Imports',
     \       'changes': l:file_code_edits,
     \   },
-    \   {}
+    \   {
+    \       'conn_id': a:conn_id,
+    \   },
```
hsanson, line 22: Add "should_save" in the options here so the files are saved and we get no errors when "nohidden" is set:

```
{
    'conn_id':  a:conn_id,
    'should_save': 1,
}
```

autoload/ale/rename.vim
```diff
@@ -84,7 +84,7 @@ function! ale#rename#HandleTSServerResponse(conn_id, response) abort
     \       'changes': l:changes,
     \   },
     \   {
-    \       'should_save': 1,
+    \       'conn_id': a:conn_id,
     \   },
```
hsanson, line 88: Same here:


```
{
    'conn_id':  a:conn_id,
    'should_save': 1,
}
```

autoload/ale/rename.vim
```diff
@@ -116,7 +116,7 @@ function! ale#rename#HandleLSPResponse(conn_id, response) abort
         \       'changes': l:changes,
         \   },
         \   {
-        \       'should_save': 1,
```
hsanson, line 119: Actually seems this was the behavior before?

jeremija:
>  But considering that "nohidden" is the default in vim/neovim and that if this option is set then the renaming fails I thinks is hard to justify the merge.

Where did you get that info from? According to the excerpt from Neovim help, it seems like `set hidden` is default.

That said, I understand why you would want to make these changes, but I'd like to at least make the saving optional, and we can warn the users in help that disabling saving won't work with `set nohidden`. This way we can ensure that the default doesn't break, and we can have the saving optional for those that use `set hidden`. WDYT?

autoload/ale/rename.vim
jeremija: Yeah it was, the reason why I changed the default is in the description of the first commit.

autoload/ale/organize_imports.vim
jeremija: Done, also added an option to disable this behavior.

autoload/ale/rename.vim
jeremija: Done, also added an option to disable this behavior.

autoload/ale/rename.vim
jeremija: Done, also added an option to disable this behavior.

hsanson:
Making it configurable should be straightforward but honestly I think allowing users to set this variable with the "hidden" option off will result in many shooting themselves in the foot.  If "hidden" is off then ALERename will throw cryptic errors and all files won't be properly changed as expected. This most likely will be seen as a bug by many even if documented in the help.

I rather set the behavior based on if hidden is on or off like:

```
{
    'should_save": !&hidden
}
```

This would tell ALE to save the buffers if "hidden" is off and don't save them if hidden is on. 
Of course also add a note in the documentation explaining this behavior in ALERename.

If you prefer to allow users to change this behavior via a config (e.g. g:ale_save_on_rename) at least make the default use the same logic based on hidden configuration. This way ALE will just work, in the sense that no errors will be thrown, for both hidden on and off.

hsanson:
@jeremija great!, thanks for the changes.  Tested with hidden and nohidden and works as expected.  Just one little detail that is annoying at best. When renaming a class that is used in several files neovim asks me to press enter many times.  Three times per file when nohidden is set and twice per file when hidden is set.  This is because the "Execute() function calls.  

I left some comments in the PR adding silent to the problematic places so this does not happen.

autoload/ale/code_action.vim
```diff
@@ -63,29 +62,29 @@ function! s:ChangeCmp(left, right) abort
...
-        let l:lines = getbufline(l:buffer, 1, '$')
-
-        " Add empty line if there's trailing newline, like readfile() does.
-        if getbufvar(l:buffer, '&eol')
-            let l:lines += ['']
-        endif
-    else
-        let l:lines = readfile(a:filename, 'b')
+    if l:buffer != l:orig_buffer
+        call ale#util#Execute('edit ' . a:filename)
```
hsanson, line 75: Add silent so neovim stops asking to press Enter to continue:

```
call ale#util#Execute('silent edit ' . a:filename)
```

autoload/ale/code_action.vim
```diff
@@ -155,46 +154,25 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
...
-        if a:should_save
-            call ale#util#Execute(':e!')
-        endif
+    call ale#util#SetBufferContents(l:buffer, l:lines)
 
-        call setpos('.', [0, l:pos[0], l:pos[1], 0])
+    call ale#lsp#NotifyForChanges(l:conn_id, l:buffer)
+
+    if l:should_save
+        call ale#util#Execute(':w!')
```
hsanson, line 169: Same here:

```
call ale#util#Execute(':silent w!')
```

autoload/ale/code_action.vim
```diff
@@ -155,46 +154,25 @@ function! ale#code_action#ApplyChanges(filename, changes, should_save) abort
...
-            execute printf(
-            \   'autocmd BufEnter <buffer=%d>'
-            \       . ' call ale#code_action#ReloadBuffer()',
-            \   l:buffer
-            \)
-        augroup END
+    call setpos('.', [0, l:pos[0], l:pos[1], 0])
+
+    if l:orig_buffer != l:buffer && bufexists(l:orig_buffer)
+        call ale#util#Execute('buf ' . string(l:orig_buffer))
```
hsanson, line 175: And here:

```
call ale#util#Execute('silent buf ' . string(l:orig_buffer))
```

hsanson (changes requested):
Just one last request and I think this is ready to merge.

autoload/ale/code_action.vim
jeremija: Done

autoload/ale/code_action.vim
jeremija: Done

autoload/ale/code_action.vim
jeremija: Done

hsanson (approved):
Thanks for the work and patience.

PR #4050, opened 2022-02-02
Treat ale_open_list integer value as a threshold

nospam2998:
Setting `ale_open_list = 1` is great when having multiple lints to address. Yet it is annoying as %#"¤ to use that setting when writing new code, as the window tends to open way too frequently.

This change makes this already numeric option possible to set to higher values, and doing so leads to only opening a list window if the number of warnings or errors equals to or exceeds the value. A typical useful value could be `ale_open_list = 2` which allows for typing a word to completion without getting a split buffer. One can imagine other thresholds being attractive for other use cases.

No change to functionality is intended when `ale_open_list` is set to the previously allowed values `0`, `1` and `on_save`.

### DIFF
--- a/autoload/ale/list.vim
+++ b/autoload/ale/list.vim
@@ -36,12 +36,22 @@ function! ale#list#IsQuickfixOpen() abort
 endfunction
 
 " Check if we should open the list, based on the save event being fired, and
-" that setting being on, or the setting just being set to `1`.
-function! s:ShouldOpen(buffer) abort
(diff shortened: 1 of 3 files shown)


nospam2998:
Branch rebased onto current master.

w0rp (approved):
This is a pretty simple change and doesn't complicate the code. I'm happy to merge this.

w0rp:
I set it to auto-merge when the checks are completed.

PR #4059, opened 2022-02-08
fix: duplicate tags in doc

pipoprods:
Hi,

Last update introduces an error when updating helptags:

```
E5113: Error while calling lua chunk: /home/sno/.config/nvim/init.lua:7: Vim(helptags):E154: Duplicate tag "g:ale_python_mypy_executable" in file /home/sno/.local/share/nvim/plugged/ale/doc/ale-python.txt
```


hsanson (approved):
Thanks for the catch and fix.

PR #4084, opened 2022-02-18
Fix 4083 - do not remove completion items that have additional text e…

### LINKED ISSUE #4083: Autocomplete ignores any completion items with additional edits unless autoimport is enabled.
For reasons foreign to me ALE drops all completion items reported by the server if they have additional text edits (see additionalTextEdits in the LSP [CompletionItem](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#completionItem) spec).

This completely renders autocompeltion useless for Java/Kotlin language servers were most (if not all) completion items have additional text edits.

Relevant code: https://github.com/dense-analysis/ale/blob/master/autoload/ale/completion.vim#L599

hsanson:
Do not remove completion items if they have additional text edits and auto-import is off. But still only apply the additional edits if auto-import is on.

Fixes #4083

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -596,16 +596,6 @@ function! ale#completion#ParseLSPCompletions(response) abort
             continue
         endif
 
-        " Don't use LSP items with additional text edits when autoimport for
-        " completions is turned off.
-        if !empty(get(l:item, 'additionalTextEdits'))
-        \&& !(
-        \   get(l:info, 'additional_edits_only', 0)
-        \   || g:ale_completion_autoimport
-        \)
-            continue
-        endif
-
         let l:doc = get(l:item, 'documentation', '')
 
         if type(l:doc) is v:t_dict && has_key(l:doc, 'value')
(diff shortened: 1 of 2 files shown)


w0rp (changes requested):
We'll have to find some other way of handling this. ALE is assuming any completion result that has some additional text is something adding an import line. With this change we'll start seeing items appear in the completion menu that require imports without importing them.

Maybe we just need to document that you have to turn the option on to use completion for certain language servers.

w0rp:
One thing I considered for this is that we could just turn the auto import option on by default and document that you might not get results if it's off.

w0rp:
Another, more complex option is to look at what the text edit is and compare it with what we'll be doing if it's not there, so if there's a redundant text edit in the results we throw it away. I can imagine there are language servers that return it when they don't need to.

hsanson:
Bit unfortunate since Java and Kotlin language servers seem to include additional text edits on every completion item they return rendering auto-completion unusable without auto-import enabled. Not seeing any auto-complete entries on my Java projects but seeing them for python/ruby led me to assume this was a bug, hence the issue and attempted fix.

I vote to enable auto-import by default and warn users that disabling it may affect auto-completion. This on the basis that ALE should just work out of the box without configuration.

I submitted another PR that does the changes #4102

w0rp:
Yeah, I figured as much. They are probably redundantly including the text when just the name of the completion item would be enough. As I say, one trick we could try when auto-import is off is comparing the changes for text edits against the changes you'd get from inserting the completion item, and if they are the same, we could include that item in the menu.

PR #4087, opened 2022-02-19
Fix 4086 - use stable command options for rustc

### LINKED ISSUE #4086: rustc linter should default to using stable interface
See #1031, #1684 and  [StackOverflow 51485765](https://stackoverflow.com/q/51485765/run-rustc-without-generating-any-files).

While the default options to rustc has been discussed before, the outcome seems a bit misinformed. The statement _people on older versions can change their options appropriately_ suggests a lack of contextual understanding. [Appendix G - How Rust is Made and “Nightly Rust”](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html) should be relevant background reading.

Features hidden behind `-Z` will always be unstable and are inappropriate to use as defaults. Quoting [rustc doc](https://doc.rust-lang.org/rustc/command-line-arguments.html):

> -Z: set unstable options

Once features are stabilized they end up accessible through a stable interface. Ending compilation after generating a mid-level intermediate representation is a more appropriate default. Anyone needing to track _+nightly_ can easily update their options.

Please consider merging [fix/use_stable_rust](https://git.netizen.se/ale/commit/?h=fix/use_stable_rust). I'm reluctant to submit a pull request as my committer and author fields for [PR#4050](https://github.com/dense-analysis/ale/pull/4050) were messed up through that process. [_For web-based Git operations_](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-email-preferences/setting-your-commit-email-address) that buggy behavior is documented as expected.

hsanson:
Fix #4086

### DIFF
--- a/test/linter/test_rustc.vader
+++ b/test/linter/test_rustc.vader
@@ -5,7 +5,7 @@ After:
   call ale#assert#TearDownLinterTest()
 
 Execute(The default command should be correct):
-  AssertLinter 'rustc', 'rustc --error-format=json -Z no-codegen -'
+  AssertLinter 'rustc', 'rustc --error-format=json --emit=mir -o /dev/null -'
 
 Execute(The options should be configurable):
   let b:ale_rust_rustc_options = '--foo'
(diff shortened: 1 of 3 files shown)


test/linter/test_rustc.vader
```diff
@@ -5,7 +5,7 @@ After:
   call ale#assert#TearDownLinterTest()
 
 Execute(The default command should be correct):
-  AssertLinter 'rustc', 'rustc --error-format=json -Z no-codegen -'
+  AssertLinter 'rustc', 'rustc --error-format=json --emit=mir -o /dev/null -'
```
nospam2998, line 8: Ooops! I missed these, sorry. Great that you caught them! Thanks!

nospam2998 (approved):
It would not surprise me if this change makes the rustc linter suddenly work for a lot of users, and that they will thus end up with duplicate lints from `rustc` and `cargo`.

I could argue that excluding this linter by default would be a sane default, as essentially everyone will get duplicates and prefer their linting through cargo. That's a different change though, as merging this still improves the situation.

hsanson:
@nospam2998 actually rustc is disabled by default. From the docs:

```
  Only cargo and rls are enabled by default. To switch to using rustc instead
  of cargo, configure |g:ale_linters| appropriately: >
```

And checking ALE's code this is indeed the case.

Issue #3160, opened 2020-05-07
There is no .deb package.


pavel-the-best:
1) This is easier to install plugin from deb package.
2) I've wanted to submit vim-ale to debian repo, but I didn't find the package.

What requirements does ale have? I wanted to create the package, but I didn't find the requirements.

nospam2998:
According to [packages.debian.org](https://packages.debian.org/vim-ale), vim-ale is available in bullseye, bookworm and sid. Details in [salsa](https://salsa.debian.org/vim-team/vim-ale). Similarly, searching [packages.ubuntu.com](https://packages.ubuntu.com/vim-ale) suggests availability in universe for hirsute, impish and jammy.

Ergo, there are .deb packages. Could this ticket thus be closed?

pavel-the-best:
Sure, that feels like a LONG time ago

Issue #1968, opened 2018-10-03
Linting/Checking fails on rust files with some shells.


stegmanh:
<!--
  This is the template for reporting ALE bugs. Make sure you try updating ALE
  to a more recent version before reporting a bug. Look through existing bug
  reports for similar issues before reporting a new one. Don't leave comments
  about new bugs in the comment section for old issues.

  Make sure to try disabling other plugins and trying to repeat your bug before
  reporting it in ALE. Some times problems can arise when two plugins are used
  together, but often your issues might be problems with other plugins.
-->

## Information

**VIM version**
NVIM v0.3.1-1-g489d32f2b                                                                                         
Build type: RelWithDebInfo  

Operating System: Ubuntu 18.04 <!-- Describe your operating system version. -->

## What went wrong

When saving a rust file that has an error, ale shows no lint/error message. When looking at the `ALEInfo` output, it appears that calls to cargo build/cargo check fail. 

I've tried solving the issue by `ale_linters/rust/cargo.vim` and changing the quoting, shell command, etc. Without any success. I've also tried changing the ale settings for rust, but they all still seem to fail at this step. Admittedly though I'm not very familiar with neovim api/vimscript.

I've also reproduced this issue on my other laptop. Same operating system, but with fresh install of neovim, ale, rust and configuration files. 

I'm happy to submit a PR if there's a fix somewhere that can be made. 

Thanks

## Reproducing the bug

<!-- Write a list of steps below. -->

1. Install Ale
2. Open up rust file and add some code that would cause cargo check/cargo build to fail
3. Notice no lints appear and an error appears in ALEInfo

Example failing code
```rust
fn main() {
    let a = 10;
    let b: String = a;
    println!("Hello, world!");
}
```

### :ALEInfo
```
Current Filetype: rust                                                                                                                                                                                                             
Available Linters: ['cargo', 'rls', 'rustc']                                                                                                                                                                                        
  Enabled Linters: ['cargo']                                                                                                                                                                                                        
 Suggested Fixers:                                                                                                                                                                                                                  
  'remove_trailing_lines' - Remove all blank lines at the end of a file.                                                                                                                                                            
  'rustfmt' - Fix Rust files with Rustfmt.                                                                                                                                                                                          
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.                                                                                                                                           
 Linter Variables:                                                                                                                                                                                                                  
[...]

Issue #1397, opened 2018-03-05
Add support for Perl 6 / Rakudo Perl 6


7ojo:
Short in short.. adding support for Perl 6 / Rakudo Perl 6 would be great!

nospam2998:
It seems Perl6 support was added back in [September 2018](https://github.com/dense-analysis/ale/commit/2b2e766dc6df4d05e58c106f51b5444be4efbcea). Could this ticket be closed, or should it have it's title rephrased to clarifying it meaning something more specific?

hsanson:
Thanks for noticing. If there is something missing I think opening a new issue is better.  We have already many issues that are no longer relevant making difficult to see the important ones.

PR #4088, opened 2022-02-19
Document the use of author rewriting tools

nospam2998:
As a new contributor to ALE, it felt great to be able to help getting issues #1397, #1968 and #3160 closed.

Overall it was good to get #4050 and #4087 merged too, but I must say it was surprising to have those commits attributed to a non-descriptive random alias plus (less surprising) one of ALE's most valued maintainers, rather than the identifying details I used when pushing the commits.

Please consider merging something like this suggested sub-chapter, making future contributors informed about the specific quirks of ALE's merge process.

Adapting the process so that these phrasings became irrelevant would of course be my preference, but that's unlikely reasonable after weighing up all interests.


w0rp:
This is good. Thanks for this. The attribution quirk is due to how GitHub handles squash merges. All contributions are squashed so I can easily build a changelog from changes without having to dig through a merge tree, which is something I'll do again soon.

PR #4089, opened 2022-02-19
remove --vanilla option

uzsolt:
With "--vanilla" option the Rscript does not read ~/.Renviron. In my case this file stores R_LIBS_USER variable which points to a user-installed R libraries (not system-wide).
My styler package is installed to my home directoy - and I think many people use install.packages(...) to his home directory (not only the default .libPaths()).
See a Q&A here: https://stackoverflow.com/questions/37970555/controversy-of-r-libs-user-path

### DIFF
--- a/autoload/ale/fixers/styler.vim
+++ b/autoload/ale/fixers/styler.vim
@@ -6,7 +6,7 @@ call ale#Set('r_styler_options', 'tidyverse_style()')
 
 function! ale#fixers#styler#Fix(buffer) abort
     return {
-    \   'command': 'Rscript --vanilla -e '
+    \   'command': 'Rscript -e '
     \       . '"suppressPackageStartupMessages(library(styler));'
     \       . 'style_file(commandArgs(TRUE), transformers = '
     \       . ale#Var(a:buffer, 'r_styler_options') . ')"'
--- a/autoload/ale/fixers/styler.vim
+++ b/autoload/ale/fixers/styler.vim
@@ -6,7 +6,7 @@ call ale#Set('r_styler_options', 'tidyverse_style()')
 
 function! ale#fixers#styler#Fix(buffer) abort
     return {
-    \   'command': 'Rscript -e '
+    \   'command': 'Rscript --no-save --no-restore --no-site-file --no-init-file -e '
     \       . '"suppressPackageStartupMessages(library(styler));'
     \       . 'style_file(commandArgs(TRUE), transformers = '
     \       . ale#Var(a:buffer, 'r_styler_options') . ')"'


hsanson (changes requested):
Thanks for the change but the tests also need to be updated with the change to be able to merge.

uzsolt:
> Thanks for the change but the tests also need to be updated with the change to be able to merge.

I see one test failed. I don't understand:

> 
>  Starting Vader: C:\testplugin\test\fixers\test_styler_fixer_callback.vader
>     (1/1) [EXECUTE] The styler callback should include custom styler options
>     (1/1) [EXECUTE] (X) {'read_temporary_file': 1, 'command': 'Rscript -e "suppressPackageStartupMessages(library(styler));style_file(commandArgs(TRUE), transformers = a_custom_option)" %t'} should be equal to {'read_temporary_file': 1, 'command': 'Rscript --vanilla -e "suppressPackageStartupMessages(library(styler));style_file(commandArgs(TRUE), transformers = a_custom_option)" %t'}

Should update this file too?

hsanson:
Yes, this PR removes the --vanilla flag from the command so the test also must be updated to reflect that change. Remove the --vanilla flag from the test case inside the test_styler_fixer_callback.vader file.

w0rp:
I happened to notice that #3744 is related while I'm finally updating the changelog after a long time.

uzsolt:
> I happened to notice that #3744 is related while I'm finally updating the changelog after a long time.
Sorry, I forgot this issue. Should I do everything or do you this? (I don't use Github deeply)

hsanson:
@uzsolt  please continue working on this PR since it is related to #3744 but is not the same. The change in #3744 removes --vanilla option from the linter while this one removes it from the fixer.

I do not use RScript but I see in #3744 that "--vanilla" option was replaced with "--no-save --no-restore --no-site-file --no-init-file". For consistency maybe would be good idea to do the same in this PR?? not sure if those options apply when using RScript as a fixer.

uzsolt:
I'm thinking about it. My opinion: the `--no-environ` should remove (so the `--vanilla` isn't good solution). 
The `--no-save`: it's okay. We don't need save the lint session. The `--no-restore` is same, we don't need restore any previously session.
The `--no-site-file` and `--no-init-file`: I'm not an expert R-user but in my case these options isn't annoying while linting/fixing.

So: the `--no-save` and `--no-restore` should use (it's clear). The `--no-site-file` and `--no-init-file` options _probably_ aren't needed while linting/fixing.

w0rp:
Update the options so the fixer and linter match some sensible options and I'll have a look.

uzsolt:
Sorry, I don't use github's so I think I did something wrong, I create a new pull request (https://github.com/dense-analysis/ale/pull/4129). I hope these patches will merge into one but no :(

Issue #4028, opened 2022-01-05
ALEImport does not work, it cannot import some basic libraries


kabeersvohra:
## Information

**VIM version**
NVIM v0.6.0                                                                           
Build type: RelWithDebInfo

Operating System: 
Ubuntu 20.04

## What went wrong

When I try and `ALEImport` on something which is successfully imported with autocomplete, it does not work. Also many basic libraries cannot be imported at all such as `datetime`

## Reproducing the bug

Video here:
https://user-images.githubusercontent.com/5859986/148258615-d827f76e-fba1-4057-94b1-19103138e219.mp4

### :ALEInfo

```
Current Filetype: python
Available Linters: ['bandit', 'cspell', 'flake8', 'flakehell', 'jedils', 'mypy', 'prospector', 'pycodestyle', 'pydocstyle', 'pyflakes', 'pylama', 'pylint', 'pylsp', 'pyre', 'pyright', 'vulture']
  Enabled Linters: ['pyright']
  Ignored Linters: []
 Suggested Fixers: 
  'add_blank_lines_for_python_control_statements' - Add blank lines before control statements.
  'autoflake' - Fix flake issues with autoflake.
  'autoimport' - Fix import issues with autoimport.
  'autopep8' - Fix PEP8 issues with autopep8.
  'black' - Fix PEP8 issues with black.
  'isort' - Sort Python imports with isort.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'reorder-python-imports' - Sort Python imports with reorder-python-imports.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'yapf' - Fix Python files with yapf.
 Linter Variables:

let g:ale_python_auto_pipenv = 0
let g:ale_python_auto_poetry = 0
let g:ale_python_pyright_config = {}
let g:ale_python_pyright_executable = 'pyright-langserver'
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = 100
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = 50
let g:ale_disable_lsp = 0
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
[...]

PR #4105, opened 2022-03-06
Investigating fixing ALEImport

yoshi1123:
Perhaps this might help fix ALEImport.

ALEImport currently does not work for me with eclipselsp. It seems to use the first character position of \<cword> to complete (in the position parameter of the textDocument/complete request message), which results in many completion items. When using the last character of \<cword> it seems to return only completion items that match \<cword>, instead of many words, perhaps all references in the source.

If using the first character matches all words in the source, then it will only be guaranteed to work (taking a long time) if the response has `{"isIncomplete": false}`. For example if the LSP response were to include all the words in the completion item list, and so it is incomplete, like this,
```
{"jsonrpc":"2.0","id":4,"result":{"isIncomplete":true, ...}
```
then there is a good chance \<cword> is not found in the results.

I have not looked at this extensively, but if what I have said above is true, then this would be a bug that this PR fixes. This PR seems to fix ALEImport for me.

---

Also, this might be a bug in the current version, which this patch would also fix:

If the cursor is on the first character of a word to import, it will not work, if using the current approach, one would have to use the "c" flag in searchpos (line 914 in `autoload/ale/completion.vim`):
```
let l:column = searchpos('\V' . escape(l:word, '/\'), 'bnc', l:line)[1]
```
instead of the current:
```
let l:column = searchpos('\V' . escape(l:word, '/\'), 'bn', l:line)[1]
```
Or use another method, such as what is in this PR.

---

Perhaps this is related to issue #4028.

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -911,7 +911,8 @@ function! ale#completion#Import() abort
     endif
 
     let [l:line, l:column] = getpos('.')[1:2]
-    let l:column = searchpos('\V' . escape(l:word, '/\'), 'bn', l:line)[1]
+    let l:column = searchpos('\V' . escape(l:word, '/\'), 'bnc', l:line)[1]
+    let l:column = l:column + len(l:word) - 1
 
     if l:column isnot 0
         let l:started = ale#completion#GetCompletions('ale-import', {
--- a/test/completion/test_ale_import_command.vader
+++ b/test/completion/test_ale_import_command.vader
@@ -187,7 +187,7 @@ Execute(ALEImport should request imports correctly for tsserver):
   \   'conn_id': 0,
   \   'request_id': 0,
   \   'source': 'ale-import',
-  \   'column': 11,
+  \   'column': 21,
   \   'line': 2,
   \   'line_length': 21,
   \   'prefix': 'missingword',
@@ -206,7 +206,7 @@ Execute(ALEImport should request imports correctly for tsserver):
   \   'conn_id': 347,
   \   'request_id': 1,
   \   'source': 'ale-import',
-  \   'column': 11,
+  \   'column': 21,
   \   'line': 2,
   \   'line_length': 21,
   \   'prefix': 'missingword',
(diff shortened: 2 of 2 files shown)


autoload/ale/util.vim
```diff
@@ -571,3 +571,31 @@ function! ale#util#ToResource(uri) abort
 
     return l:resource
 endfunction
+
+" Returns a list of matches of pattern in string.
+function! ale#util#MatchStrings(string, pattern)
```
w0rp, line 576: Let's keep the completion-specific code in the completion script so it isn't loaded for people who don't use completion.

autoload/ale/completion.vim
```diff
@@ -902,16 +902,17 @@ endfunction
...
     if empty(l:word)
         call s:message('Nothing to complete at cursor!')
 
         return
     endif
 
-    let [l:line, l:column] = getpos('.')[1:2]
-    let l:column = searchpos('\V' . escape(l:word, '/\'), 'bn', l:line)[1]
+    let l:line = getpos('.')[1]
+    let l:column = l:wordpos[2]
```
w0rp, line 915: If we need the end position of the word, can't we just use `len(expand('<cword>'))` to work out the end position based on the start position we have already?

w0rp (changes requested):

autoload/ale/completion.vim
yoshi1123: Well the vim function `matchstrpos` (used in `ale#util#GetCWordBoundary()`) returns the start and end position of the match, so we already have the end position.

The line (in GetCWordBoundary),
```
let lastpos = matchstrpos(getline('.'), escape(l:word, '/\'), laststart)
```
has `lastpos` set to [word, startpos, endpos].

We could, and then have just one return value, and change the function name `GetCWordBoundary` to `GetCWordStartColumn`.

What do you think?

autoload/ale/util.vim
yoshi1123: Yes that would be more efficient. And if the functions `MatchStrings` and `GetCWordBoundary` can be reused for other purposes, we could always put them in `util.vim`.

autoload/ale/completion.vim
w0rp: Can't we just use `len(expand('<cword>'))` to know how long the string is and use that to get the end position from the start position we already know?

yoshi1123: I think I see what you are asking. Yes, if you add 'c' to the flags for `searchpos()`, then that should get the beginning position, and we could use `len(expand('<cword>')` to get the end position. That would work.

All depends if you want to use `matchstrpos()` or `searchpos()`. Do you think there is a significant performance difference between the two?

yoshi1123:
Okay, take a look.

w0rp (commented):
This is much better. I tested with with tsserver for TypeScript and gopls for Go, and it seems to work for both. I'll update the tests to match the new column position and get this in.

autoload/ale/completion.vim
w0rp: I don't think it will matter.

w0rp (approved):

w0rp:
Cheers! :beers:

PR #4106, opened 2022-03-06
Improve pylama linter output handling

favadi:
The current pattern for matching pylama output is:

    '\v^.{-}:([0-9]+):([0-9]+): +%(([A-Z][0-9]+):? +)?(.*)$'

By default, pylama output format looks like:

    samples/pylama.py:99:101 E501 line too long (189 > 100 characters) [pycodestyle]
    samples/pylama.py:104:101 E501 line too long (189 > 100 characters) [pycodestyle]
    samples/pylama.py:11:9  f-string is missing placeholders [pyflakes]
    samples/pylama.py:12:101 E501 line too long (150 > 100 characters) [pycodestyle]

Notice the there is no `:` in the first part of the output. It looks
like the linter expects pylama to a non-default `--format` option, or
pylama changed the default format at some point. pylama support
`--format json` that will produce following output:

    {
      "source": "pycodestyle",
      "col": 101,
      "lnum": 99,
      "etype": "E",
      "message": "line too long (189 > 100 characters)",
      "filename": "samples/pylama.py",
      "number": "E501"
    },
    {
      "source": "pycodestyle",
      "col": 101,
      "lnum": 104,
      "etype": "E",
      "message": "line too long (189 > 100 characters)",
      "filename": "samples/pylama.py",
      "number": "E501"
    },
    {
      "source": "pyflakes",
      "col": 9,
      "lnum": 11,
      "etype": "E",
      "message": "f-string is missing placeholders",
      "filename": "samples/pylama.py",
      "number": ""
    },
    {
      "source": "pycodestyle",
      "col": 101,
      "lnum": 12,
      "etype": "E",
      "message": "line too long (150 > 100 characters)",
      "filename": "samples/pylama.py",
      "number": "E501"
    }

This commit introduces following changes:

* forces pylama command to always use `--format json`. Not sure if this
  will cause any issues.

* use `json_decode` instead of regex to parse the errors

### DIFF
--- a/ale_linters/python/pylama.vim
+++ b/ale_linters/python/pylama.vim
@@ -110,25 +110,26 @@ function! ale_linters#python#pylama#Handle(buffer, version, lines) abort
 
         for l:error in l:errors
             call add(l:output, {
-                        \   'lnum': l:error['lnum'],
-                        \   'col': l:error['col'],
-                        \   'code': l:error['number'],
-                        \   'type': get(l:pylint_type_to_ale_type, l:error['etype'], 'W'),
-                        \   'sub_type': get(l:pylint_type_to_ale_sub_type, l:error['etype'], ''),
-                        \   'text': printf('%s [%s]', l:error['message'], l:error['source']),
-                        \})
+            \   'lnum': l:error['lnum'],
+            \   'col': l:error['col'],
+            \   'code': l:error['number'],
+            \   'type': get(l:pylint_type_to_ale_type, l:error['etype'], 'W'),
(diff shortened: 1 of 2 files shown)


w0rp (approved):

w0rp (changes requested):
I just double-checked the changelog and the JSON format option was added late last year: https://github.com/klen/pylama/blob/develop/Changelog

To support this we'll have to check which version of pylama is being used and only use the JSON format if the version is new enough. Have a look at how we check the version for `flake8` as an example.

Issue #4085, opened 2022-02-18
ALEImport always applies first suggestion ignoring everything else.


hsanson:
When running ALEImport the LSP can return multiple options to import but ALE simply takes and applies the first one.  Would be better if ALE instead shows an auto-complete menu and allows the user to select which item to import.

Tested this with eclipselsp on a project that uses [Joda Time](https://www.joda.org/joda-time/) that has its own implementations of LocalDate and LocalDateTime classes.  When calling ALEImport on either of those classes the system.time.LocalDate gets imported without giving the option to the user to select org.joda.time.LocalDate instead.

Checking the code this seems to be fixable by modifying these parts:

 -  https://github.com/dense-analysis/ale/blob/master/autoload/ale/completion.vim#L377  logic here seems to only open the completion popup if the source is ale-manual or ale-automatic. The ale-import source used when calling ALEImport is not considered.
 - https://github.com/dense-analysis/ale/blob/master/autoload/ale/completion.vim#L9 Seems the popup menu only displays in insert mode. ALEImport is invoked in normal mode so the menu won't appear.

On my limited testing I was unable to get the popup shown by modifying these parts of the code. I did verify that eclipselsp is indeed returning multiple completion items with additional text edits for the possible imports but ALE refuses to show the popup.

w0rp:
I've marked it as an enhancement. It would be good to show a menu if there's more than one to pick from.

PR #4108, opened 2022-03-06
Add import menu for ALEImport for multiple matches

yoshi1123:
This is one possible solution. Its a menu via echo. Perhaps temporary if another type of menu is preferred.

This PR is for the second out of the two commits on this branch, #74c0760.

For me, without the first commit, ALEImport does not even work (see PR #4105). Thus, dealing with PR #4105 first might be best.

This simply displays a menu if there are multiple names to import matching the selected name. In other words, this is a possible solution for issue #4085.

I am not sure if its worth it to write a test for `ale#util#Prompt` since it gets input from the user. I could rewrite it to make it more testable if you like.

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -370,32 +370,31 @@ function! ale#completion#Show(result) abort
         return
     endif
 
-    " Replace completion options shortly before opening the menu.
-    if l:source is# 'ale-automatic' || l:source is# 'ale-manual'
-        call s:ReplaceCompletionOptions(l:source)
-
-        call timer_start(0, function('s:OpenCompletionMenu'))
-    endif
-
-    if l:source is# 'ale-callback'
-        call b:CompleteCallback(b:ale_completion_result)
-    endif
-
-    if l:source is# 'ale-import'
+    if l:source is# 'ale-import' && len(b:ale_completion_result) == 1
         call ale#completion#HandleUserData(b:ale_completion_result[0])
 
         let l:text_changed = '' . g:ale_lint_on_text_changed
 
         " Check the buffer again right away, if linting is enabled.
         if g:ale_enabled
-        \&& (
-        \   l:text_changed is# '1'
-        \   || l:text_changed is# 'always'
-        \   || l:text_changed is# 'normal'
-        \   || l:text_changed is# 'insert'
-        \)
+                    \&& (
+                    \   l:text_changed is# '1'
+                    \   || l:text_changed is# 'always'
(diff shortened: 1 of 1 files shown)


w0rp (commented):
The code just looks very messy. I can't quite figure out what it's doing, but I have a feeling there's a much easier way to do this.

yoshi1123:
Okay I pushed a commit that refactors it and adds tests. Take a look.

autoload/ale/completion.vim
```diff
@@ -382,7 +382,23 @@ function! ale#completion#Show(result) abort
...
+                        \   l:names,
+                        \   1
+                        \)
+            if l:userselection is# v:null
+                return
+            else
+                let l:index = l:userselection
+            endif
+        endif
+        call ale#completion#HandleUserData(b:ale_completion_result[l:index])
```
w0rp, line 401: You can update this now I've merged the commit with the simple fix for the position. What you can probably do here is this, merging this with the above conditions.

```vim
if l:source is# 'ale-import' && len(b:ale_completion_result) == 1
    call ale#completion#HandleUserData(b:ale_completion_result[0])
    
    let l:text_changed = '' . g:ale_lint_on_text_changed

    " Check the buffer again right away, if linting is enabled.
    if g:ale_enabled
    \&& (
    \   l:text_changed is# '1'
    \   || l:text_changed is# 'always'
    \   || l:text_changed is# 'normal'
    \   || l:text_changed is# 'insert'
    \)
        call ale#Queue(0, '')
    endif
elif l:source is# 'ale-automatic' || l:source is# 'ale-manual' || l:source is# 'ale-import'
    if l:source is# 'ale-import'
        call feedkeys('i', 'n')
    endif

    " Replace completion options shortly before opening the menu.
    call s:ReplaceCompletionOptions(l:source)
    call timer_start(0, function('s:OpenCompletionMenu'))
elif l:source is# 'ale-callback'
    call b:CompleteCallback(b:ale_completion_result)
endif
```

If there are extra possible results we should be able to just enter insert mode and show the usual completion menu, so we don't have to have code for a special menu.

w0rp (changes requested):

autoload/ale/completion.vim
yoshi1123: Okay. I changed it as you suggested. Only thing is instead of `feedkeys('i', 'n')`, I used `feedkeys('ea', 'n')`. This fixes it so it does not add new text in the middle of the word the cursor is on.

PR #4109, opened 2022-03-08
Don't use temp file for clj-kondo

christoph-frick:
clj-kondo since version 2022.03.04 has a check to for the namespace
declared in the file matching the actual file hierarchy, which now
always fails due to the file hierarchy generated for %t not matching the
file to be linted.  Since clj-kondo only reads the file, use %s instead
of %t to start it.

See: https://github.com/clj-kondo/clj-kondo/issues/1240

### DIFF
--- a/ale_linters/clojure/clj_kondo.vim
+++ b/ale_linters/clojure/clj_kondo.vim
@@ -8,7 +8,7 @@ function! ale_linters#clojure#clj_kondo#GetCommand(buffer) abort
 
     let l:command = 'clj-kondo'
     \   . ale#Pad(l:options)
-    \   . ' --lint %t'
+    \   . ' --lint %s'
 
     return l:command
 endfunction
--- a/test/linter/test_clj_kondo.vader
+++ b/test/linter/test_clj_kondo.vader
@@ -6,10 +6,10 @@ After:
 
 Execute(The default command should be correct):
   AssertLinter 'clj-kondo', 'clj-kondo'
-  \   . ' --cache --lint %t'
+  \   . ' --cache --lint %s'
 
 Execute(Extra options should be supported):
   let g:ale_clojure_clj_kondo_options = '--config ./clj-kondo/config.edn'
 
   AssertLinter 'clj-kondo', 'clj-kondo'
-  \   . ' --config ./clj-kondo/config.edn --lint %t'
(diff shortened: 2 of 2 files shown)


hsanson:
Wonder if this change will break linting for users using pervious versions of clj-kondo.  If this is the case it would be better to check the tool version and set the proper command depending on the version. You can check https://github.com/dense-analysis/ale/blob/master/autoload/ale/fixers/isort.vim#L65 that is one example on how to check version and set proper command depending on it.

christoph-frick:
Thanks for checking.  To my knowledge, clj-kondo never changed the file
(which I assume is the main reason for `%t`).

In the meantime this feature got
[removed](https://github.com/clj-kondo/clj-kondo/commit/af6b18c03ab9603a96cd33fe0b8bc33d8c934675)
again due to different problems.

I'll close this for now and if and when it comes up again I'll check
with the clj-kondo maintainers what is best.

mode89:
As we got the [confirmation](https://github.com/clj-kondo/clj-kondo/discussions/1668) that clj-kondo has never changed the source files, should this PR be reopened?

christoph-frick:
@mode89 I don't think this patch is "finished".  As stated in the kondo discussion, the advantage of running kondo on the temp-file is rapid feedback. You basically get instant kondo errors after each insert/change. If the original file would be used, you would only get this feedback after writing the file. So the best apporach would be: run kondo on tempfile, but ignore file-centric checks; run kondo on the actual file once it gets written - or something like that.

mode89:
Oh, I see.

borkdude:
Note that you can still use a temp file, but be sure to pass `--filename <filename>` with the original filename. clj-kondo uses this for linting namespace names.

borkdude:
Example from flycheck-clj-kondo:

https://github.com/borkdude/flycheck-clj-kondo/blob/ff7bed2315755cfe02ef471edf522e27b78cd5ca/flycheck-clj-kondo.el#L67

borkdude:
Also see https://github.com/mfussenegger/nvim-lint/blob/master/lua/lint/linters/clj-kondo.lua#L16 /cc @lispyclouds

borkdude:
@christoph-frick Awesome!

PR #4114, opened 2022-03-15
Make it possible to override awk --lint option

agostonbarna:
Currently, it's not possible to override the awk `--lint` option with
```viml
let g:ale_awk_gawk_options = '--lint=no-ext'
```
altough this could be useful for those who only use gawk and don't want to get these lint errors:
> FEATURE X is a gawk extension

The idea is to move the default `--lint` option before the `awk_gawk_options` in the gawk.vim code to give the custom `--lint=...` option a higher precedence.

Updated & ran the tests.


w0rp (approved):

agostonbarna:
Thanks @w0rp !

PR #3799, opened 2021-07-02
Change default of go_staticcheck_lint_package to 1

fishy:
Fixes https://github.com/dense-analysis/ale/issues/3798.

PR #4122, opened 2022-03-24
Update help for ale_go_staticcheck_lint_package

### LINKED ISSUE #3798: `go_staticcheck_lint_package` really should default to 1 instead of 0
this is a minimal example:

```
$ cat go.mod
module foo

go 1.16
$ cat type.go
package foo

type Foo int
$ cat code.go
package foo

func Bar() {
        var _ Foo
}
```

`staticcheck .` would report no error, which is expected (this is what `let g:ale_go_staticcheck_lint_package = 1` does), but `staticcheck code.go` would complain (this is what `let g:ale_go_staticcheck_lint_package = 0` does, which is the current default value):

```
$ staticcheck code.go 
code.go:4:8: undeclared name: Foo (compile)
```

I'm not sure in which scenario `go_staticcheck_lint_package = 0` is preferred.

A PR will follow shortly.

stackline:
Fixes #3798.
I have updated a help file in response to the changes in #3799.

### DIFF
--- a/doc/ale-go.txt
+++ b/doc/ale-go.txt
@@ -354,7 +354,7 @@ g:ale_go_staticcheck_options                     *g:ale_go_staticcheck_options*
 g:ale_go_staticcheck_lint_package           *g:ale_go_staticcheck_lint_package*
                                             *b:ale_go_staticcheck_lint_package*
   Type: |Number|
-  Default: `0`
+  Default: `1`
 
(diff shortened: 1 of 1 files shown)


fishy (approved):
Sorry I forgot to update the doc in my original PR.

w0rp (approved):

PR #4124, opened 2022-03-25
Set default value for insertTextFormat

JD557:
As per the spec (https://microsoft.github.io/language-server-protocol/specification#textDocument_completion), `insertTextFormat` should default to `InsertTextFormat.PlainText` when ommited.

Unfortunately, I'm not very familiar with viml develpment and couldn't run the `./run-tests` script (I keep getting an error that the image is not known, even with `--build-image`), so I ended up not writing any tests nor checking if the current tests pass...


w0rp (approved):
I added a test for this.

PR #4128, opened 2022-03-29
Add buf linter and fixer

amckinney:
This adds the `buf lint` command as a linter and the `buf format` command as a fixer. The `buf lint` linter was previously published from our own [vim-buf](https://github.com/bufbuild/vim-buf) repository [here](https://github.com/bufbuild/vim-buf/blob/main/ale_linters/proto/buf_lint.vim), whereas the `buf format` fixer is completely new. We'd like both of them to exist in `ale` now that `buf` has a stable `v1` release.

I've included a description and test for each (pattern matching from what previously exists, e.g. `clang-format`, `dartfmt`, and `gofmt`). I've also verified that both work locally by pointing `vim-plug` to my local copy of `ale`.

I've used the following `.vimrc` configuration to test locally against `.proto` files:

```vim
let g:ale_linters = {
\   'proto': ['buf-lint'],
\}
let g:ale_linters_explicit = 1

let g:ale_fixers = {
\   'proto': ['buf-format'],
\}
let g:ale_fix_on_save = 1
```

### DIFF
--- a/test/fixers/test_buf_format_fixer_callback.vader
+++ b/test/fixers/test_buf_format_fixer_callback.vader
@@ -1,0 +1,21 @@
+Before:
+  Save g:ale_proto_buf_format_executable
+
+  " Use an invalid global executable, so we don't match it.
+  let g:ale_proto_buf_format_executable = 'xxxinvalid'
+
+  call ale#test#SetDirectory('/testplugin/test/fixers')
+
+After:
+  Restore
+
+  call ale#test#RestoreDirectory()
+
+Execute(The buf-format callback should return the correct default values):
+  call ale#test#SetFilename('../test-files/proto/testfile.proto')
+
+  AssertEqual
+  \ {
+  \   'command': ale#Escape('xxxinvalid') . ' format  %t',
+  \ },
+  \ ale#fixers#buf_format#Fix(bufnr(''))
--- a/test/fixers/test_buf_format_fixer_callback.vader
+++ b/test/fixers/test_buf_format_fixer_callback.vader
@@ -16,6 +16,6 @@ Execute(The buf-format callback should return the correct default values):
 
   AssertEqual
   \ {
-  \   'command': ale#Escape('xxxinvalid') . ' format  %t',
+  \   'command': ale#Escape('xxxinvalid') . ' format %t',
   \ },
   \ ale#fixers#buf_format#Fix(bufnr(''))
--- a/test/linter/test_buf_lint.vader
+++ b/test/linter/test_buf_lint.vader
@@ -1,0 +1,22 @@
+Before:
+  call ale#assert#SetUpLinterTest('proto', 'buf-lint')
+  call ale#test#SetFilename('test.proto')
+
+After:
+  call ale#assert#TearDownLinterTest()
+
+Execute(The default command should be correct):
+  AssertLinter 'buf-lint',
+  \ ale#Escape('buf')
+  \ . ' lint'
+  \ . ' %s#include_package_files=true'
+
+Execute(The callback should include any additional options):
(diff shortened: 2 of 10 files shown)


test/linter/test_buf_lint.vader
```diff
@@ -0,0 +1,22 @@
+Before:
+  call ale#assert#SetUpLinterTest('proto', 'buf-lint')
```
hsanson, line 2: The SetUpLInterTest() function uses filename matching to setup the linter. Since the filename is "proto/buf_lint.vim" the second argument here should be "buf_lint" or ALE won't find the linter.

test/linter/test_buf_lint.vader
```diff
@@ -0,0 +1,22 @@
...
+  call ale#test#SetFilename('test.proto')
+
+After:
+  call ale#assert#TearDownLinterTest()
+
+Execute(The default command should be correct):
+  AssertLinter 'buf-lint',
```
hsanson, line 9: This first argument is the binary name so this should be "buf" instead of "buf-lint".

test/fixers/test_buf_format_fixer_callback.vader
```diff
@@ -0,0 +1,21 @@
...
+  Restore
+
+  call ale#test#RestoreDirectory()
+
+Execute(The buf-format callback should return the correct default values):
+  call ale#test#SetFilename('../test-files/proto/testfile.proto')
+
+  AssertEqual
+  \ {
+  \   'command': ale#Escape('xxxinvalid') . ' format  %t',
```
hsanson, line 19: There is an extra blank space between "format" and "%t" that causes the test to fail.

hsanson (changes requested):

test/test-files/proto/testfile.proto
```diff
@@ -0,0 +1 @@
+syntax = "proto3";
```
w0rp, line 1: You can actually make the file 0 bytes.

amckinney:
Thanks for the reviews @w0rp and @hsanson! This should be ready for another look.

amckinney:
It looks like we're seeing a lint failure for the name of the `buf-lint` linter:

```
Running custom linting rules

ale_linters/proto/buf_lint.vim: Use snake_case names for linters
``` 

Is there any way we can make an exception for this one? Given that we're migrating the linter from [vim-buf](https://github.com/bufbuild/vim-buf), this would be a breaking change for anyone that removes the dependency on `vim-buf` and instead uses `ale` directly - they would need to update the following configuration:

```
let g:ale_linters = {
\   'proto': ['buf-lint'],
\}
```

Plus, I noticed that there is already a `protoc-gen-lint` linter in here and it appears that kebab-case is allowed for it ([here](https://github.com/dense-analysis/ale/blob/d3df00b89803f1891a772c47fc8eda6a1e9e1baa/ale_linters/proto/protoc_gen_lint.vim#L21)). I searched around for an exception list (one that would include `protoc-gen-lint`), but I can't seem to find it anywhere.

hsanson:
@amckinney what  you can do is set the name using snake case and add an alias. See "cabal_ghc" for example:

```
call ale#linter#Define('haskell', {
\   'name': 'cabal_ghc',
\   'aliases': ['cabal-ghc'],
\   'output_stream': 'stderr',
\   'executable': 'cabal',
\   'cwd': '%s:h',
\   'command': function('ale_linters#haskell#cabal_ghc#GetCommand'),
\   'callback': 'ale#handlers#haskell#HandleGHCFormat',
\})
```

amckinney:
Thanks for the tip @hsanson, I added the alias and everything works on my end.

hsanson (approved):

PR #4129, opened 2022-03-30
Update test_styler_fixer_callback.vader

uzsolt:
Remove --vanilla option.


w0rp (changes requested):
This updates the test but not the fixer file to change the command, so you might want to change that too.

w0rp:
Oh, I see. You need to add this change to this PR: #4089

PR #4130, opened 2022-03-31
Fix end line number when it exceeds the file

skyfmmf:
If the end of the error exceeds the file, set it to the last line, similarly as it is done with the beginning of the error.

This issue was noticed by vim getting killed because of out of memory. The reason was that rust-analyzer delivered a very large number for the end line of an error (4294967295) due to an underflow bug (rust-analyzer/rust-analyzer#11866). Ale fixed the beginning of the error to line 1 and then tried to generate a list of the lines to highlight (https://github.com/dense-analysis/ale/blob/v3.2.0/autoload/ale/highlight.vim#L80). This list is very large, actually so large that vim was automatically killed because it ran out of memory.

While I don't think Ale should add workarounds for all bugs in language servers, I do think that this fix is reasonable generic and can help to mitigate future memory usage issues with error ranges larger than the current file.

### DIFF
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -347,6 +347,12 @@ function! ale#engine#FixLocList(buffer, linter_name, from_other_source, loclist)
 
         if has_key(l:old_item, 'end_lnum')
             let l:item.end_lnum = str2nr(l:old_item.end_lnum)
+
+            " When the error ends after the end of the file, put it at the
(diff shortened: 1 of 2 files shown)


hsanson (approved):
Thanks for the explanation and fix. Agree the fix is generic enough and is always good to protect ourselves from wrong data delievered from LSP, linters, and fixers.

Issue #4005, opened 2021-12-06
[Neovim 0.6 Diagnostic API] ALE as the source of diagnostics to Neovim's new Diagnostic module


bluz71:
Hello ALE developers,

Spawning a new from [my comment](https://github.com/dense-analysis/ale/issues/3528#issuecomment-978915077) in #3528.

Neovim 0.6 was recently released with the main feature being the new [vim.diagnostic](https://neovim.io/doc/user/diagnostic.html) API. Previously diagnostics were a part of the native LSP client; but now it's an extracted stand-alone top level module.

Linting plugins such as [nvim-lint](https://github.com/mfussenegger/nvim-lint) and [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim) are now using that API.

Can ALE also be enhanced to do the same? Does the architecture allow for that?

Aka, can ALE be a diagnostic source for display by another diagnostic engine? ALE has long had the option to be end-point for other diagnostic sources (aka CoC feeding it's diagnostics to ALE for display and navigation). But, this issue is wanting to go the other way; use ALE simply as the runner and then feed it's diagnostics to Neovim's new diagnostic module. The benefit for Neovim folks is a single source of display (signs, virtual-text, underlines) and a single set of mappings to navigate diagnostics.

Many thanks.

dbalatero:
This would be awesome to see! I even tried to switch to null-ls but something in how its architected made the linting/fixing lock up the editor on save. ALE at its core feels much more stable than these new Lua reimplementation plugins, at least at this time, so I'd love to keep using this and unify the diagnostics :)

w0rp:
I don't see why we can't support this. It looks like a few simple function calls. We'd just need to check a flag being enabled to send diagnostics to NeoVim's system in `ale#engine#SetResults` instead of calling the various functions there for settings all of the visual elements that ALE does.

w0rp:
I don't use NeoVim myself, but someone who does can open a pull request for this.

dbalatero:
> We'd just need to check a flag being enabled to send diagnostics to NeoVim's system in ale#engine#SetResults instead of calling the various functions there for settings all of the visual elements that ALE does.

Just to iterate on the design a bit before anyone codes…

Are you thinking this flag would be a "one or the other" kind of thing?

e.g.
* If the flag is `true`, send diagnostics only to Neovim
* If the flag is `false`, keep doing the normal stuff ALE does today

dbalatero:
[...]

PR #4135, opened 2022-04-01
Add support for sending ALE's output to Neovim's diagnostics API

dbalatero:
Hi @w0rp,

Here is a PR for the changes we discussed in #4005.

## Changes

* `g:ale_send_to_neovim_diagnostics = 0 | 1` as a configuration option
* When `g:ale_send_to_neovim_diagnostics` is enabled, we send ALE output to Neovim diagnostics
* A version check for Neovim 0.6+. We print an error if the config is set to `1` and the user is on an unsupported Vim version
* Documentation

## TODOs

- [ ] Fix off by 1 issue reported by @daliusd below
- [ ] Maybe good idea to add a test test_linting_sets_signs.vader or test_sign_placement.vader that tests ALE signs are disabled when neovim diagnostics are enabled.
- [ ] Also maybe add a test to test_cursor_warnings.vader that tests cursor messages is not echoed when neovim diagnostics are enabled.
- [ ] Figure out if it's OK to disable loclist/quickfix like we do currently
- [ ] Fix issue with `g:ale_virtualtext_cursor` reported below

## Screenshots

**Error on startup if you're not running Neovim 0.6+**
image

**ALE in the diagnostics API**
image

**ALE in trouble.nvim's fix list**
image

### DIFF
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -242,13 +242,15 @@ endfunction
 
 function! ale#engine#SendResultsToNeovimDiagnostics(buffer, loclist)
     if !has('nvim-0.6')
-        " We will warn the user on startup if they set
+        " We will warn the user on startup as well if they try to set
         " g:ale_send_to_neovim_diagnostics outside of a Neovim context.
         return
     endif
 
-    let SendDiagnostics = luaeval("require('diagnostics').sendAleResultsToDiagnostics")
-    call SendDiagnostics(a:buffer, a:loclist)
+    " Keep the Lua surface area really small in the VimL part of ALE,
+    " and just require the diagnostics.lua module on demand.
+    let l:SendDiagnostics = luaeval("require('diagnostics').sendAleResultsToDiagnostics")
+    call l:SendDiagnostics(a:buffer, a:loclist)
 endfunction
 
 function! s:RemapItemTypes(type_map, loclist) abort
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -184,14 +184,22 @@ endfunction
 function! ale#engine#SetResults(buffer, loclist) abort
     let l:linting_is_done = !ale#engine#IsCheckingBuffer(a:buffer)
 
-    " Set signs first. This could potentially fix some line numbers.
-    " The List could be sorted again here by SetSigns.
-    if g:ale_set_signs
-        call ale#sign#SetSigns(a:buffer, a:loclist)
-    endif
+    if g:ale_send_to_neovim_diagnostics
+        call ale#engine#SendResultsToNeovimDiagnostics(a:buffer, a:loclist)
+    else
+        " Set signs first. This could potentially fix some line numbers.
+        " The List could be sorted again here by SetSigns.
+        if g:ale_set_signs
+            call ale#sign#SetSigns(a:buffer, a:loclist)
+        endif
 
-    if g:ale_set_quickfix || g:ale_set_loclist
-        call ale#list#SetLists(a:buffer, a:loclist)
+        if g:ale_set_quickfix || g:ale_set_loclist
+            call ale#list#SetLists(a:buffer, a:loclist)
+        endif
+
+        if g:ale_set_highlights
+            call ale#highlight#SetHighlights(a:buffer, a:loclist)
+        endif
     endif
 
     if exists('*ale#statusline#Update')
@@ -199,21 +207,19 @@ function! ale#engine#SetResults(buffer, loclist) abort
         call ale#statusline#Update(a:buffer, a:loclist)
     endif
 
-    if g:ale_set_highlights
-        call ale#highlight#SetHighlights(a:buffer, a:loclist)
-    endif
-
     if l:linting_is_done
-        if g:ale_echo_cursor
-            " Try and echo the warning now.
-            " This will only do something meaningful if we're in normal mode.
-            call ale#cursor#EchoCursorWarning()
-        endif
+        if !g:ale_send_to_neovim_diagnostics
+            if g:ale_echo_cursor
+                " Try and echo the warning now.
+                " This will only do something meaningful if we're in normal mode.
+                call ale#cursor#EchoCursorWarning()
+            endif
 
-        if g:ale_virtualtext_cursor
-            " Try and show the warning now.
-            " This will only do something meaningful if we're in normal mode.
-            call ale#virtualtext#ShowCursorWarning()
+            if g:ale_virtualtext_cursor
+                " Try and show the warning now.
+                " This will only do something meaningful if we're in normal mode.
+                call ale#virtualtext#ShowCursorWarning()
+            endif
         endif
 
         " Reset the save event marker, used for opening windows, etc.
@@ -232,6 +238,17 @@ function! ale#engine#SetResults(buffer, loclist) abort
         " Call user autocommands. This allows users to hook into ALE's lint cycle.
         silent doautocmd <nomodeline> User ALELintPost
     endif
+endfunction
+
+function! ale#engine#SendResultsToNeovimDiagnostics(buffer, loclist)
+    if !has('nvim-0.6')
+        " We will warn the user on startup if they set
+        " g:ale_send_to_neovim_diagnostics outside of a Neovim context.
+        return
+    endif
+
+    let SendDiagnostics = luaeval("require('diagnostics').sendAleResultsToDiagnostics")
+    call SendDiagnostics(a:buffer, a:loclist)
 endfunction
 
 function! s:RemapItemTypes(type_map, loclist) abort
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -240,7 +240,7 @@ function! ale#engine#SetResults(buffer, loclist) abort
     endif
 endfunction
 
-function! ale#engine#SendResultsToNeovimDiagnostics(buffer, loclist)
+function! ale#engine#SendResultsToNeovimDiagnostics(buffer, loclist) abort
     if !has('nvim-0.6')
         " We will warn the user on startup as well if they try to set
         " g:ale_send_to_neovim_diagnostics outside of a Neovim context.
@@ -249,7 +249,7 @@ function! ale#engine#SendResultsToNeovimDiagnostics(buffer, loclist)
 
     " Keep the Lua surface area really small in the VimL part of ALE,
     " and just require the diagnostics.lua module on demand.
-    let l:SendDiagnostics = luaeval("require('diagnostics').sendAleResultsToDiagnostics")
+    let l:SendDiagnostics = luaeval('require("diagnostics").sendAleResultsToDiagnostics')
     call l:SendDiagnostics(a:buffer, a:loclist)
 endfunction
 
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -1938,6 +1938,22 @@ g:ale_root                                                         *g:ale_root*
   LSP linter, it will not run.
 
 
+g:ale_send_to_neovim_diagnostics             *g:ale_send_to_neovim_diagnostics*
+
+  Type: Number
+  Default: 0
+
+  If enabled, this option will disable ALE's standard UI, and instead send
+  all linter output to Neovim's diagnostics API. This allows you to collect
+  errors from nvim-lsp, ALE, and anything else that uses diagnostics all in
+  one place.
+
+  To enable this option, set the value to 1.
+
+  This option requires Neovim 0.6+, as that version introduces the diagnostics
+  API.
+
+
 g:ale_set_balloons                                         *g:ale_set_balloons*
                                                            *b:ale_set_balloons*
 
--- a/lua/diagnostics.lua
+++ b/lua/diagnostics.lua
@@ -1,0 +1,46 @@
+local module = {}
+
+local ale_type_to_diagnostic_severity = {
+  E = vim.diagnostic.severity.ERROR,
+  W = vim.diagnostic.severity.WARN,
+  I = vim.diagnostic.severity.INFO
+}
+
+module.sendAleResultsToDiagnostics = function(buffer, loclist)
+  diagnostics = {}
+
+  -- Convert all the ALE loclist items to the shape that Neovim's diagnostic
+  -- API is expecting.
+  for _, location in ipairs(loclist) do
+    table.insert(
+      diagnostics,
+      -- All line numbers from ALE are 1-indexed, but all line numbers
+      -- in the diagnostics API are 0-indexed, so we have to subtract 1
+      -- to make this work.
+      {
+        lnum = location.lnum - 1,
+        -- Ending line number, or if we don't have one, just make it the same
+        -- as the starting line number
+        end_lnum = (location.end_lnum or location.lnum) - 1,
+        -- Which column does the error start on?
+        col = (location.col or 1) - 1,
+        -- end_col does *not* appear to need 1 subtracted, so we don't.
+        end_col = location.end_col,
+        -- Which severity: error, warning, or info?
+        severity = ale_type_to_diagnostic_severity[location.type] or "E",
+        -- The error message
+        message = location.text,
+        -- e.g. "rubocop"
+        source = location.linter_name,
+      }
+    )
+  end
+
+  vim.diagnostic.set(
+    vim.api.nvim_create_namespace('ale'),
+    buffer,
+    diagnostics
+  )
+end
+
+return module
--- a/plugin/ale.vim
+++ b/plugin/ale.vim
@@ -191,6 +191,7 @@ let g:ale_popup_menu_enabled = get(g:, 'ale_popup_menu_enabled', has('gui_runnin
 let g:ale_send_to_neovim_diagnostics = get(g:, 'ale_send_to_neovim_diagnostics', 0)
 
 if g:ale_send_to_neovim_diagnostics && !has('nvim-0.6')
+    " no-custom-checks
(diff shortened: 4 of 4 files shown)


dbalatero:
Here's an example of ALE now unified with nvim-lsp's diagnostics in trouble.nvim:

image

image

dbalatero:
@bluz71 Since you were the original requester for this, would you be down to test this branch out on your end as well, and see if this config works for you?

ajukraine:
Cool stuff :) I actually never used ALE before, and started with nvim's LSP and diagnostics. But some old plugins still not supporting nvim's diagnostics, but do support ALE. And I was missing some diagnostic's features, like virtual text (partially support with `ale_virtualtext_cursor`, but not exactly what I was used to).

So I decide to switch to your branch :)

Here is first thing that jumps into my eyesight:

image

Duplication happens if `ale_virtualtext_cursor` is on.

ajukraine:
I think it behaves like because of auto command

https://github.com/dense-analysis/ale/blob/master/autoload/ale/events.vim#L142

dbalatero:
@ajukraine Is this because diagnostics _and_ ALE are both using virtualtext?

Is the right solution simply to call this out in the docs as something you might want to disable with `g:ale_virtualtext_cursor = 0`?

ajukraine:
@dbalatero yes, they both use virtual text.

But the problem is even a bit deeper. your current implementation takes care only of `ale#engine#SetResults` part of the logic. However virtual text feature is also used in `events.vim`. Which means that after your change the behaviour is not consistent. It also applies to "echo cursor" functionality.

So I think there at least 3 ways of moving forward:
 - Let ALE's users decide whether to disable ALE's virtual text functionality or not; Is OK for power users, but will be confusing for beginners
 - Make the current change consistent and follow the same logic everywhere around ALE's code base (from what I can see it should be at least repeated in `events.vim`; maybe somewhere else, needs to be investigated)
 - Make the current change consistent, but move the logic into deeper levels of code base (`virtualtext.vim` and `cursor.vim`)

dbalatero:
> So I think there at least 3 ways of moving forward:

Makes sense! I think I'd defer to @w0rp for preferences here, but I'm happy to follow any of these paths!

bluz71:
> @bluz71 Since you were the original requester for this, would you be down to test this branch out on your end as well, and see if this config works for you?

Apologies, I am not in a position to test this since I did switch over to [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim) about 4 or so months back. 

Thankfully there are other testers available.

Cheers.

mikedfunk:
Thanks for this PR!! This is a killer feature for me.

I am still using ALE due to [these reasons](https://github.com/mikedfunk/dotfiles/blob/b3acf6bd0be76f96fb25a0d917580e2f11c4355b/.config/lvim/config.lua#L1418-L1423) for phpcs and phpstan. I was maintaining a whole separate UI for ALE errors in virtual text, floating windows, sign column, tabs, etc. With your PR I can get rid of all that hacky code. 

Hopefully I can eventually switch these last few tools over to null-ls too, but not until these problems are fixed.

dbalatero:
@w0rp any feedback so far on this PR? I can address any items that jump out to you. Also curious about your thoughts on @ajukraine's comments!

hsanson (commented):
Change seems simple enough and my limited tests show it works as intended. But I still have some doubts about this:

1. Maybe good idea to add a test test_linting_sets_signs.vader or test_sign_placement.vader that tests ALE signs are disabled when neovim diagnostics are enabled.
2. Also maybe add a test to test_cursor_warnings.vader that tests cursor messages is not echoed when neovim diagnostics are enabled.
3. I see that by enabling neovim diagnostics we disable filling the quickfix/loclist. I personally use these to navigate between issues reported by ALE. Not sure if neovim diagnostics offers alternative ways to navigate the issues. If not I would say is better to let ALE populate the quickfix/loclist?
Enabling neovim diagnostics disables ALE's signs, cursor echo, quickfix|loclist,

hsanson:
Actually on further testing it seems that the cursor echo and virtual text are still enabled even with neovim diagnostics enabled:

image


The red "E" and virtual text I believe are rendered by neovim diagnostics.  The virtual text and echo at the bottom of the screenshot are rendered by ALE.

I do not use neovim diagnostics but from this screenshot I belive it handles "Signs" and "Virtual Text"?  If this is the case then I would only disable those two in ALE and leave everything else untouched (e.g. echo cursor, loclist/quicklist, highlights, etc).

daliusd:
I have tried this and it is great. However I get error that is not reproducible without this PR. It does not look like a problem of this PR but somehow it is now triggered:

1. Install `npm install -g vim-language-server`
2. Open .vimrc or any other vim language file.
3. Move cursor to any keyword:

image

Here is a fix:
```
diff --git a/autoload/ale/hover.vim b/autoload/ale/hover.vim
index a738c84..483ce92 100644
--- a/autoload/ale/hover.vim
+++ b/autoload/ale/hover.vim
@@ -231,7 +231,7 @@ function! ale#hover#HandleLSPResponse(conn_id, response) abort
             \&& (l:set_balloons is 1 || l:set_balloons is# 'hover')
                 call balloon_show(join(l:lines, "\n"))
             elseif get(l:options, 'truncated_echo', 0)
-                call ale#cursor#TruncatedEcho(join(l:lines[0], '\n'))
+                call ale#cursor#TruncatedEcho(join([l:lines[0]], '\n'))
             elseif g:ale_hover_to_floating_preview || g:ale_floating_preview
                 call ale#floating_preview#Show(l:lines, {
                 \   'filetype': 'ale-preview.message',
```

lua/diagnostics.lua
```diff
@@ -0,0 +1,46 @@
...
+      -- All line numbers from ALE are 1-indexed, but all line numbers
+      -- in the diagnostics API are 0-indexed, so we have to subtract 1
+      -- to make this work.
+      {
+        lnum = location.lnum - 1,
+        -- Ending line number, or if we don't have one, just make it the same
+        -- as the starting line number
+        end_lnum = (location.end_lnum or location.lnum) - 1,
+        -- Which column does the error start on?
+        col = (location.col or 1) - 1,
```
daliusd, line 26: Could it be that this one is not correct. Here 'trouble' has trouble:
image

dbalatero:
@hsanson Thanks for looking at the PR! Would it be possible to add Neovim 0.6+ to the test matrix in CI? If so I could rebase this branch on top of that change and add more tests in that specifically use the new 0.6 APIs that this PR is based off of.

w0rp:
I'm sure this is good. I'll have a look through soon when I can.

lua/diagnostics.lua
daliusd: column_number is -1 in `trouble/view.lua:453`. In ale `lua/diagnostics.lua` location.col is 0, lua interpreter says that `0 or 1` is 0, so your code sets column as -1.

dbalatero: Thanks! I will fix this error. Do you have a line of text that reproduces it? I'm just curious about where the error is supposed to be in your case. Something tells me it starts in column 0?

dbalatero: (a screenshot of the broken line of code in your editor would be fine)

daliusd: Here is minimal reproducible Typescript case. There will be error on line 3 because import statement is incomplete. So, you are right that it is column 0 (it is interesting that eslint and tsserver reports different column):

```
import

const x = 1;
console.log(x);
```

image

dbalatero: Last questions:

1. For `tsserver`, is that via Neovim LSP, or ALE's LSP? Just curious about the column discrepancy between tsserver + eslint.
2. For `eslint`, is that just configured with the standard ALE setup? Do you use `eslint_d` or `eslint`?

dbalatero:
@daliusd I have rebased this branch on `master` again, so it should up to date as of now.

hsanson:
> @hsanson Thanks for looking at the PR! Would it be possible to add Neovim 0.6+ to the test matrix in CI? If so I could rebase this branch on top of that change and add more tests in that specifically use the new 0.6 APIs that this PR is based off of.

Latest master already tests againt neovim 0.6.1 and this new branch #4180 that needs review tests against 0.7.

lua/diagnostics.lua
daliusd: 1. ALE's LSP.
2. Standard ALE setup. `eslint`

daliusd: This is clearly some kind of `eslint` corner case as the same code in `js` file produces different indexes so I assume this depends even on how typescript is handled. I guess it would be good not to pass `-1` as column (anyway that does not make sense).

dbalatero: Hmm… is the thing to do to wrap it in in `Math.max(0, lineNumber)` to ensure it never drops below `0`? Or is there a safer approach to handle this edge case, I wonder.

daliusd: KISS `col = math.max((location.col or 1) - 1, 0),`

w0rp:
@dbalatero So it's easy to understand, what I try to do with the tests is run them in CI with:

1. The oldest Vim version supported.
2. The oldest NeoVim version supported.
3. The latest Vim version.
4. The latest NeoVim version.

I try not to ever drop support for any versions at will, but instead _very carefully_. I'm waiting for the day when NeoVim hits 1.0. I actually won't use NeoVim myself until it does. I have become... a 30 year old boomer. I run extremely bog standard Linux software these days and rarely use anything that isn't completely stable. I'm just very conscious of how I spend my time these days. :smiley:

plugin/ale.vim
```diff
@@ -186,6 +186,15 @@ let g:ale_deno_executable = get(g:, 'ale_deno_executable', 'deno')
 " If 1, enable a popup menu for commands.
 let g:ale_popup_menu_enabled = get(g:, 'ale_popup_menu_enabled', has('gui_running'))
 
+" If 1, disables ALE's built in error display. Instead, all errors are piped
+" to Neovim's diagnostics API.
+let g:ale_send_to_neovim_diagnostics = get(g:, 'ale_send_to_neovim_diagnostics', 0)
```
w0rp, line 191: We could name this `ale_send_diagnostics_to_neovim`. It might still sound a bit strange, because we're sending data via the NeoVim diagnostics API.

I think this variable name sound a bit odd, but I can't think of a better one, so don't change it.

plugin/ale.vim
```diff
@@ -186,6 +186,15 @@ let g:ale_deno_executable = get(g:, 'ale_deno_executable', 'deno')
...
+
+if g:ale_send_to_neovim_diagnostics && !has('nvim-0.6')
+    " no-custom-checks
+    echoerr('Cannot set g:ale_send_to_neovim_diagnostics to 1 unless you are running Neovim 0.6+.')
```
w0rp, line 195:
```suggestion
    echoerr('g:ale_send_to_neovim_diagnostics should not be set to 1 unless you are running Neovim 0.6+.')
```

I think this is easier to understand.

lua/diagnostics.lua
```diff
@@ -0,0 +1,46 @@
+local module = {}
```
w0rp, line 1: In this particular case I'm okay with having a little bit of Lua code, as we might as well. Typically I have a strong preference for nearly 100% Vim Script.

lua/diagnostics.lua
w0rp: I often use "max" and "min" functions in various places with the last line number, the end of a line, and so on. Often tools just report numbers that are wrong. Look out for "off-by-one" issues.

doc/ale.txt
```diff
@@ -1938,6 +1938,22 @@ g:ale_root                                                         *g:ale_root*
...
 
+g:ale_send_to_neovim_diagnostics             *g:ale_send_to_neovim_diagnostics*
+
+  Type: Number
+  Default: 0
+
+  If enabled, this option will disable ALE's standard UI, and instead send
+  all linter output to Neovim's diagnostics API. This allows you to collect
+  errors from nvim-lsp, ALE, and anything else that uses diagnostics all in
+  one place.
```
w0rp, line 1949: We need to list out all of the options here that will be disabled by virtue of turning this one one, with links to the documentation for each. That way people will know exactly what this does.

doc/ale.txt
```diff
@@ -1938,6 +1938,22 @@ g:ale_root                                                         *g:ale_root*
   LSP linter, it will not run.
 
 
+g:ale_send_to_neovim_diagnostics             *g:ale_send_to_neovim_diagnostics*
+
+  Type: Number
+  Default: 0
```
w0rp, line 1944:
```suggestion
  Type: |Number|
  Default: `0`
```

I prefer this style of formatting. I notice that some other documentation entries like the above one aren't quite right either.

autoload/ale/engine.vim
```diff
@@ -184,36 +184,42 @@ endfunction
...
+            endif
 
-        if g:ale_virtualtext_cursor
-            " Try and show the warning now.
-            " This will only do something meaningful if we're in normal mode.
-            call ale#virtualtext#ShowCursorWarning()
+            if g:ale_virtualtext_cursor
+                " Try and show the warning now.
+                " This will only do something meaningful if we're in normal mode.
+                call ale#virtualtext#ShowCursorWarning()
```
w0rp, line 221: I think we should keep the `g:ale_echo_cursor` warnings even if the Neovim API is enabled. That way you're sure to see problems by default on lines if something goes wrong with the rest of Neovim. All that does is show a message when your cursor gets to a line, and that might not be something that Neovim will do for you. That's my suggestion.

If people don't like it, they can always add in `let g:ale_echo_cursor = 0` in addition to the new setting. We need to document whatever behaviour we go with. I think `g:ale_virtualtext_cursor` ought to be ignored when sending diagnostics to NeoVim.

w0rp (approved):
This is a welcome addition. Nudge me again when you're able to update this. Thanks for this!

dbalatero:
Thanks for the review @w0rp! Appreciate the approval but it sounds like there are changes I need to make, so you might want to just mark it as Pending changes/comments until I've addressed everything.

bluz71:
> @bluz71 Since you were the original requester for this, would you be down to test this branch out on your end as well, and see if this config works for you?

@dbalatero,

I have had time to have a look at this impressive work.

Well done; especially since I am mulling switching back to ALE from null-ls (due to some quirks of null-ls).

Though I have encountered a bug with with the `mdl` Markdown linter and this PR.

My rough ALE + Neovim diagnostics setup:

```lua
local g = vim.g

g.ale_linters = {
  markdown = {'mdl'},
}

g.ale_completion_enabled = 0
g.ale_echo_cursor = 0
g.ale_hover_cursor = 0
g.ale_lint_on_enter = 0
g.ale_lint_on_filetype_changed = 0
g.ale_lint_on_insert_leave = 0
g.ale_lint_on_save = 1
g.ale_lint_on_text_changed = 'never'
g.ale_linters_explicit = 1
g.ale_send_to_neovim_diagnostics = 1

local map = vim.keymap.set
map("n", "]d", "<cmd>lua vim.diagnostic.goto_next()<CR>")
map("n", "[d", "<cmd>lua vim.diagnostic.goto_prev()<CR>")
```

Load up a Markdown files such as [this one](https://github.com/bluz71/vim-nightfly-guicolors/blob/master/README.md) and delete the black line at line 3, this should now produce a lint warning (correctly).

Now hit `]d` to go to the next warning. That produces this failure for me:

failure

I am using Neovim 0.7.0 (official release).

My quick testing with Standard Ruby and Standard JS linters (via this PR) does not produce the same issue.

So linter specific.

daliusd:
@dbalatero hey, how are you doing? I am using your branch for a month and its awesome. Do you need helping hand with fixing minor issues?

mikedfunk:
☹️

bluz71:
Can this be re-opened?

daliusd:
> @daliusd I'm still interested in getting this merged. Sorry to everyone for not being around. If you can make the adjustments I left in my review back in May, we can merge this into master. There wasn't all that much I wanted to change. I just haven't had the time to adjust it myself and test it in NeoVim. You might have a better idea as a regular NeoVim user. (I'm still using Vim until I am pretty sure NeoVim is tagged as 1.0.0 and there's a semi-official GUI client.)

@w0rp sorry, I have migrated to null-ls.nvim. I can't help with this.

mathstuf:
I'll try and take this over…I'm interested in it enough at least.

plugin/ale.vim
mathstuf: I think for future-proofing, renaming to `ale_use_diagnostics_api` is better in case Vim proper ever gains support (and not necessarily via Lua even).

plugin/ale.vim
mathstuf: I reworded to `Setting g:ale_use_diagnostics_api to 1 requires Neovim 0.6+.`

mathstuf:
See #4345 for my resurrection of this PR. Thanks for the base @daliusd!

dbalatero:
Hey @mathstuf thanks for taking this over–life took over on my end and I lost the bandwidth to get this one done. Should I close this PR in lieu of yours?

mathstuf:
Yeah, I think that'd be fine. I wish there was a way to get those Cc'd here Cc'd on a superceding issue/PR, but that's a long-standing feature request from my end here anyways :) .

dbalatero:
@mathstuf I think you can manually tag in the folks you think would be most interested on the new PR? You can also post back here when the other PR merges optionally, even if it's closed (I think?)

dbalatero:
Closing this one in lieu of #4345 - any interested watchers should hop over there.

Issue #4141, opened 2022-04-03
Message "Press ENTER or type command to continue" for hlint appears since b3d1d6eecfb5eb08787501294979dc6ba2426c35


RST-J:
Hello,
first, please let me now details I missed to add here, if the problem is not obviously caused by the suspected change.

I recently updated ale (by pulling master) and since then observed that errors reported by hlint behave differently than before.
Now, if hlint reports an error, neovim displays the "Press ENTER or type command to continue" message beneath the hlint output in the status bar(? not sure what its called exactly, sorry).
This repeats when I navigate left and right in the identified line and certain commands do not work unless I hit enter but then message immediately reappears.

I tracked this behavior down to be introduced (in my setup at least) by this commit: https://github.com/dense-analysis/ale/commit/b3d1d6eecfb5eb08787501294979dc6ba2426c35

As I do not now anything substantial about vim plugins, I cannot judge whether this is only due to this change or must have to do with my other plugins and or some config. Can someone tell me?

Edit: I am using neovim v0.7.0-dev from the PPA on Kubuntu 20.04.4. Also using vim-airline as a plugin (amongst 32 others which I'd less suspect of being a potential interference).

taq:
I can confirm that. Even moving left or right on a warning/error line fires the message and the ENTER key request. The ENTER request must be removed.

taq:
> Hello, first, please let me now details I missed to add here, if the problem is not obviously caused by the suspected change.
> 
> I recently updated ale (by pulling master) and since then observed that errors reported by hlint behave differently than before. Now, if hlint reports an error, neovim displays the "Press ENTER or type command to continue" message beneath the hlint output in the status bar(? not sure what its called exactly, sorry). This repeats when I navigate left and right in the identified line and certain commands do not work unless I hit enter but then message immediately reappears.
> 
> I tracked this behavior down to be introduced (in my setup at least) by this commit: [b3d1d6e](https://github.com/dense-analysis/ale/commit/b3d1d6eecfb5eb08787501294979dc6ba2426c35)
> 
> As I do not now anything substantial about vim plugins, I cannot judge whether this is only due to this change or must have to do with my other plugins and or some config. Can someone tell me?
> 
> Edit: I am using neovim v0.7.0-dev from the PPA on Kubuntu 20.04.4. Also using vim-airline as a plugin (amongst 32 others which I'd less suspect of being a potential interference).

Based on that commit, seems that there's some newline chars still present. I saw the 
```vimscript
[...]

PR #4144, opened 2022-04-05
Fix 4141 - Stop press enter prompt on long diagnostic messages

hsanson:
This fixes the recent issue that prompts users to press enter when the cursor is on a line with a long error diagsotic message (#4141).

This revert the change back in #3888 that replaced echom with echon to avoid poluting the `message-history`. Unfortunatelly echon does not seem to honor the "shortmess=+T" option that truncates messages and the manual trucation based on &column did not work properly as evident by how many people were affected with #4141.

Note I tried fixing the manual truncation in different ways to keep the echon but it always failed in some way. The only sure way was to revert to using echom with the shortmess=+T and let vim take care of the truncation.

Actually long ago there was some discussion on this: https://github.com/dense-analysis/ale/pull/1987#issuecomment-428864872

> Manually truncating the text is difficult because there's no way to accurately measure how wide text will be, given different fonts.

### DIFF
--- a/autoload/ale/cursor.vim
+++ b/autoload/ale/cursor.vim
@@ -11,9 +11,9 @@ let g:ale_echo_msg_format = get(g:, 'ale_echo_msg_format', '%code: %%s')
 let s:cursor_timer = -1
 
 " A wrapper for echon so we can test messages we echo in Vader tests.
-function! ale#cursor#Echon(message) abort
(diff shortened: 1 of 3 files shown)


hsanson:
Reference: https://vim.fandom.com/wiki/Get_shortened_messages_from_using_echomsg

PR #4145, opened 2022-04-05
Handle ghc panic in haskell

lykmast:
Primitive handling for **ghc: panic!** (compiler error): Just put everything on detail and location at start of file (1,1). 

It is not very sophisticated, but the current behaviour on **ghc: panic!** is identical to not having errors.


hsanson (approved):
Thanks for the addition.

PR #4146, opened 2022-04-05
Add the buffer-local options 'b:ale_shell' and 'b:ale_shell_arguments'.

jamescherti:
This pull request adds the following buffer-local options:
- `b:ale_shell`: the buffer-local version of `g:ale_shell`.
- `b:ale_shell_arguments`: the buffer-local version of `b:ale_shell_arguments`.


hsanson (approved):
Thanks, looks like a sensible addition to have.

PR #4157, opened 2022-04-15
Fallback to summary field if detail exists but is empty in terraform linter

Sh1ftry:
Terraform linter incorrectly chooses empty detail field for message when both summary and detail fields exist in `terraform validate` output, but detail is empty. Example problematic output from `terraform validate` command:
```
{
  "format_version": "1.0",
  "valid": false,
  "error_count": 1,
  "warning_count": 0,
  "diagnostics": [
    {
      "severity": "error",
      "summary": "registry.terraform.io/hashicorp/aws: there is no package for registry.terraform.io/hashicorp/aws 4.9.0 cached in .terraform/providers",
      "detail": ""
    }
  ]
}
```
This PR fixes this by checking if detail is not empty before returning it.

### DIFF
--- a/ale_linters/terraform/terraform.vim
+++ b/ale_linters/terraform/terraform.vim
@@ -21,7 +21,12 @@ function! ale_linters#terraform#terraform#GetType(severity) abort
 endfunction
 
 function! ale_linters#terraform#terraform#GetDetail(error) abort
-    return get(a:error, 'detail', get(a:error, 'summary', ''))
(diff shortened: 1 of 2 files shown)


hsanson:
There is a linting error. Adding this linter to your ale_linters configuration would make ALE warn you about these linting issues:

```
 'vim': ['vint', 'ale_custom_linting_rules'],
```

hsanson (approved):
Looks good, thanks.

Issue #3523, opened 2021-01-06
Getting `no code action received from server` from ALECodeAction


refaelsh:
## Information

**VIM version**

NVIM v0.4.4
Build type: Release

Operating System: Arch Linux

## What went wrong

No matter what/when/how, I always get `no code action received from server` from `ALECodeAction`.

### :ALEInfo

 Current Filetype: rust
Available Linters: ['analyzer', 'cargo', 'rls', 'rustc']
  Enabled Linters: ['analyzer']
  Ignored Linters: []
 Suggested Fixers: 
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'rustfmt' - Fix Rust files with Rustfmt.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

let g:ale_rust_analyzer_config = {}
let g:ale_rust_analyzer_executable = 'rust-analyzer'
let g:ale_rust_rustfmt_executable = 'rustfmt'
let g:ale_rust_rustfmt_options = ''
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = 100
let g:ale_completion_enabled = 1
let g:ale_completion_max_suggestions = 50
let g:ale_disable_lsp = 0
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 1
let g:ale_fixers = {'*': ['remove_trailing_lines', 'trim_whitespace'], 'yaml': ['prettier'], 'rust': ['rustfmt']}
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_filetype_changed = 1
let g:ale_lint_on_insert_leave = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'normal'
let g:ale_linter_aliases = {}
let g:ale_linters = {'vim': ['vimls'], 'rust': ['analyzer'], 'yaml': ['yamllint'], 'gitcommit': ['gitlint']}
[...]

PR #4161, opened 2022-04-21
Fixes for ALECodeAction and rust-analyzer

Benabik:
Two commits that make it possible to invoke the code actions for rust-analyzer.

The first is a quick fix that removes an error where ALE submits a column of -1 on an empty line, when rust-analyzer (and the specification) expects an unsigned integer.

The second advertises the [`codeAction.codeActionLiteralSupport`](https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction) capability during initialization.  Without this, [rust-analyzer never reports any code actions](https://github.com/rust-lang/rust-analyzer/blob/f83dce0a4a70e1690fdbd8123d787fbaa6f37970/crates/rust-analyzer/src/handlers.rs#L1048-L1053).  The code already handled processing the CodeAction literals, so no further changes were needed.  It's possible the `valueSet` key should have some actual [`CodeActionKind`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionKind) values, but the spec also says having the key at all  "the client guarantees that it will handle values outside its set gracefully".

### DIFF
--- a/autoload/ale/codefix.vim
+++ b/autoload/ale/codefix.vim
@@ -458,6 +458,7 @@ function! s:ExecuteGetCodeFix(linter, range, MenuCallback) abort
     endif
 
     let l:column = min([l:column, len(getline(l:line))])
+    let l:column = max([l:column, 1])
     let l:end_column = min([l:end_column, len(getline(l:end_line))])
 
     let l:Callback = function(
--- a/autoload/ale/lsp.vim
+++ b/autoload/ale/lsp.vim
@@ -443,6 +443,11 @@ function! s:SendInitMessage(conn) abort
     \               },
     \               'codeAction': {
     \                   'dynamicRegistration': v:false,
+    \                   'codeActionLiteralSupport': {
+    \						'codeActionKind': {
+    \							'valueSet': []
+    \						}
+    \					}
     \               },
     \               'rename': {
     \                   'dynamicRegistration': v:false,


Benabik:
Oh, this fixes #3523, which is where I started looking for solutions when I had the same problem.

stevenengler:
Have also tested this and it seems to be working.

Benabik:
@sak96 's PR #4163 is an improved version of one of my commits, but the empty line commit is still useful, IMHO.  Apologies for not having fixed the tests and lint myself, I honestly spent far more time than I had figuring out what was wrong and just wanted to put this out rather than let it sit.

sak96:
@Benabik 
on "empty line commit is still useful". I check that the pr did not break rust analyzer on empty lines.

Also as per ms lsp specs mentioned [here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position) `Special values like for example -1 to denote the end of a line are not supported.` which means -1 could be used by other LSP for special purposes.

Since ale is not specifically working only for rust analyzer I don't intend to break other LSP.

Benabik:
I'd argue the other way.  That is the protocol specification and it's specifically calling out that positions are always non-negative integers and that using negative numbers is outside of specification.  If there is a non-spec server out there ALE needs to support, then an exception should be made for it while giving correct requests to all the others.

sak96:
> Do code actions still work correctly without the `let l:column = max([l:column, 1])` patch? If not, then I don't think #4163 should be merged at all. (Also if -1 indicates end-of-line, then is `let l:column = max([l:column, 1])` the correct fix for this, rather than having -1 wrap around to the line length?)

code works fine @stevenengler. There is no issue when using empty lines.
Also this pr just implements `ALECodeAction`. I can raise a pr for the fix seperately.
That way pr can be merged individually. 

codefix



Can you reproduce the issue?

sak96:
@Benabik @stevenengler was using wrong branch.
Looks like the fix is required.

will raise a separate pr with proper stuff

hsanson:
These changes are already included in 	#4163

PR #4164, opened 2022-04-22
vscode-json-languageserver support

daliusd:
VSCode JSON languageserver has schema support for linting and
completions.

I have enabled snippets support (`snippetSupport`) even if it is not
fully supported. `label` that comes with completions response can be
used as well.

### DIFF
--- a/ale_linters/json/vscodejson.vim
+++ b/ale_linters/json/vscodejson.vim
@@ -1,0 +1,14 @@
+" Author: Dalius Dobravolskas <dalius.dobravolskas@gmail.com>
+" Description: VSCode json languageserver
+
+function! ale_linters#json#vscodejson#GetProjectRoot(buffer) abort
+    return fnamemodify(bufname(a:buffer), ':h')
+endfunction
+
+call ale#linter#Define('json', {
+\   'name': 'vscodejson',
+\   'lsp': 'stdio',
+\   'executable': 'json-languageserver',
+\   'command': '%e --stdio',
+\   'project_root': function('ale_linters#json#vscodejson#GetProjectRoot'),
+\})
--- a/ale_linters/json/vscodejson.vim
+++ b/ale_linters/json/vscodejson.vim
@@ -2,7 +2,9 @@
 " Description: VSCode json languageserver
 
 function! ale_linters#json#vscodejson#GetProjectRoot(buffer) abort
-    return fnamemodify(bufname(a:buffer), ':h')
+    let l:git_path = ale#path#FindNearestDirectory(a:buffer, '.git')
+
+    return !empty(l:git_path) ? fnamemodify(l:git_path, ':h:h') : ''
 endfunction
 
 call ale#linter#Define('json', {
--- a/ale_linters/json/vscodejson.vim
+++ b/ale_linters/json/vscodejson.vim
@@ -8,7 +8,7 @@ endfunction
 call ale#linter#Define('json', {
 \   'name': 'vscodejson',
 \   'lsp': 'stdio',
-\   'executable': 'json-languageserver',
+\   'executable': 'vscode-json-languageserver',
 \   'command': '%e --stdio',
 \   'project_root': function('ale_linters#json#vscodejson#GetProjectRoot'),
 \})
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -581,9 +581,14 @@ function! ale#completion#ParseLSPCompletions(response) abort
             continue
         endif
 
-        if get(l:item, 'insertTextFormat', s:LSP_INSERT_TEXT_FORMAT_PLAIN) is s:LSP_INSERT_TEXT_FORMAT_PLAIN
+        let l:insertTextFormat = get(l:item, 'insertTextFormat', s:LSP_INSERT_TEXT_FORMAT_PLAIN)
+
+        if l:insertTextFormat is s:LSP_INSERT_TEXT_FORMAT_PLAIN
         \&& type(get(l:item, 'textEdit')) is v:t_dict
             let l:text = l:item.textEdit.newText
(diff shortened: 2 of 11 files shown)


daliusd:
Please do not merge yet. I want to investigate some alternatives.

daliusd:
Now everything is good.

daliusd:
@hsanson it is not WIP anymore 😄 Unless you have some ideas what can be improved.

daliusd:
Some examples how it looks like in action:
image

image

Schema json `person.json`:

```json
{
  "$id": "https://example.com/person.schema.json",
  "title": "Person",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string",
      "description": "The person's first name."
    },
    "lastName": {
      "type": "string",
      "description": "The person's last name."
    },
    "age": {
      "description": "Age in years which must be equal to or greater than zero.",
      "type": "integer",
      "minimum": 0
    }
  }
}
```

Sample test file:
```json
{
  "$schema": "./person.json",
  "firstName": "John",
  "lastName": "Doe",
  "age": 2
}
```

autoload/ale/completion.vim
```diff
@@ -581,9 +581,14 @@ function! ale#completion#ParseLSPCompletions(response) abort
             continue
         endif
 
-        if get(l:item, 'insertTextFormat', s:LSP_INSERT_TEXT_FORMAT_PLAIN) is s:LSP_INSERT_TEXT_FORMAT_PLAIN
+        let l:insertTextFormat = get(l:item, 'insertTextFormat', s:LSP_INSERT_TEXT_FORMAT_PLAIN)
+
+        if l:insertTextFormat is s:LSP_INSERT_TEXT_FORMAT_PLAIN
         \&& type(get(l:item, 'textEdit')) is v:t_dict
             let l:text = l:item.textEdit.newText
+        elseif l:insertTextFormat is s:LSP_INSERT_TEXT_FORMAT_SNIPPET
```
daliusd, line 589: This is there we have fake snipped support. Still it is better than nothing as `label` can be used for auto-complete.

daliusd: It is safe to do this as below we have call to `ale#completion#Filter`.

doc/ale.txt
```diff
@@ -2951,6 +2952,7 @@ documented in additional help files.
     jq....................................|ale-json-jq|
     prettier..............................|ale-json-prettier|
     spectral..............................|ale-json-spectral|
+    vscodejson............................|ale-json-vscode|
```
hsanson, line 2955: Please add an entry for this new tool in the `ale-supported-languages-and-tools.txt` and `supported-tools.md` files.

ale_linters/json/vscodejson.vim
```diff
@@ -0,0 +1,14 @@
+" Author: Dalius Dobravolskas <dalius.dobravolskas@gmail.com>
+" Description: VSCode json languageserver
+
+function! ale_linters#json#vscodejson#GetProjectRoot(buffer) abort
```
hsanson, line 4: This is not good idea. This means that for every JSON file open in vim a new vscode-json-languageserver will be spawn.  Better look for generic project files such as the presence of a .git folder to set as project root.

autoload/ale/completion.vim
hsanson: Not well versed in language servers so will have to ask @w0rp to take a look here.

hsanson (changes requested):

ale_linters/json/vscodejson.vim
daliusd: OK. Changed to git root while I think opening it for every JSON is OK as well.

doc/ale.txt
daliusd: Done.

autoload/ale/completion.vim
daliusd: OK. Let's see what @w0rp has to say. It solves problem with this LSP but I'm pretty sure other LSPs will benefit accidentally without causing any accidental harm. Proper snippet support would be even better potentially, but that's really big task IMHO.

autoload/ale/lsp.vim
```diff
@@ -412,7 +412,7 @@ function! s:SendInitMessage(conn) abort
     \               'completion': {
     \                   'dynamicRegistration': v:false,
     \                   'completionItem': {
-    \                       'snippetSupport': v:false,
+    \                       'snippetSupport': v:true,
```
hsanson, line 415: Unfortunatelly I think this needs more evaluation.  I did some tests and when this is enabled ALEImport breaks for me.
If enablinb snippetSupport is not critical for vscodejson I would recommend reverting this to get vscodejson merged and then handle adding snippetSupport on a separate issue.

daliusd: Absolutely not critical. Let me fix it.

autoload/ale/completion.vim
daliusd: Removed from this PR. Let's do this as separate PR.

hsanson (approved):
Thanks, looks good.

autoload/ale/lsp.vim
daliusd: @hsanson BTW, how ALEImport breaks for you? What filetype you were workin on? I want to reproduce the problem and see if there is a fix.

hsanson: Simple Java HelloWorld.java managed via Gradle:

```
public class HelloWorld {
    public static void main(String[] argv) {
        LocalDate today = LocalDate.now();
        System.out.println("Hello gradle world at " + today);
    }
}
```

1. Put the cursos in LocalDate class and invoke :ALEImport.
2. ALE should add the import statement "import java.time.LocalDate" at the beginning of the file.
3. With the snippetSupport enabled this does not work.

PR #3554, opened 2021-01-22
WIP - Initial attempt migrate appveyor to Github Actions.

hsanson:
After #3548 it makes sense to also migrate Appveyor to Gitub Actions.

Closes: #3553

PR #4172, opened 2022-04-30
Fix windows github action

sak96:
Fix github action for windows

Follow up for: #3554

update vader.vim:  junegunn/vader.vim#30

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,6 +88,5 @@ jobs:
       - name: Run tests
         run: |
           cd C:\testplugin
-          C:\vim\vim\vim80\vim.exe -u test\vimrc `
-            "+Vader! test/*.vader test/*/*.vader test/*/*/*.vader test/*/*/*.vader"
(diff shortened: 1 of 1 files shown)


sak96:
@hsanson @w0rp 

i would need some help for fixing failing test cases.
do let me know where to look for. it is mostly tsserver

hsanson:
@sak96 thanks for the effort but we already tried to get this to work and failed. I do not have a single windows machine I can use to debug this so not sure I can help.

sak96:
quick question: https://github.com/dense-analysis/ale/blob/34892186e44b21356c84e1c2e019f96a304c3edc/.appveyor.yml#L48

why is `test/*/*/*.vader test/*/*/*.vader` twice ?

sak96:
ok i give up don't think it is easy without win machine.

w0rp:
I appreciate the effort. We will get it to work... some day. I just don't think it's a priority for me.

PR #4173, opened 2022-05-01
add yaml actionlint support (github actions)

bretello:
Adds support for github action linter [actionlint](https://github.com/rhysd/actionlint)

### DIFF
--- a/ale_linters/yaml/actionlint.vim
+++ b/ale_linters/yaml/actionlint.vim
@@ -1,0 +1,10 @@
+" Author: bretello <bretello@distruzione.org>
+
+call ale#Set('yaml_actionlint_executable', 'actionlint')
+
+call ale#linter#Define('yaml', {
+\   'name': 'actionlint',
+\   'executable': {b -> ale#Var(b, 'yaml_actionlint_executable')},
+\   'command': function('ale#handlers#actionlint#GetCommand'),
+\   'callback': 'ale#handlers#actionlint#Handle',
+\})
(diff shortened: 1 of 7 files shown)


hsanson (changes requested):
Please add entries for the new linter to `ale-supported-languages-and-tools.txt` and `supported-tools.md`. The list of tools must be listed in alphabetic order.

ale_linters/yaml/actionlint.vim
```diff
@@ -0,0 +1,11 @@
+" Author: bretello <bretello@distruzione.org>
+
+call ale#Set('yaml_actionlint_executable', 'actionlint')
+call ale#Set('yaml_actionlint_options', '--no-color --oneline')
```
hsanson, line 4: I believe these two options are required for the handler to work properly?  If users set different options the handler pattern can break and the linter won't work.  If this is the case then these two options should be set in the command no matter what yaml_actionlint_options is set to.

bretello: Got rid of actionlint options. For the record, the pattern can be customized calling actionlint with the following syntax:
```actionlint -format '{{range $err := .}}{{$err.Filepath}}{{$err.Line}}:{{$err.Column}}: [{{$err.Kind}}] {{$err.Message}}\n{{end}}'```

**Comment by hsanson** [comment]
Sorry, you also need to add the new linter entry to ale.txt section 7.6 in the same order as in the ale-supported-languages-and-tools.txt

**Comment by bretello** [comment]
Thanks! I wasn't sure why CI was failing. It should go through now.

Just as a note for whoever stumbling upon this, this is how I use this (to avoid having this lint all yaml files):
```vim
au BufRead,BufNewFile */.github/*/*.y.?ml let b:ale_linters = {'yaml': ['actionlint'] }
```

hsanson (approved):
Thanks, looking good.

PR #4174, opened 2022-05-01
Use snippets for completion.

daliusd:
Some LSPs returns label with snippet that can be used for completion.
This way we can extract at least some value from snippets even if full snippets
support is not available.

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -581,9 +581,14 @@ function! ale#completion#ParseLSPCompletions(response) abort
             continue
         endif
 
-        if get(l:item, 'insertTextFormat', s:LSP_INSERT_TEXT_FORMAT_PLAIN) is s:LSP_INSERT_TEXT_FORMAT_PLAIN
+        let l:insertTextFormat = get(l:item, 'insertTextFormat', s:LSP_INSERT_TEXT_FORMAT_PLAIN)
+
+        if l:insertTextFormat is s:LSP_INSERT_TEXT_FORMAT_PLAIN
         \&& type(get(l:item, 'textEdit')) is v:t_dict
             let l:text = l:item.textEdit.newText
+        elseif l:insertTextFormat is s:LSP_INSERT_TEXT_FORMAT_SNIPPET
+        \&& !empty(get(l:item, 'label'))
+            let l:text = l:item.label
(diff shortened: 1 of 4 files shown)


autoload/ale/completion.vim
```diff
@@ -581,9 +581,14 @@ function! ale#completion#ParseLSPCompletions(response) abort
...
         endif
 
-        if get(l:item, 'insertTextFormat', s:LSP_INSERT_TEXT_FORMAT_PLAIN) is s:LSP_INSERT_TEXT_FORMAT_PLAIN
+        let l:insertTextFormat = get(l:item, 'insertTextFormat', s:LSP_INSERT_TEXT_FORMAT_PLAIN)
+
+        if l:insertTextFormat is s:LSP_INSERT_TEXT_FORMAT_PLAIN
         \&& type(get(l:item, 'textEdit')) is v:t_dict
             let l:text = l:item.textEdit.newText
+        elseif l:insertTextFormat is s:LSP_INSERT_TEXT_FORMAT_SNIPPET
+        \&& !empty(get(l:item, 'label'))
```
daliusd, line 590: @hsanson Previously it was `!empty(get(l:item.label))` and I believe this is the reason why it was failing for you. I do not have Java on any machine so I can't test it locally but I tested this with some other languages and it seems everything is good with `ALEImport`.

daliusd:
@w0rp this might require your attention as @hsanson says you know LSPs better than he.

autoload/ale/completion.vim
hsanson: I have been trying this out without success.  My suspicious is that once snippets are enabled the eclipselsp starts returning auto-completions and imports as snippets (e.g. insertTextFormat: 2) instead of plain (e.g. insertTextFormat: 1) and since ALE does not fully support snippets then these features stop working.

I am afraid this may also affect other LSP servers so two options:

1. Fully support snippets.
2. Allow enabling snippets only on certain LSPs.

I would prefer option 1 but I am not familiar with how mouch work that involves.

daliusd: OK. Let it be then - that's not critical for me. I am closing this PR now. Adding full support for snippets might be out of scope of Ale even (e.g. it should be combined with some snippet plugin).

PR #4175, opened 2022-05-02
VSCode LSPs for CSS, HTML and JSON

daliusd:
Switched to `vscode-langservers-extracted` as it is the one most
up-to-date.


hsanson (approved):
Thanks for the quick update.

PR #4181, opened 2022-05-03
fix docs for actionlint

bretello:
Update docs with correct `aupat` for actionlint

related: https://github.com/dense-analysis/ale/pull/4173


hsanson (approved):
Thanks for the improvements.

Issue #4168, opened 2022-04-28
Show warnings as warnings in golangci_lint


jonasrichard:
golangci_lint cannot really distinguish if a lint message is a compilation error or it is a warning, or I think one way maybe:

When a file has syntax errors, the last part of the message in parenthesis will be `typecheck` [see it here](https://golangci-lint.run/usage/linters/)

```
$ golangci-lint run main.go
main.go:3:1: expected declaration, found mport (typecheck)
mport (
^
```

Although if there is no compilation error, other sublinters kick in and those should be shown as warnings.
```
$ golangci-lint run main.go
main.go:25:6: `publish` is unused (deadcode)
func publish(port int) {
     ^
```

I think in the message collection part we can match the last word of the linter message in the parentheses, and if it is not `typecheck` we can append them as `'W'` warning.

https://github.com/dense-analysis/ale/blob/6c51bb1573f0bf5deff04508769208fd503b5ff7/ale_linters/go/golangci_lint.vim#L43

I don't know the other linters, but one-by-one those also can be improved.

hsanson:
Looks good...  you can make a PR with these changes.  Make sure to update and add new tests to `test/handler/test_gometalinter_handler.vader`.

I do not develop frequently in Golang but I would consider also marking typecheck, staticcheck, errcheck, and govet as Errors since their preset is "bug" and these are enabled by default.

hsanson:
Handled by #4182

PR #4182, opened 2022-05-03
Handle golangci_lint warning and error messages correctly

jonasrichard:
Improve golangci_lint plugin in order that it shows the error messages only if the message comes from the typecheck linter.

See the issue for details. https://github.com/dense-analysis/ale/issues/4168


hsanson (approved):
Looks good... thanks for the contribution.

hsanson (changes requested):
There seem to be some linter issues:

  - ale_linters/go/golangci_lint.vim:37 Use 'is#' instead of '==#'. 0 ==# 'foobar' is true

Make sure to enable vim linters so vim tells you if there are issues:

```
let g:ale_linters = {
  \   'vim': ['vint', 'ale_custom_linting_rules'],
\}
```

hsanson (approved):
Thanks for the contribution.

PR #4183, opened 2022-05-03
Use stdin for Selene Lua linter

deathlyfrantic:
When I added Selene as a linter, I wasn't aware it had the option to process stdin using `-`.


hsanson (approved):
Thanks. Looks good.

Issue #3870, opened 2021-08-10
Racket `omni_start_map` should be something like `\k\+`


benknoble:
Racket, being a lisp, has far more than just alpha-numeric identifiers, which is what the default pattern matches. This means when I type `file-name`, invoking completion won't give `file-name-from-path` but completions like `namespace`—it has used only the `name` part. Using `\k\+` seemed to work in cases like this, but I don't have an exhaustive test.

More broadly, and the reason this is an issue and not a PR, is I think the default should be `\k\+`—then it should be effectively scoped to the languages `iskeyword` setting, no?

I also don't understand why `$` is included in the `omni_start_map` pattern: does that only allow completions when there is nothing else on the line? Perhaps word boundaries `\<\>` would be more appropriate?

Looking for feedback here.

---

PS Why not use the lisp pattern? Well, I don't think even it covers all possible lisp or racket identifiers, and `\k\+` is so much easier to read.

w0rp:
I don't know what any of this is, but if you can create a pull request that fixes this without breaking any current functionality in languages like TypeScript, go for it.

benknoble:
I am referring to `s:omni_start_map` in `autoload/ale/completion.vim` (apologies for not being clear). The declaration is around line 139, with sole use at line 311.

It seems like the intent was to allow filetypes to specify patterns to find the start of a word for completion, with a default of
```
\v[a-zA-Z$_][a-zA-Z$_0-9]*$
```

Unfortunately, this doesn't work for some languages. It explicitly disallows `-` in words that should start completion, for example; every lisp I've ever used has `-` in names commonly (some examples are in the original post). This means that trying to complete from a word that already has `-` in it breaks.

My suggestion is to use `\k\+` (possibly with word boundaries)—then the default should work for all filetypes by leveraging their custom `iskeyword` settings (which _define_ in vim what a word is for a filetype). If a filetype has this correctly set (and, _e.g._, racket, scheme, and common lisp should), then it should "just work."

If that's no good, then I'd like to at least add `'racket': '\k\+'` or similar—the issue is that racket is actually a lot of languages, so there are lots of possible filetypes… and other lisps suffer, too, so it would be nice to add them… etc. So then you have this mechanism that controls completion (and breaks it, if not set right) but that is really hard to customize. (By really hard, I mean that the only way is to (a) edit your local copy of ALE and be careful not to lose that when updating or (b) submit a PR and wait for it to merge. That seems untenable in the long-term.)

[...]

PR #4186, opened 2022-05-05
racket: start completion in a full keyword

benknoble:
The default `omni_start_map` is too restrictive for Lisps and Schemes
like Racket, which permit hyphens (among other special characters).

As recorded in #3870, trying to complete `file-name-from-path` when
typing `file-name<C-x><C-o>` would give completions like `namespace`
because the hyphen is ignored to find the start of the word for
completion.

Now the racket filetype searches for the start using the keyword class
`\k`, which is more precise.


hsanson (approved):
Sorry for the delay.  After studying a little about  "iskeyword" and doing some tests on my setup I think this is an good improvement.  Tested with the languages I use daily (Java, Python, Ruby) and nothing breaks.

benknoble:
Thanks!

PR #4187, opened 2022-05-05
omni-compl: start completion in a full keyword

benknoble:
The default `omni_start_map` is too restrictive for Lisps and Schemes
like Racket, which permit hyphens (among other special characters).

As recorded in #3870, trying to complete `file-name-from-path` when
typing `file-name<C-x><C-o>` would give completions like `namespace`
because the hyphen is ignored to find the start of the word for
completion.

Now the default searches for the start using the keyword class
`\k`, which is more precise and configurable for each filetype without
modifying the source.


hsanson:
Closed in favor of #4186

Issue #2531, opened 2019-05-24
Support Rego 


pbnj:
**Name:** Rego
**URL:** https://www.openpolicyagent.org/docs/latest/how-do-i-write-policies/

Rego is a declarative language, inspired by [Datalog](https://en.wikipedia.org/wiki/Datalog), used in [Open-Policy-Agent](https://www.openpolicyagent.org/) Kubernetes project.

The vim syntax is provided via [vim-rego](https://github.com/tsandall/vim-rego).
The `opa` CLI itself provides a [`fmt`](https://github.com/open-policy-agent/opa/blob/fd86f74231f89ecb796cf6cd89d54400b6eb7bcf/cmd/fmt.go#L28) command for formatting Rego source files

redbmk:
Looks like there's support for rego now with `opa check` and `opa fmt`. Although I'm seeing an issue where if if I run `opa check .` there are no issues but checking an individual file shows issues (e.g. a variable defined in another file).

Seems like ALE would need to check a file in the context of surrounding files somehow?

e.g. instead of `opa check % --format json` it could `open check . --format json` and then only report errors that match that filename. or better than `.` might be to somehow find the root

nospam2998:
What is this issue about? To my eyes it reads as is about adding a new linter. Yet it appears that was done in #4199.

There are comments added after the merge of that pull request. They appear to indicate there are bugs, but that they can be worked around? I'm not really getting all the details.

Should this issue be re-titled and have its purpose clarified, or would it be time to close it given that the May 2022 merge actually delivered everything needed?

redbmk:
Yeah it does feel like the May '22 PR should have closed this.

nospam2998:
Thanks @redbmk !

Closing, as I understood this is confirmed to have been fixed by #4199.

PR #4199, opened 2022-05-12
add rego support

rzjfr:
Where are the tests? Have you added tests? Have you updated the tests? 

```
test/fixers/test_opa_fmt_fixer_callback.vader
test/linter/test_rego_opacheck.vader

```

Related issue:

https://github.com/dense-analysis/ale/issues/2531

### DIFF
--- a/doc/ale.txt
+++ b/doc/ale.txt
@@ -3127,8 +3127,8 @@ documented in additional help files.
     reason-language-server................|ale-reasonml-language-server|
     refmt.................................|ale-reasonml-refmt|
   rego....................................|ale-rego-options|
-    opafmt................................|ale-rego-opa-fmt-fixer|
     opacheck..............................|ale-rego-opa-check|
+    opafmt................................|ale-rego-opa-fmt-fixer|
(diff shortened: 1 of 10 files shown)


doc/ale.txt
```diff
@@ -3126,6 +3126,9 @@ documented in additional help files.
     ols...................................|ale-reasonml-ols|
     reason-language-server................|ale-reasonml-language-server|
     refmt.................................|ale-reasonml-refmt|
+  rego....................................|ale-rego-options|
```
hsanson, line 3129: Since you added cpell support would be nice to have cpell added here. Same for the other index documents.

hsanson (changes requested):
Looks good and complete. Just missing cspell support in the list of supported tools for rego file.

doc/ale.txt
rzjfr: @hsanson thanks for reviewing this PR, good catch, I've added the required docs now.

hsanson (approved):
Thanks for the prompt fixes.

PR #4202, opened 2022-05-14
Look for .mypy.ini when finding project root

fionn:
We already check for mypy.ini, but the fallback .mypy.ini was ignored.


hsanson (approved):
Thanks for the fix.

PR #4204, opened 2022-05-15
fixes cpplint url in supported-tools.md

krishnakumarg1984:


hsanson (approved):
Thanks, these documentation fixes are always appreciated. Impossible for one person to keep track of all these tools.

PR #4215, opened 2022-05-25
Allow customization of all floating window borders

djpohly:
Users may not necessarily want the same border character for top+bottom or left+right, so allow all eight border characters to be configured in `g:ale_floating_window_border`.

For backwards compatibility, the old rules are still applied if only six elements are given.

### DIFF
--- a/autoload/ale/floating_preview.vim
+++ b/autoload/ale/floating_preview.vim
@@ -106,18 +106,31 @@ function! s:NvimPrepareWindowContent(lines) abort
     let l:width += 2
     let l:height += 2
 
-    let l:hor          = g:ale_floating_window_border[0]
-    let l:top          = g:ale_floating_window_border[1]
-    let l:top_left     = g:ale_floating_window_border[2]
-    let l:top_right    = g:ale_floating_window_border[3]
-    let l:bottom_right = g:ale_floating_window_border[4]
-    let l:bottom_left  = g:ale_floating_window_border[5]
+    if len(g:ale_floating_window_border) == 6
+        let l:left         = g:ale_floating_window_border[0]
+        let l:right        = l:left
+        let l:top          = g:ale_floating_window_border[1]
+        let l:bottom       = l:top
+        let l:top_left     = g:ale_floating_window_border[2]
+        let l:top_right    = g:ale_floating_window_border[3]
+        let l:bottom_right = g:ale_floating_window_border[4]
+        let l:bottom_left  = g:ale_floating_window_border[5]
+    else
+        let l:top          = g:ale_floating_window_border[0]
+        let l:right        = g:ale_floating_window_border[1]
+        let l:bottom       = g:ale_floating_window_border[2]
+        let l:left         = g:ale_floating_window_border[3]
+        let l:top_left     = g:ale_floating_window_border[4]
+        let l:top_right    = g:ale_floating_window_border[5]
+        let l:bottom_right = g:ale_floating_window_border[6]
+        let l:bottom_left  = g:ale_floating_window_border[7]
+    endif
 
     let l:lines = [l:top_left . repeat(l:top, l:width - 2) . l:top_right]
 
     for l:line in a:lines
         let l:line_width = strchars(l:line)
-        let l:lines = add(l:lines, l:hor . l:line . repeat(' ', l:width - l:line_width - 2). l:hor)
+        let l:lines = add(l:lines, l:left . l:line . repeat(' ', l:width - l:line_width - 2). l:right)
     endfor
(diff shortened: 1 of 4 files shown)


djpohly:
As I made this, I wondered if it would be a even better solution to allow the user to specify a `g:ale_floating_window_options` dictionary that is merged into the options for `popup_create()`. I'd be willing to code that up if it seems better to the devs.

autoload/ale/floating_preview.vim
```diff
@@ -106,26 +106,39 @@ function! s:NvimPrepareWindowContent(lines) abort
...
     let l:height += 2
 
-    let l:hor          = g:ale_floating_window_border[0]
-    let l:top          = g:ale_floating_window_border[1]
-    let l:top_left     = g:ale_floating_window_border[2]
-    let l:top_right    = g:ale_floating_window_border[3]
-    let l:bottom_right = g:ale_floating_window_border[4]
-    let l:bottom_left  = g:ale_floating_window_border[5]
+    if len(g:ale_floating_window_border) == 6
+        let l:left         = g:ale_floating_window_border[0]
```
hsanson, line 110: I would add the new boder options to the end of the array so the first 6 entries stay compatible without any change and then modify this code to use safer array indexing:

```
let l:left                      = get(g:ale_floating_window_border, 0, '|')
let l:top                      = get(g:ale_floating_window_border, 1, '-')
let l:top_left               = get(g:ale_floating_window_border, 2, '+')
let l:top_right             = get(g:ale_floating_window_border, 3, '+')
let l:bottom_left         = get(g:ale_floating_window_border, 4, '+')
let l:bottom_right       = get(g:ale_floating_window_border, 5, '+')
let l:right                    = get(g:ale_floating_window_border, 5, l:left)
let l:bottom                = get(g:ale_floating_window_border, 5, l:bottom)
```

This not only simplies logic it also makes the code safer in case someone sets the g:ale_floating_window_border array wrong.

hsanson:
The current array  configuration is good enough I think. If we allow users to set all possible options to popup_create we must be careful dealing with differences between vim popup_create and neovim nvim_open_win functions.

djpohly:
I see - does ALE intentionally try to abstract away any differences between the two?  (Unfortunately Neovim decided to put the array in a different order from Vim when they implemented borders... do they do this to us on purpose?)

I was thinking ahead to ways I'd like to be able to customize the appearance of ALE popups, and it seemed like the option names could get unwieldy (`g:ale_floating_window_border_highlight_groups`??).  Would it make sense to group these into a dictionary if it's not given directly to either application's window function but rather translated by ALE itself to be cross-"platform"?

hsanson:
> ALE intentionally try to abstract away any differences between the two?

Not really but seems natural that ALE functions seamessly on both Vim and Neovim with the same configuration.

> do they do this to us on purpose?

Cannot say but I envision the situation will get even worse as Neovim moves to Lua and features start diverging more.

> Would it make sense to group these into a dictionary if it's not given directly to either application's window function but rather translated by ALE itself to be cross-"platform"?

Yes, makes sense to me at least.  If we are going to be adding more options to customize the look of windows then having 10 different arrays would be confusing and error prone to users.

>  I'm just interested in making the ALE popups look... better.

Great!!, any contributions are welcome.  I did some thinkering myself and my conclusion was that eyecandy is highly subjective (everyone has different preference) and it falls outside ALE's main scope.

Instead of ALE implementing eyecandy features that people can customize to their liking I think is better if ALE provides ways to integrate with other plugins that are specifically created for this like [nvim-notify](https://github.com/rcarriga/nvim-notify), and [fidget](https://github.com/j-hui/fidget.nvim).

ALE can provide bare minimum visual tools like current floating windows, preview, echo, virtual-text,  and add some callbacks that allow users to plug their own window rendering. Similar to how integration with other auto-completion plugins work currently.

Of course does not have to be a plugin. A user that knows how to use popup_create() and nvim_create_win() functions can override the callback to customize the windows to their liking.

hsanson (approved):
Thanks for the changes. Looks good.

djpohly:
> Of course does not have to be a plugin. A user that knows how to use popup_create() and nvim_create_win() functions can override the callback to customize the windows to their liking.

Thanks - I didn't even realize ALE had customizable callbacks for displaying info!  I haven't found anything in the README, but I'll dig through ale.txt and the source if necessary to see what I can learn. If you know offhand of a better explanation somewhere, let me know where to look!

hsanson:
Sorry, what I meant was that we should implement those callbacks instead of adding options for customization.  These callbacks still do not exist in ALE.

Issue #4226, opened 2022-06-08
Autocomplete of struct fields with Clangd is broken.


s-marios:
## Information

**VIM version**
:version
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Apr 21 2022 00:23:54)
Included patches: 1-4793

.vimrc file [here](https://github.com/s-marios/dotfiles/blob/master/vim/.vimrc)

Operating System: Debian Testing
Additional Info: clangd-13 (Debian clangd version 13.0.1-4)
ALE: commit f6d7aa4f0f396f0ed2d58a4ee5d0107322368701

## What went wrong

Auto-completion for fields of basic structs in C with clangd and using the omnifunc is broken.  Writing out the name of the struct and then writing "." (or "->" in case of a pointer) brings up no further suggestions.

## Reproducing the bug
I've setup a simple repository [here](https://github.com/s-marios/broken_autocomplete) which demonstrates the problem. 
First, compile the project in order to generate compile_commands.json:
```
mkdir build && cd build && cmake .. && cmake --build ./
```

Then, at the end of function main, start inserting the following:
```
p->
```
I would expect the x and y fields declared in module.h to appear. Although omnifunc gets engaged there are no suggestions. Similarly, when inserting '.' after a struct, nothing happens.

Grudgingly, I opened the same project with vscode, installed clangd and it worked flawlessly.  

Can you reproduce this behavior? If yes, what are the next steps towards fixing this?
Any help greatly appreciated!

-----

### :ALEInfo
```
 Current Filetype: c
Available Linters: ['cc', 'ccls', 'clangd', 'clangtidy', 'cppcheck', 'cpplint', 'cquery', 'cspell', 'flawfinder']
   Linter Aliases:
'cc' -> ['gcc', 'clang']
  Enabled Linters: ['cc', 'clangd', 'cppcheck', 'flawfinder']
  Ignored Linters: []
 Suggested Fixers: 
  'astyle' - Fix C/C++ with astyle.
  'clang-format' - Fix C/C++ and cuda files with clang-format.
  'clangtidy' - Fix C/C++ and ObjectiveC files with clang-tidy.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'uncrustify' - Fix C, C++, C#, ObjectiveC, ObjectiveC++, D, Java, Pawn, and VALA files with uncrustify.
 Linter Variables:

[...]

PR #4231, opened 2022-06-12
Improve struct and pointer autocompletion in C

s-marios:
This pull request addresses #4226 and comes in two patches:

1. add explicit trigger characters for triggering autocompletion in C,
2. stop the previous autocompletion operation when a new autocompletion request is sent to the LSP.

For the first patch, ALE currently does not use the trigger characters advertised by the LSP. Thus it is necessary to hard-code trigger characters. This patch introduces ```->``` in order to make autocompletion with pointers possible.

The second patch is a proof-of-concept on how ALE should behave when a newer autocompletion request has been sent to the server. Currently, if there was a previous autocomplete request, the autocomplete operation will *not* stop if the user enters the full name of a variable and then a trigger character, which results in no completion candidates *even though* ALE did talk to the LSP and retrieved completion candidates successfully (see #4226 for this description). This patch stops the previous completion operation, and ALE is able to show a new autocompletion menu with the updated, more recent completion candidates. 

Review and further comments on getting this merged are greatly appreciated!


hsanson:
I tested this with the sample code in the issue and auto-completion works fine with the chages to fix "1.".  The changes done for "2." break auto-completion and from my testing are not required. Note that I use asynccomplete for auto-completion instead of ALE's built in auto-completion.

s-marios:
@hsanson unfortunate to hear that patch 2 breaks autocompletion with asynccomplete. I do not know enough about how it interacts with ALE. Can you try with ale's built-in autocompletion function?

One more question, I tried to look into the appveyor results, but for the life of me I couldn't find which tests broke. What should I be looking for?

autoload/ale/completion.vim
```diff
@@ -998,6 +1001,10 @@ function! ale#completion#Queue() abort
 
     call ale#completion#StopTimer()
 
+    if !empty(complete_info(['mode']))
```
hsanson, line 1004: This line does not work at all.  That complete_info method does not exists. See CI errors:

```
Starting Vader: C:\testplugin\test\completion\test_completion_events.vader
    ( 1/29) [EXECUTE] ale#completion#GetCompletions should be called when the cursor position stays the same
    ( 1/29) [EXECUTE] (X) Vim(if):E117: Unknown function: complete_info
      > function CheckCompletionCalled[12]..ale#completion#Queue, line 21
    ( 2/29) [EXECUTE] ale#completion#GetCompletions should not be called if the global setting is disabled
    ( 3/29) [EXECUTE] ale#completion#GetCompletions should not be called if the buffer setting is disabled
    ( 4/29) [  GIVEN]
    ( 4/29) [EXECUTE] ale#completion#GetCompletions should not be called when the cursor position changes
    ( 4/29) [EXECUTE] (X) Vim(if):E117: Unknown function: complete_info
      > function ale#completion#Queue, line 21
    ( 5/29) [  GIVEN]
    ( 5/29) [EXECUTE] ale#completion#GetCompletions should not be called if you switch to normal mode
    ( 5/29) [EXECUTE] (X) Vim(if):E117: Unknown function: complete_info
      > function ale#completion#Queue, line 21
    ( 6/29) [  GIVEN]
    ( 6/29) [EXECUTE] Completion should not be done shortly after the CompleteDone function
    ( 6/29) [EXECUTE] (X) Vim(if):E117: Unknown function: complete_info
      > function CheckCompletionCalled[12]..ale#completion#Queue, line 21
    ( 7/29) [  GIVEN]
```

s-marios: @hsanson I ...am at a loss here? 

:help complete_info on VIM 8.2.4793 (this is whatever debian packages) comes up with the proper builtin function documentation. Only disclaimer, I have no idea when this function was introduced.

I will have to look into and understand the testing process for ALE, but in the mean time, what versions of VIM do these tests run against?

hsanson: This function does not exists in neovim 0.2 and vim 8.0 that are still supported by ALE.

We would need to figure out a way to make this work on all supported versions. Like creating a function that checks for the existence of that method before using it:

```
function ale#completion#Info(what)
    if exists('*complete_info')
        return complete_info(a:what)
    else
       // Some code that does same as complete_info 
       // in older version of vim and neovim
    endif
endfunction
```

s-marios: That's unfortunate. I'll have a look into vim 8 and see if there's an alternative way to achieve the same thing, but it's not urgent for me. I can just keep rebasing my fork. Thanks for the pointers!

s-marios: Spun up a couple of containers and looked into vim 8.0 and 8.1. Surprisingly to me, ```complete_info()``` and CTRL^X CTRL^Z (to exit autocompletion) are only available to vim 8.2 (did not check neovim at all).

An alternative would be to check ```mode()``` and if it is 'i' then issue CTRL^E to stop completion. I'll be looking into that a bit later.

In general, what's the stance of this project when it comes to using features  introduced in later versions? Is this if-else pattern common? Or for example completely dropping the else part when the functionality cannot be reproduced in earlier versions?

hsanson: If omiting the else does not break anything then I would say is ok to omit. In worst case something will work better on newer versions of vim/neovim and nothing happens on older versions.

s-marios: So, I did implement the conditional logic, looking specifically to see if ```complete_info()``` is available, and if it's not, it'll be a no-op. 

As far as I can tell, we'll get the desired functionality in vim8.2 and no changes in older versions.

As a side-note, I did try using ```mode()``` and CTRL^E for older versions of vim, and it *almost* worked: autocompletion behaved nicely in vim8.0 *but* it had side effects when I was editing other stuff (text appeared when editing comments, etc).

autoload/ale/completion.vim
```diff
@@ -148,6 +150,7 @@ let s:trigger_character_map = {
 \   'typescript': ['.', '''', '"'],
 \   'rust': ['.', '::'],
 \   'cpp': ['.', '::', '->'],
+\   'c': ['.', '->'],
```
hsanson, line 153: I think would be a good idea to honor the comment and update the ale.py accordingly.

s-marios: Done. Added it as part of the first patch

autoload/ale/completion.vim
```diff
@@ -999,6 +1008,10 @@ function! ale#completion#Queue() abort
 
     call ale#completion#StopTimer()
 
+    " Close the previous completion menu (if any), so that the newer autocompletion
+    " candidates will show up
+    call s:closePopupMenu()
```
hsanson, line 1013: I did some tests and the patch works fine (at least does not breaks anything) but I think this can be simplified.

You can call  `s:closePreviousCompletionMenu()` directly here instead of `s:closePopupMenu()` and get the same result. I tested this and works fine.  You can move the comment to the s:closePreviousCompletionMenu().

With this change you can completely remove `s:closePopupMenu()` and `s:SetupClosePopupMenu()` functions simplifying the code.

s-marios: > You can call `s:closePreviousCompletionMenu()` directly here instead of `s:closePopupMenu()` and get the same result. I tested this and works fine. 

Unfortunately, this is not the case. I just tried this on vim8.0 and, despite the abort present in the definition of ```s:closePreviousCompletionMenu()```, vim still barfs 4 errors complaining that ```complete_info``` does not exist etc. This is the same problem that the first version of the patch had.

> I did some tests and the patch works fine (at least does not breaks anything) but I think this can be simplified.
>
>  <snip...>
> You can move the comment to the s:closePreviousCompletionMenu(). 
>
>  <snip...>
> With this change you can completely remove `s:closePopupMenu()` and `s:SetupClosePopupMenu()` functions simplifying the code.

In order to remove `s:SetupClosePopupMenu()` and simplify things, we'll have to pull the `exists('*complete_info')` check inside `s:closePreviousCompletionMenu()`, meaning it'll be performed every time autocompletion takes place instead of only once during setup. I won't argue about performance overheads (esp. without hard numbers in hand) as I can see the value of simpler code.

I'll push an additional commit with the simplified code in this branch for review purposes, with the intention to squash it just before merge.

s-marios: Added commit, rebased on latest master.

hsanson (dismissed):
Thanks for the work. Looks good and tested works on my setup.

hsanson:
@s-marios seems that the new patterns need to be also added to this test: `test/python/test_deoplete_source.py`.

s-marios:
Ah, sorry about that one. Amended the test, history rewritten and rebased.

PR #4237, opened 2022-06-22
Show infos when using `dart_analyze` linter

grlduarte:
Currently the infos are not showed by ALE when working with Dart files.

- Changed the linter command to use the argument `--fatal-infos` so that when there's one info it triggers ALE,
- Updated the test routine.

Code snippet for testing:

```dart
// main.dart
class A {
  const A();
}

void accessA() {
  A a = new A();
}
```


test/handler/test_dart_analyze_handler.vader
```diff
@@ -9,20 +9,20 @@ Execute(Basic problems should be parsed correctly):
...
-  \     'col': 16,
+  \     'type': 'I',
+  \     'text': 'prefer_const_constructors: Prefer const with constant constructors.',
+  \     'lnum': 9,
+  \     'col': 10,
   \   },
   \ ],
   \ ale_linters#dart#dart_analyze#Handle(bufnr(''), [
   \ 'Analyzing main.dart...',
-  \ '  error - main.dart:5:1 - Expected to find ''}'' - expected_token',
```
hsanson, line 25: I think these changes should not affect the already exising tests. Please keep the previous tests and add the new ones.

hsanson (changes requested):

hsanson (approved):
Looks good, thanks.

PR #4247, opened 2022-07-08
Allow callbacks for floating preview popups

shaunduncan:
I'm not sure if this PR will be useful to anyone else, but I wanted a better way to customize how the floating preview popup was being displayed. The simplest way I could think to do this was to just allow a callback function that could supply custom popup options.

I have added some documentation for this and I am currently using it with vim 8.2. I haven't tried it with neovim but I will assume it doesn't work (I don't use neovim so I can't say for sure).

Just wanted to share :)

### DIFF
--- a/autoload/ale/floating_preview.vim
+++ b/autoload/ale/floating_preview.vim
@@ -7,7 +7,8 @@
 
 function! ale#floating_preview#Show(lines, ...) abort
     if !exists('*nvim_open_win') && !has('popupwin')
-        execute 'echom ''Floating windows not supported in this vim instance.'''
+        " no-custom-checks
+        echom 'Floating windows not supported in this vim instance.'
 
         return
     endif
@@ -105,18 +106,20 @@ function! s:NvimPrepareWindowContent(lines) abort
     let l:width += 2
     let l:height += 2
 
-    let l:hor          = g:ale_floating_window_border[0]
-    let l:top          = g:ale_floating_window_border[1]
-    let l:top_left     = g:ale_floating_window_border[2]
-    let l:top_right    = g:ale_floating_window_border[3]
-    let l:bottom_right = g:ale_floating_window_border[4]
-    let l:bottom_left  = g:ale_floating_window_border[5]
+    let l:left         = get(g:ale_floating_window_border, 0, '|')
+    let l:top          = get(g:ale_floating_window_border, 1, '-')
+    let l:top_left     = get(g:ale_floating_window_border, 2, '+')
+    let l:top_right    = get(g:ale_floating_window_border, 3, '+')
+    let l:bottom_right = get(g:ale_floating_window_border, 4, '+')
+    let l:bottom_left  = get(g:ale_floating_window_border, 5, '+')
+    let l:right        = get(g:ale_floating_window_border, 6, l:left)
+    let l:bottom       = get(g:ale_floating_window_border, 7, l:top)
 
     let l:lines = [l:top_left . repeat(l:top, l:width - 2) . l:top_right]
 
     for l:line in a:lines
         let l:line_width = strchars(l:line)
-        let l:lines = add(l:lines, l:hor . l:line . repeat(' ', l:width - l:line_width - 2). l:hor)
+        let l:lines = add(l:lines, l:left . l:line . repeat(' ', l:width - l:line_width - 2). l:right)
     endfor
 
     " Truncate the lines
@@ -124,7 +127,7 @@ function! s:NvimPrepareWindowContent(lines) abort
         let l:lines = l:lines[0:l:max_height]
     endif
 
-    let l:lines = add(l:lines, l:bottom_left . repeat(l:top, l:width - 2) . l:bottom_right)
+    let l:lines = add(l:lines, l:bottom_left . repeat(l:bottom, l:width - 2) . l:bottom_right)
 
     return [l:lines, l:width, l:height]
 endfunction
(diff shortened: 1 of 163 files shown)


michaelorr:
Yesssss. I love the floating window but having it in the upper right corner (for example) is way better than covering up the context around the issue. This allows for reasonable defaults but also full customization, love it

hsanson:
This sounds very similar to what #4221 wants to achieve.

autoload/ale/floating_preview.vim
```diff
@@ -170,7 +172,14 @@ function! s:VimCreate(options) abort
     \        get(g:ale_floating_window_border, 5, '+'),
     \    ],
     \    'moved': 'any',
-    \    })
+    \    }
+
+    " allow custom popup opts
+    if exists('g:ale_floating_preview_popup_opts')
```
hsanson, line 178: I would allow this to be either a function name or a dictionary. Then we can check do something like:

```vimscript
if exists('g:ale_floating_preview_popup_opts')
    if type(g:ale_floating_preview_popup_opts) is# v:t_dict
        let l:popup_opts = extend(l:popup_opts, g:ale_floating_preview_popup_opts)
    elseif exists('*' . g:ale_floating_preview_popup_opts)
        let l:popup_opts = extend(l:popup_opts, function(g:ale_floating_preview_popup_opts)())
    endif
endif
```

Note that:

  1. I am extending the defaults instead of replacing them. This allows users to set some of the options without need to redefine them all.
  2. I check if the function does exists before invoking it.

doc/ale.txt
```diff
@@ -1234,6 +1234,17 @@ g:ale_floating_preview                                 *g:ale_floating_preview*
   |g:ale_detail_to_floating_preview| to `1`.
 
 
+g:ale_floating_preview_popup_opts             *g:ale_floating_preview_popup_opts*
+
+  Type: |String|
+  Default: `''`
+
```
hsanson, line 1241: Make clear that his option takes effect only in vim and not neovim.  Also if we are going to support both funcref and dict update the type too.

doc/ale.txt
```diff
@@ -1234,6 +1234,17 @@ g:ale_floating_preview                                 *g:ale_floating_preview*
...
+  The string name of a function that will be called when |g:ale_floating_preview| is
+  enabled to get options to pass to |popup_create|. This function should return a
+  |Dictionary| compatible with |popup_create|. If none is supplied, default options
+  are used.
+
```
hsanson, line 1246: Maybe good idea add an example configuration using dictionary showing some of the things that can be customized.

hsanson (changes requested):

hsanson:
I think this kind of new functionality should be implemented in a way compatible with both vim and neovim. Unfortuntelly I do not have resources (time) to try figuring out a common ground that would allow this to work equally on both. 

Also would be nice if we spend some time trying to figure out how to write tests for this functionality to ensure future changes in ALE, vim, or neovim do not break it.

autoload/ale/floating_preview.vim
shaunduncan: Allowing for either a dictionary or a function is a great idea, I'll add that. Originally I wrote this using extend like you have suggested, but I wanted to be able to specify only the options that I wanted. I think this is more specific to my usage, so extending the default options is probably a better idea for most users.

On checking if the funcref actually exists, in my setup I am referencing an autoload function that fails the exists() check because it hasn't been loaded yet. Example:

```vim
let g:ale_floating_preview_popup_opts = 'myplugin#get_popup_opts'
```

Do you have a suggestion on how best to handle this? I'm happy to adjust my setup if you prefer to keep the exists check.

shaunduncan:
@hsanson I've included changes based on your suggestions. I am not well-versed enough on vim-vader tests to write some otherwise I would be happy to do so. I did, however, add the callback for showing the neovim floating preview. I am not a regular neovim user so it is somewhat difficult to test, but by doing a quick neovim check it appeared to honor my configuration.

autoload/ale/floating_preview.vim
hsanson: Nice, I tested this and works nicely in Neovim with a dictionary and a funcref.  

The idea of checking the funcref is that if the user mistakenly sets something wrong we don't get ALE spitting cryptic messages.  Something like this would do and even cover other cases like setting a number instead of a string or funcref:

```vimscript
function! s:GetPopupOpts() abort
    if !exists('g:ale_floating_preview_popup_opts')
        return {}
    endif

    let l:ref = g:ale_floating_preview_popup_opts

    if type(l:ref) is# v:t_dict
        return l:ref
    endif

    if type(l:ref) is# v:t_string
        try
            return function(l:ref)()
        catch E700
        endtry
    endif

    return {}
endfunction
```

shaunduncan: Perfect! I've modified this now to try/catch if it's a string.

hsanson (approved):
Thanks, looks good.

hsanson:
CI failed with some linter issues. Please add ale custom linter to your config so vim informs you of these issue:

```
'vim': ['ale_custom_linting_rules']
```

shaunduncan:
@hsanson I've made corrections based on the lint errors. For some reason though adding the vim lint configuration never showed any errors for me. Not sure why. Hopefully I've made the changes correctly.

hsanson (approved):
Thanks for the contribution.

PR #4250, opened 2022-07-13
Fix 4249 - Revert change to stop compl menu.

hsanson:
In #4231 some code was added to stop the completion menu if any when
opening a new one. This resulted in an issue in Vim that fills the
buffer with Ctrl-Z characters when deleting to the end of a line in a
position that triggers auto-completion.

Since auto-completion seems to work fine on all my tests I am reverting
this specific change.


franckinux:
Horacio, the commit fixes the issue. Thanks.

s-marios:
Ah, I'm really sorry to see this reverted, and honestly speaking I'm surprised this occurs in the first place. 

Vim 8.1 should not have the ```complete_info``` function defined, and it is even stranger that it somehow got invoked in normal mode, in an infinite loop nonetheless. Anyway, apologies for this mess!

PR #4259, opened 2022-07-20
Prevent buffering of job output and excessive polling

### LINKED ISSUE #4098: gopls stops working if you write sufficiently complex code in Go 1.18
## Information

**VIM version**

VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Aug 29 2020 14:11:37)
Included patches: 1-1539 

Operating System: Linux

## What went wrong

If you write very complex code that causes `gopls` to freak out something goes wrong with the messages between ALE and `gopls` and you can't see errors in your editor until the code is modified and you restart the server with `:ALEStopAllLSPs`. It might be that the server is panicking with code it can't understand and ALE need to handle the server crashing better, or that the messages `gopls` is sending aren't understood by ALE.

## Reproducing the bug

Reproducing this is complicated. I'm going to have a look at it.

### :ALEInfo

N/A

liskin:
When 'close_cb' is set for job_start(), but out_cb or err_cb isn't, vim buffers data instead of dropping it (in case someone wanted to read and process it in close_cb), and additionally polls for new data every 10 milliseconds, causing excessive wakeups and CPU usage. Since we don't read the data anywhere outside of out_cb/err_cb, any LSP that prints an error to stderr triggers this and vim keeps spinning until :ALEStopAllLSPs.

Fix this by always setting both callbacks, thus dropping any data we're not interested in.

See https://github.com/vim/vim/issues/10758 for an upstream report of the excessive polling. It's possible this is intentional, I dunno.

Fixes: b42153eb1786 ("Fix #4098 - Clear LSP data when servers crash")

> Where are the tests? Have you added tests? Have you updated the tests? Read the comment above and the documentation referenced in it first. Write tests!

I don't think this is easily unit-testable. :-(

### DIFF
--- a/autoload/ale/job.vim
+++ b/autoload/ale/job.vim
@@ -250,10 +250,16 @@ function! ale#job#Start(command, options) abort
 
         if has_key(a:options, 'out_cb')
             let l:job_options.out_cb = function('s:VimOutputCallback')
+        else
+            " prevent buffering of output and excessive polling in case close_cb is set
+            let l:job_options.out_cb = {->0}
         endif
 
         if has_key(a:options, 'err_cb')
             let l:job_options.err_cb = function('s:VimErrorCallback')
+        else
+            " prevent buffering of output and excessive polling in case close_cb is set
+            let l:job_options.err_cb = {->0}
         endif
 
         if has_key(a:options, 'exit_cb')


liskin:
In https://github.com/dense-analysis/ale/commit/b42153eb1786fbff1c926e60c23ba82b0108f0cc#commitcomment-78967108 I also mentioned it'd be nice to not just set a dummy err_cb, but actually collect that output and present it in :ALEInfo. I looked at it briefly and it seems :ALEInfo is currently designed to only show output for completed jobs: https://github.com/dense-analysis/ale/blob/f2837b5802c14e0fcdddf1af7caab0fb96858589/autoload/ale/history.vim#L57-L62
I guess it'll be fine if the stderr output is remembered in ale/lsp.vim's s:connections and `ale#history#RememberOutput` called to update it whenever err_cb fires, but it's a bit of a hack and the output from a long-running LSP might get really large. Any ideas? Shall I implement the hacky way?

PR #4263, opened 2022-07-23
Add support for Dune

apeschar:
This adds a fixer that runs `dune format` on `dune` files.

`dune` files are build specifications for the Dune build system for OCaml.


hsanson (changes requested):
Please add an entry for  the new fixer in the list of supported tools in ale.txt file.

hsanson (approved):
Thanks, looks good.

PR #4268, opened 2022-07-28
Add support for syntax_tree fixer

molawson:
This is the library that now powers prettier/plugin-ruby but is also stands on its own: https://github.com/ruby-syntax-tree/syntax_tree

I added tests based on existing examples, but much of this was new for me.  I'm happy to tweak or update any of this to fit within the rest of this plugin.  I have also tested it locally and it's working as I'd expect for me.


hsanson (approved):
Thanks, looks good.

PR #4271, opened 2022-08-02
Add vim 9.0 and remove 8.0

### LINKED ISSUE #4242: Update tests to use vim 9
Now with vim 9 released we should update the CI to test ALE against it and to fix any breaking issues introduced in vim 9.

- https://github.com/vim/vim/releases/tag/v9.0.0000

hsanson:
Replaces vim 8.0 with vim 9.0 in the test scripts.

Closes #4242

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,8 +26,8 @@ jobs:
     strategy:
       matrix:
         vim-version:
-          - '--vim-80-only'
           - '--vim-82-only'
+          - '--vim-90-only'
           - '--neovim-02-only'
           - '--neovim-07-only'
           - '--linters-only'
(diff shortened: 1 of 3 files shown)


w0rp (changes requested):
Let's keep Vim 8.0 support and replace 8.2 in the test suite with 9.0. No reason to remove 8.0 support yet.

hsanson:
Ok, changed the PR to keep 8.0 and remove 8.2.

RyanSquared:
Vim 8.2 has some significant features over Vim 8.0, and is the version that I'm personally using. Is there any reason to not be testing against vim 8.2 and instead continue to support Vim 8.0?

hsanson:
In the theorical world testing the min and max supported versions should ensure inbetween versions also work.  Except for some breaking changes between vim 8.0 and 8.2 this has been the case so far.

Issue #4285, opened 2022-08-17
`:ALERename` does not pick up ruby methods with a trailing question mark


erikw:
## Information

**VIM version**
```
NVIM v0.7.2
Build type: Release
```

Operating System: macOS 12.5

## What went wrong

I have a ruby method like this

```ruby
def whatever?
   # return false or true
end
```
* I put the text cursor inside `whatever?`, say on the `h`
* I type `:ALERename<CR>`
* I'm prompted with `New name: whatever`

Here we see already that the rename command did not pick up the trailing question mark. If I go on to rename this what will happen is that references are renamed, but not the actual method definition (the current line).

From how it looks to me, the rename command is not aligned with what are [valid method names](https://docs.ruby-lang.org/en/3.0/syntax/methods_rdoc.html#label-Method+Names) in Ruby. The issue could of course be with one of the tools configured for ruby here with Ale: solargraph & rubocop.

### :ALEInfo
```
 Current Filetype: ruby
Available Linters: ['brakeman', 'cspell', 'debride', 'rails_best_practices', 'reek', 'rubocop', 'ruby', 'solargraph', 'sorbet', 'standardrb']
   Linter Aliases:
'sorbet' -> ['srb']
  Enabled Linters: ['ruby', 'solargraph']
  Ignored Linters: []
 Suggested Fixers: 
  'prettier' - Apply prettier to a file.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'rubocop' - Fix ruby files with rubocop --auto-correct.
  'rufo' - Fix ruby files with rufo
  'sorbet' - Fix ruby files with srb tc --autocorrect.
  'standardrb' - Fix ruby files with standardrb --fix
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

let g:ale_ruby_rails_best_practices_executable = 'rails_best_practices'
let g:ale_ruby_rails_best_practices_options = ''
let g:ale_ruby_ruby_executable = 'ruby'
let g:ale_ruby_solargraph_executable = 'solargraph'
let g:ale_ruby_solargraph_options = {}
 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = 100
let g:ale_completion_enabled = 1
let g:ale_completion_max_suggestions = 50
let g:ale_disable_lsp = 0
[...]

PR #4286, opened 2022-08-17
Put `:ALEInfo` in collapsible for bug reports

erikw:
To make the bug reports a bit more readable. See an example of usage
here: #4285


hsanson (approved):
Nice, thanks.

Issue #2962, opened 2020-01-09
Allow virtual text to always appear for all issues


iocentos:
Just a small suggestion. 
I really like the virtual text option when I am on the line that has the issue. Currently you can choose between the following by setting the option `g:ale_virtualtext_cursor`

0 - completely disable the feature
1 - show virtual text one line at a time only when cursor is on the faulty line

There could be an extra option to allow to show virtual text for all issues in a file. In this case virtual text is always on for all issues. We would not need to hover or go to the line to actually see the error.

bapjiws:
Is there any update on this one? Would be really neat to have it for the entire file all at once, like,say, [vim-lsp](https://github.com/prabirshrestha/vim-lsp#virtual-text) has it 🙂

malukenho:
That would be pretty nice!

malukenho:
I just made it work in a hacky way https://gist.github.com/malukenho/6392a69c55f3f1daff4ab4bfb7d56654

PR #4289, opened 2022-08-23
Implement buffer-wide virtual text support

vimpostor:
This implements buffer-wide virtual text as detailed in #2962, both for vim and neovim.

You can activate it with `let g:ale_virtualtext_cursor = 2`.

Note that sometimes there is a problem, where very similar warnings appear
twice as virtual text, but this is not a bug in the implementation, rather it's
a problem of using multiple linters that return very similar warnings.

## Demo
Screenshot_20220823_204435

### DIFF
--- a/autoload/ale/engine.vim
+++ b/autoload/ale/engine.vim
@@ -203,6 +203,10 @@ function! ale#engine#SetResults(buffer, loclist) abort
         call ale#highlight#SetHighlights(a:buffer, a:loclist)
     endif
 
+    if g:ale_virtualtext_cursor == 2
+        call ale#virtualtext#SetTexts(a:buffer, a:loclist)
+    endif
+
     if l:linting_is_done
         if g:ale_echo_cursor
             " Try and echo the warning now.
@@ -210,7 +214,7 @@ function! ale#engine#SetResults(buffer, loclist) abort
             call ale#cursor#EchoCursorWarning()
         endif
 
-        if g:ale_virtualtext_cursor
+        if g:ale_virtualtext_cursor == 1
             " Try and show the warning now.
             " This will only do something meaningful if we're in normal mode.
             call ale#virtualtext#ShowCursorWarning()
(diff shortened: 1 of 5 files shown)


hsanson (commented):
Can verify this works in Neovim 0.7 and does not break anything as far as I can see.  Again would be better if someone (not the PR author) can check this works fine in their environment using Vim.

RyanSquared:
I believe this happens because `s:hl_list` is only set when `s:emulate_virt` is *not* set, but the following is run as a test: `if s:emulate_virt && s:last_virt != -1`. When opening a file, `s:emulate_virt && s:last_virt != -1` is not true, so it goes to the second expression, which relies on `s:hl_list`. This can be fixed by checking `!s:emulate_virt && !empty(s:hl_list)`, as it will skip the second half of the expression if the first half fails.

RyanSquared:
This also completely breaks all virtualtext for Vim using emulated virtual text. I am unsure if the buffer-wide virtualtext should be supported or not, but I would prefer it to not break vim 8.x setups.

vimpostor:
> This also completely breaks all virtualtext for Vim using emulated virtual text. I am unsure if the buffer-wide virtualtext should be supported or not, but I would prefer it to not break vim 8.x setups.

Yeah, sorry I only tested with vim 9 and neovim, will fix that asap

WhiteBlackGoose:
>  Again would be better if someone (not the PR author) can check this works fine in their environment using Vim.

Works fine! Thank you all, people who make it!

image

info:
```
NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by runner@fv-az316-460
```

jeffdn:
This works great for me:

image

```
ViM v9.0.350
```

hsanson (approved):
Thanks for the contribution and to those that verified it working.

hsanson:
@vimpostor there is a merge conflict. Appreciated if you can take a look into it.

vimpostor:
On Tue, Sep 06, 2022 at 03:41:09AM -0700, Horacio Sanson wrote:
***@***.*** there is a merge conflict. Appreciated if you can take a 
>look into it.

I fixed the merge conflict.

WhiteBlackGoose:
Thanks!

Issue #4290, opened 2022-08-24
ALE v8.x is broken after commit d93bc2baf


RyanSquared:
## Information

**VIM version**

VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Apr 18 2022 19:26:30)
Included patches: 1-3995

Operating System: 

## What went wrong

Error detected while processing function ale#virtualtext#ShowCursorWarning[17]..ale#virtualtext#Clear:
line   15:
E968: Need at least one of 'id' or 'type'

## Reproducing the bug

Using branch master after commit: d93bc2baf

Edit any file; after moving around, the message should happen at every cursor move.

### :ALEInfo
<details>
  <summary>Expand</summary>
```
 Current Filetype: fugitiveblame
Available Linters: []
  Enabled Linters: []
  Ignored Linters: []
 Suggested Fixers: 
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = v:null
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = v:null
let g:ale_disable_lsp = 0
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'E'
let g:ale_echo_msg_format = '[%linter%] %s [%severity%:%code%]'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'W'
let g:ale_enabled = 0
let g:ale_fix_on_save = 0
let g:ale_fixers = {'rust': ['rustfmt']}
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_filetype_changed = 1
let g:ale_lint_on_insert_leave = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'normal'
let g:ale_linter_aliases = {}
let g:ale_linters = {'html': [], 'rust': ['analyzer'], 'cpp': ['clang', 'g++'], 'python': ['flake8']}
let g:ale_linters_explicit = 0
let g:ale_linters_ignore = {}
[...]

PR #4291, opened 2022-08-24
Fix virtual text for vim 8.2

vimpostor:
See #4290

We need to guard the if condition with another check for whether we are running vim 9, otherwise it is possible to enter vim 9 code paths even when we are using vim 8.2.


RyanSquared (approved):
This does work, fixes #4290.

hsanson (approved):
Thanks for the fix.

PR #4293, opened 2022-08-24
omni: find last racket keyword for completion

benknoble:
Otherwise it finds the first keyword, which is usually not relevant to the
cursor position, and incorrectly calculates the completion position.


benknoble:
cc @hsanson 

Is there something wrong with the NeoVim 06 CI?

hsanson:
@benknoble 
> Is there something wrong with the NeoVim 06 CI?

Sorry, the problem is my lack of Github knowledge.  I finally found how to remove that check requirement. We now check for neovim 0.7 instead.

hsanson (approved):
I do not use racket so I trust this has been tested by the author.

PR #3257, opened 2020-07-30
Fix Astyle Project Option Detection when Placed with Source Files

jhlink:
# Summary

In response to the [bug](https://github.com/dense-analysis/ale/pull/3231#issuecomment-665806694) reported by @cjappl, submitting a fix to resolve Issue #3256. 

## Changes

- Add a test case detecting `.astylerc` when placed along a cpp source file. 
  - Created a **test_cpp_project** dir containing `.astylerc` and `dummy.cpp` to support the test case.
- Add the bug fix [proposal](https://github.com/dense-analysis/ale/pull/3231#issuecomment-666000907) by @hsanson to resolve issue. 
- Fixed test cases to reflect correct output from bug fix.

Closes Issue #3256

PR #4295, opened 2022-08-28
Make `astyle` under `:ALEFix` operate on buffer

ankitpati:
Previously, `:ALEFix` made `astyle` operate on the disk file, and populated the buffer with the results, potentially losing work.

This PR regresses #3257, but that appears to be a bug with `astyle` itself, not with `ale`, so it should be fixed there anyways. I feel avoiding data-loss should be a higher goal than working around `astyle` bugs.

Thanks for your awesome work!

### DIFF
--- a/autoload/ale/fixers/astyle.vim
+++ b/autoload/ale/fixers/astyle.vim
@@ -49,11 +49,9 @@ endfunction
 function! ale#fixers#astyle#Fix(buffer) abort
     let l:executable = ale#fixers#astyle#Var(a:buffer, 'executable')
     let l:proj_options = ale#fixers#astyle#FindProjectOptions(a:buffer)
-    let l:command = ' --stdin=' . ale#Escape(expand('#' . a:buffer))
 
     return {
     \   'command': ale#Escape(l:executable)
     \     . (empty(l:proj_options) ? '' : ' --project=' . l:proj_options)
-    \     . l:command
     \}
 endfunction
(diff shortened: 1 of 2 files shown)


hsanson (commented):
Change looks good to me but I do not use this tool so would be great if @jhlink  and @skorokithakis could test this change.

jhlink (changes requested):
Tested, and see no change. 

Still, I'm against this change. Without any data *proving* that astyle formats code with some loss into a buffer, this is an unneeded, unsolicited change. 

Further, if the bug is a problem with astyle, why does ALE need to accommodate? 
Regressing support because a dependency has a bug makes little sense. 

If data loss is a concern, use another tool like clang. I will wait for astyler maintainers to release a fix.

ankitpati:
@jhlink Steps to reproduce:

1. Ensure `astyle` is the only ALE fixer defined for C.
2. Open a `.c` file with valid C code.
3. Add a line of valid C code anywhere in the file.
4. Run `:ALEFix`
5. Observe that the added line has vanished.

hsanson:
Took some time to re-read the related discussions and test this PR and unfortunately applying this PR reverts the initial issue.  If we do not use the --stdin flag then astyle fails to find and use any .astylerc or _astylerc configuration files within the project.   This would be a breaking change so I cannot approve this PR unless alternative solution for the .astylerc/_astylerc file issue is also included.

jhlink:
Thanks for helping with the analysis @hsanson !

PR #4300, opened 2022-09-04
Allow to pass options to the buf linter

gou4shi1:
I need to pass `--disable-symlinks` to the `buf lint` command.


hsanson (approved):
Thanks, looks good to me.

PR #4304, opened 2022-09-08
Fix typo in plug mapping for `ale_go_to_implementation_in_vsplit`

mrpinsky:
This is just a typo fix, and none of the other plug mappings are tested, so I'm not sure how to test them. Happy to add that on if you can point me in the right direction.


hsanson (approved):
Thanks, nice catch.

PR #4306, opened 2022-09-09
Deno: find project root using deno.json/deno.jsonc

dirkjonker:
Deno LSP automatically detects config files named deno.json or deno.jsonc since version 1.18.

For Deno 1.18+ this means that ALE no longer needs to resolve the project root. However, removing the project root logic from ALE means breaking changes for people that are still using an older version. Adding deno.json to the list of looking files to look for will keep the behavior consistent and compatible with the Deno config file naming convention.

See also:
https://deno.com/blog/v1.18#auto-discovery-of-the-config-file


hsanson (approved):
Thanks, looks good.

PR #4308, opened 2022-09-10
chore(volar): update volar executable name

BBOOXX:
see: johnsoncodehk/volar#876
see: https://github.com/johnsoncodehk/volar/blob/master/packages/vue-language-server/package.json#L11


hsanson (approved):
Looks good, thanks.

PR #4310, opened 2022-09-12
Add support for Microsoft's DSL Bicep

carlsmedstad:
The compilation command 'bicep build' catches compilation errors as well as providing some lint warnings.

Repository for Bicep: https://github.com/Azure/bicep

### DIFF
--- a/ale_linters/bicep/bicep.vim
+++ b/ale_linters/bicep/bicep.vim
@@ -5,7 +5,7 @@ let g:ale_bicep_bicep_executable =
 \   get(g:, 'ale_bicep_bicep_executable', 'bicep')
 
 let g:ale_bicep_bicep_options =
-\   get(g:, 'ale_bicep_bicep_options', 'build --outfile /dev/null')
+\   get(g:, 'ale_bicep_bicep_options', '')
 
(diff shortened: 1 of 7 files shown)


ale_linters/bicep/bicep.vim
```diff
@@ -0,0 +1,52 @@
+" Author: Carl Smedstad <carl.smedstad at protonmail dot com>
+" Description: bicep for bicep files
+
+let g:ale_bicep_bicep_executable =
+\   get(g:, 'ale_bicep_bicep_executable', 'bicep')
+
+let g:ale_bicep_bicep_options =
+\   get(g:, 'ale_bicep_bicep_options', 'build --outfile /dev/null')
```
hsanson, line 8: Some doubts about these default options:

1. If users blindly set this variable then the linter most likely stop woking if they do not add "build --outfile /dev/null" also. Maybe better to force these options always in the ale_linters#bicep#bicep#Command() function?
2. Not sure "/dev/null" would work on Windows platforms.

carlsmedstad: Good points. I have hopefully addressed both of them.

hsanson (approved):
Thanks, looks good.

PR #4311, opened 2022-09-14
bicep: Lint files on disk instead of buffer

carlsmedstad:
I discovered that references to other Bicep files (modules) will be broken if running on a temporary file in a different location. I've found no way of providing an alternate path when invoking the command.

Sorry for not catching this earlier.


carlsmedstad:
Fixed :)

hsanson (approved):
Thanks for the fix.

PR #4313, opened 2022-09-18
Fix jsonnet lint pattern

user827:
The current version does not prefix the output with ERROR anymore e.g:
`hoi.jsonnet:5:3-8 Expected a comma before next field`


hsanson (changes requested):
this linter lacks handler tests. Would be good to add some test/handler/test_jsonnet_handler.vader test file that verifies that errors with old and new pattern work.

Also wonder if this change also affects the related jsonnetfmt linter.

PR #4315, opened 2022-09-21
Pass input via stdin and filename as arg to clj-kondo

christoph-frick:
This allows using linters like `:namespace-name-mismatch`.

This replaces: https://github.com/dense-analysis/ale/pull/4109


hsanson (approved):
Thanks, looks good.

PR #4318, opened 2022-09-23
cc: use '-x c*-header' for header files for C and C++ linters.

### LINKED ISSUE #4096: Linting C/C++ headers - unused variables, non-declared, wrong types...
I quote the opening message of a discussion I started a few days ago:

> I have been struggling for a while trying to find out how to make ALE not to mark variables in C headers as unused, non-declared, etc, while they are being used by the .c files , declared in other headers that are called by the main .c files, etc.

> I am trying to use this plugin in Neovim and testing it with Suckless' DWM. The project compiles fine, so it does not seem logical to me that ALE marks those lines in headers as wrong. GCC is being used both for compiling and linting (no other linters installed), and from what I can see in :ALEInfo, GCC seems to be called with the current opened .h file as argument. This explains to me why all those errors appear, as GCC is trying to compile the header standalone. This is what I deduce.

> I would like to know if there is a way to configure ALE so it parses the output of the command that compiles the whole project instead of only the currently opened header. Generating compile_commands.json with bear and setting g:ale_c_parse_compile_commands has helped solving some flags issues, but not this one with the headers. Even using compdb to include headers in compile_commands.json.

Just to add some more info, the project on which I am experiencing this issue is [my custom build of DWM](https://github.com/JuanClBl/dwm). As [the original project](https://git.suckless.org/dwm/files.html), both are compiled using `make`, so it would be nice if ALE can be configured to read the project's `Makefile`, and use `make`'s output to lint the currently opened header file instead of trying to compile it standalone. Moreover, when opening a main source file like `dwm.c`, ALE properly follows all included headers and does not give any variable or type non declared errors.

nicopauss:
Fix  #4096

When linting an header file in C or C++, `-x c-header` or `-x c++-header` should be used instead of `-x c` or `-x c++`.

Using `-x c` or `-x c++` for headers files can lead to unused variables and functions marked as static inlined as seen in #4096.

Using `-x c-header` or `-x c++-header` solve these issues.

The list of file extensions that are considered as header files can be configured with the variables `g:ale_c_cc_header_exts` and `g:ale_cpp_cc_header_exts`.

### DIFF
--- a/ale_linters/c/cc.vim
+++ b/ale_linters/c/cc.vim
@@ -3,6 +3,7 @@
 
 call ale#Set('c_cc_executable', '<auto>')
 call ale#Set('c_cc_options', '-std=c11 -Wall')
+call ale#Set('c_cc_header_exts', ['h'])
 
 function! ale_linters#c#cc#GetExecutable(buffer) abort
     let l:executable = ale#Var(a:buffer, 'c_cc_executable')
@@ -22,6 +23,8 @@ endfunction
 function! ale_linters#c#cc#GetCommand(buffer, output) abort
     let l:cflags = ale#c#GetCFlags(a:buffer, a:output)
     let l:ale_flags = ale#Var(a:buffer, 'c_cc_options')
+    let l:header_exts = ale#Var(a:buffer, 'c_cc_header_exts')
+    let l:lang_flag = ale#c#GetLanguageFlag(a:buffer, l:header_exts, 'c')
 
     if l:cflags =~# '-std='
         let l:ale_flags = substitute(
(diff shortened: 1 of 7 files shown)


hsanson (approved):
Thanks, looks good.

nicopauss:
Oh sorry about that. I though I have tested it with GCC, but I must have messed my config while testing and only tested with clang.
I'll try to make a fix ASAP.
In the meanwhile, you can set `g:cpp_cc_header_exts` to `[]`: `let g:ale_cpp_cc_header_exts = []`

nicopauss:
Fix here: #4334

PR #4319, opened 2022-09-26
Add fixer "css-beautify". Adjust "html-beautify."

Spixmaster:
The test for this new feature (`./run-tests  test/fixers/test_css_beautify_fixer_callback.vader`) works.

### DIFF
--- a/doc/ale-supported-languages-and-tools.txt
+++ b/doc/ale-supported-languages-and-tools.txt
@@ -120,7 +120,7 @@ Notes:
 * CSS
   * `VSCode CSS language server`
   * `cspell`
-  * css-beautify
+  * `css-beautify`
   * `csslint`
   * `fecs`
   * `prettier`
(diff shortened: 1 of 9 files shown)


doc/ale-supported-languages-and-tools.txt
```diff
@@ -120,6 +120,7 @@ Notes:
 * CSS
   * `VSCode CSS language server`
   * `cspell`
+  * css-beautify
```
hsanson, line 123: Please add the backticks "`" around as the other entries.

hsanson (changes requested):

doc/ale-supported-languages-and-tools.txt
Spixmaster: Thanks for the hint. It is done.

hsanson (approved):
Thanks.

Issue #4317, opened 2022-09-23
:ALEInfo misses fixer "dprint" for TOML files


Spixmaster:
:ALEInfo in TOML files prints these fixers:
```
 Suggested Fixers:
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
```

"dprint" is missing.

PR #4320, opened 2022-09-27
Added file types for dprint.

Spixmaster:
dprint can format more file types. I added them.


Spixmaster:
#4317

hsanson (approved):
Thanks, looks good.

PR #4325, opened 2022-10-04
Fix 4324 - add neovim 0.8 test

hsanson:


Stealthii:
I would suggest adding NeoVim 0.8, rather than replacing 0.7, as many installations will be continuing to use 0.7 until their plugins and supporting scripts are all confirmed working.

w0rp:
> I would suggest adding NeoVim 0.8, rather than replacing 0.7, as many installations will be continuing to use 0.7 until their plugins and supporting scripts are all confirmed working.

We test against a minimum and maximum supported version, and I don't expect anything to break in 0.7 which isn't also broken in 0.8, or anything that doesn't work in 0.7 that wouldn't also be something that doesn't work in 0.2. At least not anything major enough that it requires the CI to check it specifically.

w0rp (approved):

w0rp:
One of those smoke tests is flaky in Vim 9.0 and I don't think we need it any more. I'll remove it in another commit.

w0rp (approved):

PR #4323, opened 2022-10-03
Add support for tfsec Terraform linter

koka831:
This PR addresses https://github.com/dense-analysis/ale/issues/4229

PR #4326, opened 2022-10-04
fix: wrong config name

koka831:
fix https://github.com/dense-analysis/ale/pull/4323#discussion_r986944816


hsanson (approved):
Thanks for the quick fix.

PR #4334, opened 2022-10-07
cc: fix using '-x c*-header' for header files with GCC.

### LINKED ISSUE #4096: Linting C/C++ headers - unused variables, non-declared, wrong types...

nicopauss:
Fix #4096 again.

Gcc does not support `x c*-header` when using `-` as input filename, which is what ALE does.

Rework the feature to only use `-x c*-header` flag when using Clang and not GCC.

The feature is now also controlled with the variable `g:ale_c_cc_use_header_lang_flag`.

### DIFF
--- a/doc/ale-c.txt
+++ b/doc/ale-c.txt
@@ -133,7 +133,30 @@ g:ale_c_cc_options                                         *g:ale_c_cc_options*
   Type: |String|
   Default: `'-std=c11 -Wall'`
 
-  This variable can be change to modify flags given to the C compiler.
+  This variable can be changed to modify flags given to the C compiler.
+
+
+g:ale_c_cc_use_header_lang_flag               *g:ale_c_cc_use_header_lang_flag*
+                                              *b:ale_c_cc_use_header_lang_flag*
+  Type: |Number|
+  Default: `-1`
+
+  By default, ALE will use `'-x c-header'` instead of `'-x c'` for header files
+  when using Clang.
+
+  This variable can be changed to manually activate or deactivate this flag
+  for header files.
+
+  - When set to `-1`, the default beviour is used, `'-x c-header'` is used with
+    Clang and `'-x c'` is used with other compilers.
+  - When set to `0`, the flag is deactivated, `'-x c'` is always used
+    independently of the compiler.
+  - When set to `1`, the flag is activated, `'-x c-header'` is always used
+    independently of the compiler.
+
+  Gcc does not support `'-x c-header'` when using `'-'` as input filename,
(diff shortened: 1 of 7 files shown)


doc/ale-c.txt
```diff
@@ -133,18 +133,39 @@ g:ale_c_cc_options                                         *g:ale_c_cc_options*
...
+
+  By default, ALE will use `'-x c-header'` instead of `'-x c'` for header files
+  when using Clang.
+
+  This variable can be changed to manually activate or deactivate this flag
+  for header files.
+
+  When set to `-1`, the default beviour is used.
+  When set to `0`, the flag is deactivated and `'-x c'` is used.
+  When set to `1`, the flag is activated and `'-x c-header'` is used.
```
hsanson, line 152: Looking at the implementation it seems that "-1" (default behavior) sets the `-x c-header` option for header files when using Clang while "1" sets the `-x c-header` option for header files always?

Not everyone will look into the implementation so adding "always independently of compiler" at the end of the "1" option description would be good.

hsanson (changes requested):

hsanson (approved):
Thanks for the fix.

PR #4335, opened 2022-10-08
Add a generic fixer for removing duplicate blank lines

benfranklin79:
Just started using ALE and thought this might be a useful option to have


benfranklin79:
Will look into adding some tests

PR #4338, opened 2022-10-14
fix: make autoimport fixer read from file instead of stdin

ErikBjare:
Enables use of `--ignore-init-modules` flag introduced in https://github.com/lyz-code/autoimport/pull/228


ErikBjare:
Finally got the CI to pass, hope there's nothing else :)

hsanson:
Thanks for the work.  I can see autoimport supports a [pyproject.toml](https://lyz-code.github.io/autoimport/#configuration) configuration file and this change will probably cause the tool being unable to find it and break things for some users.

This is due to the use of temporary files that are not located in the same place where the configuration is expected to be found by the tool.

I recommend testing if this change allows autoimport work with configuration files before it is merged.

Issue #4301, opened 2022-09-05
Add support for ruff linter for python


TheEdgeOfRage:
**Name:** ruff
**URL:** https://github.com/charliermarsh/ruff

Ruff is a new Python linter written in Rust that promises to be 10-100x faster than other linters (They provide benchmarks that seem to prove them correct). It seems like an interesting project and might be worth adding support for in Ale.

ssokolow:
I do effectively all my linting through ALE, so this would be valuable to me.

I'd contribute a patch myself but, at the moment, I've had to put a hold on all new things to focus on clearing out backlogged mess. Someday maybe, given how helpful it'd be to have that speedup.

mfurquimdev:
This was already merged. Should this issue be closed?

dvzrv:
@w0rp: It would be awesome if you could tag a release for this.

We have this plugin available on Arch Linux (and I have just added ruff to the list of official packages as well), but currently the integration is unfortunately not there.
Thanks a lot!

hsanson:
Ruff is already supported by ALE. Closing.

dvzrv:
@hsanson Yes, but for this to arrive in downstream packages, it needs a release. Could you create one? That would be very much appreciated! :)

hsanson:
I see.  Releases are done by @w0rp so we would have to wait until @w0rp has some time to do the release.  Sorry.

jgb:
@yining could it be that there is a bug in https://github.com/dense-analysis/ale/pull/4347 ?
The first time I try to do :ALEFix on a newly opened python file, I get the error `There is no fixer named ruff`.
When I then do an :ALELint, and then retry :ALEFix, the fix works fine.
Repeated this with several python files. Fix won't work until I first run a lint. After the first lint, fix works fine.

My ALE config:

let g:ale_linters = { 'python': ['ruff', ], }
let g:ale_fixers = { '*': ['remove_trailing_lines', 'trim_whitespace'], 'python': ['ruff', 'black'], }

yining:
@jgb could you please help check if the PR #4393 fixed the bug? thanks.

PR #4347, opened 2022-10-30
add: support for ruff as a Python linter and fixer

yining:
this commit adds ruff as both a Python linter and fixer, together with some tests and documentation.

ruff repo: https://github.com/charliermarsh/ruff

this pull request addresses #4301


hsanson (approved):
Thanks, looks ok.

PR #4351, opened 2022-11-02
add: support config option for checkmake linter

yining:
`checkmake` by default checks config file _"in the same folder it's
executed in"_[^0] unless `--config` option is set.

This commit allows setting the `--config` option with an option
variable (with documentation updated).

[^0]: [configuration section on checkmake manual](https://github.com/mrtazz/checkmake/blob/main/man/man1/checkmake.1.md#configuration)


yining:
any feedback or question on this PR?

hsanson (approved):
Looks ok. Thanks.

Issue #4248, opened 2022-07-08
Hadolint Linting fails if executable does not support --no-color flag


AubreySLavigne:
## Information

**VIM version**

    VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jul 18 2021 11:42:29)
    macOS version - x86_64

**Operating System**

    macOS Monterey
    Version 12.4

**hadolint version**

    Haskell Dockerfile Linter 1.19.0-no-git

Note that this version does not container a flag `--no-color`

**ALE version**

    commit 5ad4fdd583116ec253aaf43e2d457cd3e3aa5587

## What went wrong

When linting a file named `Dockerfile` containing errors, ALE does not flag these errors.

Reviewing `:ALEInfo`, the `hadolint` executable failed with exit code 1.

In the shell, hadolint succeeds when run manually:

    » hadolint Dockerfile
    Dockerfile:4 DL4006 Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check

But the command fails when using the command generated by ALE:

```
» zsh -c 'hadolint --no-color - < Dockerfile'
Invalid option `--no-color'

Usage: hadolint [-v|--version] [-c|--config FILENAME] [-f|--format ARG] 
                [DOCKERFILE...] [--ignore RULECODE] 
                [--trusted-registry REGISTRY (e.g. docker.io)]
  Lint Dockerfile for errors and best practices
```

The [`l:opts` variable in hadolint.vim](https://github.com/dense-analysis/ale/blob/master/ale_linters/dockerfile/hadolint.vim#L105) is hardcoded to `'--no-color -'`. The linter's config in ALE does not support changing this value.

## Reproducing the bug

1. Write a Dockerfile with an error.
2. Save the file to initiate ALE's linting.
3. No errors are reported, so it appears the file is fine.
4. Commit and push the file to trigger build pipeline
5. Pipeline fails during linting stage

### :ALEInfo

```
 Current Filetype: dockerfile
Available Linters: ['dockerfile_lint', 'hadolint']
  Enabled Linters: ['hadolint']
  Ignored Linters: []
 Suggested Fixers: 
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

[...]

Issue #2875, opened 2019-11-02
hadolint: Honor config when using docker image


varac:
When using the hadolint docker image, the [hadolint global config file](https://github.com/hadolint/hadolint#configure) won't be evalutated.
I propose to add sth like `ale_dockerfile_hadolint_docker_opts` which lets you [configure hadolint ](https://github.com/hadolint/hadolint#how-to-use) using command-line options.

matthewarmand:
Took a stab at adding this over at #4353, if anyone interested wants to take a peek

PR #4353, opened 2022-11-04
Add 'dockerfile_hadolint_options' config setting

matthewarmand:
This PR adds the ability for users to configure command line arguments to send to `hadolint` using the `ale_dockerfile_hadolint_options` setting. This setting is respected for running `hadolint` via the executable and in Docker.

The most direct issue this would fix is #2875. It's similar to a solution described in #4248, but my assumption is that `--no-color` is probably necessary or desired for Ale to parse the output and display it in Vim; and besides,  `--no-color` has been supported by `hadolint` for the last 17 versions (one a major bump), [first being introduced about 20 months ago now](https://github.com/hadolint/hadolint/releases/tag/v1.23.0), so I would guess there isn't a strong desire on Ale's part to support those older versions of `hadolint`. 

> Where are the tests? Have you added tests? Have you updated the tests? Read the comment above and the documentation referenced in it first. Write tests!
>

I made the necessary modifications to existing tests, and added a couple more for the positive cases using the new config setting. I also updated the docs and ran all the tests/linters before committing.

### DIFF
--- a/test/test_dockerfile_hadolint_linter.vader
+++ b/test/test_dockerfile_hadolint_linter.vader
@@ -19,6 +19,7 @@ After:
   Restore
   silent! unlet b:ale_dockerfile_hadolint_use_docker
   silent! unlet b:ale_dockerfile_hadolint_docker_image
+  silent! unlet b:ale_dockerfile_hadolint_options
 
 
 Execute(linter honors ..._use_docker correctly):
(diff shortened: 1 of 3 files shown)


test/test_dockerfile_hadolint_linter.vader
```diff
@@ -55,15 +56,30 @@ Execute(command is correct when using docker):
   let b:ale_dockerfile_hadolint_use_docker = 'always'
 
   AssertEqual
-  \ "docker run --rm -i hadolint/hadolint hadolint --no-color -",
+  \ "docker run --rm -i hadolint/hadolint hadolint  --no-color -",
```
matthewarmand, line 59: This isn't really my favorite change to need to make, but this was necessary for a one-line change in the actual source if I wasn't going to make a giant `if else` to see if arguments were provided and we needed an extra space or not

hsanson (approved):
Thanks, looks good.

PR #4354, opened 2022-11-07
Add raco_fmt fixer for Racket files

jmcantrell:
I did not see a racket doc to add this to.

Yes, I did add a test file :)


hsanson (approved):
Looks good....  too bad there is no proper racket documentation.

PR #4358, opened 2022-11-14
Markdown: allow passing custom executable for markdownlint

### LINKED ISSUE #3750: Add support for markdownlint-cli2
**Name:** markdownlint-cli2
**URL:** https://github.com/davidanson/markdownlint-cli2

This is the linter CLI behind the VS Code markdownlint plugin, so it's in heavy usage. It supports inheritance of rule configuration files, which original markdownlint-cli does not. It could be as simple as allowing current "markdownlint" linter to specify the executable and allowing user to point to markdownlint-cli2 binary instead, but as of now I can't find a documented way or option to do that.

goetzc:
Should fix #3750.


hsanson (approved):
Thanks, looks good.

PR #4365, opened 2022-11-22
Fix bug in sqlfluff implementation & implement fixer support

carlsmedstad:
Made a mistake in my initial implementation that is corrected here.

Also added support for using sqlfluff as a fixer.


hsanson (approved):
Looks good, thanks .

PR #4376, opened 2022-12-02
Add Ruby linter with Steep

### LINKED ISSUE #3254: Add support for Steep (Ruby)
**Name:** steep
**URL:** https://github.com/soutaro/steep

Steep is a type checker for Ruby 2, using RBS files (Ruby type signature) that will be a core part of Ruby 3.

Direct CLI output:

```
$ steep check
lib/phone.rb:46:0: MethodDefinitionMissing: module=::Phone, method=self.countries (class Phone)
```

LSP support:

```
steep langserver
```

lloeki:
Fixes #3254

### DIFF
--- a/ale_linters/ruby/steep.vim
+++ b/ale_linters/ruby/steep.vim
@@ -1,0 +1,18 @@
+call ale#Set('ruby_steep_executable', 'steep')
+call ale#Set('ruby_steep_options', '')
+
+function! ale_linters#ruby#steep#GetCommand(buffer) abort
+    return '%e langserver --log-level=debug'
+endfunction
+
+call ale#linter#Define('ruby', {
+\   'name': 'steep',
+\   'lsp': 'stdio',
+\   'language': 'ruby',
+\   'executable': {b -> ale#Var(b, 'ruby_steep_executable')},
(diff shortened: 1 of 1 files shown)


lloeki:
I tried to add some tests but I'm not very proficient with this whole thing, so it's taking me a bit of time.

lloeki:
Another thing I noticed: one can have `Steepfile`s in a directory below where Gemfile is, and then this linter should use that instead. Indeed when there's no `Steepfile` at all at the top it fails.

e.g this fails:

```
./Gemfile
./Gemfile.lock
./gh-701/kwargs.rb
./gh-701/Steepfile
./gh-701/kwargs.rbs
./gh-708
./gh-708/foo.rb
./gh-708/Steepfile
./gh-708/foo.rbs
```

and this works:

```
./gh-701/kwargs.rb
./gh-701/Steepfile
./gh-701/Gemfile
./gh-701/Gemfile.lock
./gh-701/kwargs.rbs
./gh-708
./gh-708/foo.rb
./gh-708/Steepfile
./gh-708/Gemfile
./gh-708/Gemfile.lock
./gh-708/foo.rbs
```

Issue #4362, opened 2022-11-20
Add support for refurb to modernize Python codebases 


aurelg:
**Name:** refurb
**URL:** https://github.com/dosisod/refurb

`refurb` is a tool for refurbishing and modernizing Python codebases. It would be extremely useful to have notifications on the fly, while writing/refactoring code.

JonZeolla:
@hsanson should this be closed now, based on https://github.com/dense-analysis/ale/commit/cad8f8e68338221e157adb1fffbe8eb16c430fe0?

PR #4379, opened 2022-12-04
add: support for refurb as a Python linter

yining:
this commit adds refurb as a Python linter, together with some tests and documentation. it should fix issue: #4362

refurb repo: https://github.com/dosisod/refurb


hsanson (approved):
Thanks. Nice and complete PR.

PR #4393, opened 2022-12-21
fix: ruff not registered as fixer

yining:
this commit fixes the issue reported at: https://github.com/dense-analysis/ale/issues/4301#issuecomment-1359289391

it also had some code refactored and
tests added for ruff as fixer (missing in PR #4347).


jgb:
@yining I can confirm that this PR fixes the problem I mentioned in #4301 !
Thanks a lot, I'm enjoying ale + ruff :+1:

hsanson (approved):
Thanks for the quick fix.

PR #4394, opened 2022-12-22
nix: handle versions past 2.9

Goorzhel:
Same problem as #4045, with Nix now at 2.12.


hsanson (approved):
Thanks for keeping nix updated.

w0rp:
It would be nice to flip the logic around so we check for older `nix` versions and default to the newer command. The oldest version that supports the newer command string came out a while ago now.

Goorzhel:
> check for older nix versions and default to the newer command

Damn, I should've thought of this. The only other idea I had was to write a proper version comparison function, but I'm not nearly good enough at Vimscript for that. New PR later today.

PR #4398, opened 2022-12-25
nix: durably handle versions before 2.4

Goorzhel:
What I should have done in #4394. Idea: @w0rp


hsanson (approved):
Thanks for the quick update.

PR #4403, opened 2022-12-28
Add support for llvm-mc as an assembly linter

uidops:
This is a fork of `asm-gcc`


hsanson (changes requested):
Please add a test for the handler too in `test/handler` folder.

hsanson:
There seem to be some linting errors.  I recommend you enable the `ale_custom_linting_rules` linter for `vim` files in your config so ALE informs you of this errors in vim/neovim.

PR #4414, opened 2023-01-07
fix(ale_linters: python: ruff): add --stdin-filename if version >= 0.0.69

actionless:

yes


yining:
Good catch, thanks for the fix. I forgot to add `'--stdin-filename'` to linter after adding that to fixer.

hsanson (approved):
Thanks for the fix.

Issue #4340, opened 2022-10-19
Add support for pycln


w0rp:
**Name:** pycln
**URL:** https://github.com/hadialqattan/pycln

pylcn is a simple tool for removing unused import statements from Python code. We could run `pycln -c %t` as the command to look for problems in files to have it function as a linter. `pycln %t` should most likely work as a fixer command to remove the imports. I don't think there's anything about the files on disk it will need to know about other than maybe where a configuration file is.

hsanson:
Added in #4415

PR #4415, opened 2023-01-07
add: pycln as a python linter and fixer

yining:
this commit adds pycln as a Python linter and fixer, together with some tests and documentation. It addresses #4340

pycln repo: https://github.com/hadialqattan/pycln


hsanson (approved):
Thanks for the contribution.

PR #4421, opened 2023-01-14
run-tests: download images only on x86-64

apeschar:
When running the tests on aarch64, the run-tests script tries to download a pre-built image that is built for x86-64, and thus does not run.

This change adds a check for the Docker daemon host platform and only downloads the image if it will run.

Furthermore, the image dependency testbed/vim:24 is also built unless the platform is x86_64, since it is also only provided for this platform.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
-FROM testbed/vim:24
+ARG TESTBED_VIM_VERSION=24
+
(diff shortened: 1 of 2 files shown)


w0rp (approved):

w0rp:
Cheers! :beers:

Issue #3279, opened 2020-08-10
Add support for LSP textDocument/formatting as a fixer


cuviper:
**Name:** LSP `textDocument/formatting`
**URL:** https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_formatting

Some language servers can format your code -- it would be nice to use this for `ALEFix`.

See also #2718 for `textDocument/onTypeFormatting`.

OJFord:
Not sure if you want a separate issue for it (or each instance of it) @w0rp, but it's a more general issue that language servers can be used for formatting, but aren't included (I'm not even sure if they can be, since the I/O mechanism is different?) as available fixers.

I was just looking at `pyls`, which has a `black` plugin, but as far as I'm aware this can't be used in ale today; it's needed instead to use `pyls` as the linter, and `black` (the independent executable) as the fixer.

rgossiaux:
ALE seems to be in a kind of half-state here where it supports a few important LSP features directly (eg go to definition) but is missing some others like this one. For me this feature is critical and without it I need to use a different LSP client.

Is the intention longer-term for ALE to become a more full-featured LSP client, supplanting something like vim-lsp? Or is the intention to leave the LSP client features to other plugins and to just focus on being a hub for diagnostics/fixers from many sources?

w0rp:
The features that are supported or not are pretty much based on whatever I use personally, or whatever people implement support for with clean enough code in a pull request. I'll add in anything that looks good enough.

I've also set up ALE so any other plugin can feed in lists of data for diagnostics so you can use ALE in combination with anything that does that. I tend to add in more LSP features very slowly as I'm not much of an IDE guy. (I care more about getting the results of linting/static analysis, and quick-fixing them.) That's been best implemented by coc.nvim, and I put it in the README here: https://github.com/dense-analysis/ale#faq-coc-nvim I'd like it if there were plugins for bridging any of them. No extra support is needed in ALE for that, people just need to write some Vim scripts which use the existing API. (See `:help ale-lint-other-sources`)

I tend to only care about formatting personally when I can also run CI checks to complain when formatting hasn't been done, so an entire team of programmers is forced to make all code look the same, but in very permissive way that doesn't have your code constantly flip-flopping between different formats. Otherwise you just get random people in a team pushing different changes to files, which causes a bunch of merge conflicts.

[...]

PR #4427, opened 2023-01-22
Adding a fixer implementation for gopls

enckse:
Adds an implementation of a fixer for `gopls` (which is going to be similar input/output to `goimports` but built-in to `gopls`)

I've done a pass at trying to get this setup properly (at a minimum it works locally) but given that there is already a `gopls` implementation for other aspects of ale if I have done anything out of bounds or irregular, etc. - let me know and I'm happy to update my PR.


neersighted (approved):
This LGTM given that we support `gopls` as a linter, and https://github.com/dense-analysis/ale/issues/3279 is unsolved.

PR #4430, opened 2023-01-26
State ALE works with vim 8.0 and above

hsanson:
Readme stated ALE works with vim 8 and many thought it meant it works only with vim 8 and not with vim 9.


w0rp (approved):

w0rp:
Cheers! :beers:

PR #4462, opened 2023-02-23
change the default linters of rust

yuesong-feng:
issue#4461

RLS has been deprecated and is no longer supported. It has been replaced with [rust-analyzer](https://rust-analyzer.github.io/).
Now, rust-analyzer is the official LSP for rust. It is necessary to change the default rust linter of ALE too.

### DIFF
--- a/autoload/ale/linter.vim
+++ b/autoload/ale/linter.vim
@@ -33,7 +33,7 @@ let s:default_ale_linter_aliases = {
 "
 " No linters are used for plaintext files by default.
 "
-" Only cargo and rls are enabled for Rust by default.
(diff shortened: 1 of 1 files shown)


nospam2998:
How about also completely removing rls with this PR, since it is deprecated and no longer possible to use?

Please feel free to look at [label: old tool, is_closed](https://github.com/dense-analysis/ale/issues?q=label%3A%22old+tool%22+is%3Aclosed) for inspiration from previous clean-ups of old tools.

w0rp:
> How about also completely removing rls with this PR, since it is deprecated and no longer possible to use? 

We can take it out of the default list and keep the linter definition around for a good while. People with old Rust installs can still use it if they really want to.

PR #4471, opened 2023-03-07
Added instructions for pathogen install

TheBearodactyl:


hsanson (commented):
Thanks, looks ok.

hsanson (approved):

w0rp:
I took them away because I figure "everyone should just be using the built-in package management anyway, and I don't know anyone who uses Pathogen," but I'll leave it there if people are using it for ALE.

Issue #4472, opened 2023-03-08
Support column alignment for virtual text.


systemmonkey42:
The default virtual text setup causes multiple lines of errors/warning to `cuddle` with the ends of the buffer text.
I find this unsightly and a visual distraction.    In languages like 'c' where a single mistype can generate a page of errors on
every line, the resulting visual mess is worse than useless.

I would like to suggest two changes
1. Filter errors in virtualtext#SetTex() so only ONE error is added per buffer line.
2. Modify virtualtext#ShowMessage() to support left aligning all messages at a specific column.

As a test, I added something like the following to test virtualtext.vim around line 177
```vim
 silent! call prop_add(l:line, 0, {
          \   'type': l:hl_group,
          \   'text':  ' ' . l:msg,
          \   'bufnr': a:buffer,
          \   'text_padding_left': 50-strdisplaywidth(getline(l:line)),
 \})
```

This force all warning to align at the 50th column.  Lines already longer than 50 columns had their warnings automatically.

A similar modification to virtualtext#SetText()
```vim
let l:seen = {}

for l:item in a:loclist
    if l:item.bufnr == a:buffer
        let l:line = max([1, l:item.lnum])
        if !has_key(l:seen,l:line)
           call ale#virtualtext#ShowMessage(a:buffer, l:item)
           let l:seen[l:line] = 1
        endif
    endif
endfor
```

This eliminates the mess caused by having multiple errors/warnings per line (only the first was usually of any use).

With these two changes, I find the virtual text feature much more usable.

Thanks

w0rp:
> the resulting visual mess is worse than useless

You don't need to make hyperbolic statements like these.

I'll have a read of your pull request.

w0rp:
I recommend reading this issue: https://github.com/dense-analysis/ale/issues/4208

Generally I find there are linters, and especially a few worst offenders, where with a little bit of VimL  we can make the output a lot better, and more efficient to process. Some tools output an incredible number of errors if just one thing is wrong, especially in larger files. This slows down Vim a lot, and the output isn't helpful at all. One of my many ideas, and I plan on implementing everything mentioned in that issue, is to alter the output of select tools to produce more meaningful information telling you what is actually wrong with a file instead of screaming at you that everything is wrong.

systemmonkey42:
> You don't need to make hyperbolic statements like these.

My spologies.  I certainly wasn't criticising you or your work.

[...]

PR #4473, opened 2023-03-08
Virtualtext formatting and alignment fixes

systemmonkey42:
Example implementation of the enhancement described in #4472 

This include minimum and maximum columns expected for messages to start, and filtering to ensure only
one error appears per line of code.

### DIFF
--- a/autoload/ale/virtualtext.vim
+++ b/autoload/ale/virtualtext.vim
@@ -28,6 +28,11 @@ let g:ale_virtualtext_prefix =
 " Controls the milliseconds delay before showing a message.
 let g:ale_virtualtext_delay = get(g:, 'ale_virtualtext_delay', 10)
 
+" Controls the positioning of virtualtext
+let g:ale_virtualtext_column = get(g:, 'ale_virtualtext_column', 0)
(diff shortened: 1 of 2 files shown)


w0rp (changes requested):
Could you do the following?

1. Add videos or GIFs into the PR description to show a "before" and "after" for the changes.
2. Localise the new blocks of code into a new function or two to stop functions getting any larger.
3. Add comments in code to explain what the new code is doing.
4. Add Vader tests for code where you can. See: `test/test_virtualtext.vader`.
5. Change text alignment to only be applied when a new settings are enabled. I may want this on or off by default.
6. Add documentation for the new settings and how they work.
7. Set default values for new settings.

Aligning text is a good idea. In my editor, I want all warnings and errors to appear at the end of lines, but not so far to the right that they don't fit on the screen. If you see a file full of errors... fix the file so that doesn't happen, or turn the linter off.

w0rp:
The GIFs look good. :+1:

The other issue I linked in the issue covers cleaning up spam from linters.

hsanson (approved):
Good improvement to visuals and I have confirmed it does not break things at least on my side.  We can think about tests latter.

Issue #4060, opened 2022-02-08
Mark error on failure to lint


nospam2998:
With ale's default configuration it enables every known linter and opportunistically attempts to run them all, silently failing in most cases.  While this yolo approach might be a reasonable out of the box, a more trusted ale experience can be obtained by explicitly configuring working linters only on known filetypes.

When using such a tightly controlled ale configuration, one might want to get immediate visual feedback whenever any linter fail to execute. Preferably a linter-like error on the first line of the buffer. Such functionality must be hidden behind a configuration option defaulting to disabled, of course. I made an attempt at <https://git.netizen.se/ale/commit/?h=wip/lint_fail_err>, but gave up and settled with the following work-around in my vimrc instead:

```vim
function! ALELinterProblem()
  let l:history = ale#history#Get(bufnr(''))
  if match(l:history, "'status': 1")
    echom l:history
    sign define missing_linter text=!! texthl=Error
    exe ':sign place 1 line=1 name=missing_linter'
  endif
endfunction

augroup ALEMissingLinter
  autocmd User ALELintPost unsilent call ALELinterProblem()
augroup END
```

While the triggering on the ALELintPost event kind of addresses the problem, having support in ale for reporting linter failure in a noticeable manner would be preferred.

Does the ALE team agree this functionality could at all be desired? If so, would anyone be able to give any hints on how to add it more completely and cleanly than on my branch linked above?

nospam2998:
No comment at all for the first thirteen months?

It should really be the responsibility of ale to report failures when failing to act as configured. I'm now tempted to claim this is a bug, rather than an enhancement. Linting becomes very unreliable when ale silently ignores this fatal error.

I've rebased the branch on top of current master, and am opening a PR for this.

nospam2998:
Varying response times is something I can relate to on a very personal level, and I do know that you're _good at life_ @w0rp. You're rightly making time for other interesting things apart from ale.

PR #4474 got killed by **stale** it seems (and reading [#719][stale719] horrified me!).

[...]

PR #4474, opened 2023-03-08
wip: Add error on lint execution failure

nospam2998:
Draft PR for #4060. Feedback is needed.


w0rp (changes requested):
I like the idea of ALE eating its own dog food to use the external sources API itself for showing errors on executable failures. I like also that the setting is off by default.

You could undraft this, add tests for the changes, and add documentation for the new setting.

PR #4475, opened 2023-03-08
Clear virtualtext no matter how enabled

living180:
When toggling ALE off, clear the virtualtext even when `g:ale_virtualtext_cursor` is `'all'` / `2`, not just when it is `'current'` / `1`.

### DIFF
--- a/autoload/ale/toggle.vim
+++ b/autoload/ale/toggle.vim
@@ -14,7 +14,7 @@ function! s:DisablePostamble() abort
         call ale#highlight#UpdateHighlights()
     endif
 
-    if g:ale_virtualtext_cursor is# 'current' || g:ale_virtualtext_cursor == 1
+    if g:ale_virtualtext_cursor isnot# 'disabled' && g:ale_virtualtext_cursor != 0
         call ale#virtualtext#Clear(bufnr(''))
     endif
 endfunction


hsanson (changes requested):
Would be great if the tests in `test_ale_toogle.vader` are modified to test for this, in specific the `ALEToggle should reset everything and then run again` test.

Not familiar with virtualtext implementation, specially since it has some differences between Vim and NeoVim implementations so cannot provide too much insights. Maybe @w0rp can help here.

Adding a method to `virtualtext.vim` that returns the number of virtual texts in the buffer could help to assert for the count  before and after ALE is disabled/enabled.

living180:
Unfortunately, I'm just tweaking the feature, and before I submitted the PR, I looked for relevant tests and didn't find any. :slightly_frowning_face:  ALE appears to use three virtualtext implementations, Neovim, Vim-native for vim-9.0.0297 and newer, and emulation for older versions of Vim.  For determining whether a buffer contains ALE virtualtexts, it looks like for Neovim, the `nvim_buf_get_extmarks()` function might be relevant, for Vim-native, `prop_list()` is probably the best option, while for emulation checking the value of `s:last_virt` is probably what should be used.  I'm by no means a Vimscript expert, but could probably hack together a function that reports whether or not a buffer contains ALE virtualtexts.  But actually getting the tests right for all three implementations is not something I'm confident I could do.

w0rp (approved):

w0rp:
Cheers! :beers:

Issue #4327, opened 2022-10-04
Rust analyzer config is ignored


stevenengler:
## Information

**VIM version**

VIM - Vi IMproved 8.1 (2018 May 18, compiled Sep 19 2022 04:59:57)
Included patches: 1-2269, 3612, 3625, 3669, 3741

Ubuntu 20.04

## What went wrong

I'm attempting to disable the `buildScripts` and `procMacro` options for rust analyzer. In my vimrc I have:

```vim
let g:ale_rust_analyzer_executable = '/home/user/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/rust-analyzer'
let g:ale_rust_analyzer_config = {'cargo': {'buildScripts': {'enable': v:false}}, 'procMacro': {'enable': v:false}}
let g:ale_linters = {'rust': ['analyzer'], 'c': ['clangd'], 'cpp': ['clangd']}
let g:ale_completion_enabled = 1
let g:ale_enabled = 0
```

According to https://github.com/dense-analysis/ale/issues/3350#issuecomment-706020061, this should be the correct syntax for the rust analyzer config. The options I'm using are specified at https://rust-analyzer.github.io/manual.html#configuration.

Rust analyzer works, but is running the build scripts even though I set `ale_rust_analyzer_config` to disable them.

## Reproducing the bug

1. Use the above config.
2. Open a file from a rust project.
3. See that the build script was run.

### :ALEInfo

```
 ALEInfo
 Current Filetype: rust
Available Linters: ['analyzer', 'cargo', 'cspell', 'rls', 'rustc']
 Current Filetype: rust
Available Linters: ['analyzer', 'cargo', 'cspell', 'rls', 'rustc']
  Enabled Linters: ['analyzer']
  Ignored Linters: []
 Suggested Fixers:
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'rustfmt' - Fix Rust files with Rustfmt.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:
let g:ale_rust_analyzer_config = {'cargo': {'buildScripts': {'enable': v:false}}, 'procMacro': {'enable': v:false}}
let g:ale_rust_analyzer_executable = '/home/user/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/rust-analyzer'
 Global Variables:
let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = v:null
[...]

Issue #832, opened 2017-08-07
Implement project configuration files


w0rp:
Some people want to configure ALE differently for different projects. Support should be added for loading a project configuration file before linting or fixing is done. It should work like so.

1. Add an option for enabling per-project configuration files.
2. Add an option for changing the project file configuration name.
3. When project configuration files are enabled, and before fixing or linting is done, `:source` the configuration file to run whatever VimL code is contained within.

A variable might need to be set for the configuration file like `g:ale_active_buffer` for determining which buffer is being checked or fixed, as the current buffer (`bufnr('')`) might be a different buffer from the buffer being checked or fixed.

andrew-vimmer:
Should it? As long as the plugin can react on change of global variables while being active the [local vimrc](https://github.com/embear/vim-localvimrc) plugin solves the problem.

w0rp:
Yeah, that'll do the job. I won't implement this then. If anyone else asks how to handle this, I'll put it in the FAQ.

offero:
I just came here looking for the same way to have ale use eslint for 1 project and healthier for another.

w0rp:
You can use one of many Vim plugins that implement project configuration files and buffer-specific options in ALE. (Those that begin with `b:`). You can also try using `g:ale_pattern_options` to set options based on regular expression matches for paths.

Flimm:
Neovim has a way to run local `.nvim.lua`, `.nvimrc` or `.exrc` files in the current directory, if the file is in the trust list. See `:help 'exrc'`. Vim also has the `exrc` option, but it doesn't seem to have a way to mark files as trusted, so Vim's implementation is considered dangerous and insecure, as mentioned in Vim's `:help 'exrc'`.

Here's an example for Neovim, to disable ALE's automated fixing in a certain project. Create a file named `.nvim.lua` in your current directory, with this line:

    vim.g.ale_fix_on_save = 0

When you run Neovim in this directory for the first time, you will be asked whether you want to mark `.nvim.lua` in this directory as trusted or not. If you trust it, the hash of the file is saved. If the file changes contents, you will be asked again if you trust it or not.

PR #4479, opened 2023-03-11
Rust : Look for analyzer.json

yoann9344:
Add option `g:ale_rust_analyzer_use_local_config`. When set, it will `FindNearestFile` _analyzer.json_, then it will make an union with the global user config `g:ale_rust_analyzer_config`.

The point is to detect and use config by project.

### DIFF
--- a/ale_linters/rust/analyzer.vim
+++ b/ale_linters/rust/analyzer.vim
@@ -3,6 +3,7 @@
 
 call ale#Set('rust_analyzer_executable', 'rust-analyzer')
 call ale#Set('rust_analyzer_config', {})
+call ale#Set('rust_analyzer_use_local_config', 0)
 
 function! ale_linters#rust#analyzer#GetCommand(buffer) abort
     return '%e'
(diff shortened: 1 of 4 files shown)


yoann9344:
I used `readfile` and `json_decode`, is there any problem with asynchronous tasks ?

yoann9344:
Maybe it fixes https://github.com/dense-analysis/ale/issues/4327 I haven't check `g:ale_rust_analyzer_config`, before making the modifications. But it works fine after the MR (with and without local's config activated).

ale_linters/rust/analyzer.vim
```diff
@@ -26,10 +27,28 @@ function! ale_linters#rust#analyzer#GetProjectRoot(buffer) abort
     return ''
 endfunction
 
+function! ale_linters#rust#analyzer#GetConfig(buffer) abort
+    let l:config = copy(ale#Var(a:buffer, 'rust_analyzer_config'))
+
+    if ale#Var(a:buffer, 'rust_analyzer_use_local_config')
+        let l:config_local_path = ale#path#FindNearestFile(a:buffer, 'analyzer.json')
```
hsanson, line 34: As mentione in previous command I rather have this be a filename rather than a boolean. Is more flexible in that users can set any name for the config to match their projects conventions.

Check if the file is readable and if not then we assume no configuration file is present.

ale_linters/rust/analyzer.vim
```diff
@@ -3,6 +3,7 @@
 
 call ale#Set('rust_analyzer_executable', 'rust-analyzer')
 call ale#Set('rust_analyzer_config', {})
+call ale#Set('rust_analyzer_use_local_config', 0)
```
hsanson, line 6: Prefer if this is called  `rust_analyzer_config_file` and accepts a string with the name of the desired config file with default to "rust-analyzer.json".   

From what I understand the rust-analyzer does not define any official configuration file so I expect diferent project to use different names.

doc/ale-rust.txt
```diff
@@ -60,7 +60,17 @@ g:ale_rust_analyzer_config                         *g:ale_rust_analyzer_config*
...
+
+
+g:ale_rust_analyzer_use_local_config     *g:ale_rust_analyzer_use_local_config*
+                                         *b:ale_rust_analyzer_use_local_config*
+  Type: |Number|
+  Default: `0`
+
+  When set to 1, search for the closest config `analyzer.json` for
+  `rust-analyzer`. If found, use it to extend *g:ale_rust_analyzer_config*.
+  It is usefull for using automatically a different config by project.
```
hsanson, line 74: Would be nice to have minimal config example and a link to the official docs:

  - https://rust-analyzer.github.io/manual.html#configuration

hsanson (changes requested):

ale_linters/rust/analyzer.vim
yoann9344: But we need a try/catch anyway, 'cause the json content might be mis-formatted, in this case to would flood the the buffer.

ale_linters/rust/analyzer.vim
hsanson: I think this is overcomplicating things and may have performance impact. Searching files recursively can be heavy depending on how deep the folder structure is.

I would recommend leaving this to single file. Users can employ other methods or plugins to create configurations specifc for each project they work on (see discussion https://github.com/dense-analysis/ale/issues/832).

PR #4480, opened 2023-03-13
More virtual text work

living180:
The change I made for # #4475 did not work correctly because when Vim compares a non-numeric string with a number using `!=`, it converts the string to `0` before comparing.  Thus the second half of the clause (`g:ale_virtualtext_cursor != 0`) was always false if `g:ale_virtualtext_cursor` was `'all'` or `'current'`.  Fix by using `isnot#`, which does not coerce strings to numbers before comparing.

Also update to use `nvim_buf_set_extmark()` when available in preference to the now-deprecated `nvim_buf_set_virtual_text()`.

Finally add some tests to ensure that virtual texts get displayed when ALE is enabled and cleared when ALE is disabled.

### DIFF
--- a/autoload/ale/toggle.vim
+++ b/autoload/ale/toggle.vim
@@ -14,7 +14,9 @@ function! s:DisablePostamble() abort
         call ale#highlight#UpdateHighlights()
     endif
 
-    if g:ale_virtualtext_cursor isnot# 'disabled' && g:ale_virtualtext_cursor != 0
+    if g:ale_virtualtext_cursor isnot# 'disabled'
+    \&& g:ale_virtualtext_cursor isnot# 0
+    \&& g:ale_virtualtext_cursor isnot# '0'
         call ale#virtualtext#Clear(bufnr(''))
     endif
 endfunction
--- a/autoload/ale/virtualtext.vim
+++ b/autoload/ale/virtualtext.vim
@@ -253,3 +253,39 @@ function! ale#virtualtext#SetTexts(buffer, loclist) abort
         endif
     endfor
 endfunction
+
+if get(g:, 'ale_private_testing_mode', 0)
+    function! ale#virtualtext#IsSupported() abort
+        return s:has_virt_text
+    endfunction
+
+    function! ale#virtualtext#IsEmulated() abort
+        return s:emulate_virt
+    endfunction
+
+    function! ale#virtualtext#CountTexts(buffer) abort
+        if !s:has_virt_text
+            return 0
+        elseif s:emulate_virt
+            return s:last_virt != -1
+        elseif has('nvim')
+            if exists('*nvim_buf_get_extmarks')
+                return len(nvim_buf_get_extmarks(
+                \   a:buffer,
+                \   s:ns_id,
+                \   0,
+                \   -1,
+                \   {'details': 0}
+                \))
+            else
+                throw 'Cannot determine the number of virtual texts'
+            endif
+        else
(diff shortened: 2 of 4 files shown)


autoload/ale/virtualtext.vim
```diff
@@ -245,3 +253,37 @@ function! ale#virtualtext#SetTexts(buffer, loclist) abort
...
+
+function! ale#virtualtext#IsSupported() abort
+    return s:has_virt_text
+endfunction
+
+function! ale#virtualtext#IsEmulated() abort
+    return s:emulate_virt
+endfunction
+
+function! ale#virtualtext#CountTexts(buffer) abort
```
w0rp, line 265: Last move this code to the test file instead, as it's just for tests.

test/test_ale_toggle.vader
```diff
@@ -74,6 +75,57 @@ Before:
     return l:results
   endfunction
 
+  function! TestVirtualtext(buffer, disabled, current, all)
```
w0rp, line 78: Add `delfunction` in `After` so it doesn't leak into other tests.

w0rp (changes requested):
Looks pretty good to me. Just needs some test tweaks.

test/test_ale_toggle.vader
living180: Done.  Thanks for the feedback!

autoload/ale/virtualtext.vim
living180: OK, so this is my first foray into Vimscript, so it is very possible that I'm missing something.  But all three of these functions access script-local variables, so I don't know of a good way to move them into another file.  If there is a clean way to handle this that is being used in other tests, could you point me to an example I could emulate?

autoload/ale/toggle.vim
```diff
@@ -14,7 +14,9 @@ function! s:DisablePostamble() abort
         call ale#highlight#UpdateHighlights()
     endif
 
-    if g:ale_virtualtext_cursor isnot# 'disabled' && g:ale_virtualtext_cursor != 0
+    if g:ale_virtualtext_cursor isnot# 'disabled'
+    \&& g:ale_virtualtext_cursor isnot# 0
+    \&& g:ale_virtualtext_cursor isnot# '0'
```
living180, line 19: So I had a question about this `if` clause.  While this is definitely an edge case, if within a single session a user had `g:ale_virtualtext_cursor` enabled, then set it to `'disabled'`, and then ran `ALEDisable`, the virtual texts would not be cleared.  Would it be better to call `ale#virtualtext#Clear()` uncoditionally?

autoload/ale/virtualtext.vim
w0rp: I'll see myself if I can eliminate the need for these being in the file here. I don't like to overburdern the scripts with things just for tests, but sometimes it's the only way. I'm pretty sure we can rewrite the same conditions as above into tests or simplify them in tests with version checks, given that we only test against certain versions in Docker anyway, and we can probably just get all available marks instead of specific ones in tests. Something like that.

living180: OK, so I just pushed up a new version with an alternate approach:  now the functions are only defined if the `g:ale_private_testing_mode` variable is set to true, and that variable is only defined in `test/vimrc`.  That way the functions don't have to be modified, but the global namespace is kept clean when not testing.  Thoughts?

PR #4488, opened 2023-03-28
remove `--enable-all` from golangci-lint options

v3s1e:
When using `golangci-lint`, configured via `.golangci.yml` with a custom ruleset, ALE does not report any errors and the command it executes exits with exit code 3:

```
Command History:
(executable check - success) golangci-lint
(finished - exit code 3) ['/Users/user/.vim/shell', '-c', 'cd ''/path'' && ''golangci-lint'' run ''file.go'' --enable-all']
<<<NO OUTPUT RETURNED>>>
```

The reasoning behind the exit code is that `--enable-all` cannot be combined with a `.golangci.yml` file that defines its own ruleset:

```
$ cd /path && golangci-lint run file.go --enable-all
ERRO Running error: can't combine options --enable-all and --enable errcheck
```

The proposed change is to remove the default `--enable-all` because every project uses different rules.

### DIFF
--- a/ale_linters/go/golangci_lint.vim
+++ b/ale_linters/go/golangci_lint.vim
@@ -1,7 +1,7 @@
 " Author: Sascha Grunert <mail@saschagrunert.de>
 " Description: Adds support of golangci-lint
 
-call ale#Set('go_golangci_lint_options', '--enable-all')
+call ale#Set('go_golangci_lint_options', '')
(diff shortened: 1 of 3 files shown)


hsanson (approved):
Thanks, looks like sensible change  considering that flag is deprecated:

  - https://github.com/golangci/golangci-lint/issues/1878#issuecomment-809646180

PR #4491, opened 2023-04-02
Fix 4490 - Fix pyright not running with poetry

hsanson:
These seems to fix a problem where pyright is not being invoked when using poetry or pipenv.

### DIFF
--- a/ale_linters/python/pyright.vim
+++ b/ale_linters/python/pyright.vim
@@ -65,7 +65,7 @@ endfunction
 function! ale_linters#python#pyright#GetCommand(buffer) abort
     let l:executable = ale_linters#python#pyright#GetExecutable(a:buffer)
     let l:exec_args = l:executable =~? 'pipenv\|poetry$'
-    \   ? ' run pyright'
+    \   ? ' run pyright-langserver'
     \   : ''
     let l:env_string = ''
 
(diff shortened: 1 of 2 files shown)


hsanson:
@w0rp original code used only "pyright" when invoking it via poetry/pipenv but my tests indicates it must use "pyright-serverlang" instead.  I do not use these tools frequently so please review if this change won't break things.

peterzhu1992:
Hi @hsanson confirms it fixed the problem:
```
(executable check - success) pipenv
(started) ['/usr/bin/bash', '-c', 'cd ''<>'' && ''pipenv'' run pyright-langserver --stdio']
```
Thanks.

evmorov:
`pyright` doesn't work for me in the main repo when I use it with `pipenv`. My config:

```
let g:ale_python_auto_pipenv = 1
let g:ale_python_pyright_auto_pipenv = 1
let g:ale_python_pyright_use_global = 0
```

But this PR fixes my problem. Now I have:

```
(executable check - success) pipenv
(started) ['/bin/bash', '-c', 'cd ''/mypath'' && ''pipenv'' run pyright-langserver --stdio']
```

Thanks!

hsanson:
Since issue seems to affect many and I got confirmation this fixes it I am skipping due review process and merging.

PR #4492, opened 2023-04-03
Fix error from ansible-lint versions >=6.11.0.

SkrrtBacharach:
The JSON output format of ansible-lint has changed since 6.11.0. Issue locations can have either a 'positions' or a 'lines' member, rather than just a 'lines' member as it was before. This fix checks which member is present, and passes that member name to subsequent dictionary lookups.

The error was caused by the following change:
https://github.com/ansible/ansible-lint/pull/2897


hsanson:
There seems to be linter issues.  Please fix those.

SkrrtBacharach:
> There seems to be linter issues. Please fix those.

Should be fixed now.

hsanson (approved):
Looks good, thanks.

PR #4494, opened 2023-04-04
Replace `solhint` default formatter

hbarcelos:
# TL;DR

`solhint` dropped the `compact` formatter. Use `unix` formatter instead.

# Investigation

A [recent change](https://github.com/protofire/solhint/blob/master/CHANGELOG.md#340---2023-02-17) in `solhint` dropped the built-in `eslint` formatters in favor of their own.

I am not a maintainer of `solhint`, so I don't have the full context, but apparently [they decided to not include the `compact` formatter](https://github.com/juanpcapurro/solhint/tree/4bd4d1f7cada0f30f691cb8e833816183dcc1491/lib/formatters), which was used in ALE.

Looking at the available formatters, I decided to stick with `unix`, since it's a well known format. There was a small issue differentiating linter issues from syntax errors, because the outputs are pretty similar.

I had to resort to ~dark sorcery,~ a negative look-behind for "Parse error" in the message to be able to distinguish between them. If I'm being honest, the parsing regexes are a bit frail right now, but they do the job.

### DIFF
--- a/autoload/ale/handlers/solhint.vim
+++ b/autoload/ale/handlers/solhint.vim
@@ -18,29 +18,30 @@ function! ale#handlers#solhint#Handle(buffer, lines) abort
     " /path/to/file/file.sol: line 1, col 10, Error - 'addOne' is defined but never used. (no-unused-vars)
     let l:output = []
 
-    let l:lint_pattern = '\v^[^:]+: line (\d+), col (\d+), (Error|Warning) - (.*) \((.*)\)$'
+    " let l:lint_pattern = '\v^[^:]+: line (\d+), col (\d+), (Error|Warning) - (.*) \((.*)\)$'
+    let l:lint_pattern = '\v^[^:]+:(\d+):(\d+): %(Parse error: )@<!\ze(.*)\s+\[(Error|Warning)\/([^\]]+)\]$'
 
     for l:match in ale#util#GetMatches(a:lines, l:lint_pattern)
-        let l:isError = l:match[3] is? 'error'
+        let l:isError = l:match[4] is? 'error'
         call add(l:output, {
         \   'lnum': l:match[1] + 0,
         \   'col': l:match[2] + 0,
-        \   'text': l:match[4],
+        \   'text': l:match[3],
         \   'code': l:match[5],
         \   'type': l:isError ? 'E' : 'W',
         \})
     endfor
 
-    let l:syntax_pattern = '\v^[^:]+: line (\d+), col (\d+), (Error|Warning) - (Parse error): (.*)$'
+    " let l:syntax_pattern = '\v^[^:]+: line (\d+), col (\d+), (Error|Warning) - (Parse error): (.*)$'
+    let l:syntax_pattern = '\v^[^:]+:(\d+):(\d+): Parse error: (.*)\s+\[Error\]$'
 
     for l:match in ale#util#GetMatches(a:lines, l:syntax_pattern)
-        let l:isError = l:match[3] is? 'error'
         call add(l:output, {
         \   'lnum': l:match[1] + 0,
         \   'col': l:match[2] + 0,
-        \   'text': l:match[5],
-        \   'code': l:match[4],
-        \   'type': l:isError ? 'E' : 'W',
+        \   'text': l:match[3],
+        \   'code': 'Parse error',
+        \   'type': 'E',
         \})
     endfor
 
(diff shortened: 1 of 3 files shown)


autoload/ale/handlers/solhint.vim
```diff
@@ -18,29 +18,30 @@ function! ale#handlers#solhint#Handle(buffer, lines) abort
...
         \   'col': l:match[2] + 0,
-        \   'text': l:match[4],
+        \   'text': l:match[3],
         \   'code': l:match[5],
         \   'type': l:isError ? 'E' : 'W',
         \})
     endfor
 
-    let l:syntax_pattern = '\v^[^:]+: line (\d+), col (\d+), (Error|Warning) - (Parse error): (.*)$'
+    " let l:syntax_pattern = '\v^[^:]+: line (\d+), col (\d+), (Error|Warning) - (Parse error): (.*)$'
```
hsanson, line 35: I understand solhint no longer supports the older formatter but this change is too recent and some users may be still stuck using older versions and be required to use the current formatter.  In this cases this would be a breaking change on ALE's side.  Instead of replacing the previous pattern is better to support both.

Simply keep the previous code intact and add a second pass over the lines using the new pattern adding the matches to the output array.

autoload/ale/handlers/solhint.vim
```diff
@@ -94,5 +95,5 @@ function! ale#handlers#solhint#GetCommand(buffer) abort
 
     return ale#node#Executable(a:buffer, l:executable)
     \   . (!empty(l:options) ? ' ' . l:options : '')
-    \   . ' --formatter compact %s'
+    \   . ' --formatter unix %s'
```
hsanson, line 98: Again there may be users tha cannot upgrade their solhint version and need the "compact" formatter so this cannot be easily removed.  Rather modify this linter to use "ale#semver#RunWithVersionCheck()" and use either compact or unix formatter depending on the version of solhint tool.

A good example on how to do this is the `ansible_lint.vim` linter.

test/handler/test_solhint_handler.vader
```diff
@@ -51,15 +51,14 @@ Execute(The solhint handler should parse linter error messages correctly):
   \   },
   \ ],
   \ ale#handlers#solhint#Handle(bufnr(''), [
-  \   'contracts/Bounty.sol: line 1, col 17, Warning - Compiler version must be fixed (compiler-fixed)',
```
hsanson, line 54: Instead of modifying these tests add new ones with the new pattern.  ALE should be able to parse and report both the older and newer patterns.

test/linter/test_solhint.vader
```diff
@@ -2,7 +2,7 @@ Before:
   call ale#assert#SetUpLinterTest('solidity', 'solhint')
   runtime autoload/ale/handlers/solhint.vim
 
-  let b:args = ' --formatter compact %s'
```
hsanson, line 5: Modify the test to check command is correct depending on the version of solhint used. You can see the `test_ansible_lint.vader` test for example on how to do this.

hsanson (changes requested):

autoload/ale/handlers/solhint.vim
hbarcelos: Hey... Sorry about the long delay.
Apparently `solhint` added the `compact` formatter back after I made this PR.
However, the `unix` formatter is available in every single version, so these changes would not break the integration.

Apparently, only `v3.4.0` is missing the `compact` format though, so it might be the case that this PR is not required anymore.

w0rp:
I can't re-open this pull request unless it's rebased again, but I would like to. The key to getting this problem solved is to chain the command with `ale#semver#RunWithVersionCheck` like a few other linters do, and to change the command we run based on the version string we can read from whatever `--version` option `solhint` has. This can be used for either linters or fixers, as ALE has a very simple way of deferring execution kind of like a JavaScript `Promise`. For the `callback` function for handling the output, we can just apply both regular expression strings to the lines of output and make it possible to read the output of either command.

w0rp:
I had a comment I purged after I read more carefully. This is something we could have merged much sooner, as `solhint` versions going back to `v1.1.10` way back in 2018 have the `unix` formatter option. That sounds about as safe as it gets for a non-breaking change for users. I'll prepare a new commit now crediting you which fixes this and also cleans up the code a bit. The location of the `solhint` code in the `handlers` directory isn't needed, and the code is overly verbose and lacks tests for the command and executable.

PR #4496, opened 2023-04-05
Add support for Bicep when installed as a plugin to Azure CLI

carlsmedstad:
The compiler for Microsoft's DSL Bicep can be installed both independently and as a plugin to Azure CLI. The latter is probably how most people install it.

The program output is the same but Azure CLI wraps the arguments and has a slightly different interface, hence I opted to copy the old linter and modify it to match the plugin arguments.

Not entirely sure  this is the right approach to this, feedback appreciated

### DIFF
--- a/ale_linters/bicep/az_bicep.vim
+++ b/ale_linters/bicep/az_bicep.vim
@@ -1,0 +1,57 @@
+" Author: Carl Smedstad <carl.smedstad at protonmail dot com>
+" Description: az_bicep for bicep files
+
+let g:ale_bicep_az_bicep_executable =
+\   get(g:, 'ale_bicep_az_bicep_executable', 'az')
+
+let g:ale_bicep_az_bicep_options =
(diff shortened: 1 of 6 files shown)


carlsmedstad:
In order to not produce any output files in the users project, a separate output directory must be specified. Are there any examples of this in other linters? Creating a directory `/tmp/` would be an option.

hsanson:
Looks good... if it works for you remove the WIP and I can review/merge this.

There is a `ale#command#CreateDirectory(buffer)` function that creates temporary directories with unique names. Maybe can be used to keep the output files?

carlsmedstad:
Realized I could indeed use the same way of discarding the output files as the other linter. Also fixed some other things as described in the commit message in d773a512cececb30d84309ca192686980655bf53.

From my point of view this is ready for review/merge :)

hsanson (approved):
Thanks, looks good to me.

PR #4501, opened 2023-04-12
Add fourmolu fixer

pbrisbin:
Fourmolu is aversion of Ormolu that supports configuration. This fixer
was modeled after the Ormolu one, but using the "stack executable"
approach of the Brittany and Stylish Haskell fixers.


hsanson (approved):
Thanks, looks good.

PR #4169, opened 2022-04-29
Add support for selene Lua linter

deathlyfrantic:

PR #4516, opened 2023-04-30
Add support for 'Selene' linter in Luau

xactlyblue:
Adds built-in support for the Selene linter to work with luau as well. Selene allows for linting *.luau files by using "luau" as a base in your Selene configuration.

### DIFF
--- a/ale_linters/luau/selene.vim
+++ b/ale_linters/luau/selene.vim
@@ -1,0 +1,48 @@
+" Copied from ..\lua\selene.vim
+
+call ale#Set('luau_selene_executable', 'selene')
+call ale#Set('luau_selene_options', '')
+
+function! ale_linters#luau#selene#GetCommand(buffer) abort
+    return '%e' . ale#Pad(ale#Var(a:buffer, 'luau_selene_options'))
+    \   . ' --display-style=json -'
+endfunction
+
+function! ale_linters#luau#selene#Handle(buffer, lines) abort
+    let l:output = []
+
+    for l:line in a:lines
+        " as of version 0.17.0, selene has no way to suppress summary
+        " information when outputting json, so stop processing when we hit it
+        " (PR for this here: https://github.com/Kampfkarren/selene/pull/356)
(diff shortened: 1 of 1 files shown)


ale_linters/luau/selene.vim
```diff
@@ -0,0 +1,48 @@
...
+    \   . ' --display-style=json -'
+endfunction
+
+function! ale_linters#luau#selene#Handle(buffer, lines) abort
+    let l:output = []
+
+    for l:line in a:lines
+        " as of version 0.17.0, selene has no way to suppress summary
+        " information when outputting json, so stop processing when we hit it
+        " (PR for this here: https://github.com/Kampfkarren/selene/pull/356)
```
Kampfkarren, line 17: This was merged a while ago

xactlyblue: Not sure about that one, literally just copied over the lua linter and changed all lua references to luau since selene works for both.

hsanson (changes requested):
There are two ways I would approach this:

1. Simplest Way

In the linter definition simply use `ale_linters#lua#selene#GetCommand` and `ale_linters#lua#selene#Handle` methods. No need to create different luau versions of these if the code is exactly the same.

This also avoids the need to duplicate the tests etc.  The disadvantage is that you won't be able to set different executable or options for luau. If this is a problem then is ok to keep the GetCommand function separate for Lua and Luau.  In this case make sure to add a test for the Luau GetCommand function.

2. Leave the Luau GetCommand and Handle methods as they are but pleas add respective tests.  Check the Selene PR for details on the tests:

  - https://github.com/dense-analysis/ale/pull/4169/files


Also since this is a different file type make sure to add entries in these files:

  - supported-tools.md
  - doc/ale.txt
  - doc/ale-supported-languages-and-tools.md

Finally if we are keeping the luau_selene_executable and luau_selene_options then create a doc/ale-luau.txt  similar to  doc/ale-lua.txt that documents those configuration options.

PR #4521, opened 2023-05-18
🐛 Fix a Ruby deprecation warning in the ERB linter

purinkle:
Before, the ERB linter used positional arguments. Newer versions of Ruby have deprecated this method signature. We fixed the linter to use keyword arguments.


hsanson (approved):
Thanks. Looks good.

PR #4443, opened 2023-02-03
Add deadnix linter

apeschar:
[deadnix](https://github.com/astro/deadnix) detects unused variables in Nix code.

Screenshot 2023-02-03 at 13 46 21

PR #4528, opened 2023-05-28
Mention deadnix in supported tools

n8henrie:
Deadnix support was added in
https://github.com/dense-analysis/ale/pull/4443 but it seems not to have
been mentioned in the lists of supported tools.


hsanson (approved):
Good catch, thanks for the fix.

PR #4530, opened 2023-06-07
Update cairo linter to Cairo 1.0

0xHyoga:
Updated handler test for Cairo. This new version is compatible with the latest Cairo 1.0 release. Recently the language got upgraded and there is no point including the old Cairo, so i just replaced the whole ale_linters/cairo/starknet.vim file. Thanks!

### DIFF
--- a/ale_linters/cairo/sierra.vim
+++ b/ale_linters/cairo/sierra.vim
@@ -1,0 +1,54 @@
+" Author: 0xHyoga <0xHyoga@gmx.com>
+" Description: Report Starknet compile to sierra errors in cairo 1.0 code
+
+call ale#Set('cairo_sierra_executable', 'starknet-compile')
+call ale#Set('cairo_sierra_options', '')
+
+function! ale_linters#cairo#sierra#Handle(buffer, lines) abort
+    " Matches patterns like the following:
+    " Error: Expected ';' but got '('
+    "    --> /path/to/file/file.cairo:1:10:)
+    let l:pattern = '\v(error|warning): (.*)$'
+    let l:line_and_column_pattern = '\v\.cairo:(\d+):(\d+)'
+    let l:output = []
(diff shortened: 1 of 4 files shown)


hsanson (changes requested):
Changes to support Cairo 1.0 look good but I am not entirely sure we can just remove previous version without potentially breaking someones setup. Specially if this new version is very recent I can imagine there will be many projects still in migration process.

I would recommend we leave old patterns and only add a separate loop to check for the new patterns.

0xHyoga:
> Changes to support Cairo 1.0 look good but I am not entirely sure we can just remove previous version without potentially breaking someones setup. Specially if this new version is very recent I can imagine there will be many projects still in migration process.
> 
> I would recommend we leave old patterns and only add a separate loop to check for the new patterns.

The previous version of Cairo is deprecated, and should not be used any longer. To not break the setup of users still using old Cairo, i have added a new linter called Sierra (which is Cairo 1.0) under the Cairo folder. It would be up to users now to declare which linter to use, (starknet.vim for old and sierra.vim for new cairo) or can just use both.

0xHyoga:
Have also added the tests for both handlers:

new cairo 1.0 : /test/handler/test_sierra_handler.vader
old cairo: /test/handler/test_starknet_handler.vader

0xHyoga:
Wanted to comment on why I chose to create two linters instead of adding another loop:

With the current setup, in case a user has not upgraded then ALE will **only show the old linter** (Starknet) errors, and if they have then ALE will **only show the new linter** (Sierra) errors as the compiler uses different patterns. This avoids unnecessary code and loops in the new linter and helps us maintain it further. @hsanson

Thanks

hsanson (approved):
Sorry for the delay and thanks for the contribution.

PR #4531, opened 2023-06-07
fix: run opacheck on entire dir for more context

### LINKED ISSUE #2531: Support Rego

redbmk:
Running `opa check` on a single file can report errors that might not exist when considering other files in the same directory

Adds to #4199 to help close #2531

### DIFF
--- a/ale_linters/rego/opacheck.vim
+++ b/ale_linters/rego/opacheck.vim
@@ -11,7 +11,7 @@ function! ale_linters#rego#opacheck#GetCommand(buffer) abort
     let l:options = ale#Var(a:buffer, 'rego_opacheck_options')
 
     return ale#Escape(ale_linters#rego#opacheck#GetExecutable(a:buffer))
-    \   . ' check %s --format json '
+    \   . ' check . --format json '
     \   . (!empty(l:options) ? ' ' . l:options : '')
 endfunction
 
--- a/test/linter/test_rego_opacheck.vader
+++ b/test/linter/test_rego_opacheck.vader
@@ -8,9 +8,9 @@ After:
 
 Execute(The default command should be correct):
   AssertLinter 'opa',
-  \ ale#Escape('opa') . ' check %s --format json '
+  \ ale#Escape('opa') . ' check . --format json '
 
 Execute(The default command should be overridden):
   let b:ale_rego_opacheck_executable = '/bin/other/opa'
   AssertLinter '/bin/other/opa',
-  \ ale#Escape('/bin/other/opa') . ' check %s --format json '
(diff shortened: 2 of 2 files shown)


ale_linters/rego/opacheck.vim
```diff
@@ -11,7 +11,7 @@ function! ale_linters#rego#opacheck#GetCommand(buffer) abort
     let l:options = ale#Var(a:buffer, 'rego_opacheck_options')
 
     return ale#Escape(ale_linters#rego#opacheck#GetExecutable(a:buffer))
-    \   . ' check %s --format json '
+    \   . ' check . --format json '
```
w0rp, line 14: Because the command can start asynchronously, this could fire from a different buffer from the buffer you started on. Luckily, I added a quick way to get the directory of a file with `fnamemodify`-like syntax a while ago.

```suggestion
    \   . ' check %s:h --format json '
```

Try this instead, and let me know if it works.

w0rp (changes requested):

ale_linters/rego/opacheck.vim
w0rp: I noted in the documentation for `ale#linter#Define` that by default ALE doesn't set a working directory for commands as that can actually break a lot of things. Another thing you could try is adding `'cwd': '%s:h'` to the arguments below, and then `.` here should work. If you grep the codebase, a lot of linters explicitly set the work directory to be the directory of the file. Setting `cwd` instead may or may not work better here, it's better to just run the linter and see what happens.

redbmk: It's been a while since I've done anything with `opa` or `rego` but I just tried this out. For a file `%s` throws some errors due to missing context from the rest of the directory, while both `.` and `%s:h` seem to work fine.

I can see in `:ALEInfo` that using `%s:h` shows the full path to the directory and that seems simpler to me than setting the cwd first.

w0rp (approved):

w0rp:
Cheers! :beers:

PR #4532, opened 2023-06-09
floatwin: use win_execute in NeoVim for commands when it exists

BarrensZeppelin:
ALEs floating windows are broken in NeoVim 0.10 after https://github.com/neovim/neovim/pull/23711.
The `CursorMoved` event from switching windows is now sent after `s:NvimShow` exits, which triggers the autocommand to close the window.
`win_execute` is available in NeoVim, so we can use a similar approach as the one used in `s:VimShow` to avoid switching windows altogether.


w0rp (approved):
Looks like it should do the job. Cheers! :beers:

Issue #4533, opened 2023-06-10
Go: remove support for gometalinter and golint


rodrigc:
* https://github.com/alecthomas/gometalinter#go-meta-linter mentions that gometalinter has been deprecated/archived, in favor of golangci-lint
* https://github.com/golang/lint has been archived/frozen

These older linters do not support newer Go language features like [generics](https://gobyexample.com/generics).

I suggest that:
* golint be removed from the default linters enabled by ale
* enable golangci-lint by default in ale (replacing golint)
* remove gometalinter support in ale
* remove golint support support in ale

PR #4535, opened 2023-06-10
Remove golint support

rodrigc:
golint has been deprecated and was archived in 2019

See: https://github.com/golang/lint , where it mentions that `go vet` and `staticcheck` should be used instead.

Fixes: #4533


hsanson (approved):
Thanks.

PR #4540, opened 2023-06-21
fixed parsing errors when certain options are used in glslangValidator

JoseGRuiz:
I noticed that the linter stopped working when I attempted to use the -V option for checking code that will be used with Vulkan. After a bit of testing I realized that the pattern that is currently checked is incompatible with using certain flags. This is because the error reporting by glslangValidator is different when using at least these flags {-V, -G, --target-env}. For example without these flags set, output for error reporting looks like "ERROR: 0:5: 'foo' : undeclared identifier" and with any of the previously mentioned flags it looks like "ERROR: filename:5: 'foo' : undeclared identifier". Similar but not quite accepted by the pattern currently used, one downside is we no longer get a column number provided in the new report, although in my testing the column number reported is always 0 so I don't consider it very valuable as it is. However, with a bit of modification to the pattern matching code and no longer recording column number I managed to get both the, no options report, and the report with the aforementioned options set, to correctly report errors and be useful via ALE.

### DIFF
--- a/ale_linters/glsl/glslang.vim
+++ b/ale_linters/glsl/glslang.vim
@@ -17,13 +17,14 @@ function! ale_linters#glsl#glslang#Handle(buffer, lines) abort
     " Matches patterns like the following:
     "
     " ERROR: 0:5: 'foo' : undeclared identifier
-    let l:pattern = '^\(.\+\): \(\d\+\):\(\d\+\): \(.\+\)'
+    " or when using options like -V or -G or --target-env
+    " ERROR: filename:5: 'foo' : undeclared identifier
+    let l:pattern = '^\(.\+\): \(.\+\):\(\d\+\): \(.\+\)'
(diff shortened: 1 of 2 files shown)


hsanson (changes requested):
Appreciated if tests for the different error formats are added to `test/handler/test_glslang_handler.vader`.

JoseGRuiz:
I added a test that should cover the output from options -V and -G at least in format. I also noticed by using this patch for a while, that option such as -V should be used with '-o /dev/null', or whatever is appropriate on the target system. Unfortunately this glslangValidator program doesn't have an option to send output to stdout or dry-run. But if it is not used it is not a huge deal it just compiles the shader and saves it under a default name such as frag.spv. Still this is probably not the desired behavior for some users so if you have a better solution please suggest it.

hsanson (approved):

PR #4541, opened 2023-06-22
Fix project file search for haskell hls

alt-romes:
Previously, we would append the candidate root files to the output paths from `ale#path#Upwards(expand(#buffer:p:h))` which do not include a trailing forwardslash (/).
These would form invalid paths which would never trigger `filereadable`.

Furthermore, we need to return `l:path . '/'` because in `ale#haskell#hls#GetProjectRoot` we take `:h` based on the project root file. In the case a root file was found, we want to keep the directory it was found in, so we need to identify the path as a directory (again, with the forward slash)

The correct identification of root files allows us to use LSP in multi package projects, and in projects with a custom hie.yaml, such as GHC!

TODO
- [x] Tests


alt-romes:
I would be happy to add a test for this but I'm not sure in which category this falls, despite the documentation.

The test should
* Create a directory with some root file (e.g. `cabal.project`)
* Call `ale_linters#haskell#hls#FindRootFile`
* Result should match directory the `cabal.project` file is in, with a trailing slash.
* Call `ale_linters#haskell#hls#GetProjectRoot`
* Result should match directory the `cabal.project` file is in.

hsanson:
There are some tests for finding project roots that can serve as example on how to test this:

- ./test/test_swift_find_project_root.vader
- ./test/test_python_find_project_root.vader
- ./test/test_maven_find_project_root.vader
- ./test/test_gradle_find_project_root.vader
- ./test/test_ant_find_project_root.vader

alt-romes:
Thanks! I've added the tests accordingly.

This is ready to go.

alt-romes:
The test is failing because the "find project root" functions don't seem to be in scope for neovim-02-only.
I don't know why that's the case.

Help!

w0rp:
@alt-romes I've reworked this into a new commit crediting you now.

Cheers! :beers:

alt-romes:
Thanks!

Issue #3621, opened 2021-03-08
echomsg line causes abrupt Visual -> Normal mode change


cachance7:
## Information

**VIM version**

```
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Dec  3 2020 02:03:05)
macOS version
Included patches: 1-2050
Compiled by Homebrew
```

Operating System: Mac OSX 10.15.7 (19H2)

## What went wrong

Sometimes while highlighting text in a file in Vim using Visual mode, Vim will just abruptly switch to Normal mode and lose my highlight. I traced the line that shows up in the messages bar to Ale, specifically this one: https://github.com/dense-analysis/ale/blob/master/autoload/ale/cursor.vim#L40

If I disable that line, Vim no longer exhibits the buggy behavior.

## Reproducing the bug

1. Switch to Visual mode and start highlighting. 
2. Sometimes vim would abruptly leave Visual mode and dump me back into Normal mode. The following line would show up in the messages bar `:'<,'>echomsg l:message` when that happened.

### :ALEInfo

```

 Current Filetype: typescriptreact
Available Linters: ['deno', 'eslint', 'standard', 'tslint', 'tsserver', 'typecheck', 'xo']
  Enabled Linters: ['deno', 'eslint', 'standard', 'tslint', 'tsserver', 'typecheck', 'xo']
  Ignored Linters: []
 Suggested Fixers: 
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

 Global Variables:

let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = 0
let g:ale_command_wrapper = ''
let g:ale_completion_delay = v:null
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = v:null
let g:ale_disable_lsp = 0
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = '%code: %%s'
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_history_enabled = 1
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = 0
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
[...]

PR #4560, opened 2023-07-22
Don't echom if not in normal mode

amadeus:
I've been running into this issue for a couple years now: https://github.com/dense-analysis/ale/issues/3621 where often times I would make a visual selection only to be interrupted and booted out of visual mode back into normal mode and finally decided to do some digging into how to fix it. 

w0rp suggested a fix which I've attempted to apply to the `ale#cursor#Echom` method. Not entirely sure if this is the right way to fix it, but using it for a day it definitely hasn't repro'd for me since


w0rp (approved):

w0rp:
I'll merge this and see if this somehow causes anything to go wrong.

w0rp:
Cheers! :beers:

PR #4574, opened 2023-08-02
Fix error loading "text" syntax for hover

liskin:
rust-analyzer sometimes returns a hover result with language being "text", but there's no syntax/text.vim, so this would fail with:

    Error detected while processing function <SNR>150_VimOutputCallback[6]..<lambda>8[1]..ale#lsp#HandleMessage[30]..ale#hover#HandleLSPResponse[42]..ale#floating_preview#Show[13]..<SNR>161_VimShow:
    line   13:
    E484: Cannot open file syntax/text.vim

Only including the file when it actually exists fixes this.

(Note that I had to substitute javascript for typescript in `test/test_hover_parsing.vader` because there's no `syntax/typescript.vim` in vim 8.0 and neovim 0.2.)

> comment above and the documentation referenced in it first. Write tests!
>

There are some, yeah.

### DIFF
--- a/autoload/ale/hover.vim
+++ b/autoload/ale/hover.vim
@@ -70,6 +70,18 @@ function! s:ConvertLanguageName(language) abort
     return a:language
 endfunction
 
+" Cache syntax file (non-)existence to avoid calling globpath repeatedly.
+let s:syntax_file_exists_cache = {}
+
+function! s:SyntaxFileExists(syntax_file) abort
+    if !has_key(s:syntax_file_exists_cache, a:syntax_file)
+        let s:syntax_file_exists_cache[a:syntax_file] =
+        \   !empty(globpath(&runtimepath, a:syntax_file))
+    endif
+
+    return s:syntax_file_exists_cache[a:syntax_file]
+endfunction
+
 function! ale#hover#ParseLSPResult(contents) abort
     let l:includes = {}
     let l:highlights = []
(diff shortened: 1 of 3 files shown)


autoload/ale/hover.vim
```diff
@@ -160,10 +160,11 @@ function! ale#hover#ParseLSPResult(contents) abort
                 let l:language = s:ConvertLanguageName(l:language)
 
                 if !empty(l:language)
-                    let l:includes[l:language] = printf(
-                    \   'syntax/%s.vim',
-                    \   l:language,
-                    \)
+                    let l:syntax_file = printf('syntax/%s.vim', l:language)
+
+                    if !empty(globpath(&runtimepath, l:syntax_file))
```
w0rp, line 165: Let's write a script function `s:SyntaxFileExists` and store the result in script variable so we only run `globpath`  the first time we encounter a particular language. That will stop us scanning the filesystem each time we want to show a message. We wouldn't show highlights for new syntax files until Vim is restarted, but I think that's a low cost to pay.

w0rp (changes requested):

autoload/ale/hover.vim
liskin: Done: https://github.com/dense-analysis/ale/compare/36ace32666f2385a580c3b04193a6dc60e183cb5..04c123524426955bc8992aaf9db805f46746cbe5

w0rp (approved):

w0rp:
Cheers! :beers:

PR #4583, opened 2023-08-09
feat(terraform): add snyk linter for terraform

pbnj:
# To Do:

- [ ] add tests
- [ ] add docs

### DIFF
--- a/ale_linters/terraform/snyk.vim
+++ b/ale_linters/terraform/snyk.vim
@@ -26,7 +26,7 @@ function! ale_linters#terraform#snyk#Handle(buffer, lines) abort
                 call add(l:output, {
                 \   'filename': l:result.targetFilePath,
                 \   'lnum': l:iac_issues.lineNumber,
-                \   'text': l:iac_issues.iacDescription.issue,
+                \   'text': l:iac_issues.iacDescription.issue . ale#Pad('.') . ale#Pad(l:iac_issues.iacDescription.impact) . ale#Pad('.') . ale#Pad(l:iac_issues.iacDescription.resolve),
                 \   'code': l:iac_issues.publicId,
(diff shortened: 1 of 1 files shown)


ale_linters/terraform/snyk.vim
```diff
@@ -0,0 +1,61 @@
...
+"      https://github.com/snyk/cli
+
+call ale#Set('terraform_snyk_options', '')
+call ale#Set('terraform_snyk_executable', 'snyk')
+
+let s:separator = has('win32') ? '\' : '/'
+
+function! ale_linters#terraform#snyk#Handle(buffer, lines) abort
+    let l:output = []
+    let l:json = ale#util#FuzzyJSONDecode(a:lines, {})
```
w0rp, line 13:
```suggestion
    let l:json = ale#util#FuzzyJSONDecode(a:lines, [])
```

I think you might need a List as the default here.

ale_linters/terraform/snyk.vim
```diff
@@ -0,0 +1,61 @@
...
+endfunction
+
+" Construct command arguments to snyk with `terraform_snyk_options`.
+function! ale_linters#terraform#snyk#GetCommand(buffer) abort
+    let l:cmd = '%e iac test'
+
+    let l:opts = ale#Var(a:buffer, 'terraform_snyk_options')
+
+    if !empty(l:opts)
+        let l:cmd .= ale#Pad(l:opts)
```
w0rp, line 47: `ale#Pad` returns an empty string if the input is empty, so you can take the condition away and put the whole command in a `return` statement at the end. You can write` --json` into the command before the other options without `ale#Pad`.

w0rp (changes requested):
I'll check again when you've finished this.

PR #4584, opened 2023-08-11
fix(yaml): make actionlint respect config

pbnj:
When using [custom labels](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-labels-with-self-hosted-runners) for GitHub runners in Workflow files, `actionlint` complains that the label is invalid or unknown. `actionlint` allows for defining these custom labels in its [configuration file `.github/actionlint.yaml`](https://github.com/rhysd/actionlint/blob/main/docs/config.md#configuration-file). However, the `actionlint` ale linter is currently configured to lint in a temporary directory, thus losing context of the configuration file.

This PR removes the need to lint GitHub Workflow files in a temporary directory and, instead, lints GitHub Workflows in-place.


w0rp (approved):

w0rp:
Cheers! :beers:

PR #4585, opened 2023-08-11
Support format xml file which haven't yet persisted at the disk

jiz4oh:
Hi, xmllint is invalid if the buffer is not saved for now. and the pr can fix it


w0rp (commented):
Does `xmllint` use the path to the file as part of its linting at all? If so, one thing we could try here is using the temporary filename if the buffer hasn't been saved yet, and the path to the file if it has been saved. We can have tests for each case.

jiz4oh:
@w0rp, fine, have updated the pr according your suggestion

w0rp (approved):
Cheers! 🍻

PR #4594, opened 2023-08-19
Add support for Packwerk

ymap:
Packwerk (https://github.com/Shopify/packwerk) is a Ruby gem used to enforce boundaries and modularize Rails applications.


w0rp (approved):
Cheers! :beers:

PR #4597, opened 2023-08-23
fix(neovim): ignore unrelated diagnostics

haoqixu:
In some cases, linters may produce unrelated diagnostics for the buffer. For example, when enabling `g:ale_go_golangci_lint_package`, the whole Go package will be checked instead of only the current file.

If `g:ale_use_neovim_diagnostics_api` is enabled, we will see unrelated diagnostics belonging to other files/buffers.

This PR filters out the unrelated diagnostics before calling the `vim.diagnostic.set` api.


w0rp (changes requested):
Looks pretty good. Just one comment.

w0rp (approved):

w0rp:
Cheers! :beers:

PR #4602, opened 2023-09-04
go/golangci_lint.vim: set package linting to 1

extemporalgenome:
The present default of 0 causes excessive and unnecessary typecheck errors in all but the most trivial (single-file) packages. The standard Go unit of code is a directory, not a file, and optimizing for processing a single file provides negligible savings in processing time for a typical project, yet creates a lot of noise for the user.

Since those users that will specifically want single-file processing are less common, it's arguably reasonable for them to configure this back to zero in their vimrc.


w0rp (changes requested):
I'm happy to change the default. We just need to update our tests and documentation to reflect the change.

PR #4617, opened 2023-09-26
fix tempfile for phpmd, to be able to use phpmd 2.14.0

shadowwa:
phpmd, until version 2.13.0 was taking the file to test as first argument (%s) and ignored the last argument (%t)
since version 2.14.0 if a last argument is given, phpmd exits with this message:
```
Can't find the custom report class: cleancode,codesize,controversial,design,naming,unusedcode
```
Moreover, phpmd is checking the filename hence show diagnostic for saved buffer only.

With this modification, phpmd 2.14.0 is able to run correctly and the diagnostic is done on the temp file provided by Ale and is refreshed even if the buffer is not yet saved.

### DIFF
--- a/ale_linters/php/phpmd.vim
+++ b/ale_linters/php/phpmd.vim
@@ -7,9 +7,9 @@ let g:ale_php_phpmd_executable = get(g:, 'ale_php_phpmd_executable', 'phpmd')
 let g:ale_php_phpmd_ruleset = get(g:, 'ale_php_phpmd_ruleset', 'cleancode,codesize,controversial,design,naming,unusedcode')
 
 function! ale_linters#php#phpmd#GetCommand(buffer) abort
-    return '%e %s text'
+    return '%e %t text'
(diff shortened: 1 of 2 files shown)


w0rp:
How does phpmd behave in older versions if supplied with these arguments? If it works just as well for versions that go back a few years, we can make this change.

If this changes breaks older versions, we can support both versions by chaining the command with a version check. Have a search for `ale#semver` in the codebase. Note that `AssertLinter` can accept a `List` for the expected output to check every command in a chain that is run to test version commands.

shadowwa:
I've tried with the earliest versions available on packagist.org:
- phpmd 1.5.0 (2013-07-26) works fine out of the box with theses arguments.
- phpmd 1.4.0 (2012-09-07) works fine with theses arguments (but since ruleset **cleancode** was introduce in phpmd 1.5.0,  you just have to remove **cleancode** from _g:ale_php_phpmd_ruleset_ option in _.vimrc_, if you want to test it).

w0rp (approved):

w0rp:
Cheers! :beers:

Issue #4487, opened 2023-03-28
ALE eslint cannot find config


theahura:
## Information

**VIM version**
VIM - Vi IMproved 9.0 (2022 Jun 28, compiled May 10 2022 08:40:37)          
Included patches: 1-749   

Operating System: 
Ubuntu 20.04.3 LTS

## What went wrong

Apologize if this is not the right place for this! Could be considered a configuration issue, not sure.

I switched from Syntastic to ALE recently and noticed that ALE was not able to find my organization's `eslint.config.js` file. The organization has a folder structure like so:
```
foo/
  node_modules/
    eslint/
  packages/
    app1/
    app2/
    app3/
      eslint.config.js
```
Basically we have a monorepo with multiple apps and hoist shared dependencies (like eslint) to the top level, but individual apps can have their own lint configurations among other things. 

ALE says it cannot find the config:
```
Oops! Something went wrong! :(

ESLint: 8.35.0

ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:

  npm init @eslint/config

ESLint looked for configuration files in /home/user/foo/packages/app3/src and its ancestors. If it found none, it then looked in your home directory.
```
When I run ALEInfo, I see the following:
```
(finished - exit code 2) ['/bin/bash', '-c', 'cd ''/home/user/foo'' && ''/home/user/foo/node_modules/eslint/bin/eslint.js''
-f json --stdin --stdin-filename ''/home/user/foo/packages/app3/src/index.js'' < ''/tmp/vGyg5eP/4/index.js''']
<<<OUTPUT STARTS>>>
Oops! Something went wrong! :(
ESLint: 8.35.0
ESLint couldn't find a configuration file. To set up a configuration file f
or this project, please run:
    npm init @eslint/config
ESLint looked for configuration files in /home/user/foo/packages/app3/src and its ancestors. If it found none, it then looked in your home
directory.
If you think you already have a configuration file or if you need more help
, please stop by the ESLint Discord server: https://eslint.org/chat
<<<OUTPUT ENDS>>>
```
The reason I don't think this is an eslint bug is because I can run eslint just fine in my app package. For what its worth, Syntastic too did not have any issues. And yet, through ALE, eslint is unable to find any config. I also tried running that eslint command in my own bash terminal and it just hangs, not sure whats up with that. If I remove the `--stdin --stdin-filename` and just run eslint on the file in question, it works fine in bash, but again not in ALE

Would appreciate any help!

### :ALEInfo
[...]

PR #4637, opened 2023-10-19
fix(eslint): handle monorepo setup for eslint

bodymindarts:
The heuristic for finding the correct Cwd when executing eslint does not work in the context of a monorepo setup where the `.eslintrc.js` may not live in the root of the repo but in one of the leaf packages ie. each package may have its own settings.

This PR fixes the behaviour by setting Cwd to the closest config path instead of the bin path.

Fixes https://github.com/dense-analysis/ale/issues/4487


w0rp:
I'll have a think about this. ESLint keeps changing over the years and this change could break something.

w0rp:
This looks like the correct behaviour to me. You'll have to fix the tests. See `:help ale-dev`.

PR #4639, opened 2023-10-20
Fix #4251 --file-filter option removed for cppcheck handler

### LINKED ISSUE #4251: cppcheck does not find file + solution
## Information

**VIM version**

VIM - Vi IMproved 9.0

Operating System: 
Arch Linux

## What went wrong

cppcheck does not work although the other linters and fixers for C++ projects work. The command `ALEInfo` shows the following relevant error message.
```sh
(executable check - success) cppcheck
(finished - exit code 1) ['/bin/zsh', '-c', 'cd ''/home/matheus/c++/tools-cpp'' && ''cppcheck'' -q --language=c++ --template=''{file}:{line}:{column}: {severity}:{inconclusive:inconclusive:} {message} [{id}]\\n{code}'' --project=''build/compile_commands.json'' --file-filter=''Tools.cpp'' --enable=style ''/tmp/vDHHcxp/2/Tools.cpp''']

<<<OUTPUT STARTS>>>
cppcheck: error: could not find any files matching the filter.
<<<OUTPUT ENDS>>>
```

## Solution
I even found the solution but am not able of writing Vim Script. **At the beginning of --file-filter there needs to be a "*".** The following command works.
```sh
cppcheck -q --language=c++ --template="{file}:{line}:{column}: {severity}:{inconclusive:inconclusive:} {message} [{id}]\\n{code}" --project="build/compile_commands.json" --file-filter="*Tools.cpp" --enable=style /tmp/
```

## Reproducing the bug

1. Install this plugin.
2. Create a C++ project and the suitable compile_commands.json.
3. Open a file.

juanmitaboada:
--file-filter option has been removed because using it prevents cppcheck from processing the file if it is in a different path

### DIFF
--- a/autoload/ale/handlers/cppcheck.vim
+++ b/autoload/ale/handlers/cppcheck.vim
@@ -48,9 +48,8 @@ function! ale#handlers#cppcheck#GetCompileCommandsOptions(buffer) abort
     let [l:dir, l:json_path] = ale#c#FindCompileCommands(a:buffer)
 
     " By default, cppcheck processes every config in compile_commands.json.
-    " Use --file-filter to limit to just the buffer file.
     return !empty(l:json_path)
-    \   ? '--project=' . ale#Escape(l:json_path[len(l:dir) + 1: ]) . ' --file-filter=' . ale#Escape(bufname(a:buffer))
+    \   ? '--project=' . ale#Escape(l:json_path[len(l:dir) + 1: ])
     \   : ''
 endfunction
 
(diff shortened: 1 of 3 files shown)


w0rp:
I'll have to have a think about this before I merge it. It's probably right.

w0rp:
Projects often have a structure such as `src/*.cpp` for all source files. Running `cppcheck` on the entire directly could be slow without a file filter, which I belive is the reason the filter is in place. I think the solution the problem here is the following:

1. Edit the linter so it runs in a specific `cwd`.
2. Keep `--filte-filter` and so it to a path relative to the `cwd` that was selected for the linter.

ALE has functions in `ale#path` that can be used for this purpose. You can do that and update the tests to test for it.

Thanks for looking into this!

Issue #4609, opened 2023-09-13
Add support for `ruff format`


JaRoSchm:
**Name:** ruff
**URL:** https://github.com/astral-sh/ruff/

In their last release ruff added a new formatter option. The style is intended to be compatible with black.
The documentation of the formatter can be found here: https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_formatter/README.md

treyhunner:
`ruff format` was in alpha. It's now in beta and [the ruff team has deemed it production-ready](https://astral.sh/blog/the-ruff-formatter).

It would be great to see support for this!

crimsonknave:
#4645 added the `ruff_format` fixer. This issue should be resolved.

JaRoSchm:
Thanks!

PR #4645, opened 2023-10-27
Adding support for ruff formatter

crimsonknave:
Adds support for the new ruff formatter as requested in #4609. I took the existing ruff fixer and made a version that runs the `format` command. That way the user can opt into either or both of them.

This is built on top of #4347 and #4393 and contains only a few tweaks (the version checking was removed because the formatter was added well after the version in question).


w0rp (dismissed):
Nice. Looks fine to me.

Cheers! 🍻

w0rp:
Have a look at the CI issues.

w0rp (approved):
Cheers! :beers:

Issue #4593, opened 2023-08-18
`ormolu` fixer doesn't work out of the box


eahlberg:
## Information

**VIM version**
VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Jan 01 1980 00:00:00)
Included patches: 1-1441

Operating System: 
NixOS 23.05.589.a558f7ac29f (Stoat)

`ormolu` version:
```
> ormolu --version
ormolu 0.5.0.1 UNKNOWN UNKNOWN
using ghc-lib-parser 9.2.7.20230228
```

`vim-ale` version:
I believe it is `2023-08-05` ([this commit](https://github.com/dense-analysis/ale/commit/15cbc0e912acf31d8185dbef00473e9555591350)). From [this nix channel](https://github.com/NixOS/nixpkgs/commit/ef8288935ba859fc3b30632fa6e04705f81b9c2a) (updated an hour ago at time of writing).

## What went wrong
Ormolu fixer doesn't format the file. The command history contains the following line (note exit code 9). See `:ALEInfo` section below for full output.
```
(finished - exit code 9) ['/run/current-system/sw/bin/bash', '-c', '''ormolu'' < ''/tmp/vKSqUsU/5/About.hs''']
```

## Reproducing the bug
1. Enable the `ormolu` fixer, open a Haskell source file which isn't formatted according to `ormolu`'s rules and save and trigger the formatting.
2. Observe that the source file wasn't formatted.

### :ALEInfo
<details>
  <summary>Expand</summary>

 Current Filetype: haskell
Available Linters: ['cabal_ghc', 'cspell', 'ghc', 'ghc_mod', 'hdevtools', 'hie', 'hlint', 'hls', 'stack_build', 'stack_ghc']
   Linter Aliases:
'cabal_ghc' -> ['cabal-ghc']
'ghc_mod' -> ['ghc-mod']
'stack_build' -> ['stack-build']
'stack_ghc' -> ['stack-ghc']
  Enabled Linters: ['hlint']
  Ignored Linters: []
 Suggested Fixers: 
  'brittany' - Fix Haskell files with brittany.
  'floskell' - Fix Haskell files with floskell.
  'fourmolu' - A formatter for Haskell source code.
  'hfmt' - Fix Haskell files with hfmt.
  'hindent' - Fix Haskell files with hindent.
  'hlint' - Refactor Haskell files with hlint.
  'ormolu' - A formatter for Haskell source code.
[...]

PR #4654, opened 2023-11-20
Fix Haskell `ormolu` fixer

eahlberg:
1. Fix https://github.com/dense-analysis/ale/issues/4593 by enabling `ormolu`'s `--stdin-input-file` option by default.
2. Switch to the same design as other Haskell fixers (see https://github.com/dense-analysis/ale/pull/1851), which AFAIU is to improve support for `stack`. Note: I haven't been able to test this since I haven't found a way to make `stack` work on my setup (`NixOS` and `cabal`).

### DIFF
--- a/autoload/ale/fixers/ormolu.vim
+++ b/autoload/ale/fixers/ormolu.vim
@@ -11,7 +11,7 @@ function! ale#fixers#ormolu#ApplyFixForVersion(buffer, version) abort
     let l:executable = ale#fixers#ormolu#GetExecutable(a:buffer)
     let l:options = ale#Var(a:buffer, 'haskell_ormolu_options')
 
-    if ale#semver#GTE(a:version, [0, 3, 0])
+    if ale#semver#GTE(a:version, [0, 8, 0])
         let l:args = ' --stdin-input-file %s'
     else
         let l:args = ' %s'
--- a/autoload/ale/fixers/ormolu.vim
+++ b/autoload/ale/fixers/ormolu.vim
@@ -1,12 +1,20 @@
 call ale#Set('haskell_ormolu_executable', 'ormolu')
 call ale#Set('haskell_ormolu_options', '')
 
-function! ale#fixers#ormolu#Fix(buffer) abort
+function! ale#fixers#ormolu#GetExecutable(buffer) abort
     let l:executable = ale#Var(a:buffer, 'haskell_ormolu_executable')
+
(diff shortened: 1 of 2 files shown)


autoload/ale/fixers/ormolu.vim
```diff
@@ -1,12 +1,20 @@
...
+    let l:executable = ale#fixers#ormolu#GetExecutable(a:buffer)
     let l:options = ale#Var(a:buffer, 'haskell_ormolu_options')
 
     return {
-    \   'command': ale#Escape(l:executable)
-    \       . (empty(l:options) ? '' : ' ' . l:options),
+    \   'command': l:executable
+    \       . (empty(l:options) ? '' : ' ' . l:options)
+    \       . ' --stdin-input-file '
+    \       . ale#Escape(@%),
```
w0rp, line 18: `@%` may not be the current file at the time the linter runs, as this code could be executed from a different buffer. Instead get the filename using the `a:buffer` variable. ALE has a shorthand for this where you can use `%s` in the command string. See `:help ale-command-format-strings`.

Have a look at changelogs and make sure the `--stdin-input-file` argument is available in old enough `ormolu` versions. If it's not available in older versions, you can make this fixer function still compatible with older versions with the `ale#semver` functions and a version check. You can grep the codebase for other fixers that run version checks.

w0rp (changes requested):

autoload/ale/fixers/ormolu.vim
eahlberg: Sorry for being slow to respond but thanks for the feedback!

> @% may not be the current file at the time the linter runs, as this code could be executed from a different buffer. Instead get the filename using the a:buffer variable. ALE has a shorthand for this where you can use %s in the command string. See :help ale-command-format-strings.

👍

> Have a look at changelogs and make sure the --stdin-input-file argument is available in old enough ormolu versions. If it's not available in older versions, you can make this fixer function still compatible with older versions with the ale#semver functions and a version check. You can grep the codebase for other fixers that run version checks.

`ormolu` is a Haskell program which uses [Haskell PVP](https://pvp.haskell.org/) which AFAIU means the `ale#semver` functions won't work. I couldn't find usage of PVP version checks in the codebase, but I guess one option would be to create a similar `ale#pvp` module. Does that sound reasonable or do you know of a better way to solve this?

PR #4656, opened 2023-11-21
Ruff use json-lines output format

0xhtml:
Using the json-lines output format allows for setting of the end_line, end_col and code field of the handle output.

Additionally, the first letter of the code is used to determine the type field.

### DIFF
--- a/ale_linters/python/ruff.vim
+++ b/ale_linters/python/ruff.vim
@@ -47,7 +47,7 @@ function! ale_linters#python#ruff#GetCommand(buffer, version) abort
 
     " NOTE: ruff version `0.0.69` supports liniting input from stdin
     " NOTE: ruff version `0.1.0` deprecates `--format text`
-    return ale#Escape(l:executable) . l:exec_args
+    return ale#Escape(l:executable) . l:exec_args . ' -q'
     \   . ale#Pad(ale#Var(a:buffer, 'python_ruff_options'))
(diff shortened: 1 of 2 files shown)


w0rp (approved):
I check other linters more carefully for changes to arguments, but this is probably good and I'll take this. I don't yet consider `ruff` to be stable enough to be used in my day job or for Dense Analysis code, but I'm looking forward to the future where after a few more bug fixes it can completely replace `flake8` and others for me, because the improvements to execution time are great.

Cheers! 🍻

ale_linters/python/ruff.vim
```diff
@@ -47,22 +47,25 @@ function! ale_linters#python#ruff#GetCommand(buffer, version) abort
...
-        \   'lnum': l:match[1] + 0,
-        \   'col': l:match[2] + 0,
-        \   'text': l:match[3],
+        \   'lnum': l:item.location.row,
+        \   'col': l:item.location.column,
+        \   'end_lnum': l:item.end_location.row,
+        \   'end_col': l:item.end_location.column - 1,
+        \   'code': l:item.code,
+        \   'text': l:item.message,
+        \   'type': l:item.code =~ '\vE\d+' ? 'E' : 'W',
```
w0rp, line 68:
```suggestion
        \   'type': l:item.code =~? '\vE\d+' ? 'E' : 'W',
```

Case sensitivity needs to be explicit, or it can be altered by configuration. The linter points this out. I'll commit the patch.

w0rp (approved):

PR #4661, opened 2023-11-24
Fix chktex highlighting wrong column when using tabs instead of spaces

### LINKED ISSUE #723: Wrong elements highlighted when using tabs instead of spaces (at least on LaTeX)
This seems a bit obvious but I did not find any related reports. I am suspecting I may have missed something or have something wrong in my .vimrc. Here is my .vimrc: https://github.com/jackgu1988/dotfiles/blob/master/vimrc

When I am using tabs on LaTeX and there is an error or a warning on a line, the wrong element is highlighted. For example here, the opening parenthesis bracket should have been highlighted because there is no space before it (line 88):

untitled

When I replace tabs with spaces it works just fine.
Is it my fault, or is it a bug?

Jorengarenar:
Fixes #723

chktex implemented [bug #56486: Allow setting options on the command line][1]
Thanks to that we can tell it to treat tab character as of one space width, i.e. one char.
That means, after we translate the output back to Vim columns, we get correct numbers.

[1]: https://savannah.nongnu.org/bugs/?56486

### DIFF
--- a/ale_linters/tex/chktex.vim
+++ b/ale_linters/tex/chktex.vim
@@ -16,6 +16,8 @@ function! ale_linters#tex#chktex#GetCommand(buffer) abort
     let l:command = ale#Var(a:buffer, 'tex_chktex_executable')
     " Avoid bug when used without -p (last warning has gibberish for a filename)
     let l:command .= ' -v0 -p stdin -q'
+    " Avoid bug of reporting wrong column when using tabs (issue #723)
+    let l:command .= ' -s TabSize=1'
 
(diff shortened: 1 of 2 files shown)


w0rp (dismissed):
Looks like we might be missing a test for the `chtex` command, or one might be broken by this. Could you add a test or edit one? See `:help ale-dev`.

Jorengarenar:
Yeah, there's no command test. I'll try writing one then

hsanson (approved):
Seems ok to me.

PR #2477, opened 2019-05-06
Add clangcheck linter for C

nganhkhoa:
This issue is stalled even though it is simple #976.

Document is to be added later on.

PR #4662, opened 2023-11-24
Add clang-check linter for C

### LINKED ISSUE #976: Add clangcheck to the pure c linters
As of today, I added `clang-check` support only for cpp files, because I didn't feel the need for it in c files, and because all [the example on llvm's website](http://clang.llvm.org/docs/ClangCheck.html) did not hint towards use for C files.

I don't know if the tools are really meant to be used for C (but I see no reason why not), so porting the tool to C is mostly harmless I guess.

TODO : 
- Create the `ale_linters/c/clangcheck.vim` file from the `cpp` version
- Change the last call to define to add a c linter instead of a cpp linter
- Change the naming of the options accordingly
- Write the doc

I'll do it when I have time and/or if it becomes really important somehow. But this will surely comes after finishing with #725

Jorengarenar:
Closes #976 

A continuation of abandoned PR #2477

### DIFF
--- a/ale_linters/c/clangcheck.vim
+++ b/ale_linters/c/clangcheck.vim
@@ -1,0 +1,38 @@
+" Author: gagbo <gagbobada@gmail.com>
+"       : luibo <ng.akhoa98@gmail.com>
+"       : Jorengarenar <jorengarenar@outlook.com>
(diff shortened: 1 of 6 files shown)


w0rp:
Cheers! :beers:

PR #4669, opened 2023-12-02
Add support for Scarb in `cairo` files

swan-of-bodom:
Added support for Cairo files when used with Scarb projects (package manager).

test: `/test/handler/test_scarb_handler.vader`

thanks

### DIFF
--- a/autoload/ale/handlers/cairo.vim
+++ b/autoload/ale/handlers/cairo.vim
@@ -1,0 +1,41 @@
+" Author: 0xhyoga <0xhyoga@gmx.com>,
+" Description: This file implements handlers specific to Cairo
+"
+function! ale#handlers#cairo#HandleCairoErrors(buffer, lines) abort
+    " Matches patterns like the following:
+    " Error: Expected ';' but got '('
+    "    --> /path/to/file/file.cairo:1:10:)
+    let l:pattern = '\v(error|warning): (.*)$'
+    let l:line_and_column_pattern = '\v\.cairo:(\d+):(\d+)'
+    let l:exclude_pattern = '\vcould not compile.*'
+    let l:output = []
+
+    for l:line in a:lines
+        let l:match = matchlist(l:line, l:pattern)
+
+        if len(l:match) == 0
+            let l:match = matchlist(l:line, l:line_and_column_pattern)
+
+            if len(l:match) > 0
+                let l:index = len(l:output) - 1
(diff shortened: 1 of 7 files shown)


doc/ale-supported-languages-and-tools.txt
```diff
@@ -101,6 +101,7 @@ Notes:
   * `gcc` (`cc`)
   * `uncrustify`
 * Cairo
+  * `scarb`
```
w0rp, line 104: Mark the tool as requiring the file to be saved here and in Markdown. See other examples.

autoload/ale/handlers/cairo.vim
```diff
@@ -0,0 +1,41 @@
+" Author: 0xhyoga <0xhyoga@gmx.com>,
+" Description: This file implements handlers specific to Cairo
+"
+function! ale#handlers#cairo#HandleCairoErrors(buffer, lines) abort
```
w0rp, line 4: If the function isn't generally useful other than for `scarb`  you could put this in the `scarb.vim` file. Either way is good.

doc/ale-cairo.txt
```diff
@@ -13,3 +26,5 @@ g:ale_cairo_starknet_executable               *g:ale_cairo_starknet_executable*
   Overrides the starknet-compile binary after installing the cairo-language.
 
   For more information read 'https://starknet.io/docs/quickstart.html'
+
+===============================================================================
```
w0rp, line 30: Check other files to see if this should be there or not. Follow the most common example. I can't remember what is best.

w0rp (changes requested):

doc/ale-supported-languages-and-tools.txt
swan-of-bodom: Done!

autoload/ale/handlers/cairo.vim
swan-of-bodom: I'd rather have it separate as it could be useful for future Cairo/Scarb tools. Could we leave this here?

doc/ale-cairo.txt
swan-of-bodom: From what I can tell in `/doc/` all files end with a separator so i just left it as is.

autoload/ale/handlers/cairo.vim
w0rp: Yeah.

w0rp:
Cheers! :beers:

Issue #2991, opened 2020-02-04
Add support for rubyfmt


lloeki:
**Name:** rubyfmt
**URL:** https://github.com/penelopezone/rubyfmt/tree/v0.2.0

`rubyfmt` is like `gofmt`, but for Ruby.

Similarly to `gofmt`, the command takes either input as stdin or a filename as first argument, and returns the formatted code on stdout.

```
$ echo 'def foo; 42; end' > foo.rb
$ rubyfmt foo.rb; echo $?            
def foo
  42
end
0
```

When an error is encountered, there is output on stderr and the exit code is nonzero:

```
$ echo 'def foo 42 end' > foo.rb
$ rubyfmt foo.rb > /dev/null; echo $?
Got a parse error while reading #<File:foo.rb (closed)>
true
bailing with exit code 1
1
```

PR pending for Homebrew: https://github.com/Homebrew/homebrew-core/pull/49775

nospam2998:
It appears support for rubyfmt was added in #4678, almost 2 years ago.

Would you @lloeki please consider closing this, or add a comment in case there is any reason to keep it open?

lloeki:
Thanks, closing.

PR #4678, opened 2023-12-11
add: support for `rubyfmt` as `ruby` fixer

yining:
[rubyfmt](https://github.com/fables-tales/rubyfmt) is a formatter for
 `ruby` code.

This commit adds support for `rubyfmt` as a `ruby` fixer (#2991), together with some tests and documentation.


hsanson (approved):

PR #4681, opened 2023-12-15
Add wiki link to ShellCheck linting

blmaier:
The ShellCheck integrations wiki suggests providing users with more info by linking back to the ShellCheck wiki.

https://www.shellcheck.net/wiki/Integration#linking-to-the-wiki


hsanson (approved):
LGTM, thanks.

Issue #4372, opened 2022-12-01
phpstan filename mappings not respected when running via docker


adam-ja:
<!--
  This is the template for reporting ALE bugs. Make sure you try updating ALE
  to a more recent version before reporting a bug. Look through existing bug
  reports for similar issues before reporting a new one. Don't leave comments
  about new bugs in the comment section for old issues.

  Make sure to try disabling other plugins and trying to repeat your bug before
  reporting it in ALE. Some times problems can arise when two plugins are used
  together, but often your issues might be problems with other plugins.
-->

## Information

**VIM version**

<!-- Paste just the first two lines of :version here. -->
NVIM v0.9.0-dev-377+gc9adbcafa                                                                                                                                                                                     
Build type: RelWithDebInfo 

Operating System: <!-- Describe your operating system version. -->
Ubuntu 20.04.5 LTS
## What went wrong

<!-- Describe what went wrong here. Be specific. -->

I have configured ale to run phpstan via my docker container with the following vim config:
```viml
let b:ale_php_phpstan_executable = '/home/adam/code/docker-phpstan.sh'
let b:ale_php_phpstan_use_global = 1
let b:ale_filename_mappings = {
\ 'phpstan': [
\   ['/home/adam/code/my-project', '/var/www'],
\ ],
\}
```
and this shell script:
```sh
#!/usr/bin/zsh
exec docker exec -i my_container vendor/bin/phpstan "$@"
```

When I put a simple error in a file, the `ALEInfo` output (see below) shows that phpstan is running and the error is being detected. However, when ale tries to read the output I get this error:

> Error detected while processing function <SNR>216_NeoVimCallback[35]..function <SNR>216_NeoVimCallback[29]..<lambda>53[1]..<SNR>213_ExitCallback[28]..<SNR>143_HandleExit[24]..ale_linters#php#phpstan#Handle:     
line    8:                                                                                                                                                                                                         
E716: Key not present in Dictionary: "/home/adam/code/my-project/modules/SomeModule/Jobs/SomeJob.php"

The issue seems to be that the phpstan output has the file keyed at `/var/www`:
```json
{"totals":{"errors":0,"file_errors":1},"files":{"/var/www/modules/SomeModule/Jobs/SomeJob.php":{"errors":1,"messages":[{"message":"If condition is always false.","line":47,"ignorable":true}]}},"errors":[]}
```
[...]

PR #4685, opened 2023-12-17
Add file mapping to phpstan

EdwinHuijsing:
This should solve issue #4372 "phpstan filename mappings not respected when running via docker".

Vim 8 caused a problem in one test, because the error vim returns is not the same as the other three.


EdwinHuijsing:
I just rerun the docker test `./run-tests -q` and no problems where reported.
"continuous-integration/appveyor/pr" runs on windows, looks like there is the cause of the test problem.
Will the patch work on windows? If the mapping is filled correct, i see no reason why it would give an problem

Any ideas who to fixes this? I can't test on windows (i don't have windows), so i am out of options.
I need a way to give the buffer a fixed full path, i have search before but i can't find way to give the buffer a fixed full path.

Regards,
Edwin

EdwinHuijsing:
Yesterday while working on an other linter, i found that we do not need the filename.
So i need to review the changes i did, locally i now have an version running that only
added two lines and changes one line.

More testing is needed before i can push the changes

Regards,
Edwin

hsanson (approved):
LGTM.

PR #4692, opened 2023-12-21
Add end_col and end_lnum to ShellCheck

blmaier:
ShellCheck supports a JSON format mode which includes an 'endLine' and
'endColumn' field.

We must use the newer 'json1' format as it properly treats tabs as a
single character. 'json1' was not supported until v0.7.0 in 2019[1], so
we maintain support for the older GCC based format.

[1] https://github.com/koalaman/shellcheck/blob/v0.9.0/CHANGELOG.md?plain=1#L121

This also contains a second commit which adds ShellCheck wiki links, to match my other PR https://github.com/dense-analysis/ale/pull/4681


hsanson (approved):
Thanks, looks good.

PR #4696, opened 2023-12-25
Gleam: Support for gleam language server

jpalardy:
This adds support for [gleamlsp](https://github.com/gleam-lang/gleam), which is shipped with the Gleam compiler/executable.

This PR is heavily influenced by other similar PRs for simple LSPs.

This is also my first PR for ALE, so I'm willing to fix and iterate as needed.

### DIFF
--- a/ale_linters/gleam/gleamlsp.vim
+++ b/ale_linters/gleam/gleamlsp.vim
@@ -1,0 +1,18 @@
+" Author: Jonathan Palardt https://github.com/jpalardy
+" Description: Support for Gleam Language Server
+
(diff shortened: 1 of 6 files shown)


jpalardy:
In retrospect, I don't know if `gleam_lsp` would have been more idiomatic. (rather than `gleamlsp`)

Thoughts?

jpalardy:
I've got some commits for `gleam format` fixer, which I can:
- add to this PR
- put in a separate PR

Let me know

hsanson (approved):
Thanks for the contribution.

PR #4697, opened 2023-12-25
Update zls.vim

coffeebe4code:
I was playing around with zig, and noticed that completion suggestion/hover do not work in ALE.

 I have not read ale-dev, and don't know how to run a plug .vim plugin locally, however based on looking at tsserver this might work. My thought process was that zls did not originally have this functionality when the initial zig linter was added.
 
 Please approve at your own risk. I won't have time to run tests, or figure out how to do this all locally for a while.


coffeebe4code:
apologies "stdio" is a valid option based on how does this server respond. rust-analyzer also has this option.

Issue #4698, opened 2023-12-26
Markdown hover text without language spec keep code fences


oliverralbertini:
I have an LSP ([pyre code-navigation](https://github.com/facebook/pyre-check)) returning code-fenced hover contents without a language specified, e.g.

```
"response": {"contents": "```\nFoobar\n```"}
```

Other LSPs may respond like this:

```
"response": {"contents": "```python\nFoobar\n```"}
```

Currently that means the backticks/code fences are not removed, so I see in the message bar 3 backticks, and if I hover into a floating window, I get the code + code fences, e.g.

````
```
Foobar
```
````

## Information

**VIM version**

```
:version
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Feb 09 2023 00:00:00)
Included patches: 1-2637
```

Operating System: CentOS9

## What went wrong

See above description

## Reproducing the bug

Requires an LSP that doesn't include a language hint in the code fence.

PR #4699, opened 2023-12-26
[hover] ParseLSPResult sets language 'text' for missing spec

oliverralbertini:
I have an LSP that is returning markdown code blocks on Hover with no language specified, e.g.

````
```
Foobar
```
````

As a result, you get "```" in the message line which is not that useful.

I made the regex to catch the first code fence accept empty language as well, and if it's empty, we set it to "text".

This makes it so that LSPs that return no language still produce legible restuls on the message line.

Fixes dense-analysis/ale#4698


hsanson (approved):
Thanks.

Issue #4657, opened 2023-11-22
ALE does not display any signs in NeoVim


kirsa:
## Information

**VIM version**

NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1699524327

Operating System:

MacOS 14.0 (23A344)

## What went wrong

Updated ALE via VimPlug update and now ALE does not display any signs in the gutter.

## Reproducing the bug

Open a python file with errors and no sigs are displayed. 

1. Open an empty file with .py extention

2. Add an obvious python error

3. Save file

4. See no signs showing in gutter

5. Run :sign place group=* to confirm that no ALE signs are active

### :ALEInfo

 Current Filetype: python
Available Linters: ['bandit', 'cspell', 'flake8', 'flakehell', 'jedils', 'mypy', 'prospector', 'pycln', 'pycodestyle', 'pydocstyle', 'pyflakes', 'pylama', 'pylint', 'pylsp', 'pyre', 'pyright', 'refurb', 'ruff', 'unimport', 'vulture']
   Linter Aliases:
'jedils' -> ['jedi_language_server']
  Enabled Linters: ['flake8', 'mypy', 'pydocstyle', 'pylint']
  Ignored Linters: []
 Suggested Fixers:
  'add_blank_lines_for_python_control_statements' - Add blank lines before control statements.
  'autoflake' - Fix flake issues with autoflake.
  'autoimport' - Fix import issues with autoimport.
  'autopep8' - Fix PEP8 issues with autopep8.
  'black' - Fix PEP8 issues with black.
  'isort' - Sort Python imports with isort.
  'pycln' - remove unused python import statements
  'pyflyby' - Tidy Python imports with pyflyby.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'reorder-python-imports' - Sort Python imports with reorder-python-imports.
  'ruff' - A python linter/fixer for Python written in Rust
  'ruff_format' - Fix python files with the ruff formatter.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
  'yapf' - Fix Python files with yapf.
 
 Linter Variables:
let g:ale_python_auto_pipenv = 0
let g:ale_python_auto_poetry = 0
let g:ale_python_auto_virtualenv = 0
let g:ale_python_flake8_auto_pipenv = 0
let g:ale_python_flake8_auto_poetry = 0
let g:ale_python_flake8_change_directory = 'project'
let g:ale_python_flake8_executable = 'flake8'
[...]

PR #4704, opened 2024-01-05
Use different group names for signs and virtual text

IngoMeyer441:
Since Neovim commit [c4afb9788c4f139eb2e3b7aa4d6a6a20b67ba156](https://github.com/neovim/neovim/commit/c4afb9788c4f139eb2e3b7aa4d6a6a20b67ba156), the sign API uses extmarks internally. Virtual text is already rendered using extmarks. ALE uses the same group name for both signs and virtual text and as a result, both are placed in the same extmark group. Since ALE deletes all extmarks in the virtual text group after all signs have been placed, no signs are ever shown. This PR fixes this by renaming the sign group from `ale` to `ale_signs`.

Fixes issue #4657.

### DIFF
--- a/autoload/ale/sign.vim
+++ b/autoload/ale/sign.vim
@@ -161,7 +161,7 @@ endfunction
 
 function! s:GroupCmd() abort
     if s:supports_sign_groups
-        return ' group=ale '
+        return ' group=ale_signs '
     else
         return ' '
     endif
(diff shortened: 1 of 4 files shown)


IngoMeyer441:
The tests seem to pass now...

hsanson (approved):
Seems to work without issues. Thanks.

PR #4705, opened 2024-01-08
Add full support for biome

### LINKED ISSUE #3280: Add support for Rome
**Name:** Rome frontend
**URL:** https://romefrontend.dev/

Rome is a linter, compiler, bundler, and more for JavaScript, TypeScript, JSON, HTML, Markdown, and CSS. At the current beta release only the linter is availble.

### LINKED ISSUE #4674: Add support for biome
**Name:** biome
**URL:** https://github.com/biomejs/biome

Because JavaScript cannot survive without yet another way of doing things: there's yet another way of doing things. Biome offers support for fixing and formatting code via the CLI and is mostly compatible with `prettier`. It offers a language server for checking code. It shouldn't be difficult to integrate it with ALE.

toxeus:
ALE currently has received very limited support for biome. Neither its linting capabilities nor its auto-formatter are exposed to ALE. Instead, the fixer implements a "lint-auto-fix" for a few lints that are auto-fixable by biome which is inconsistent with the behavior of other fixers that implement auto-format.  

Closes #3280


autoload/ale/fix/registry.vim
```diff
@@ -37,6 +37,11 @@ let s:default_registry = {
 \       'suggested_filetypes': ['bib'],
 \       'description': 'Format bib files using bibclean.',
 \   },
+\   'biome': {
+\       'function': 'ale#fixers#biome',
```
lanker, line 41:
```suggestion
\       'function': 'ale#fixers#biome#Fix',
```

hsanson (approved):
LGTM, thanks.

autoload/ale/fixers/biome.vim
```diff
@@ -1,17 +1,10 @@
...
+    let l:options = ale#Var(a:buffer, 'biome_options')
 
     return {
-    \   'command': (has('win32') ? (ale#Escape(l:node) . ' ') : '')
-    \       . ale#Escape(l:executable)
-    \       . ' check --apply'
-    \       . ale#Pad(l:options)
-    \       . ' %t',
-    \   'read_temporary_file': 1,
+    \   'command': '%e format'
```
lanker, line 6: Sorry for commenting on an already merged PR, but to me it seems like `%e` is not substituted with the executable when I run this. Switching to use `l:executable` directly works fine. Is `%e` supposed to work in this context?

ddominguez: @lanker I created a PR for this issue. #4733

PR #4710, opened 2024-01-20
Gleam: Support 'gleam format' as fixer

jpalardy:
This PR adds support for [gleam format](https://gleam.run/writing-gleam/command-line-reference/#format) which is shipped with the Gleam compiler/executable.

This is a follow up to #4696


hsanson (approved):
Looks good, thanks.

PR #4717, opened 2024-02-02
PHPactor works with file mapping

EdwinHuijsing:
PHPactor did not work on remote machines like docker.

The PHPactor constructor needs the remote path, this is now handled.

Documentation updated.


hsanson (approved):
LGTM, thanks.

PR #4726, opened 2024-02-23
Set working directory for Elvis linter

dmitrivereshchagin:
Most of the time it works to assume that the current working directory is the root of the project.  However, this is not the case for Rebar3 checked out dependencies, for example.

It's also worth noting that because of the way Elvis handles file patterns, and because directories in configuration are relative to the project root, the path supplied to command must be also relative.


hsanson (approved):
Thanks, do not use Elvis so cannot test but looks good.

PR #4732, opened 2024-03-01
Fix: ruff version `0.3.0` cli breaking changes

yining:
This commit appends `check` to the ruff executable if the version of specified ruff executable is `>= 0.3.0`, as ruff version `0.3.0` deprecates `ruff <path>` in favor of `ruff check <path>`:

https://github.com/astral-sh/ruff/releases/tag/v0.3.0


hsanson (approved):
Thanks for the fix.

PR #4733, opened 2024-03-01
Fix biome formatter

ddominguez:
I believe a bug was introduced to the biome fix command in this [PR](https://github.com/dense-analysis/ale/pull/4705/files#diff-619c20c541c9e30cad363d4db4b49a42025ee3dd5bd52703c62b5074f41d76d8R6)

This update adds the correct executable to the biome#Fix command.

* Added a fixer test


hsanson (approved):
Thanks for the fix.

Issue #4642, opened 2023-10-24
Delay when leaving insert mode using Esc


flux77:
## Information

**VIM version**
```
VIM - Vi IMproved 9.0 (2022 Jun 28, compiled May 04 2023 10:24:44)
Included patches: 1-1378, 1499
```

Operating System: Debian GNU/Linux 12 (bookworm)

Full contents of `~/.vimrc`:

```
packadd! ale
```

## What went wrong

Since commit dd3abf1a, ALE has caused <kbd>Esc</kbd> to leave insert mode slowly. Prior to that commit, Vim would leave insert mode instantly.

## Reproducing the bug

1. Start Vim.
2. Enter insert mode using <kbd>i</kbd>.
3. Leave insert mode using <kbd>Esc</kbd>.
4. Observe that there is a noticeable delay when leaving insert mode in step 3.

### :ALEInfo
<details>
  <summary>Expand</summary>
  
  
  
<pre>
 Current Filetype: 
Available Linters: []
  Enabled Linters: []
  Ignored Linters: []
 Suggested Fixers:
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 
 Global Variables:
let g:ale_cache_executable_check_failures = v:null
let g:ale_change_sign_column_color = v:null
let g:ale_command_wrapper = v:null
let g:ale_completion_delay = v:null
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = v:null
let g:ale_disable_lsp = 0
let g:ale_echo_cursor = 1
let g:ale_echo_msg_error_str = 'Error'
let g:ale_echo_msg_format = v:null
let g:ale_echo_msg_info_str = 'Info'
let g:ale_echo_msg_warning_str = 'Warning'
let g:ale_enabled = 1
let g:ale_fix_on_save = 0
let g:ale_fixers = {}
let g:ale_history_enabled = 1
let g:ale_info_default_mode = 'preview'
let g:ale_history_log_output = 1
let g:ale_keep_list_window_open = v:null
let g:ale_lint_delay = 200
let g:ale_lint_on_enter = 1
let g:ale_lint_on_filetype_changed = 1
let g:ale_lint_on_insert_leave = 1
let g:ale_lint_on_save = 1
let g:ale_lint_on_text_changed = 'normal'
let g:ale_linter_aliases = {}
let g:ale_linters = {}
[...]

PR #4738, opened 2024-03-10
Possible fix for issue #4642: use ModeChanged events instead of InsertLeave emulation

zlogic:
ALE is my favourite way of using LSPs and linters, the only thing that might need improvement is https://github.com/dense-analysis/ale/issues/4642. Any time I press <kbd>Esc</kbd> and start typing or moving around the document (muscle memory :) ), Vim beeps and  Tried switching to NeoVim and Helix, only to return back to Vim+ALE :)

I've done some research/testing during this weekend, and it looks like Vim 9 has some issues (a timeout reset?) if a timer is started from an `InsertEnterEvent`.
`InsertEnter` is called _before_ switching into Insert mode, so `mode()` might potentially return "normal" mode if the timer is triggered too early.
I was able improve the situation by [specifying a shorter timeoutlen](https://stackoverflow.com/a/13485315), but that doesn't seem like an ideal solution. When <kbd>Esc</kbd> or`\<C-[\>` is pressed, Vim waits some time to see if it's an escape code for something else (like an Arrow Up key).

`ModeChanged` looks like a more reliable way to detect an "exit insert mode" event and is a lot simpler (doesn't need a timer). Also, it can detect some other transitions like `\<C-o\>` in insert mode.
The `ModeChanged` event is available in:
* [Vim 8.2.3430](https://github.com/vim/vim/commit/f1e8876fa2359b572d262772747405d3616db670)
* [NeoVim 0.7.0](https://github.com/neovim/neovim/commit/69bd1e4e36fa3e3f604740c92b15b13141520822)

so the current tests are covering both `InsertMode` emulation and the new `ModeChanged` events.

With this PoC, the situation has improved significantly on my machine - no more beeping, and exiting Insert mode is instantaneous.

### DIFF
--- a/autoload/ale/events.vim
+++ b/autoload/ale/events.vim
@@ -100,6 +100,10 @@ if !exists('s:insert_leave_timer')
     let s:insert_leave_timer = -1
 endif
 
+" True if the ModeChanged event exists.
+" In this case, ModeChanged will be used instead of InsertLeave emulation.
+let s:mode_changed_exists = exists('##ModeChanged')
+
 function! ale#events#EmulateInsertLeave(buffer) abort
     if mode() is# 'n'
         call timer_stop(s:insert_leave_timer)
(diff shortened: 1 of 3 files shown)


zlogic:
Or it might make sense to just make the `InsertLeave` emulation optional (configurable), perhaps some developers might use <kbd>Esc</kbd> to run the linter and other event handlers, but use `\<C-c\>` to temporarily exit Insert mode without any possible side-effects/callbacks.

w0rp (approved):
This is fantastic, and I love this. 😄  Let's run this this and see how this behaves over time.

I'll probably follow up with a documentation update.

Cheers! :beers:

PR #4743, opened 2024-03-13
Add support for sqlfluff 3.0.0

shadowwa:
As stated in the changelog of sqlfluff:
"the original fields of line_pos and line_no have been renamed to start_line_pos and start_line_no, to distinguish them from the new fields starting end_*"

The version of sqlfluff is checked to continue to works with older versions.


hsanson (approved):
LGTM, thanks.

PR #4746, opened 2024-03-22
Added fix subcommand options

kevinvanleer:
php-cs-fixer command line options are ordered. Options that appear after the main command are applied to the main command. Options that appear after the subcommands are applied to the subcommands. Baseline behavior will not apply a custom configuration. 

```
// Does not work
php-cs-fixer --config ./custom-config.php fix target-file.php
```

That option must be applied to the fix subcommand. 

```
// Works!
php-cs-fixer fix --config ./custom-config.php target-file.php
```

This change enables a user to fix options (like --config) in addition to base command options (like -nq).

After this change a user can set:

```
  let g:ale_php_cs_fixer_fix_options = '--config ' . fnamemodify(findfile('.php-cs-fixer.php', './;'), ':p')
```

in `init.vim` and php-cs-fixer will behave like eslint. A more sophisticated fixer definition might try to apply this strategy for finding the configuration before falling back to default behavior if no configuration is found.


hsanson (approved):
Thanks.

PR #4755, opened 2024-04-15
Implement listing all returned results for LSP textDocument/implements

bgarber:
This PR  has the intention of implementing the list feature for `:ALEGoToDefinition` and `:ALEGoToImplementation`. Specially `:ALEGoToImplementation`. There are times a LSP may return multiple results for the `textDocument/implementation` command. This is true when we try to find implementations of an abstract class or interface. The way ALE is implemented, only the first result is returned.

Take this example.

foobar.go:
```go
type Foobarer interface {
    Do() bool
}
```

foobar_a.go
```go
type FoobarA struct {
   foo int
}

func (f *FoobarA) Do() bool {
   fmt.Printf("This foobar is int (%d)\n", f.foo)
   return true
}
```

another_foobar.go
```go
type AnotherFoobar struct {
   foo bool
}

func (f *AnotherFoobar) Do() bool {
   fmt.Printf("This foobar is bool (%v)\n", f.foo)
   return f.foo
}
```

When trying to find implementations of the `Do` function, it will only find the one from `AnotherFoobar` struct (because the LSP returns in alphabetic order). But I could want the other one. This change fixes this behavior.

Captura de tela 2024-04-14 221711

### DIFF
--- a/autoload/ale/definition.vim
+++ b/autoload/ale/definition.vim
@@ -68,6 +68,15 @@ function! ale#definition#HandleTSServerResponse(conn_id, response) abort
             if empty(l:item_list)
                 call ale#util#Execute('echom ''No definitions found''')
             elseif len(l:item_list) == 1
+                let l:filename = l:item_list[0].filename
+                if get(l:options, 'open_in') is# 'quickfix'
+                    let l:line = l:item_list[0].lnum
+                    let l:column = l:item_list[0].col
+                else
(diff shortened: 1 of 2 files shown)


bgarber:
Please, let me know if this is good to go. Thank you!

w0rp (approved):
This looks good to me. Sorry for the delay.

Cheers! :beers:

PR #4761, opened 2024-04-20
Update dart analysis_server command

macuyler:
In 2021 the dart team added a new sub-command `language-server` to replace the original `./snapshots/analysis_server.dart.snapshot --lsp` convention for starting the language server.

The original convention does not work with the latest versions of dart.

* [Dart SDK Language Server Documentation](https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/lsp_spec/README.md)
* https://github.com/dart-lang/sdk/commit/c224cc2e0d4cd8fc536c21ee963a0254d18a27bb

### DIFF
--- a/doc/ale-dart.txt
+++ b/doc/ale-dart.txt
@@ -27,6 +27,16 @@ g:ale_dart_analysis_server_executable   *g:ale_dart_analysis_server_executable*
   This variable can be set to change the path of dart.
 
 
+g:ale_dart_analysis_server_enable_language_server
+                            *g:ale_dart_analysis_server_enable_language_server*
+                            *b:ale_dart_analysis_server_enable_language_server*
+  Type: |Number|
+  Default: `0`
+
+  When set to `1`, ALE will use the new `dart language-server` command to
+  launch the language server.
+
+
 ===============================================================================
 dart-analyze                                                 *ale-dart-analyze*
 
--- a/doc/ale-dart.txt
+++ b/doc/ale-dart.txt
@@ -31,10 +31,11 @@ g:ale_dart_analysis_server_enable_language_server
                             *g:ale_dart_analysis_server_enable_language_server*
                             *b:ale_dart_analysis_server_enable_language_server*
   Type: |Number|
-  Default: `0`
+  Default: `1`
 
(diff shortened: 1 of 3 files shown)


w0rp:
Can we add an option to switch the command back to the old version for those who need to keep using it? Changing right away is a little bit of a breaking change, but it's been unstable so I'm okay with an option to change it back and using the stable command right away.

macuyler:
@w0rp I added the option to switch back to the old version. Let me know if there is anything else I can do to help!

doc/ale-dart.txt
```diff
@@ -27,6 +27,17 @@ g:ale_dart_analysis_server_executable   *g:ale_dart_analysis_server_executable*
...
+g:ale_dart_analysis_server_enable_language_server
+                            *g:ale_dart_analysis_server_enable_language_server*
+                            *b:ale_dart_analysis_server_enable_language_server*
+  Type: |Number|
+  Default: `1`
+
+  When set to `1`, ALE will use the new `dart language-server` command to
+  launch the language server. When set to `0`, ALE will use the original
+  `./snapshots/analysis_server.dart.snapshot --lsp` command.
+
```
hsanson, line 39: Maybe adding the dart version from where the new command is necessary would help people decide if they need to change this flag or not. Exampl:

```
When set to `1`, ALE will use the new `dart language-server` command, 
available from Dart version XX.XX.XX to launch the language server. When
set to `0`, ALE will instead use the deprecated `./snapshots/analysis_server.dart.snapshot --lsp`
command used by older versions of Dart.

macuyler: I have updated the documentation to state that dart version `2.16.0` is when the new command was released.

* https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#2160---2022-02-03
* https://github.com/dart-lang/sdk/blob/2.16.0/pkg/analysis_server/tool/lsp_spec/README.md#running-the-server
* https://github.com/dart-lang/sdk/blob/2.15.1/pkg/analysis_server/tool/lsp_spec/README.md#running-the-server

hsanson (approved):
Thanks, looks good to me.

Issue #4754, opened 2024-04-13
[Biome] Linter not usable


Spixmaster:
From what I see, the language server is started but I personally need `biome lint`.

PR #4763, opened 2024-04-24
fix(biome): run all enabled biome fixers

redbmk:
- based on biome config, will format, lint, and/or sort imports
- adds variable to apply unsafe fixes, disabled by default

fixes: #4754

### DIFF
--- a/autoload/ale/fixers/biome.vim
+++ b/autoload/ale/fixers/biome.vim
@@ -1,10 +1,11 @@
 function! ale#fixers#biome#Fix(buffer) abort
     let l:executable = ale#handlers#biome#GetExecutable(a:buffer)
     let l:options = ale#Var(a:buffer, 'biome_options')
+    let l:apply = ale#Var(a:buffer, 'biome_fixer_apply_unsafe') ? '--apply-unsafe' : '--apply'
 
     return {
     \   'read_temporary_file': 1,
-    \   'command': ale#Escape(l:executable) . ' format'
+    \   'command': ale#Escape(l:executable) . ' check ' . l:apply
     \       . (!empty(l:options) ? ' ' . l:options : '')
     \       . ' %t'
     \}
--- a/autoload/ale/handlers/biome.vim
+++ b/autoload/ale/handlers/biome.vim
@@ -4,6 +4,7 @@
 call ale#Set('biome_executable', 'biome')
 call ale#Set('biome_use_global', get(g:, 'ale_use_global_executables', 0))
 call ale#Set('biome_options', '')
+call ale#Set('biome_fixer_apply_unsafe', 0)
 
 function! ale#handlers#biome#GetExecutable(buffer) abort
     return ale#path#FindExecutable(a:buffer, 'biome', [
--- a/doc/ale-typescript.txt
+++ b/doc/ale-typescript.txt
@@ -27,6 +27,14 @@ g:ale_biome_use_global                                 *g:ale_biome_use_global*
   See |ale-integrations-local-executables|
 
 
+g:ale_biome_fixer_apply_unsafe                 *g:ale_biome_fixer_apply_unsafe*
+                                               *b:ale_biome_fixer_apply_unsafe*
+  Type: |Number|
+  Default: `0`
+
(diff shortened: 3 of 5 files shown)


redbmk:
@w0rp I have a few other fixes queued up for biome, I just need to make sure I have tests for them and everything. Should I make it all part of one PR or keep them more atomic with separate PRs?

Spixmaster:
@redbmk I highly doubt that waiting this long for a response is worth it. Maybe you can just handle it how you like.

I appreciate your work!

hsanson:
Smaller self-contained PRs are easier to review given limited time maintainers have. Also easier to revert in case something breaks.

autoload/ale/fixers/biome.vim
```diff
@@ -1,9 +1,15 @@
 function! ale#fixers#biome#Fix(buffer) abort
     let l:executable = ale#handlers#biome#GetExecutable(a:buffer)
     let l:options = ale#Var(a:buffer, 'biome_options')
+    let l:apply = '--apply'
+
+    let l:unsafe = ale#Var(a:buffer, 'biome_fixer_apply_unsafe')
+    if (l:unsafe is 1 || l:unsafe is v:true)
```
hsanson, line 7: No need to check both `1` and `v:true`.  Simply using  `if  (l:unsafe)` should suffice the check.

autoload/ale/fixers/biome.vim
```diff
@@ -1,9 +1,15 @@
...
+        let l:apply = '--apply-unsafe'
+    endif
 
     return {
-    \   'command': ale#Escape(l:executable) . ' format'
+    \   'command': ale#Escape(l:executable) . ' check ' . l:apply
     \       . (!empty(l:options) ? ' ' . l:options : '')
```
hsanson, line 13: Since we want to run only the formater, maybe is good idea to also add `--linter-enabled=false` to the command so linter is not executed?

Spixmaster: From what I understand, everything is intended that the user specificed in his configuration file. Both formatting and linting can be set there.

redbmk: Yeah I would prefer to run the fixer for both the linter and the formatter at the same time. WIthout this change it will only run the formatter.

One could add `--linter-enabled=false` to the options if they only wanted to fix the formatting, but there's a bug in `g:ale_biome_options`. It will apply both to the fixer and the LSP, but they take pretty different arguments and biome will bail if you give invalid ones. I think those should be split out into `ale_biome_lsp_options` and `ale_biome_fixer_options` or something

hsanson: I see, agree that is better to have different options for fixer and linter. Can be handled on a separate PR.

hsanson:
There is a linter issue that causes the test to fail.  After that is fixed I can merge this:

```
./autoload/ale/fixers/biome.vim:7 Blank line required before `if`.
```

redbmk:
> There is a linter issue that causes the test to fail. After that is fixed I can merge this:
> 
> ```
> ./autoload/ale/fixers/biome.vim:7 Blank line required before `if`.
> ```

@hsanson Cool I just pushed up a fix that included your oneliner suggestion. That's a lot cleaner, thanks! I checked that linters are passing locally now

redbmk:
Not sure if it would be an issue in the test runners but I noticed if I run the linter and then fixer tests I get an error. I think if it runs fixers first then linters, it wouldn't show up, but the fix I just pushed lets it work either way.

Spixmaster:
What it the issue with the commits. Why do not the tests finish? I would like to inspect the output but can only call the one of "continuous-integration/appveyor/pr".

hsanson (dismissed):
LGTM, thanks

hsanson:
@redbmk if you could fix the conflict I can merge this.

redbmk:
> LGTM, thanks

looks like I need to fix merge conflicts with the other one now. I'll try to do that asap

redbmk:
@hsanson should be good now

hsanson (approved):
Thanks

PR #4764, opened 2024-04-25
Match filename in xvlog linter and add tip

gastmaier:
Match the xvlog stdout for the filename to properly filter out the errors per file.
Suggest using Vivado generated compile.sh file to obtain the xvlog options.

For reference, the sane way of using xvlog is along these lines:
```
let g:ale_verilog_xvlog_options = '
\ --incr --relax \
\ -L uvm -L axi_vip_v1_1_15 -L smartconnect_v1_0 -L clk_vip_v1_0_3 -L axi4stream_vip_v1_1_15 -L rst_vip_v1_0_5 -L xilinx_vip
\ -prj ./path/to/xsim/system_tb_vlog.prj'
```
Room for improvement:
The `.prj` uses relative paths and expects xvlog to be called from its path.
As is, I'm manually "fixing" the relative paths in the `.prj` to point from the cwd to xsim.
Not sure if it's worth a more elaborated solution (e.g. search for the `compile.sh`, set cwd to its location, and extract the xvlog options.

### DIFF
--- a/ale_linters/verilog/xvlog.vim
+++ b/ale_linters/verilog/xvlog.vim
@@ -11,13 +11,16 @@ endfunction
 function! ale_linters#verilog#xvlog#Handle(buffer, lines) abort
     "Matches patterns like the following:
     " ERROR: [VRFC 10-1412] syntax error near output [/path/to/file.v:5]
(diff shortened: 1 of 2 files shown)


w0rp:
I'll close this PR as it's been around for a while now. If you come back with some tests for the changes here, we can merge this.

Issue #4768, opened 2024-05-05
On Windows 11, shell=pwsh/powershell continues breaking ALE lint


wunewww:
## Information

**VIM version**

```
VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Jan  2 2024 23:53:44)
MS-Windows 64-bit console version
```

Operating System: 

```
Edition	Windows 11 Pro
Version	23H2
Installed on	‎17/‎06/‎2023
OS build	22631.3527
Experience	Windows Feature Experience Pack 1000.22700.1003.0
```

## What went wrong

On Windows 11, Vim 9.1, when `set shell=pwsh.exe` or `set shell=powershell.exe`, ale will excute commands in wrong format.

To be specific, I am using *vale* to lint my markdown file. when `set shell=cmd.exe`, ale will excute CMD like (excerpted from `:ALEInfo`)

```
cmd /s/c "vale --output=JSON  C:\Users\NIUZHA~1\AppData\Local\Temp\VMQ43BA.tmp\README.md"
```

when `set shell=pwsh.exe`, it will be

```
cmd /s/c "''vale'' --output=JSON  ''C:\Users\NIUZHA~1\AppData\Local\Temp\VA2D6D5.tmp\README.md''"
```

which could not be recognized by cmd.exe

## Reproducing the bug

1. On windows 11, vim 9.1, `set shell=pwsh.exe`, open vim and trigger ale.
2. `:ALEInfo` shows above, and `(finished - exit code 1)`

### :ALEInfo

Error log appears at the bottom.

<details>
  <summary>Expand</summary>
  
  
  
 Current Filetype: markdown
Available Linters: ['alex', 'cspell', 'languagetool', 'markdownlint', 'marksman', 'mdl', 'proselint', 'redpen', 'remark_lint', 'textlint', 'vale', 'writegood']
   Linter Aliases:
'remark_lint' -> ['remark-lint']
'writegood' -> ['write-good']
  Enabled Linters: ['alex', 'cspell', 'languagetool', 'markdownlint', 'marksman', 'mdl', 'proselint', 'redpen', 'remark_lint', 'textlint', 'vale', 'writegood']
  Ignored Linters: []
 Suggested Fixers:
  'dprint' - Pluggable and configurable code formatting platform
  'pandoc' - Fix markdown files with pandoc.
  'prettier' - Apply prettier to a file.
  'remark-lint' - Fix markdown files with remark-lint
[...]

PR #4769, opened 2024-05-06
Fix: Resolving cmd.exe Construction Error in Ale on Windows

wunewww:
this fixes problem described in #4768 , by rewrite win32 part of `ale#job#PrepareCommand` . The original code appears to have an issue with its string concatenation expression.


wunewww:
idk why `fix\test_ale_fix.vader` - `Post-processing should work` failed. I am not so familiar with it. I may need some help. Or the original code is deliberate for this test...

Issue #4752, opened 2024-04-10
ALE linting .tsx as .ts files when using Biome 


jbadavis:
## Information

**VIM version**

VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Feb 11 2024 17:19:45)
macOS version - arm64

Operating System: Mac OS 13.2.1

## What went wrong

I've added and configured [Biome](https://biomejs.dev/) for my React/Typescript project and I suspect ALE/Biome is linting my `.tsx` files as `typescript` files rather than `typescriptreact` files. 

## Reproducing the bug

1. Clone https://github.com/jbadavis/jbadavis.dev.
2. Checkout `feat/biome`.
3. Open `main.tsx` and see the linting errors.

### :ALEInfo
<details>
  <summary>Expand</summary>
  
  
  

 Current Filetype: typescriptreact
Available Linters: ['biome', 'cspell', 'deno', 'eslint', 'standard', 'tslint', 'tsserver', 'typecheck', 'xo']
  Enabled Linters: ['biome']
  Ignored Linters: []
 Suggested Fixers:
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
</details>

PR #4773, opened 2024-05-13
fix(biome): send correct language to lsp proxy

redbmk:
Since Biome understands `typescriptreact` and `javascriptreact` as languages, we can send the `filetype` to the LSP, rather than only sending `typescript` for both `ts` and `tsx` files, or `javascript` for `js` and `jsx` files.

fixes: #4752

--- 

I moved the `ale_typescript_biome.vader` tests to just `ale_biome.vader` since it covers multiple languages, and I added some tests for this case. I also fixed what seemed like an issue with the biome tests where if the linters run first then the fixers fail. Now it saves the variables and resets them before the tests.

Linting and biome vader tests both pass locally for me

### DIFF
--- a/ale_linters/javascript/biome.vim
+++ b/ale_linters/javascript/biome.vim
@@ -4,6 +4,7 @@
 call ale#linter#Define('javascript', {
 \   'name': 'biome',
 \   'lsp': 'stdio',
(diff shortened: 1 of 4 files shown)


w0rp (approved):
I may change the tests a little later based on specific Vader things but I think this will work. 

Cheers! :beers:

PR #4774, opened 2024-05-13
fix(biome): find root when using biome.jsonc

redbmk:
Since biome supports either `biome.json` or `biome.jsonc` config files, we need to look for both when searching for the LSP project root. We can also look for a package.json or .git folder to use. This uses mostly the same logic as deno.

---- 
I added some documentation for the new option and explained the new logic. This is pretty much copied from the deno documentation (and code) with some minor tweaks. Probably could refactor to some shared code at some point, just not sure where to put it.

I also added a couple tests and made sure tests and linting pass locally.

This will have minor conflicts with #4763 and #4773 but I wanted to keep the commits fairly atomic and the conflicts will be quick to clean up.

PR #4775, opened 2024-05-13
fix(biome): use temp file instead of stdin

redbmk:
biome handles utf8 characters differently between files and stdin, and in some cases can replace emojis with ascii characters when using stdin

refs: biomejs/biome#2604

---

This is pretty small but I made sure to update the relevant test. Again, it will have minor conflicts with #4774, #4763, and #4773, but I wanted to keep the fixes focused.

Vader tests and linters are passing locally.

I have another change pending to add biome support for json and jsonc and clean up the options, but it sort of depends on some of these other PRs so I'll wait until these are resolved.

### DIFF
--- a/autoload/ale/fixers/biome.vim
+++ b/autoload/ale/fixers/biome.vim
@@ -3,8 +3,9 @@ function! ale#fixers#biome#Fix(buffer) abort
     let l:options = ale#Var(a:buffer, 'biome_options')
 
     return {
+    \   'read_temporary_file': 1,
     \   'command': ale#Escape(l:executable) . ' format'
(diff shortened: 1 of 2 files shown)


hsanson:
Having a hard time figuring out why the Github actions seem to be stuck without any visible way to retry them....  maybe @w0rp can help here.

redbmk:
@hsanson I think an approval might kick off the tests

> At least 1 approving review is required by reviewers with write access.

Could be that it's stale or something though? I'll try pulling in the latest and updating my branch

hsanson (approved):
LGTM, thanks

PR #4779, opened 2024-05-15
Add support for astro

ghsang:
eslint and prettier setup for astro


ghsang:
I will reopen after fixing the failed tests

PR #4780, opened 2024-05-15
Add support for astro

ghsang:
esint, prettier support for astro


w0rp (approved):
I'll merge this one. I might take the default linter list away later, as we could just enable "all" astro linters by default.

w0rp:
Cheers! :beers:

PR #4793, opened 2024-06-17
Update xvlog.vim

zhangt123456:
when using xvlog in Vivado 2019.1 for SystemVerilog, you should add the option '-sv'.

### DIFF
--- a/ale_linters/verilog/xvlog.vim
+++ b/ale_linters/verilog/xvlog.vim
@@ -7,6 +7,7 @@ if &filetype == 'systemverilog'
     call ale#Set('verilog_xvlog_options', '-sv')
 else
     call ale#Set('verilog_xvlog_options', '')
+endif
 
 function! ale_linters#verilog#xvlog#GetCommand(buffer) abort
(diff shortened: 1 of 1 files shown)


hsanson:
Also there is a linter error:

> ale_linters/verilog/xvlog.vim:6:14: Use robust operators `==#` or `==?` instead of `==` (see Google VimScript Style Guide (Matching))

ale_linters/verilog/xvlog.vim
```diff
@@ -2,7 +2,12 @@
 " Description: Adds support for Xilinx Vivado `xvlog` Verilog compiler/checker
 
 call ale#Set('verilog_xvlog_executable', 'xvlog')
-call ale#Set('verilog_xvlog_options', '')
+
+if &filetype == 'systemverilog'
+    call ale#Set('verilog_xvlog_options', '-sv')
+else
+    call ale#Set('verilog_xvlog_options', '')
+endif
```
w0rp, line 10: The `-sv` option should be added dynamically in the `GetCommand` function based on the the filetype for `a:buffer`. I'll close this PR and you can open another in future with this and a test for the change.

PR #4799, opened 2024-06-25
fix!(biome): allow passing options to `biome check`

redbmk:
The only option available to biome's `lsp-proxy` command used for linting is `--config-path`. However, we are using ALE to find and set the project root, and have a way to manually override, so that is no longer necessary.

The LSP proxy also used the `g:ale_biome_options` config, which is shared with the fixer's `check` command, but `lsp-proxy` will throw an error if unknown options are included, making it so that option is only useful to set the project root.

BREAKING CHANGE: We are no longer passing options to the biome LSP proxy, but we can still set the project root with
`g:ale_biome_lsp_project_root`.

---

I added a test that shows you can pass arbitrary options along to `biome check` and updated the documentation.

I had thought about continuing to pass `g:ale_biome_options` to the LSP for backwards compatibility and t hen creating a new `g:ale_biome_fixer_options` that get added additionally to the fixer, but that seemed a lot more complex. Considering the only shared option is `--config-path`, which is also the only LSP option, people either were only using that option or their LSP would have been broken.

The downside to this approach is that if someone is relying on `--config-path` they may need to update to use `ale_biome_lsp_project_root` instead, so maybe a deprecation path is better. @w0rp @hsanson, what do you think?

### DIFF
--- a/ale_linters/javascript/biome.vim
+++ b/ale_linters/javascript/biome.vim
@@ -6,6 +6,6 @@ call ale#linter#Define('javascript', {
 \   'lsp': 'stdio',
 \   'language': function('ale#handlers#biome#GetLanguage'),
 \   'executable': function('ale#handlers#biome#GetExecutable'),
-\   'command': function('ale#handlers#biome#GetCommand'),
+\   'command': '%e lsp-proxy',
 \   'project_root': function('ale#handlers#biome#GetProjectRoot'),
 \})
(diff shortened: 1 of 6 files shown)


hsanson:
No strong opinions here. As long as there are alternatives users can easily implement to get same functionality as before I see no reason to stop this PR.

hsanson (approved):
Thanks for the update.

PR #4802, opened 2024-07-01
Fix for ruff fixer for ruff version >=0.5.0

TBourton:
As of [ruff 0.5.0](https://github.com/astral-sh/ruff/releases/tag/0.5.0) the deprecated `ruff <path>` has now been fully removed in favour of `ruff check <path>`. This caused the `ruff` fixer to fail as it was still using the old style CLI command.


hsanson (approved):
Thanks, LGTM.

PR #4805, opened 2024-07-08
feat(biome): add support for json and jsonc

redbmk:
- adds biome linter for json and jsonc
- enables biome by default for json and jsonc
- suggests biome as a fixer for json and jsonc

----

I added some documentation to json and jsonc linting that just point to the typescript docs as the options and funcationality are all the same. This is what biome and deno are already doing in the javscript docs.

I didn't add any tests because I'm not really sure what I would be testing. Are there examples of testing that certain linters or fixers are enabled/suggested?


w0rp (dismissed):
Cheers! 🍻

w0rp:
One CI test failed, so have a look at that.

hsanson (approved):
LGTM, thanks.

PR #4811, opened 2024-07-31
Add option to use rebar3 fmt instead of standalone erlfmt

yosida95:
The current implementation of the fixer for erlfmt, an Erlang code formatter, invokes a standalone `erlfmt` command. However, the standalone `erlfmt` command is not distributed by default. Instead, erlfmt is often used as a plugin for Rebar3, the de facto build tool for Erlang.

This patch introduces a new option `g:ale_erlang_erlfmt_use_rebar`, which is disabled by default, allowing the use of `rebar3 fmt` instead.

Additionally, this patch fixes a bug that caused the command to read the original source code rather than the stdin.

### DIFF
--- a/autoload/ale/erlang.vim
+++ b/autoload/ale/erlang.vim
@@ -1,0 +1,9 @@
+" Author: Kohei YOSHIDA - https://github.com/yosida95
+" Description: Helper functions for Erlang tools
+
+call ale#Set('erlang_rebar_executable', 'rebar3')
+call ale#Set('erlang_rebar_use_global', get(g:, 'ale_use_global_executables', 0))
+
+function! ale#erlang#GetRebarExecutable(buffer) abort
+    return ale#path#FindExecutable(a:buffer, 'erlang_rebar', ['rebar3'])
+endfunction
--- a/autoload/ale/erlang.vim
+++ b/autoload/ale/erlang.vim
@@ -1,9 +1,0 @@
-" Author: Kohei YOSHIDA - https://github.com/yosida95
-" Description: Helper functions for Erlang tools
-
-call ale#Set('erlang_rebar_executable', 'rebar3')
-call ale#Set('erlang_rebar_use_global', get(g:, 'ale_use_global_executables', 0))
-
-function! ale#erlang#GetRebarExecutable(buffer) abort
-    return ale#path#FindExecutable(a:buffer, 'erlang_rebar', ['rebar3'])
-endfunction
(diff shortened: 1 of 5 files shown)


autoload/ale/erlang.vim
```diff
@@ -0,0 +1,9 @@
+" Author: Kohei YOSHIDA - https://github.com/yosida95
+" Description: Helper functions for Erlang tools
+
+call ale#Set('erlang_rebar_executable', 'rebar3')
+call ale#Set('erlang_rebar_use_global', get(g:, 'ale_use_global_executables', 0))
+
+function! ale#erlang#GetRebarExecutable(buffer) abort
```
w0rp, line 7: There's a runtime cost in loading this file, and we only have this one function and call it from one place. Let's just move this into the fixer file instead, to avoid this runtime cost.

w0rp (changes requested):
Looks pretty good! See my comment here.

autoload/ale/erlang.vim
yosida95: Done

yosida95:
@w0rp Thank you for your feedback! I have addressed the points you raised. Could you please review the changes again?

autoload/ale/fixers/erlfmt.vim
```diff
@@ -4,18 +4,49 @@
...
+function! ale#fixers#erlfmt#GetRebarExecutable(buffer) abort
+    return ale#path#FindExecutable(a:buffer, 'erlang_rebar', ['rebar3'])
+endfunction
+
 function! ale#fixers#erlfmt#Fix(buffer) abort
     let l:options = ale#Var(a:buffer, 'erlang_erlfmt_options')
-    let l:executable = ale#fixers#erlfmt#GetExecutable(a:buffer)
+    let l:use_rebar = ale#Var(a:buffer, 'erlang_erlfmt_use_rebar')
+
+    if l:use_rebar
```
dmitrivereshchagin, line 24: It would be clearer if the two cases were handled in separate functions, like this:

    function! ale#fixers#erlfmt#Fix(buffer) abort
        return ale#Var(a:buffer, 'erlang_erlfmt_use_rebar3')
        \   ? s:FixWithRebar3(a:buffer)
        \   : s:FixWithEscript(a:buffer)
    endfunction


However, it seems to me that from user's perspective it would be better to add separate `rebar3_fmt` fixer.

w0rp:
I'll close this until it can be reworked and rebased a little. It might make sense to make rebar3 by default, if that makes sense.

PR #4825, opened 2024-08-30
Improve support for python package manage: pipenv, poetry and uv

diegoholiveira:


hsanson (approved):
Great contribution, thanks.

PR #4828, opened 2024-09-02
Return floating_preview window id

idbrii:
Make `ale#floating_preview#Show` more similar to `popup_create` and return the id of the window so it's easy to set the filetype of the resulting buffer.

This doesn't affect any other ale code. It just makes it easier to use ALE's library function in my vim config.


hsanson (approved):
LGTM

PR #4834, opened 2024-09-24
syntax_tree: no need to write and read temporary file

doits:
The `format` option prints out to `stdout`, so there is no need to write and read back the temporary file.


hsanson (approved):
LGTM, thanks.

PR #4847, opened 2024-10-22
Add fixer for json: python's json.tool

### LINKED ISSUE #4314: Add JSON fixer using 'python -m json.tool'
Implement something similar to the following code:

ftplugin/json.vim:

`
let b:ale_fixers = ['json_python_fixer]
let b:ale_json_python_fixer_options = '--indent 2'
`

autoload/ale/fixers/json_python_fixer.vim:
```
" Description: Fix JSON files with "python -m json.tool"
call ale#Set('json_python_fixer_options', '')

function! ale#fixers#json_python_fixer#Fix(buffer) abort
  let l:command = 'python'

  if !executable(l:command)
    echo 'ALEFix: ' . l:command . ' executable not found'
    return {'command': 'REM'}
  endif

  let l:command .= ' -m json.tool'

  let l:options = ale#Var(a:buffer, 'json_python_fixer_options')
  if len(l:options)
    let l:command .= ' ' . l:options
  endif

  return { 'command': l:command . ' -'}
endfunction
```

idbrii:
Resolves #4314. 

Add a fixer that's built into python for json formatting. Include a couple arguments in docs to make these features more discoverable.

Uses stdin-based fixing so you don't need to save the file to fix.

Unlike most python commands (i.e., pyflakes), this one doesn't seem to include a standalone executable. I can't find any other ale modules that get a python path, so this fixer has its own variable:
```
call ale#Set('json_pytool_executable', 'python')
```

### DIFF
--- a/doc/ale-json.txt
+++ b/doc/ale-json.txt
@@ -84,6 +84,40 @@ g:ale_json_fixjson_use_global                   *g:ale_json_fixjson_use_global*
 
 
 ===============================================================================
+pytool                                                        *ale-json-pytool*
+
+Use python's json.tool module to reformat json.
+
+g:ale_json_pytool_executable                     *g:ale_json_pytool_executable*
+                                                 *b:ale_json_pytool_executable*
+
+  Type: |String|
+  Default: `'python'`
+
+  The python executable that run to use its json.tool module. This fixer
+  requires python 3, which includes the json module.
+
+g:ale_json_pytool_options                           *g:ale_json_pytool_options*
+                                                    *b:ale_json_pytool_options*
+
+  Type: |String|
+  Default: `''`
+
+  These options are passed to the json.tool module. Example: >
+    let g:ale_json_pytool_options = '--sort-keys --indent 2'
+<  See docs for all options:
+    https://docs.python.org/3/library/json.html#module-json.tool
+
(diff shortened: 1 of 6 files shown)


idbrii:
Please don't close this PR.

idbrii:
That should fix lint errors.

doc/ale-json.txt
```diff
@@ -83,6 +83,40 @@ g:ale_json_fixjson_use_global                   *g:ale_json_fixjson_use_global*
   See |ale-integrations-local-executables|
 
 
+===============================================================================
+pytool                                                        *ale-json-pytool*
```
hsanson, line 87: I think you need to make sure this name matches the one in the ToC: `json.tool` or the linter tests won't pass.

hsanson (changes requested):

doc/ale-json.txt
idbrii: Ah, I forgot to add it to the toc!

idbrii:
Ah, finally figured out toc lint issue: the validation pattern doesn't handle `.` in the name. Used `pytool` instead of `json.tool` to resolve.

hsanson (approved):
Thanks for the contribution.

PR #4851, opened 2024-11-02
Use iconv to convert message encoding on Windows

cyyever:


w0rp:
I'll close this until there's a description of the bug this is fixing and tests for the change.

PR #4852, opened 2024-11-02
Add erlang-ls as an alias for erlang_ls linter

dmitrivereshchagin:
The project itself is often referred to as Erlang LS, so erlang-ls would be a suitable alias.


hsanson (approved):
:+1:

PR #4858, opened 2024-11-06
When using `actionlint` look for & use a config file

beahues:
Actionlint supports a config file and it lives in a very searchable path, as the only files it acts on are in the `.github` directory already.

Look for an `actionlint.yml` and `.yaml` in that path, and use the config if its there.


hsanson (approved):
Looks good. thanks.

Issue #4765, opened 2024-04-27
json linting with jq doesn't report errors correctly


glennie:
## Information

**VIM version**

VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Jan 11 2024 20:38:16)
Included patches: 1-16

Operating System: Linux (Debian GNU/Linux trixie/sid)

## What went wrong

When editing json files using jq for linting, ALE doesn't report the errors correctly.

1. Enabled jq for in g:ale_linters for json
2. Edited a json file with errors

As you can see the details below, jq returns `jq: parse error: Expected separator between values at line L, column C` for each error. But, in file `ale_linters/json/jq.vim`(line 8), the regexp to catch the errors/warning is `let s:pattern = '^parse error: \(.\+\) at line \(\d\+\), column \(\d\+\)$'` (note the **^** in the regexp). Therefore, jq's output is not properly parsed.

using `let s:pattern = 'parse error: \(.\+\) at line \(\d\+\), column \(\d\+\)$'` (without **^** in the regexp) gives me expected results, but I'm not sure whether it is correct way to fix it.

### :ALEInfo
<details>
  <summary>ALEInfo Ouput</summary>

 Current Filetype: json
Available Linters: ['cspell', 'eslint', 'jq', 'jsonlint', 'spectral', 'vscodejson']
  Enabled Linters: ['jq']
  Ignored Linters: []
 Suggested Fixers:
  'clang-format' - Fix C, C++, C#, CUDA, Java, JavaScript, JSON, ObjectiveC and Protobuf files with clang-format.
  'dprint' - Pluggable and configurable code formatting platform
  'fixjson' - Fix JSON files with fixjson.
  'jq' - Fix JSON files with jq.
  'prettier' - Apply prettier to a file.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 
 Linter Variables:
let g:ale_json_jq_executable = 'jq'
let g:ale_json_jq_filters = '.'
let g:ale_json_jq_options = '. --slurp --monochrome-output'
 
 Global Variables:
let g:ale_cache_executable_check_failures = 0
let g:ale_change_sign_column_color = 1
let g:ale_command_wrapper = ''
let g:ale_completion_delay = v:null
let g:ale_completion_enabled = 0
let g:ale_completion_max_suggestions = 20
let g:ale_disable_lsp = 'auto'
[...]

PR #4862, opened 2024-11-13
Fix linting with jq (#4765)

benjos1234:
With the 1.6 version of jq the error message start with "parse error". 
With the last version of jq the error message start with "jq: parse error". 
Fix it by using a regular expression that works in both cases.

I run the test and it works.


hsanson (changes requested):
Appreciated if you can add a test to `test/handler/test_jq_handler.vader ` that tests the new format.

hsanson (approved):
Thanks for the fix .

PR #4863, opened 2024-11-15
Properly handle optional end_line_no/end_line_pos in sqlfluff

Coacher:
end_line_no/end_line_pos are optional. Example SQL:
```SELECT NULL FROM {{ a_jinja_templated_table }};```

`sqlfluff lint --dialect ansi --format json` gives the following error
among others:
```
{"start_line_no": 2, "start_line_pos": 9, "code": "TMP", "description": "Undefined jinja template variable: 'a_jinja_templated_table'", "name": "", "warning": false}
```
As one can see there is noe end_line_no/end_line_pos.

### DIFF
--- a/ale_linters/sql/sqlfluff.vim
+++ b/ale_linters/sql/sqlfluff.vim
@@ -52,16 +52,19 @@ function! ale_linters#sql#sqlfluff#Handle(buffer, version, lines) abort
 
     if ale#semver#GTE(a:version, [3, 0, 0])
         for l:violation in get(l:json, 'violations', [])
-            call add(l:output, {
+            err = {
             \   'filename': l:json.filepath,
             \   'lnum': l:violation.start_line_no,
-            \   'end_lnum': l:violation.end_line_no,
             \   'col': l:violation.start_line_pos,
-            \   'end_col': l:violation.end_line_pos,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: ale_linters/sql/sqlfluff.vim [hunk 1]
```diff
@@ -52,16 +52,19 @@ function! ale_linters#sql#sqlfluff#Handle(buffer, version, lines) abort
...
             \   'lnum': l:violation.start_line_no,
-            \   'end_lnum': l:violation.end_line_no,
             \   'col': l:violation.start_line_pos,
-            \   'end_col': l:violation.end_line_pos,
             \   'text': l:violation.description,
             \   'code': l:violation.code,
             \   'type': 'W',
-            \})
+            \}
+            if has_key(l.violation, 'end_line_no')
```
hsanson, line 63: Missing endifs?  Tests fail with:

```
   (1/2) [EXECUTE] The sqlfluff handler should handle basic warnings with version older than 3.0.0
    (1/2) [EXECUTE] (X) Vim(endfor):E171: Missing :endif:         endfor
      > function ale_linters#sql#sqlfluff#Handle, line 30
 
```

hsanson (changes requested):

PR #4865, opened 2024-11-15
Fix python linting/formatting error when in virtual environment

kwalter94:
Python fixer black fails when vim is running in a virtual environment that's located in a path containing text `poetry` with `python_auto_poetry = 0`. An example would be a path like `~/poetry_in_motion/.venv/bin/black` . The cause of this is the regular expression `poetry\|pipenv\|uv$` which matches `poetry` and `pipenv` if they appear anywhere in the virtualenv path. The trailing `$` in the regex only binds to `uv` only.

This is the error I am getting in `ALEInfo` when running black:
```
(finished - exit code 2) ['/bin/sh', '-c', 'cd ''/home/user/tmp/ale_test/poetry_in_motion'' && ''/home/user/tmp/ale_test/poetry_in_motion/.venv/bin/black'' run black - < ''/tmp/nvim.user/4ou7OE/2/formatting_test.py''']
```
Looks like the bug exists in other places too, so I have fixed them as well.

### DIFF
--- a/ale_linters/python/bandit.vim
+++ b/ale_linters/python/bandit.vim
@@ -45,7 +45,7 @@ function! ale_linters#python#bandit#GetCommand(buffer) abort
         endif
     endif
 
-    let l:exec_args = l:executable =~? 'pipenv\|poetry\|uv$'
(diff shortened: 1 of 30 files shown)


hsanson (approved):
Nice catch, thanks.

PR #4867, opened 2024-11-26
Properly handle optional end_line_no/end_line_pos in sqlfluff

Coacher:
end_line_no/end_line_pos are optional. Example SQL: `SELECT NULL FROM {{ a_jinja_templated_table }};`

`sqlfluff lint --dialect ansi --format json` gives the following error among others:
```
{"start_line_no": 1, "start_line_pos": 21, "code": "TMP", "description":
"Undefined jinja template variable: 'a_jinja_templated_table'", "name":
"", "warning": false}
```

As one can see there is no end_line_no/end_line_pos.


hsanson (approved):
Thanks for the changes.

PR #4868, opened 2024-11-27
Add erlfmt fixer to the registry and use it with stdin

dmitrivereshchagin:
Until the fixer is added to the registry, it cannot be used without additional configuration. And using the full path to the file being edited results in the loss of unsaved changes.

The tests are slightly refactored and expanded.


hsanson (approved):
Looks good, thanks.

PR #4345, opened 2022-10-28
diagnostics: support sending ALE output to Neovim's diagnostics API

mathstuf:
Closes #4005

This resurrects #4135. I'll roll with it locally to test it out more (especially to see if that off-by-one is actually fixed or not). Other than test suite bits, the review comments from there have been addressed.

Pulling the TODO item from the original PR description:

- [x] Fix off by 1 issue reported by @daliusd below
- [x] Maybe good idea to add a test test_linting_sets_signs.vader or test_sign_placement.vader that tests ALE signs are disabled when neovim diagnostics are enabled.
- [x] Also maybe add a test to test_cursor_warnings.vader that tests cursor messages is not echoed when neovim diagnostics are enabled.
- [x] Figure out if it's OK to disable loclist/quickfix like we do currently
- [x] Fix issue with g:ale_virtualtext_cursor reported below

No tests yet (beyond updating the existing config one for the new option).

PR #4872, opened 2024-11-28
fix: consider the original config for signs

delphinus:
It always overwrites signs' config this code below, and therefore, all user settings are ignored.

https://github.com/dense-analysis/ale/blob/65b49c1b8172d0ab1b08ffe8fdcabb93fc1328df/lua/ale/diagnostics.lua#L68

With this patch, ALE considers user settings and overwrites `priority` only if needed.

----

The feature for retrieving the current config values for diagnostics has been implemented in https://github.com/neovim/neovim/pull/17045. This patch is included in Neovim v0.7.0 and higher, so this fix cannot work with v0.6.x -- the minimum version ALE supports. For v0.6.x, I did a small fix (1485bab).

### DIFF
--- a/lua/ale/diagnostics.lua
+++ b/lua/ale/diagnostics.lua
@@ -68,7 +68,13 @@ module.sendAleResultsToDiagnostics = function(buffer, loclist)
   if set_signs == 1 and sign_priority then
     -- If signs are enabled, set the priority for them.
     local local_cfg = { priority = sign_priority }
-    local global_cfg = vim.diagnostic.config().signs
+    -- NOTE: vim.diagnostic.config() -- retrieving the current config values
+    -- fails in Neovim older than v0.7.0.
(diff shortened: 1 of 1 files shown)


delphinus:
@w0rp The current minimum version to support is Neovim 0.6.0. I found this was determined in #4345 because 0.6.0 is the first version that implemented `vim.diagnostic`.

Can you bump up this to 0.7.0 or higher? I want to fix this problem in this PR, but it needs the feature in 0.7.0 (details are written in PR's description).

hsanson:
@delphinus will try to tackle this during holidays:  https://github.com/dense-analysis/ale/issues/4887

w0rp:
I updated the minimum supported Neovim version to 0.7, as enough time has passed. This code could probably be simplified now if you want to try. I'll merge this as-is to address the issue.

Cheers! :beers:

PR #4873, opened 2024-11-30
fix: stylua should have cwd and --stdin-filepath

dcai:
add `cwd` and `--stdin-filepath` to stylua, so that stylua could pickup the config file correctly.


dcai:
Hi @hsanson could you please review this? thank you

hsanson (approved):
Seems ok, thanks for the contribution.

PR #4882, opened 2024-12-18
Add podman to launch test

Nounours-Logan:
Users can set the DOCKER environnement variable to select Docker or Podman to run the tests.

The tests can run with Docker and Podman.


hsanson (approved):
Thanks.

PR #4884, opened 2024-12-22
xmllint fixer must read from stdin - not buffer filename

cyberang3l:
The current xmllint fixer reads and formats the file that a buffer is associated with from disk instead of accepting input from stdin. This has the side effect that if the filename is changed in the buffer, but not saved yet, the fixer discards all the pending changes and replaces the buffer contents with the formatted text from the file contents on disk.

https://gnome.pages.gitlab.gnome.org/libxml2/xmllint.html

### DIFF
--- a/autoload/ale/fixers/xmllint.vim
+++ b/autoload/ale/fixers/xmllint.vim
@@ -7,15 +7,8 @@ call ale#Set('xml_xmllint_indentsize', 2)
 
 function! ale#fixers#xmllint#Fix(buffer) abort
     let l:executable = ale#Escape(ale#Var(a:buffer, 'xml_xmllint_executable'))
-    let l:filename = bufname(a:buffer)
 
-    if empty(l:filename)
-        let l:filename = '%t'
-    else
-        let l:filename = ale#Escape(l:filename)
-    endif
-
-    let l:command = l:executable . ' --format ' . l:filename
+    let l:command = l:executable . ' --format'
 
     let l:indent = ale#Var(a:buffer, 'xml_xmllint_indentsize')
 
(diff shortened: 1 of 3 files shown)


hsanson:
Tests must be also updated to match this change:

```
Starting Vader: C:\testplugin\test\fixers\test_xmllint_fixer_callback.vader
    (1/4) [EXECUTE] The xmllint callback should return the correct default command with unpersisted buffer
    (1/4) [EXECUTE] (X) {'command': '/path/to/xmllint --format -'} should be equal to {'command': '/path/to/xmllint --format %t'}
    (2/4) [EXECUTE] The xmllint callback should return the correct default command
    (2/4) [EXECUTE] (X) {'command': '/path/to/xmllint --format -'} should be equal to {'command': '/path/to/xmllint --format C:\testplugin\test\fixers\dummy.txt'}
    (3/4) [EXECUTE] The xmllint callback should include the XMLLINT_INDENT variable
    (3/4) [EXECUTE] (X) {'command': 'set "XMLLINT_INDENT=  " && /path/to/xmllint --format -'} should be equal to {'command': 'set "XMLLINT_INDENT=  " && /path/to/xmllint --format C:\testplugin\test\fixers\dummy.txt'}
    (4/4) [EXECUTE] The xmllint callback should include additional options
    (4/4) [EXECUTE] (X) {'command': '/path/to/xmllint --format - --nonet'} should be equal to {'command': '/path/to/xmllint --format C:\testplugin\test\fixers\dummy.txt --nonet'}
```

cyberang3l:
Thank you @hsanson. Tests updated. Do I need to do something to re-trigger the CI?

hsanson (approved):
Thanks for the contribution.

PR #4885, opened 2024-12-28
fix ale_python_auto_virtualenv to correctly set virtualenv env vars

rmuir:
First, thank you for ALE!

According to the documentation, `ale_python_auto_virtualenv` should automatically set environment variables for commands, but previously the variables were not set completely or correctly.

Before:
  `PATH` variable was expanded to include `/path/to/venv`
After:
  `PATH` variable is expanded to include `/path/to/venv/bin`
  `VIRTUAL_ENV` variable is set to `path/to/venv`

This mimics exactly what the `activate` scripts do, and allows the configuration knob to work as expected.

For example, after this change, `jedi-language-server` can be installed globally (instead of inside every venv), and it will "just work" (e.g. find references to dependencies in the venv) when editing a file in a project that uses a venv, because the correct variables are set.

### DIFF
--- a/autoload/ale/python.vim
+++ b/autoload/ale/python.vim
@@ -108,12 +108,17 @@ function! ale#python#AutoVirtualenvEnvString(buffer) abort
 
     if !empty(l:venv_dir)
         let l:strs = [ ]
+        " venv/bin directory
+        let l:pathdir = join([l:venv_dir, s:bin_dir], s:sep)
 
         " expand PATH correctly inside of the appropriate shell.
+        " set VIRTUAL_ENV to point to venv
         if has('win32')
-            call add(l:strs, 'set PATH=' . ale#Escape(l:venv_dir) . ';%PATH% && ')
+            call add(l:strs, 'set PATH=' . ale#Escape(l:pathdir) . ';%PATH% && ')
+            call add(l:strs, 'set VIRTUAL_ENV=' . ale#Escape(l:venv_dir) . ' && ')
         else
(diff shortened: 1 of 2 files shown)


rmuir:
Example `ALEInfo` output with this change:

Before:
```
(started) ['/usr/bin/bash', '-c', 'PATH=''/home/rmuir/workspace/beer/api/.venv''":$PATH" ''jedi-language-server''']
```
as you can see, PATH is not set correctly as it doesn't include `bin` directory. the `VIRTUAL_ENV` variable is missing.

After:
```
(started) ['/bin/bash', '-c', 'PATH=''/home/rmuir/workspace/beer/api/.venv/bin''":$PATH" VIRTUAL_ENV=''/home/rmuir/workspace/beer/api/.venv'' ''jedi-language-server''']
```

rmuir:
Converted to draft until I fix the test to expect the correct data.

rmuir:
I'll leave it as a separate discussion if the default value for this variable should stay as `0`, as I don't want to complicate this PR.

IMO, when working with a python virtualenv, you should always set these variables, same as if the user did `activate`.

You can sometimes get away with just `$env/bin/linter`, but not always. So it is best to set the `PATH` and `VIRTUAL_ENV` to ensure linter functions correctly. A good example is `ansible-lint` in a virtualenv, it invokes `ansible-playbook --syntax-check` which will not work with the current approach, as it won't be in the `PATH`.

hsanson (approved):
Looks good to me, thanks.

PR #4886, opened 2024-12-28
Yara yls support

TcM1911:
This adds Yara support to ALE using Avast's language server for yara:
https://avast.github.io/yls/

A `.git` folder is used to determine the project_root so this means the
Yara rules must be in a git repo for the integration to work.

The server only have 1 optional argument (-v, --verbose). Since this is
the case, no additional configuration options are available.

### DIFF
--- a/ale_linters/yara/yls.vim
+++ b/ale_linters/yara/yls.vim
@@ -1,0 +1,17 @@
+" Author: TcM1911
+" Description: A language server for Yara.
+
+call ale#Set('yara_yls_executable', 'yls')
+
+function! ale_linters#yara#yls#FindProjectRoot(buffer) abort
+   let l:project_root = ale#path#FindNearestDirectory(a:buffer, '.git')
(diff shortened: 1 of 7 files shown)


hsanson (changes requested):
Please add an entry for the new linter in `doc/ale-supported-languages-and-tools.txt` file and this is missing a `callback` function to handle the linter output.

TcM1911:
@hsanson, I've added an entry to `doc/ale-supported-languages-and-tools.txt`. With regards to the "missing callback" function, I don't know what you are referring to. I'm not seeing any references to it in the dev doc. I used Zig `zls` as a template for this implementation. So with regards to tests, the `yls` linter tests are essentially testing the same as the linter tests for `zls`.

w0rp:
I fixed the documentation and tests so the CI checks pass. The `callback` isn't needed actually for LSP linters, just for tools where we have to parse the output.

Cheers! :beers:

PR #4888, opened 2024-12-29
lint: treat bash as sh

benknoble:
Vim version 9.1.0965 [1] in commit b9b762c21f (patch 9.1.0965: filetype: sh filetype set when detecting the use of bash, 2024-12-27) broke out bash as a standalone filetype (whose runtime support already redirects through the shell runtime support).

As a consequence, ALE didn't lint filetype=bash well. Teach it to consider bash an alias of sh (until and unless it is desirable to have bash-specific linters, at which point we may need to duplicate some linters like shellcheck or find another way to mark them enabled for bash).

Something like

    runtime! ale_linters/sh/*.vim

in `ale_linters/bash/redirect_to_sh.vim` doesn't work because the linters are defined with ale#linter#Define('sh', …).

[1]: https://github.com/vim/vim/commit/b9b762c21f2b61e0e7d8fee43d4d3dc8ecffd721

### DIFF
--- a/autoload/ale/linter.vim
+++ b/autoload/ale/linter.vim
@@ -11,6 +11,7 @@ let s:linters = {}
 " NOTE: Update the g:ale_linter_aliases documentation when modifying this.
 let s:default_ale_linter_aliases = {
 \   'Dockerfile': 'dockerfile',
+\   'bash': 'sh',
 \   'csh': 'sh',
 \   'javascriptreact': ['javascript', 'jsx'],
 \   'plaintex': 'tex',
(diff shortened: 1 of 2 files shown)


benknoble:
I believe the Vim patch has also landed in Neovim.

saccarosium:
> I believe the Vim patch has also landed in Neovim.

Indeed it landed with: https://github.com/neovim/neovim/pull/31749

hsanson:
@benknoble Since there is a fix already in vim and neovim and the fact that users can define their own aliases, I think this PR is not needed and can be closed?

benknoble:
> @benknoble Since there is a fix already in vim and neovim and the fact that users can define their own aliases, I think this PR is not needed and can be closed?

@hsanson I may have missed it : which fix is available ?

benknoble:
The changes in Vim have been reverted, so I will close this.

PR #4892, opened 2025-01-12
Fix cwd setting to use .stylua.toml if stylua.toml is not found

ponko2:
Now checks `.stylua.toml` to set the `cwd` if `stylua.toml` is not found.


hsanson (approved):
Thanks for the fix.

PR #4900, opened 2025-01-29
fix: memory-limit option for phpstan

kevinquinnyo:
This might only be a problem for newer phpstan versions (2.1.1 here).

If you try to run `phpstan` the way ale will when it builds the option, you will get something like:

```
The "--memory-limit" option requires a value.
```

It wants you to use `--memory-limit=-1` instead.

Updated existing test, but we should probably check to make sure phpstan is ~~backwards~~forward compatible with this `--key=val` option format if this is in fact a change to how that option is parsed, and otherwise make some kind of shim for both version?


hsanson (approved):
Thanks for the fix.

PR #4902, opened 2025-02-03
fix(biome): update biome fixer supports css and graphql

dcai:
Update registry so that biome could support formatting css and graphql


hsanson (approved):
Thanks, looks good.

PR #4917, opened 2025-03-05
Fix 4913 - use JSON output for golangci-lint instead of regexps.

hsanson:
golangci-lint linter uses regular expressions for parsing linter output that is fragile given that golangci-lint is a meta linter that supports many others with different outputs. Fortunately it supports JSON output since at least version 1.7 from 2018  that is more stable and easier to parse.

This PR modifies ALE golangci-lint handler to parse JSON output.


w0rp:
Thanks @hsanson. I built https://github.com/dense-analysis/pie this week to track down potential duplicate issue and this patch fixes a bug that's been around for years. 😄

PR #4909, opened 2025-02-16
Add support for the [djlinter](https://www.djlint.com/)

vds2212:
This code aim to propose a linter for the djlinter from: [djlinter](https://www.djlint.com/)

PR #4918, opened 2025-03-09
Add djlint fixer for various HTML template formats

AdrianVollmer:
* Supported formats:
    - html
    - htmlangular
    - htmldjango
    - jinja
    - handlebars
    - nunjucks
    - gohtmltmpl

* Add doc entries

* Add basic vader test

---

I was a bit confused about how the doc entries relate to file types. I hope it's okay the way I did it.

Also, see #4909, where `djlint` was added as a linter, but for `html` only.

### DIFF
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -98,6 +98,11 @@ let s:default_registry = {
 \       'suggested_filetypes': ['dhall'],
 \       'description': 'Standard code formatter for the Dhall language and removing dead code',
 \   },
+\   'djlint': {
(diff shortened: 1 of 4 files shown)


hsanson (changes requested):
Appreciated if you can add additional tests to the fixer callback for the different filetypes supported. You can see an example in the `test_prettier_fixer_callback.vader` test that sets the filetype before calling the callback and verifies the correct command is generated.  Also test the case where --profile is set in the html_djlint_options.

AdrianVollmer:
Rolled into https://github.com/dense-analysis/ale/pull/4920

PR #4921, opened 2025-03-11
feat: remove logic for unsupported older Neovim

delphinus:
ref https://github.com/dense-analysis/ale/pull/4872

Now it has dropped the support for Neovim 0.6 or older. This commit removes the logic that has become unneeded.

Revert "fix: work with older version of Neovim"

This reverts commit 9b9c74373dc12a8872f46c44366dda9fc49ba5ba.


w0rp (approved):
Thanks, looks good!

PR #4922, opened 2025-03-12
Fix docs for variable names

simmel:
I suspect there was some copy pasta/sed going on.

The code is correct though, just the docs that don't reflect the code.


w0rp (approved):

w0rp:
Cheers! :beers:

PR #4923, opened 2025-03-13
Support for Zlint as Zig linter

kython28:
This PR adds support for [Zlint](https://github.com/DonIsaac/zlint) as a linter for Zig files in ALE.


w0rp (changes requested):
Looks good! Could you add a test case for the linter command itself? We should be able to test what executable and arguments we use.

test/handler/test_zlint_handler.vader
```diff
@@ -0,0 +1,57 @@
...
+  \   },
+  \ ]
+
+  " Get actual output
+  let actual_output = ale_linters#zig#zlint#Handle(buffer, input_lines)
+
+  " Assert equality
+  AssertEqual expected_output, actual_output
+
+Execute(The zlint executable and command should be configured correctly):
```
hsanson, line 45: Appreciated if you can follow convention and move this test to a separate file inside `tests/linter` folder.

hsanson (approved):
Thanks for the contribution.

w0rp (approved):

PR #4932, opened 2025-03-22
Integrate ALE with Neovim LSP Client & Lua

w0rp:
This pull request is a major move toward goals stated in #4815. The changes here accomplish the following.

1. ALE will use Neovim's LSP client in Neovim 0.8 and up by default.
2. ALE can be configured via `ale.setup` and `ale.setup.buffer` functions in Lua.
3. All ALE settings have documentation updated to describe how to configure ALE easily either in Vim script or in Lua.
4. Various ALE settings have been tweaked so Boolean values work equally well as Numbers, to make configuring ALE easier for Neovim users.
5. Various basic ALE functions have been mirrored in one way or another in Lua for easier configuration in Lua.
6. Cover all new and previous Lua code with tests with [busted](https://lunarmodules.github.io/busted/).

This is a major leap forward for integrating ALE much better with built-in Neovim tools, Neovim plugins, and for making it easy to configure ALE in Neovim user configurations that are purely written in Lua.

Issue #3600, opened 2021-02-20
Implement the new LSP diagnostic pull model


w0rp:
# What is the bug?

1. You do just about anything with a language server, such as completion
2. You have to send the document updates to the language server
3. Without the "pull model" the client (ALE) has no control over when diagnostics are sent
4. You can't just "ignore" diagnostics "until you are ready," you have to show them when they come
5. Users perceive that "linting" is happening when they don't want it to, which annoys users

We fix it by implementing the pull model and forcing servers to adapt, plus being careful about when we send document updates.

# History

https://github.com/microsoft/vscode/issues/117042

A proposal under review will allow LSP clients to explicitly request diagnostics from language servers, instead of receiving them at random. This will fix issues such as #3555 where you can receive diagnostics due to ALE having to tell servers that changes have been made to files for features like completion to work, even though you don't want them yet.

If/when this proposal is adopted and the protocol is updated, we should implement support for it in ALE so newer versions of servers automatically stop causing this problem for users over time.

**Update August 2024**: This became part of the Language Server Protocol standard, and now depending on the language server, we can support this in ALE.

# Implementation

1. Implement the pull model!
2. Limit the number of document updates based on updates that are strictly necessary for functionality only. (Hard to manage!)

hsanson:
Still proof of concept: https://github.com/microsoft/vscode-languageserver-node/blob/main/protocol/src/common/proposed.diagnostics.md#L1

w0rp:
Someone nudge me when this is implemented in the actual spec, and I'll support it.

w0rp:
@hsanson They actually did it in the past couple of days! The suggestion I made for a "pull" model for diagnostics instead of a "push" model as an option has been written into the protocol spec. Now we can implement support for it in ALE as the preferred option, so you only see diagnostics in response to an ALE lint cycle.

w0rp:
There's probably an actual server that implements the pull model now I can test against. I pinned this issue so I can try to remember to implement this.

towerpark:
> There's probably an actual server that implements the pull model now I can test against.

rust-analyzer has just [implemented the pull model](https://rust-analyzer.github.io/thisweek/2024/10/28/changelog-257.html), so maybe it could be used for testing if you're going to work on this feature.

w0rp:
After much time, I have merged a wide series of changes which integrates ALE with Neovim's native LSP client and adds support for the diagnostics pull model in both Vim script and also via Neovim's native LSP client.

PR #4933, opened 2025-03-23
Implement LSP diagnostic pull model

w0rp:
A follow up to #4932, and finally implements #3600 after all of this time.

Pyright only just implemented pull diagnostics [in the 1.1.397 release](https://github.com/microsoft/pyright/releases/tag/1.1.397) published mere days ago, and I used that language server to test these changes. I tested this with Vim 9, Neovim 0.8, and Neovim 0.10. Neovim 0.10+ is needed for pull diagnostics support. In my tests this seems to work just fine, and ALE seems to receive diagnostics a little bit faster than before.

The biggest plus of this is that when servers implement pull diagnostics support they'll never send diagnostics to ALE until ALE asks for them, and ALE will be able to track when servers are busy checking files.

### DIFF
--- a/autoload/ale/lsp.vim
+++ b/autoload/ale/lsp.vim
@@ -47,6 +47,7 @@ function! ale#lsp#Register(executable_or_address, project, language, init_option
         \       'definition': 0,
         \       'typeDefinition': 0,
         \       'implementation': 0,
+        \       'pull_model': 0,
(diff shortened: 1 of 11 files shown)


w0rp:
@hsanson This is a follow-up to the last PR, and further enhances LSP support so we can at last explicitly request diagnostics and track when servers are busy checking files.

PR #4934, opened 2025-03-24
fix(biome): for biome fixer, update format to replace deprecated args

dcai:
According to biome cli reference: https://biomejs.dev/reference/cli/#biome-format

`--apply` argument has been deprecated, should use use `--write` instead


w0rp (approved):

w0rp:
Cheers! :beers:

PR #4939, opened 2025-03-29
Remove hadolint rule code from location list entry text

dmitrivereshchagin:
The code is already passed with the code key of location list entry. Before this change, when using the default message format, it appeared twice in the description of each location list entry and twice in each echo message.


hsanson (approved):
Thanks, looks good.

PR #4942, opened 2025-03-30
Always use safe file local variables for erlang-mode fixer

dmitrivereshchagin:
File local variables in Emacs are used in a way similar to Vim modelines.  For example, at the end of the file you might find something like the following:

    %% Local Variables:
    %% erlang-indent-level: 2
    %% something-weird: t
    %% End:

The `erlang-indent-level` variable in this list instructs the Erlang mode to use two columns per indentation level.  But since the `something-weird` variable is likely unknown, both may be ignored.

By default, Emacs in batch mode ignores all variable/value pairs if it encounters at least one that is not known to be safe.  Setting `enable-local-variables` to `:safe` tells Emacs to use only safe values and ignore the rest.

### DIFF
--- a/autoload/ale/fixers/erlang_mode.vim
+++ b/autoload/ale/fixers/erlang_mode.vim
@@ -17,10 +17,11 @@ let s:variables = {
 \}
 
 function! ale#fixers#erlang_mode#Fix(buffer) abort
-    let emacs_executable =
+    let l:emacs_executable =
     \   ale#Var(a:buffer, 'erlang_erlang_mode_emacs_executable')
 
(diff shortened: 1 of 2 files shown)


w0rp (commented):
I'm not too familiar with Erlang. Will setting local variables as safe possibly open the door to arbitrary code execution, or are they completely safe to use? Vim doesn't enable modelines for the root user by default for that reason.

Not that ALE itself is the most secure tool in the world anyway by default.

hsanson (approved):
Looks good, thanks.

PR #4952, opened 2025-04-17
Add support for the Lean 4 LSP, served by Lake.

benjamindblock:
Adds support for the [Lean 4](https://github.com/leanprover/lean4) LSP, which is served by [Lake](https://github.com/leanprover/lean4/blob/master/src/lake/README.md) (Lean Make), the build system and package manager for the Lean language.


hsanson (approved):

PR #4954, opened 2025-04-22
Add initialization_options to Sorbet LSP linter

Tonkpils:
I started getting an error in neovim when opening ruby files with the Sorbet linter.

```
Error executing vim.schedule lua callback: ...nvimEllFDd/usr/share/nvim/runtime/lua/vim/lsp/client.lua:548: RPC[Error] code_name = InvalidParams, message = "Unable to deserialize LSP request: Error deserializing JSON message: Expected field `InitializeParams.initializationOptions` to have value of type `object`, but had value `[]`."
stack traceback:
        [C]: in function 'assert'
        ...nvimEllFDd/usr/share/nvim/runtime/lua/vim/lsp/cl…
```

Turns out Lua's uses the same underlying `table` data type for both arrays and dictionaries. Whenever we pass nothing or `{}` to neovim's LSP client, it gets serialized as an `[]`.  [LSP defines the initializationOptions as anything](https://github.com/neovim/neovim/pull/33551#issuecomment-2819014845). Unfortunately Sorbet is too strict and raises an error when anything other than an object is passed which causes the linter to fail.

Using a predefined object will ensure that Sorbet always receives an object instead of an array. 
This also allows users of this to set their own LSP initialization options 

I'm thinking of following up with a PR to Sorbet's LSP implementation to relax that requirement a bit more or at least not crash in this situations but I figured this would be a good start.

### DIFF
--- a/ale_linters/ruby/sorbet.vim
+++ b/ale_linters/ruby/sorbet.vim
@@ -1,6 +1,7 @@
 call ale#Set('ruby_sorbet_executable', 'srb')
 call ale#Set('ruby_sorbet_options', '')
 call ale#Set('ruby_sorbet_enable_watchman', 0)
+call ale#Set('ruby_sorbet_initialization_options', { 'highlightUntyped': v:false })
 
 function! ale_linters#ruby#sorbet#GetCommand(buffer) abort
     let l:executable = ale#Var(a:buffer, 'ruby_sorbet_executable')
@@ -21,6 +22,6 @@ call ale#linter#Define('ruby', {
 \   'language': 'ruby',
 \   'executable': {b -> ale#Var(b, 'ruby_sorbet_executable')},
 \   'command': function('ale_linters#ruby#sorbet#GetCommand'),
-\   'project_root': function('ale#ruby#FindProjectRoot')
+\   'project_root': function('ale#ruby#FindProjectRoot'),
+\   'initialization_options': {b -> ale#Var(b, 'ruby_sorbet_initialization_options')}
 \})
-
--- a/doc/ale-ruby.txt
+++ b/doc/ale-ruby.txt
@@ -267,6 +267,20 @@ g:ale_ruby_sorbet_enable_watchman
   to files from outside of vim. Defaults to disable watchman because it
   requires watchman to be installed separately from sorbet.
 
+                               *ale-options.ruby_sorbet_initialization_options*
+                                     *g:ale_ruby_sorbet_initialization_options*
+                                     *b:ale_ruby_sorbet_initialization_options*
+ruby_sorbet_initialization_options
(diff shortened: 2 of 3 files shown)


Tonkpils:
@w0rp 👋 after a bit of a discussion in my Sorbet PR https://github.com/sorbet/sorbet/pull/8796, it seems like the sensible thing is for ALE to not predefine an empty dictionary and instead skip sending the field altogether to the server. I'm happy with leaving the default presented in this PR but let me know if there's a different approach you'd like to take

ale_linters/ruby/sorbet.vim
```diff
@@ -1,6 +1,7 @@
 call ale#Set('ruby_sorbet_executable', 'srb')
 call ale#Set('ruby_sorbet_options', '')
 call ale#Set('ruby_sorbet_enable_watchman', 0)
+call ale#Set('ruby_sorbet_initialization_options', { 'highlightUntyped': v:false })
```
hsanson, line 4: By giving users the ability to set `ruby_sorbet_initialization_options` we are opening the possibility of the same issue to happen if the user explicitly sets this option to an empty dict or array. I recommend we clearly document in the help that if this option is set to an empty dictionary that sorbet lsp will fail to initialize.

Other option is to define a `l:default_init_options` variable with the default options, leave `ruby_sorbet_initialization_options` default to `{}` and then extend it with the `l:default_init_options` before passing it to the LSP:

```vimscript
let l:default_options = { 'highlightUntyped': v:false }
let l:init_options = extend(l:default_options, ale#Var(a:buffer, 'ruby_sorbet_initialization_options), "force")

....  pass l:init_options to the lsp definition
```

Tonkpils: Good callout, I'll add documentation to communicate the behavior of passing an empty object to the initialization options.

I don't have strong opinions about doing the extension route with default options. I can go ahead and add those if you think it won't be surprising for a user to set an empty value and have some default value end up being passed out. Either way, they would see an error pop up unless we default.

hsanson: @Tonkpils I don't have strong opinions either. If you add the note to the docs I think this should be ready to merge.

Tonkpils: All done https://github.com/dense-analysis/ale/pull/4954/commits/4171cebcad8c505c91a25a3c218fb84e71c3fd73 let me know if this is what you had in mind or if I there's a better way to word it 🙇

Tonkpils:
👋 @w0rp any thoughts on this?

w0rp:
I think Appveyor got stuck here so this never got merged. I'll force it in and fix Appyeyor if it fails later.

PR #4972, opened 2025-05-08
[python] add pyrefly.vim

oliverralbertini:
https://github.com/facebook/pyrefly

The pyre project has evolved to pyrefly. This adds a pyrefly.vim and test suite along with the necessary stub file for the binary and adds pyrefly entries in docs and other metadata files.

pyrefly is different from pyre in that the LSP doesn't need to have a CWD from which to run, so that makes things a bit simpler.

### DIFF
--- a/ale_linters/python/pyrefly.vim
+++ b/ale_linters/python/pyrefly.vim
@@ -1,0 +1,57 @@
+" Author: oliverralbertini <oliver.albertini@gmail.com>
+" Description: A performant type-checker supporting LSP for Python 3 created by Facebook
+
+call ale#Set('python_pyrefly_executable', 'pyrefly')
+call ale#Set('python_pyrefly_use_global', get(g:, 'ale_use_global_executables', 0))
+call ale#Set('python_pyrefly_auto_pipenv', 0)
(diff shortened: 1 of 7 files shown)


connernilsen:
Hey @oliverralbertini, thank you so much for working on this! Would we be able to keep `pyre.vim` around for now, since we're not fully ready to deprecate it yet? Love to see your support here! 😁

oliverralbertini:
Besides the unit tests, I checked that this works by installing pyrefly with

```
brew install pipx
pipx install firefly
```

And loaded up a file in vim:

```
def foo(myint: int)->str:
    return str(myint)

foo('hi')
```

with this in my vimrc:

```
let g:ale_linters = {
  "\ 'rust': ['rustc', 'rls'],
  \ 'hack': ['hack', 'hhast'],
  \ 'python': ['pyrefly', 'ruff'],
  \ 'sh': ['shellcheck'],
  \ 'bash': ['shellcheck'],
  \ }
```

image

connernilsen (approved):
Not sure if my approval here is going to help, but want to support however I can!

hsanson (approved):
LGTM, thanks for the contribution.

PR #4975, opened 2025-05-15
Support astgrep

mathstuf:
---
Support `ast-grep`. Note that it isn't really associated with a specific language, so I made a new "generic" group for it.

### DIFF
--- a/ale_linters/c/astgrep.vim
+++ b/ale_linters/c/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for c files
+
+call ale#handlers#astgrep#Define('c')
--- a/ale_linters/cpp/astgrep.vim
+++ b/ale_linters/cpp/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for cpp files
+
+call ale#handlers#astgrep#Define('cpp')
--- a/ale_linters/cs/astgrep.vim
+++ b/ale_linters/cs/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for cs files
+
+call ale#handlers#astgrep#Define('cs')
--- a/ale_linters/css/astgrep.vim
+++ b/ale_linters/css/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for css files
+
+call ale#handlers#astgrep#Define('css')
--- a/ale_linters/elixir/astgrep.vim
+++ b/ale_linters/elixir/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for elixir files
+
+call ale#handlers#astgrep#Define('elixir')
--- a/ale_linters/go/astgrep.vim
+++ b/ale_linters/go/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for go files
+
+call ale#handlers#astgrep#Define('go')
--- a/ale_linters/haskell/astgrep.vim
+++ b/ale_linters/haskell/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for haskell files
+
+call ale#handlers#astgrep#Define('haskell')
--- a/ale_linters/html/astgrep.vim
+++ b/ale_linters/html/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for html files
+
+call ale#handlers#astgrep#Define('html')
--- a/ale_linters/java/astgrep.vim
+++ b/ale_linters/java/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for java files
+
+call ale#handlers#astgrep#Define('java')
--- a/ale_linters/javascript/astgrep.vim
+++ b/ale_linters/javascript/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for javascript files
+
+call ale#handlers#astgrep#Define('javascript')
--- a/ale_linters/json/astgrep.vim
+++ b/ale_linters/json/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for json files
+
+call ale#handlers#astgrep#Define('json')
--- a/ale_linters/kotlin/astgrep.vim
+++ b/ale_linters/kotlin/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for kotlin files
+
+call ale#handlers#astgrep#Define('kotlin')
--- a/ale_linters/lua/astgrep.vim
+++ b/ale_linters/lua/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for lua files
+
+call ale#handlers#astgrep#Define('lua')
--- a/ale_linters/nix/astgrep.vim
+++ b/ale_linters/nix/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for nix files
+
+call ale#handlers#astgrep#Define('nix')
--- a/ale_linters/php/astgrep.vim
+++ b/ale_linters/php/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for php files
+
+call ale#handlers#astgrep#Define('php')
--- a/ale_linters/python/astgrep.vim
+++ b/ale_linters/python/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for python files
+
+call ale#handlers#astgrep#Define('python')
--- a/ale_linters/ruby/astgrep.vim
+++ b/ale_linters/ruby/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for ruby files
+
+call ale#handlers#astgrep#Define('ruby')
--- a/ale_linters/rust/astgrep.vim
+++ b/ale_linters/rust/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for rust files
+
+call ale#handlers#astgrep#Define('rust')
--- a/ale_linters/scala/astgrep.vim
+++ b/ale_linters/scala/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for scala files
+
+call ale#handlers#astgrep#Define('scala')
--- a/ale_linters/sh/astgrep.vim
+++ b/ale_linters/sh/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for sh files
+
+call ale#handlers#astgrep#Define('sh')
--- a/ale_linters/solidity/astgrep.vim
+++ b/ale_linters/solidity/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for solidity files
+
+call ale#handlers#astgrep#Define('solidity')
--- a/ale_linters/swift/astgrep.vim
+++ b/ale_linters/swift/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for swift files
+
+call ale#handlers#astgrep#Define('swift')
--- a/ale_linters/tsx/astgrep.vim
+++ b/ale_linters/tsx/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for tsx files
+
+call ale#handlers#astgrep#Define('tsx')
--- a/ale_linters/typescript/astgrep.vim
+++ b/ale_linters/typescript/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for typescript files
+
+call ale#handlers#astgrep#Define('typescript')
--- a/ale_linters/yaml/astgrep.vim
+++ b/ale_linters/yaml/astgrep.vim
@@ -1,0 +1,4 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: astgrep for yaml files
+
+call ale#handlers#astgrep#Define('yaml')
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -22,6 +22,11 @@ let s:default_registry = {
 \       'suggested_filetypes': ['apkbuild'],
 \       'description': 'Fix policy violations found by apkbuild-lint in APKBUILDs',
 \   },
+\   'ast-grep': {
+\       'function': 'ale#fixers#astgrep#Fix',
+\       'suggested_filetypes': ['c', 'cpp', 'cs', 'css', 'elixer', 'go', 'haskell', 'html', 'java', 'javascript', 'json', 'kotlin', 'lua', 'nix', 'php', 'python', 'ruby', 'rust', 'scala', 'sh', 'solidity', 'swift', 'tsx', 'typescript', 'yaml'],
+\       'description': 'Apply ast-grep rules.',
+\   },
 \   'autoimport': {
 \       'function': 'ale#fixers#autoimport#Fix',
 \       'suggested_filetypes': ['python'],
--- a/autoload/ale/fixers/astgrep.vim
+++ b/autoload/ale/fixers/astgrep.vim
@@ -1,0 +1,18 @@
+" Author: Ben Boeckel <github@me.benboeckel.net>
+" Description: Fix files with `ast-grep`.
+
+call ale#Set('astgrep_executable', 'ast-grep')
+call ale#Set('astgrep_scan_options', '--update-all')
+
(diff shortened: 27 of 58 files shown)


mathstuf:
Looks like to get a fixer, some more work is needed. But this is a good start at least.

mathstuf:
`astgrep` is quite grumpy when it tries to get used when there's not a configuration around. Is there a way to disable it if `ale_linters#generic#astgrep#GetProjectRoot` comes back with an empty string?

mathstuf:
Another thing is that I have a local hack to append `generic` as a filetype to inject it into all buffers:

```
diff --git i/autoload/ale.vim w/autoload/ale.vim
index 2f46bce5c..9f9a2048f 100644
--- i/autoload/ale.vim
+++ w/autoload/ale.vim
@@ -96,7 +96,7 @@ endfunction
 function! s:Lint(buffer, should_lint_file, timer_id) abort
     " Use the filetype from the buffer
     let l:filetype = getbufvar(a:buffer, '&filetype')
-    let l:linters = ale#linter#Get(l:filetype)
+    let l:linters = ale#linter#Get(l:filetype . '.generic')
 
     let l:ignore_config = ale#Var(a:buffer, 'linters_ignore')
     let l:disable_lsp = ale#Var(a:buffer, 'disable_lsp')

```

I don't imagine this is the most acceptable way to do this, but some way to have filetype-independent checkers seems useful (e.g., I can imagine spell checkers joining the ranks here; I see some special handling for `cspell` in ALE today?).

mathstuf:
Updated the `GetProjectRoot` to accurately detect the path based on logic in other linters. It now works without complaint when no configuration file exists.

mathstuf:
Hmm. I'm seeing a lot of `ast-grep lsp` instances hanging off of my `nvim` where it is used. Is there something I might be missing here?

rymdbar:
There are a couple of other tools (e.g. cspell, write-good) which are more or less generic already. I have not really thought about this, but it seems like an appropriate question to ask is whether this would better be handled like them? Wouldn't it make sense to be able to use astgrep with some filetypes, but not with others?

I believe the tool ought to be added to [`supported-tools.md`](https://github.com/dense-analysis/ale/blob/edccdfa/supported-tools.md) too.

mathstuf:
>  Wouldn't it make sense to be able to use astgrep with some filetypes, but not with others?

I suppose it could be limited to the languages that `ast-grep` actually compiles in tree-sitter support for. I don't see a way to coax it from a binary though (`completions` doesn't list them anywhere either).

rymdbar:
> I suppose it could be limited to the languages that `ast-grep` actually compiles in tree-sitter support for.

My intention was to question whether adding a new "generic" group is the right design, or if keeping granular control by making ast-grep configurable per filetype as the already existing (e.g. cspell, writegood) general tools are.

I'm not saying this pull-request necessarily does it wrong. Just stating that I noticed it being done differently. Thus wishing to understand whether you have some reasoning behind adding a globally enabled tool rather than mimicking how similar things appear to have been done in the past.

If doing like with pre-existing generic tools, I envision going through [the language crate](https://github.com/ast-grep/ast-grep/blob/main/crates/language/src/lib.rs) once and adding a linter definition per language. I assume new languages are added infrequently enough that updating ale would be a reasonable chore.

Untested, and mainly to illustrate my thinking. A question for all interested parties: Would transforming the suggested linter into a handler, and doing something like this be desirable?

```sh
for lang in $( sed -n 's/^mod \(.*\);/\1/p' < ../ast-grep/crates/language/src/lib.rs ); do
cat > ale_linters/$lang/astgrep.vim << EOF
" Author: Ben Boeckel <redacted>
" Description: astgrep for $lang files

call ale#linter#Define('$lang', {
\   'name': 'astgrep',
\   'lsp': 'stdio',
\   'executable': function('ale#handlers#astgrep#GetExecutable'),
\   'project_root': function('ale#handlers#astgrep#GetProjectRoot'),
\   'command': function('ale#handlers#astgrep#GetCommand'),         
\})                                               
EOF 
done
```

From the nitpicking department, a possible confusion point is with tree-sitter. I understand ast-grep is using it internally. Thus both vim and neovim interface that tool through ale in the same way. At the same time I believe it is a well-known fact that only one out of the two editors itself uses tree-sitter, which might lead to misconceptions. If you're the master of words who is able to add a single word or two to make ast-grep's description express this subtle detail more clearly, then feel welcome to add those to the documentation. If not, please ignore this comment. The description is correct as is, and ale is targeting users who can question their own misunderstandings.

mathstuf:
I don't know what the answer is. I don't like the idea of duplicating the block umpteen times. Maybe we make an `ale#linter#astgrep#Define('$lang')` call that at least shares all the other stuff?

I made the `generic` bit since it seemed like a way to get things that can handle "any" filetype working. If it is preferred to do per-language files, I can do that (though I'd like to add a helper API for it).

mathstuf:
I pulled the typo fix out into #5021.

As for the change here, I figured out what needed to be done for a fixer via #5022. I also refactored out the "generic" support into a module that can enable it for a given language. This is plumbed through via the `g:ale_astgrep_linter_languages` list variable that creates a linter for `ast-grep` on startup for the listed languages.

mathstuf:
Gah, the fixer just replaces the contents with the tool output. I'll need to investigate more.

rymdbar:
Apologies for not replying immediately.

## Implementation discussion

> I don't know what the answer is.

Neither do I, but mimicing how things have been done in the past might be a relatively easy way towards a merge.

I could get the previous version of this working when patching `l:linters` to always contain `'.generic'`. This version with `g:ale_astgrep_linter_languages` I couldn't get to work at all. Not even with a fresh clean `~/.vim/` folder. Is it working for you?

> I made the `generic` bit since it seemed like a way to get things that can handle "any" filetype working.

As far as I understand astgrep doesn't handle any filetype. According to [List of Languages with Built-in Support][langs] its about twenty-four or so. While the _ale_linters/_ folder has  156 entries and there are 414 different filetypes in _…vim/vim91/ftplugin/_. Still I frequently edit several other filetypes which require ftplugins from other sources than vim-runtime.

[langs]: https://ast-grep.github.io/reference/languages.html

> If it is preferred to do per-language files, I can do that (though I'd like to add a helper API for it).

I have pushed the branch [pr-4975-classic][classic]. How does that one look to you?

[classic]: https://github.com/dense-analysis/ale/compare/master...rymdbar:ale:pr-4975-classic

Unfortunately there is a mixup between sh and bash in ale, but treating them as the same should be good enough. Previously tsx didn't exist at all, so I created a new folder for it.

> I don't like the idea of duplicating the block umpteen times.

The definition on the above branch is one line per filetype, since you added the helper. Not too bad, is it? Given that it saves unnecessarily searching for that astgrep-file for 95% of filetypes. It would be great with a way of applying [DRY][] to the still missing documentation. As far as I know nothing in the code base exists to help with that.

[DRY]: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

## Actual functionality
How can one use this? When merely reading I got the impression ast-grep brought in some quite impressive ast-aware code rewriting functionality, but when trying it out it seems that is not actually exposed through the LSP? At least to my eyes the [LSP Server][lsp] section seems to say only diagnostics and the fixer are there. That's great if it helps some people, of course.

The response to _initialize_ is:

```json
{
    "jsonrpc": "2.0",
    "result": {
        "capabilities": {
            "codeActionProvider": {
                "codeActionKinds": [
                    "quickfix.ast-grep",
                    "source.fixAll.ast-grep"
                ],
                "resolveProvider": true
            },
            "executeCommandProvider": {
                "commands": [
                    "ast-grep.applyAllFixes"
                ]
            },
            "hoverProvider": true,
            "textDocumentSync": 1
        },
        "serverInfo": {
            "name": "ast-grep language server"
        }
    },
    "id": 2
}
```

Attempting `:ALECodeAction` anywhere only seems to result in:

```json
{
    "jsonrpc": "2.0",
    "result": null,
    "id": 3
}
```

As for the fixer config, shouldn't it actually better be interfaced using the LSP instead?

[lsp]: https://ast-grep.github.io/guide/tools/editors.html#lsp-server

mathstuf:
Found `read_temporary_file` for the fixer. That makes it work as I'd expect.

> I could get the previous version of this working when patching l:linters to always contain '.generic'. This version with g:ale_astgrep_linter_languages I couldn't get to work at all. Not even with a fresh clean ~/.vim/ folder. Is it working for you?

Not that easily…the load order is not correct I think. It doesn't work if I set it in `.nvimrc` or a [`.lvimrc`](https://github.com/embear/vim-localvimrc) file. But if I call `ale#astgrep#ApplyLanguages()` after startup, it does work.

> As far as I understand astgrep doesn't handle any filetype.

That is true…I think the `generic` thing is not suitable (especially for a tool that actually needs dedicated support for each language in order to properly support it).

> I have pushed the branch [pr-4975-classic](https://github.com/dense-analysis/ale/compare/master...rymdbar:ale:pr-4975-classic). How does that one look to you?

Yeah, that looks good. It also avoids the awkwardness of `ale_astgrep_linter_languages`. I'll squash that in and remove the `linter_languages` cruft.

> Given that it saves unnecessarily searching for that astgrep-file for 95% of filetypes.

That is a good point.

> How can one use this?

You have to write some rules that you want to work on. A simple example for the CMake codebase is in [this MR](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/10839). Though these rules are disabled by default (`severity: 'off'`), [This MR](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/10790) adds more complex rules, but ones that are always active. With the LSP, I get inline diagnostics about things `ast-grep` finds (with default-enabled rules at least).

> As for the fixer config, shouldn't it actually better be interfaced using the LSP instead?

Ideally, yes. But AFAICT, the LSP-as-linter shows inline diagnostic notes but I don't know how to act on them. Is there some way to access "fix" behavior of linters? On that note, when a compilation generates [fixits](https://stackoverflow.com/questions/49748996/apply-clangs-fix-it-hints-automatically-from-command-line), I'd love to be able to filter that back into the same ALE workflow (almost like a build-log-based-LSP I guess?). If that could be hooked up to [SARIF](https://sarifweb.azurewebsites.net/) as well…ooh that'd be quite the pipeline.

rymdbar:
> > How can one use this?
>
> A simple example for the CMake codebase is in [this MR]

These are for whatever language CMake is written in, not the CMake language itself?

> You have to write some rules that you want to work on.

I might need to get this explained to me like I'm five. Would be glad if you could walk me through it.

The ast-guide has a [Lint Rule][lint-rule] chapter, which appears to give an example for javascript. Thus I try creating this **hello.js** in an empty folder and run `ast-grep new`, pressing Enter all through the options given.

```javascript
#!/usr/bin/env node

console.log("World");
```

Then I configure .vimrc to enable `astgrep` for `ft=javascript` and create a **rules/hello-rule.yaml**.

```yaml
---
id: hello-rule
language: javascript
rule:
  pattern: console.log($GREET)
fix: console.log('Hello ' + $GREET)
message: Greet properly, please
severity: error
note: |
  Care should be taken to be correct when greeting someone, or something.
```

Running `ast-grep scan` then results in a seemingly successful invocation with the fix suggested:

```
hello.js
help[hello-rule]: 
@@ -0,3 +0,3 @@
1 1│ #!/usr/bin/env node
2 2│ 
3  │-console.log("World");
  3│+console.log('Hello ' + "World");
```

However I can't get the severity, message or note showing up from `ast-grep scan`. No matter what I set `--report-style` to. Presumably the rule file isn't fully correct?

[lint-rule]: https://ast-grep.github.io/guide/project/lint-rule.html

Opening up **hello.js** with vim + ALE and saving LSP communication with `:call ch_logfile(…)` gives communication, but no diagnostics:

<details>
  <summary>initialize, ALE → ast-grep</summary>

```json
{
    "method": "initialize",
    "jsonrpc": "2.0",
    "id": 2,
    "params": {
        "initializationOptions": {},
        "rootUri": "file:///tmp/ale_fjMAeK",
        "capabilities": {
            "workspace": {
                "workspaceFolders": false,
                "configuration": false,
                "symbol": {
                    "dynamicRegistration": false
                },
                "applyEdit": false,
                "didChangeConfiguration": {
                    "dynamicRegistration": false
                }
            },
            "textDocument": {
                "documentSymbol": {
                    "dynamicRegistration": false,
                    "hierarchicalDocumentSymbolSupport": false
                },
                "references": {
                    "dynamicRegistration": false
                },
                "publishDiagnostics": {
                    "relatedInformation": true
                },
                "rename": {
                    "dynamicRegistration": false
                },
                "completion": {
                    "completionItem": {
                        "snippetSupport": false,
                        "commitCharactersSupport": false,
                        "preselectSupport": false,
                        "deprecatedSupport": false,
                        "documentationFormat": [
                            "plaintext",
                            "markdown"
                        ]
                    },
                    "contextSupport": false,
                    "dynamicRegistration": false
                },
                "synchronization": {
                    "didSave": true,
                    "willSaveWaitUntil": false,
                    "willSave": false,
                    "dynamicRegistration": false
                },
                "codeAction": {
                    "codeActionLiteralSupport": {
                        "codeActionKind": {
                            "valueSet": []
                        }
                    },
                    "dynamicRegistration": false
                },
                "diagnostic": {
                    "dynamicRegistration": true,
                    "relatedDocumentSupport": true
                },
                "typeDefinition": {
                    "dynamicRegistration": false
                },
                "hover": {
                    "dynamicRegistration": false,
                    "contentFormat": [
                        "plaintext",
                        "markdown"
                    ]
                },
                "implementation": {
                    "dynamicRegistration": false,
                    "linkSupport": false
                },
                "definition": {
                    "dynamicRegistration": false,
                    "linkSupport": false
                }
            }
        },
        "rootPath": "/tmp/ale_fjMAeK",
        "processId": 2628
    }
}
```
</details>

<details>
  <summary>initialize, ALE ← ast-grep</summary>

```json
{
    "jsonrpc": "2.0",
    "result": {
        "capabilities": {
            "codeActionProvider": {
                "codeActionKinds": [
                    "quickfix.ast-grep",
                    "source.fixAll.ast-grep"
                ],
                "resolveProvider": true
            },
            "executeCommandProvider": {
                "commands": [
                    "ast-grep.applyAllFixes"
                ]
            },
            "hoverProvider": true,
            "textDocumentSync": 1
        },
        "serverInfo": {
            "name": "ast-grep language server"
        }
    },
    "id": 2
}
```
</details>


<details>
  <summary>initialized, ALE → ast-grep</summary>

```json
{
    "method": "initialized",
    "jsonrpc": "2.0",
    "params": {}
}
```
</details>

<details>
  <summary>textDocument/didOpen, ALE → ast-grep</summary>

```json
{
    "method": "textDocument/didOpen",
    "jsonrpc": "2.0",
    "params": {
        "textDocument": {
            "uri": "file:///tmp/ale_fjMAeK/hello.js",
            "version": 2,
            "languageId": "javascript",
            "text": "#!/usr/bin/env node\n\nconsole.log(\"World\");\n"
        }
    }
}
```
</details>

<details>
  <summary>window/logMessage, ALE ← ast-grep</summary>

```json
{
    "jsonrpc": "2.0",
    "method": "window/logMessage",
    "params": {
        "message": "server initialized!",
        "type": 3
    }
}
```
</details>

<details>
  <summary>window/logMessage, ALE ← ast-grep</summary>

```json
{
    "jsonrpc": "2.0",
    "method": "window/logMessage",
    "params": {
        "message": "file opened!",
        "type": 3
    }
}
```
</details>

<details>
  <summary>workspace/workspaceFolders, ALE ← ast-grep</summary>

```json
{
    "jsonrpc": "2.0",
    "method": "workspace/workspaceFolders",
    "params": null,
    "id": 0
}
```
</details>

<details>
  <summary>textDocument/hover, ALE → ast-grep</summary>

```json
{
    "method": "textDocument/hover",
    "jsonrpc": "2.0",
    "id": 3,
    "params": {
        "textDocument": {
            "uri": "file:///tmp/ale_fjMAeK/hello.js"
        },
        "position": {
            "character": 0,
            "line": 2
        }
    }
}
```
</details>

<details>
  <summary>textDocument/hover, ALE ← ast-grep</summary>

```json
{
    "jsonrpc": "2.0",
    "result": null,
    "id": 3
}
```
</details>

<details>
  <summary>window/logMessage, ALE ← ast-grep</summary>

```json
{
    "jsonrpc": "2.0",
    "method": "window/logMessage",
    "params": {
        "message": "Get Hover Notes",
        "type": 4
    }
}
```
</details>

You do get diagnostics, right? Would it be easy for you to make this simple scenario with **hello.js** and **hello-rule.yaml** work?

> > As for the fixer config, shouldn't it actually better be interfaced using the LSP instead?
> 
> Ideally, yes. But AFAICT, the LSP-as-linter shows inline diagnostic notes but I don't know how to act on them. Is there some way to access "fix" behavior of linters?

[LSP Server][lsp] states:

> Server capability
>  * publish diagnostics
>  * Fix diagnostic code action

My expectation is that a fix action ought to show up if running `:ALECodeAction` on a line with a diagnostic. Does it do that for you?

[lsp]: https://ast-grep.github.io/guide/tools/editors.html#lsp-server

mathstuf:
> These are for whatever language CMake is written in, not the CMake language itself?

Yes. These are rules for CMake's own C++ code.

> You do get diagnostics, right? Would it be easy for you to make this simple scenario with hello.js and hello-rule.yaml work?

Yes.

Screenshot From 2025-09-02 15-25-10

Alas, I'm not *good* at `ast-grep` yet. I fumbled my way through the rules CMake has to get where these are.

> My expectation is that a fix action ought to show up if running :ALECodeAction on a line with a diagnostic. Does it do that for you?

I'm not seeing it :/ . `:ALEFix` shows `ast-grep`. `:ALECodeAction` shows "No code actions received from server".

rymdbar:
> Screenshot From 2025-09-02 15-25-10

I see the screenshot, yet wonder how it works... Could you perhaps have a non-lsp setup to use ast-grep linger in your configs? Maybe you could post your full output from `:ALEInfo`?

> Alas, I'm not _good_ at `ast-grep` yet. I fumbled my way through the rules CMake has to get where these are.

My results with CMake sources are as with javascript. Running `ast-grep scan` in the terminal works, but through ale it is all silent.

I brought this question to ast-grep's [discussions][], and then looked enough at the sources to be very surprised the lsp config appears to be working for you. Could you please have a look at my linked answer and double check on your side?

Depending on whether the lsp communication is broken or not, can be made to work, this pull request might need another path.

[discussions]: https://github.com/ast-grep/ast-grep/discussions/2211#discussioncomment-14290289

mathstuf:
```
 Current Filetype: cpp
Available Linters: ['astgrep', 'cc', 'ccls', 'clangcheck', 'clangd', 'clangtidy', 'clazy', 'cppcheck', 'cpplint', 'cquery', 'cspell', 'flawfinder']
   Linter Aliases:
'cc' -> ['gcc', 'clang', 'g++', 'clang++']
  Enabled Linters: ['astgrep', 'cc', 'ccls', 'clangcheck', 'clangd', 'clangtidy', 'clazy', 'cppcheck', 'cpplint', 'cquery', 'cspell', 'flawfinder']
  Ignored Linters: []
 Suggested Fixers:
  'ast-grep' - Apply ast-grep rules.
  'astyle' - Fix C/C++ with astyle.
  'clang-format' - Fix C, C++, C#, CUDA, Java, JavaScript, JSON, ObjectiveC and Protobuf files with clang-format.
  'clangtidy' - Fix C/C++ and ObjectiveC files with clang-tidy.
  'uncrustify' - Fix C, C++, C#, ObjectiveC, ObjectiveC++, D, Java, Pawn, and VALA files with uncrustify.
```

rymdbar:
Update: We've seen fixes in ast-grep triggered by the linked discussion, as well as have been gifted #5044 which I haven't had the chance to look in detail at yet. I only saw that it likely needs  added test cases.

rymdbar:
Given that this PR effectively depends on #5044, maybe you'd want to transform this into a _Draft_ until it gets unblocked @mathstuf?

w0rp:

hsanson (approved):
Thanks for the contribution.

PR #4990, opened 2025-06-11
Modify call to javac to put options in temp file instead of on command line to prevent problems with command line being too long

MrPottery:
This branch was created to address the problem of getting '**Command too long**' errors from javac becase the classpath is too long.
The change instead writes the javac options to a file in the temporary directory that ALE creates and builds the javac command to use the @<options_file> method of specifying the options.

These changes allowed me to use ALE with my Java projects where previously it failed due to the javac command line being longer than the ~8K allowed by Windows.


w0rp:
If this is a way forward, and I'll need to think about this idea, we should use the functions in ALE for managing temporary files. Have a look at the documentation for them. I'd link to them, but I'm currently on a train. 😃

w0rp:

Issue #3481, opened 2020-12-04
gopls doesn't show any errors in gopath mode


preslavmihaylov:
## Information

**VIM version**
NVIM v0.4.4
Build type: Release

Operating System: 
MacOS

## What went wrong

gopls doesn't work when I make projects in my gopath. When I make them with a go module, everything is nice and dandy.

## Reproducing the bug
```
cd /tmp
mkdir foo
cd foo
go mod init github.com/foo/bar
vim main.go

# add the following contents
package main

import "fmt"

func main() {
	fmt.Prinln("vim-go")
}
```

After a second, ALE nicely shows an error on the proper place:
![image](https://user-images.githubusercontent.com/7642330/101185070-190a6400-365a-11eb-8ec5-f317439268a1.png)

Do the same exercise but in gopath mode:
```
cd $GOPATH/src/github.com
mkdir -p foo/bar && cd foo/bar
vim main.go

# add the same content as before
```

This time, ALE doesn't print anything...
![image](https://user-images.githubusercontent.com/7642330/101185295-5cfd6900-365a-11eb-943d-29f1cbfe2406.png)

However, if I explicitly run `:GoDiagnostics` as I have `vim-go` installed, I get the expected error in the quickfix window -> so it's not a gopls issue:
![image](https://user-images.githubusercontent.com/7642330/101185370-77374700-365a-11eb-9a70-b39c96c6b34f.png)

### :ALEInfo
```

 Current Filetype: go
Available Linters: ['bingo', 'gobuild', 'gofmt', 'golangci-lint', 'golint', 'gometalinter', 'gopls', 'gosimple', 'gotype', 'govet', 'golangserver', 'revive', 'staticcheck']
   Linter Aliases:
'gobuild' -> ['go build']
'govet' -> ['go vet']
  Enabled Linters: ['gofmt', 'golint', 'gopls']
  Ignored Linters: []
 Suggested Fixers: 
  'gofmt' - Fix Go files with go fmt.
  'goimports' - Fix Go files imports with goimports.
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

let g:ale_go_go_executable = 'go'
let g:ale_go_golangci_lint_executable = 'golangci-lint'
let g:ale_go_golangci_lint_options = '--enable-all'
let g:ale_go_golangci_lint_package = 0
let g:ale_go_golint_executable = 'golint'
[...]

PR #4993, opened 2025-06-25
Prefer ale_root option when detecting roots

w0rp:
## Summary
- add `ale#linter#GetRoot` and use it when starting LSP linters
- use `ale_root` first in `ale#python#FindProjectRoot`
- update docs to describe the new root lookup behaviour
- adjust tests and add new test for python root option

## Testing
- `./run-tests --fast` *(fails: docker not found)*

------
https://chatgpt.com/codex/tasks/task_b_685c4b409780832aaeb2154798173c38

### DIFF
--- a/ale_linters/python/pylsp.vim
+++ b/ale_linters/python/pylsp.vim
@@ -36,7 +36,7 @@ function! ale_linters#python#pylsp#GetCwd(buffer) abort
     \   'name': 'pylsp',
     \   'project_root': function('ale#python#FindProjectRoot'),
     \}
-    let l:root = ale#lsp_linter#FindProjectRoot(a:buffer, l:fake_linter)
+    let l:root = ale#linter#GetRoot(a:buffer, l:fake_linter)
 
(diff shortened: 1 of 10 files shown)


rymdbar:
Problems, as well as confusion, with getting the correct project root happens often enough for this to warrant both a technical change such as the one in this pull-request and some updates to the documentation.

I might be confused myself and mixing contexts, but figure it's likely a good idea to add this quote of myself from #3481 here:

> Adding a sentence or section to the documentation about the need for the for a project root to be found would likely be a good idea too. This might be a common theme applying to the documentation for most language servers.

> Expecting everyone to have read :help ale-options.root and realise it is what is needed is probably unrealistic.

My mind is a bit fuzzy and my notes aren't making it crystal clear, but I think once the project roots are discovered the same way for all language servers, what needs to be done is to go through documentation for them all and make sure it mentions the requirement for a root to be found and the ability to hard code the root if desired.

Issue #4989, opened 2025-06-10
Add support for fortitude


nbehrnd:
**Name:** fortitude
**URL:** https://github.com/PlasmaFAIR/fortitude

PlasmaFAIR's fortitude is fast syntax checker and linter for contemporary Fortran.  Inspired by ruff (for Python), it helps both to identify oddities in the source code, as well as to optionally fix them for the better by a set of community based rules.  The tool's action can be adjusted to the user's needs (for instance about rules gradually implemented).

Fortitude already is recognized for instance by the community of [fortran-lang.org](https://fortran-lang.discourse.group/t/fortitude-a-fortran-linter/8834).  Its efficiency to work reliably well and fast was presented by its main developers Liam Pattinson and Peter Hill e.g., by their talk on May 22nd, 2025 at a Cambridge University Research Software Engineering Seminar ([youtube recording](https://www.youtube.com/watch?v=gAdgqfq-yxU)).  I think it would be nice addition to Ale, complementary to the present checks with gcc.

PR #5004, opened 2025-07-14
Add support for fortitude (fortran linter)

gomfol12:
Hi,
I added support for fortitude. Requested here #4989.
I have also added a simple test and run the docker container, it seems ok.
But some linting tests fail, and I can not figure out why.

These are the tests:
```
diff supported-tools.md and doc/ale-supported-languages-and-tools.txt tables
and
Look for table of contents issues
```

Let me know if I forgot something else.

### DIFF
--- a/test/linter/test_redpen.vader
+++ b/test/linter/test_redpen.vader
@@ -2,8 +2,9 @@ Before:
   " This is just one example of a language using the linter.
   call ale#assert#SetUpLinterTest('markdown', 'redpen')
 
-  unlet! g:ale_redpen_options
-  call ale#Set('redpen_options', '')
+  Save g:ale_redpen_options
+
+  let g:ale_redpen_options = ''
 
 After:
   call ale#assert#TearDownLinterTest()


ale_linters/fortran/fortitude.vim
```diff
@@ -0,0 +1,59 @@
...
+        return []
+    endtry
+
+    if empty(l:errors)
+        return []
+    endif
+
+    let l:output = []
+
+    for l:error in l:errors
```
w0rp, line 32: We have a helper function for decoding JSON potentially split over many lines and returning a default value you can try here.

```suggestion
    let l:output = []

    for l:error in ale#util#FuzzyJSONDecode(a:lines, [])
```

doc/ale.txt
```diff
@@ -3482,6 +3482,7 @@ documented in additional help files.
   fortran.................................|ale-fortran-options|
     gcc...................................|ale-fortran-gcc|
     language_server.......................|ale-fortran-language-server|
+    fortran...............................|ale-fortran-fortitude|
```
w0rp, line 3485: The tests will probably fail on the list of supported tools here because they are expected to be sorted alphabetically.

w0rp (changes requested):
Looks pretty good. I suggested a couple of fixes.

gomfol12:
Thanks for the feedback. I applied the suggested fixes and the test are passing now.

w0rp (dismissed):
This is good. I'll add a test for the linter itself also after merging this. Cheers! :beers:

test/handler/test_fortitude_handler.vader
```diff
@@ -0,0 +1,74 @@
+Before
```
w0rp, line 1:
```suggestion
Before:
```

The colon was missing here so this broke. I should be able to fix this by applying the suggestion.

w0rp:
I have actually merged this now after fixing the failing test and adding another. Cheers! 🍻

PR #5006, opened 2025-07-15
Add configurable options support for redpen linter ✨

### LINKED ISSUE #5005: Add support for config options for reopen
`redpen` allows setting CLI options such as `-c <path/to/config/file>`. Currently ale offers no way to set that.
I'd like to see a `ale_redpen_options` variable that allows us to specify these options.

Gidgidonihah:
Redpen linter previously had hardcoded command-line options and duplicated linter definitions across all supported file types (asciidoc, markdown, review, rst, tex, text). This refactoring centralizes the linter definition and adds support for user-configurable options via g:ale_redpen_options.

Key changes:
- Created ale#handlers#redpen#DefineLinter() to eliminate code duplication
- Added ale#handlers#redpen#GetCommand() to support configurable options
- All file types now use shared configuration and command building
- Added comprehensive test coverage for option handling
- Updated documentation for all affected file types

This allows users to customize redpen behavior with additional command-line options while maintaining backward compatibility and reducing maintenance overhead.

Closes #5005


autoload/ale/handlers/redpen.vim
```diff
@@ -1,5 +1,21 @@
...
+function! ale#handlers#redpen#ResetOptions() abort
+    call ale#Set('redpen_options', '')
+endfunction
+
+" Reset the options so the tests can test how they are set.
+call ale#handlers#redpen#ResetOptions()
+
+function! ale#handlers#redpen#GetCommand(buffer) abort
+    let l:options = ale#Var(a:buffer, 'redpen_options')
+    let l:filetype = getbufvar(a:buffer, '&filetype')
```
w0rp, line 13: `&filetype` could be set to multiple types at once, for example `markdown.special` or some other construction. You can instead define the command with a lambda `{b -> string}` And you can apply `ale#Pad(ale#Var(b, 'redpen_options')` as `ale#Pad` will prefix `' '` if the value of the options variable is non-empty. In place of `l:filetype` here you can put in the value of `a:filetype` below from `ale#handlers#redpen#DefineLinter`.

autoload/ale/handlers/redpen.vim
```diff
@@ -1,5 +1,21 @@
 " Author: rhysd https://rhysd.github.io
 " Description: Redpen, a proofreading tool (http://redpen.cc)
+"
+function! ale#handlers#redpen#ResetOptions() abort
+    call ale#Set('redpen_options', '')
+endfunction
+
+" Reset the options so the tests can test how they are set.
+call ale#handlers#redpen#ResetOptions()
```
w0rp, line 9: Remove this code and just call `ale#Set` at the top of the file, which sets a default value. You might want to read the implementation of `ale#Set` in `autoload/ale.vim`.

w0rp (changes requested):

Issue #4973, opened 2025-05-14
Add emoji to indicate LSP in supported-tools.md


F-Kearney:
As someone new to vim and lsp's it was a bit confusing to figure out what was a lsp, linter, or fixer from the supported-tools page. So I thought adding an emoji to represent the lsp's (say :zap:) would be a good way to help people that don't know. I have an updated md file that adds the :zap: emoji if the lsp is listed in Microsoft's lsp repo (https://microsoft.github.io/language-server-protocol/implementors/servers/). If that's wanted I can upload it? Otherwise loving ALE so far!

rymdbar:
As can be seen in my comment to [#3471](https://github.com/dense-analysis/ale/issues/3714#issuecomment-2778447631), I obviously find this to be a good idea:

> One idea would be to add a 💬 emoji after each tool which is a language server.

My thinking was that the speech bubble represented language, but please feel free to convince me the flash emoji better describes it.

Pull requests is the way to make changes happen. In case you have an actual implementation, do upload it and there is a chance it'll improve for everyone.

ghost:
Yes I think the speech bubble makes more semantic sense, I'll swap it to that. Making a PR now.

PR #5011, opened 2025-07-22
Appended :speech_balloon: emoji to any line containing an lsp

F-Kearney:
Adding :speech_balloon: to indicate that thing is a language server protocol. (#4973)

The script used to check if something is an lsp is pulling from Microsoft's list of lsp's (https://microsoft.github.io/language-server-protocol/implementors/servers/) and also if the word contains "lsp" anywhere. This should have picked up most valid matches but there may be a few stragglers left.

### DIFF
--- a/supported-tools.md
+++ b/supported-tools.md
@@ -14,15 +14,16 @@ formatting.
 
 **Legend**
 
-|      Key      |             Definition                                            |
-| ------------- | ----------------------------------------------------------------- |
-| :floppy_disk: | May only run on files on disk (see: `help ale-lint-file-linters`  |
-|   :warning:   | Disabled by default                                               |
+|      Key       |             Definition                                            |
+| -------------- | ----------------------------------------------------------------- |
+|:speech_balloon:| Language Server Protocol (LSP)                                    |
+| :floppy_disk:  | May only run on files on disk (see: `help ale-lint-file-linters`  |
+|   :warning:    | Disabled by default                                               |
 
 ---
 
 * Ada
-  * [ada_language_server](https://github.com/AdaCore/ada_language_server)
+  * [ada_language_server](https://github.com/AdaCore/ada_language_server) :speech_balloon:
   * [cspell](https://github.com/streetsidesoftware/cspell/tree/main/packages/cspell)
   * [gcc](https://gcc.gnu.org)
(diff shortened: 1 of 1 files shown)


rymdbar:
Having those 💬 sure looks beautiful!

That's an impressive work you've done going through the list. As you state there may be a few stragglers left and I'd say it seems like a hard task to get it fully correct through a purely manual process. Unassisted by technology, it likely becomes impossibly to keep the list correct as new tools are added.

Luckily we do have the source code which already contains information on which tools are LSPs. What would you think about adding a script in the [`test/script/`][scripts] folder, and make it run in CI just like the other scripts in that directory?

I started hacking together a [`check-lsp-emojis`][fork] script on my fork. Far from ready, but arguably already usable for quality assurance of this PR.

When running it, I see that rust-analyzer is missing its 💬 and that cargo is misclassified to have one. Rust is a language I happen to know well enough be sure the script gets those two right.

Apart from those, there are a few more differences to be addressed. While some might be mistakes, it is not unlikely that there are tools which may operate both as LSPs and as single-shot linters. In case ale only use them in the latter form, creating issues to add the LSP mode as a new tools might be a welcome outcome of this exercise.

In the pushed version of the script all underscores are replaced with dashes. That was added to filter out false positives. Perhaps the right thing to do is to update the documentation to have tools mentioned with their correct "spelling". If instead it turns out the linter definitions has gotten the separator signs wrong, correcting them ought to only be done if at the same time adding the previous name as an alias.

Would you be up for running the script and do a round of re-classification?

[fork]: https://github.com/rymdbar/ale/tree/add-lsp-emoji/test/script/check-lsp-emojis
[scripts]: https://github.com/rymdbar/ale/tree/add-lsp-emoji/test/script

rymdbar:
> In the pushed version of the script all underscores are replaced with dashes.

Only after posting the comment I realize this is due to [`:help ale-dev`][doc] and thus works as intended.

> Use `snake_case` names for linter names, so they can be used as part of variable names. You can define `aliases` for linters, for other names people might try to configure linters with.

Thus we do not wish to update linter names in the source code, even if their official names contain dashes.

[doc]: https://github.com/dense-analysis/ale/blob/3d68ec78/doc/ale-development.txt#L126

w0rp (approved):
I'll take this. Cheers! :beers:

PR #5012, opened 2025-07-23
linter/scala/metals: add build.mill as project root detect file

woshiluo:
If you use mill as build tool, it will use `build.mill` as build configure file instead of `build.sc`. I belive we should also add `build.mill` for `potential_roots`.


rymdbar:
Should this also be added to [ale-scala.txt][doc] which currently reads:

> By default the project root is found by searching upwards for `build.sbt`, `build.sc`, `.bloop` or `.metals`. If the project root is elsewhere, you can override the project root directory.

?

[doc]: https://github.com/dense-analysis/ale/blob/3d68ec78/doc/ale-scala.txt#L35

woshiluo:
> Should this also be added to [ale-scala.txt](https://github.com/dense-analysis/ale/blob/3d68ec78/doc/ale-scala.txt#L35) which currently reads:
> 
> > By default the project root is found by searching upwards for `build.sbt`, `build.sc`, `.bloop` or `.metals`. If the project root is elsewhere, you can override the project root directory.
> 
> ?

Thanks for mentioned that, I have added this in the newest commit.

w0rp:
Cheers! 🍻

PR #5015, opened 2025-07-24
add: support project-local `ansible-lint`

yining:
this commit adds support for project-local `ansible-lint` (via `uv`, `poetry`, and `pipenv`) when using the system-wide installed one is not possible or not desirable.

### DIFF
--- a/ale_linters/ansible/ansible_lint.vim
+++ b/ale_linters/ansible/ansible_lint.vim
@@ -2,9 +2,42 @@
 " Description: ansible-lint for ansible-yaml files
 
 call ale#Set('ansible_ansible_lint_executable', 'ansible-lint')
+call ale#Set('ansible_ansible_lint_auto_pipenv', 0)
+call ale#Set('ansible_ansible_lint_auto_poetry', 0)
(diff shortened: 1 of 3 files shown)


rymdbar:
> when using the system-wide installed one is not possible or not desirable.

Isn't such configuration what `b:ale_ansible_ansible_lint_executable` and its global twin already are for?

I fully lack authority to accept or reject PR:s, yet I feel my criticism might be appropriate and helpful. With the python community's track record, aren't chances they soon deprecate both poetry and uv to replace it with their next new world order?

Wouldn't this fit better in a separate, new or existing,  poetry and/or uv centric plugin? One which could integrate with ale merely by setting the existing variable to the value required by the discovered environment?

I personally have a UsePipx() function in my `.vimrc`. It runs `pipx environment` to find the applicable `$PIPX_LOCAL_VENVS`, messes with `$MYPYPATH` and what not. Should that be added to ale too? For all tools written in python and for all possible ways someone might wish to edit python projects?

w0rp (approved):
This is how other Python linters are set up, so I'm happy to set it up this way. Cheers! :beers:

PR #5010, opened 2025-07-22
Handle ftp.vim.org (NXDOMAIN)

rymdbar:

PR #5016, opened 2025-07-25
Update oelint pattern

offa:
Recent oelint versions (6.0+) include a `[branch:…]` at the end of each line, which can break parsing.

### DIFF
--- a/ale_linters/bitbake/oelint_adv.vim
+++ b/ale_linters/bitbake/oelint_adv.vim
@@ -17,7 +17,7 @@ function! ale_linters#bitbake#oelint_adv#Command(buffer) abort
 endfunction
 
 function! ale_linters#bitbake#oelint_adv#Handle(buffer, lines) abort
-    let l:pattern = '\v^(.+):(.+):(.+):(.+):(.+)$'
+    let l:pattern = '\v^(.+):(.+):(.+):(.+):(.+)( \[branch:.+)$'
     let l:output = []
 
     for l:match in ale#util#GetMatches(a:lines, l:pattern)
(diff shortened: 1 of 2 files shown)


offa:
The Appveyor test failure is not related to this PR:

> Invoke-WebRequest : The remote name could not be resolved: 'ftp.vim.org'

rymdbar:
> The Appveyor test failure is not related to this PR:
> 
> > Invoke-WebRequest : The remote name could not be resolved: 'ftp.vim.org'

Appveyor tests are failing for all PR:s, yes. My suggestion for fixing that problem is in PR #5010.

offa:
Thanks for the hint @rymdbar.

w0rp (changes requested):
I have fixed the AppVeyor test issues now if you rebase again.

Could you upload the pattern here so it matches both old and new versions of output? You should be able to add an additional test and keep the old one. I'd update the text for the new and old tests to mention the versions we're testing for.

offa:
@w0rp PR rebased and the handler works with both formats now.

ale_linters/bitbake/oelint_adv.vim
```diff
@@ -37,6 +37,10 @@ function! StripAnsiCodes(line) abort
     return substitute(a:line, '\e\[[0-9;]\+[mK]', '', 'g')
 endfunction
 
+function! RemoveBranch(line) abort
```
w0rp, line 40: This function can pollute the global namespace and might conflict with a function a use has defined. I'll merge this and use a function name specific to this file.

w0rp (approved):

PR #5027, opened 2025-08-14
Make phpactor configurable and documented

rymdbar:
Add php_phpactor_executable for phpactor and make it configurable through php_phpactor_config.

When phpactor was first added in 3b8ff65, it had no configuration. Thus the documentation for it could be completely blank. That is why it now seems like an initial addition


w0rp:
Cheers! :beers:

PR #5032, opened 2025-08-18
tombi: support `toml_tombi_offline` flag for offline usage

mathstuf:
---
Cc: @rymdbar


hsanson (approved):
Thanks for the contribution and @rymdbar thanks for the review.

PR #5033, opened 2025-08-19
ale: add `FindNearestFileOrDirectory` function

mathstuf:
Some anchor paths can be files or directories (e.g., `.git` is a directory normally, but a file for worktrees). Add a function that finds either type of path and returns the nearest.

---
Cc: @rymdbar


mathstuf:
If this API is suitable, I'll add commits to migrate `.git` searching to use this to improve `git worktree` support ALE-wide. Separate commits should help bisection/reverting for anything that it causes problems for (e.g., the tool itself doesn't properly support worktrees).

rymdbar:
> Cc: @rymdbar

Apologies for the delay looking at this. Looks good to me, and apparently to w0rp too.

I ran grep and navigated the sources a bit, pushed #5038 and created #5039. Would you still be up for going through the list of tools and adapt them to use your new function? I might have a go at a few more of the tools I'm actively using.

Somewhat related, I also filed #5040 which is more of an idea than something I'm sure anyone (me included) is fully committed to.

Issue #5035, opened 2025-08-29
Add support for expert


pmonson711:
**Name:** expert
**URL:** https://github.com/elixir-lang/expert/

Expert is a brand new LSP for the Elixir language. I'll put in a PR shortly.

PR #5036, opened 2025-08-29
add support for expert

pmonson711:
Hopefully this should cover it #5035.

I wasn't sure if/how I should cover the call to `ale#path#FindNearestExecutable` with tests, I'm happy to find examples and add them if wanted/needed.


rymdbar (commented):
It seems I missed okaying the updated changes. Apologies for that.

Looks good now (when already merged).

pmonson711:
No worries, also I've noticed the expert API has updated for some of the nightly releases. I'll put in a new Pr with the version detection and new switch to use stdio with the LSP.

PR #5038, opened 2025-08-31
Use FindNearestFileOrDirectory, perl languageserver

rymdbar:
Update to use mentioned function to find both .git/ directories and .git work-tree files.

### DIFF
--- a/ale_linters/perl/languageserver.vim
+++ b/ale_linters/perl/languageserver.vim
@@ -9,7 +9,7 @@ function! ale_linters#perl#languageserver#GetProjectRoot(buffer) abort
     " Makefile.PL, https://perldoc.perl.org/ExtUtils::MakeMaker
     " Build.PL, https://metacpan.org/pod/Module::Build
     " dist.ini, https://metacpan.org/pod/Dist::Zilla
-    let l:potential_roots = [ 'Makefile.PL', 'Build.PL', 'dist.ini', '.git' ]
+    let l:potential_roots = [ 'Makefile.PL', 'Build.PL', 'dist.ini' ]
 
     for l:root in l:potential_roots
         let l:project_root = ale#path#FindNearestFile(a:buffer, l:root)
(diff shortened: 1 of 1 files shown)


rymdbar:
Tests are failing due to supposedly:

> ./ale_linters/perl/languageserver.vim:20 Blank line required after `let`

However, there **is** a blank line [after `let`][l21] (or rather before `if`). The [documentation][docs] does not appear to mention anything about any non-existing required empty line under these circumstances, nor does vint indicate it. Is the test-suite misidentifying this as a false positive? is the documentation incomplete? or am I reading something incorrectly?

[l21]: https://github.com/dense-analysis/ale/pull/5038/commits/d95ac8ff#diff-182614f5c139a16569433e23a43fddfaa912e0c5e6ed8b702b1a9b471fcd0201R21

[docs]: https://github.com/dense-analysis/ale/blob/528e2595/doc/ale-development.txt#L88

hsanson:
@rymdbar this  error is from the ale_custom_linting_rules linter.  I recommend you enable it in your configuration so vim/nvim warns you about these issues:

```lua
linters = {
   vim = { "vint", "ale_custom_linting_rules" },
}
```

The error is misleading since it says "after" but I can get  the warning to dissapear by adding a blank line before that line.

rymdbar:
> @rymdbar this error is from the ale_custom_linting_rules linter.

Yes? As already covered in [previous comment](https://github.com/dense-analysis/ale/pull/5038#issuecomment-3240120118).

> The error is misleading since it says "after" but I can get the warning to dissapear by adding a blank line before that line.

So you're suggestion, @hsanson, is that we unblock this PR by updating the documentation? Please feel free to suggest how the new wording should be.

I'm more getting the hunch this is caused by a false positive. If it indeed is correct, what is the motivation for requiring an empty line at this position?

ale_linters/perl/languageserver.vim
```diff
@@ -18,6 +18,11 @@ function! ale_linters#perl#languageserver#GetProjectRoot(buffer) abort
             return fnamemodify(l:project_root . '/', ':p:h:h')
         endif
     endfor
+    let l:project_root = ale#path#FindNearestFileOrDirectory(a:buffer, '.git')
```
hsanson: @rymdbar simply add a blank line before this line to get the linter check pass.

Issue #5037, opened 2025-08-31
A ale lsp startup bug.or just my bad?


shigma1379:
## Information

**VIM version**

NVIM v0.11.3                                                                                                            
Build type: Release   

Operating System: Win10 22H2 

## What went wrong

days ago ,When I edit c or cpp file in a git project.A error display below : "Client clangd:C:\Users\shigma quit with exit code 1 and signal 0. Check log for errors: C:/Users/shigma/AppData/Local/nvim-data/lsp.log "
Then I set the log level and checked lsp.log, this what I got (The last line translated):
<details>
  <summary>lsp.log</summary>
"[START][2025-08-31 01:22:13] LSP logging initiated
[INFO][2025-08-31 01:22:13] .../vim/lsp/rpc.lua:658	"Starting RPC client"	{ cmd = { "cmd", "/s/c/", "clangd" }, extra = {} }
[...]

PR #5043, opened 2025-09-04
fix issue 5037. "/s/c/" to "/s/c"

shigma1379:
a bug on windows.
link: https://github.com/dense-analysis/ale/issues/5037


hsanson (approved):
Cannot test this but looks simple enough and I assume author tested it.

PR #5045, opened 2025-09-09
Add pymarkdown fixer

AdrianVollmer:


rymdbar:
I suggest this to be merged.

According to pymarkdown's [docs][], using tabs is one failure which this tool is capable to act as a fixer for.

[docs]: https://pymarkdown.readthedocs.io/en/latest/user-guide/#fix-mode-failure-correction

Thus preparing a file can be done with:

```sh
printf '%s\n\n\t%s\n' '# Heading' 'Indented by tab' > PR5045.md
```

Configuring ale for this PR and using the fixer sure converts the tab to spaces.

The documentation updates makes sense and there are added test cases which seem reasonable at the glance I gave. All looks good to me.

PR #5049, opened 2025-09-19
Add --editor-mode flag when invoking rubocop

maxjacobson:
Since RuboCop 1.61 (released in February 2024), RuboCop accepts an `--editor-mode` flag which improves editor integrations like ale.

Some of RuboCop's auto-corrections can be surprising or annoying to run on save. When RuboCop is running via an LSP or when the `--editor-mode` flag is passed, it will understand that it is running in an editor, and it will hold off on making changes that might be surprising or annoying.

For example, if I write

```ruby
def call
  results = some_process
end
```

This has an unused variable, and RuboCop will remove the unused variable when you run it. However, if you're in the middle of editing, you may not want it to remove that unused variable, because you may be about to add a usage of it.

More context:

- PR which introduced it: https://github.com/rubocop/rubocop/pull/12682
- Release notes for 1.61: https://github.com/rubocop/rubocop/releases/tag/v1.61.0
- Docs: https://docs.rubocop.org/rubocop/1.80/configuration.html#contextual
- Issue where I learned about this flag: https://github.com/rubocop/rubocop/issues/14539

In order to make sure this isn't a breaking change for ale users who use an older version of RuboCop, this implementation only enables the `--editor-mode` flag when the version of RuboCop is >= 1.61.0.

In the meantime, I am opting into this behavior by setting this option:

```vim
let g:ale_ruby_rubocop_options = "--editor-mode"
```

So I appreciate that this seam was already introduced.

### DIFF
--- a/autoload/ale/fixers/rubocop.vim
+++ b/autoload/ale/fixers/rubocop.vim
@@ -1,7 +1,6 @@
 call ale#Set('ruby_rubocop_options', '')
 call ale#Set('ruby_rubocop_auto_correct_all', 0)
 call ale#Set('ruby_rubocop_executable', 'rubocop')
-call ale#Set('ruby_rubocop_editor_mode', '1')
 
 " Rubocop fixer outputs diagnostics first and then the fixed
 " output. These are delimited by a "=======" string that we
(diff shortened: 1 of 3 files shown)


maxjacobson:
Opening as a draft because I see I ought to write some tests for this... I expect all of the existing tests should still pass, but it should be possible to introduce a new test for the change in behavior. I just need to figure out how to run them 😅

maxjacobson:
All right, I've added some tests to this one. Feedback very welcome!

hsanson (changes requested):
There is no need to make this a breaking change. Ale has `ale#semver#RunWithVersionCheck` helper method that allows to set different commands based on the tool version.

Check [golangci-lit fixer]( https://github.com/dense-analysis/ale/blob/4217461c4859c4fac31ce3d2e22574218f244cf5/autoload/ale/fixers/golangci_lint.vim#L42C12-L42C42) for an example or search for that function in ALE's code as there are many examples using it.

maxjacobson:
> There is no need to make this a breaking change. Ale has `ale#semver#RunWithVersionCheck` helper method that allows to set different commands based on the tool version.

Thanks for bringing this to my attention! I've updated the implementation in https://github.com/dense-analysis/ale/pull/5049/commits/b4b228c86872723a025b7a1bb60cf61028c54ff9. What do you think?

maxjacobson:
:+1: thanks for your help!

PR #5050, opened 2025-09-25
Adding --no-color flag to stylelint args

stevematney:
At least some editors don't work on linting because of colors in the `stylelint` output. This should resolve that.

### DIFF
--- a/ale_linters/css/stylelint.vim
+++ b/ale_linters/css/stylelint.vim
@@ -6,7 +6,7 @@ call ale#Set('css_stylelint_use_global', get(g:, 'ale_use_global_executables', 0
 
 function! ale_linters#css#stylelint#GetCommand(buffer) abort
     return '%e ' . ale#Pad(ale#Var(a:buffer, 'css_stylelint_options'))
-    \   . '--no-color --stdin-filename %s'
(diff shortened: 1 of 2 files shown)


rymdbar:
Its always great to see bugfix pull-requests!

However, tools should not require such flags when operating on stdin/stdout. That really ought to be a channel without CSI commands. Where is the bug report towards stylelint? Could you please report and attempt getting the problem fixed at the source first, before allowing their issue to cascade into unrelated pieces of software?

In case the stylelint team are unable or unwilling to fix their issue, I guess  one could probably add this workaround to ale. Before doing so, one must consider that this fixer integration presumably once worked and that backwards compability is a goal here. If flags are added in later versions, the [`ale#semver#RunWithVersionCheck`][semver] helper method is typically used to keep setups working for users who might be stuck on older versions.

[semver]: https://github.com/dense-analysis/ale/blob/4217461/autoload/ale/semver.vim#L37

A quick look shows that `--no-color` became documented in https://github.com/stylelint/stylelint/issues/1479, but really belongs to some chalk dependency. How that stuff works requires a bit of effort to figure out though, as the [cheeseshop][] 404:s.

[cheeseshop]: https://pypi.org/project/chalk/

PR #5051, opened 2025-10-03
biome: use --stdin-file-path for fixer

apeschar:
This PR changes the Biome fixer to use stdin with --stdin-file-path instead of a temporary file.

By using --stdin-file-path instead of a temporary file, Biome knows where the file is and applies the inclusion/exclusion rules from the configuration file appropriately.

The tests have been updated and all tests pass.

### DIFF
--- a/autoload/ale/fixers/biome.vim
+++ b/autoload/ale/fixers/biome.vim
@@ -1,12 +1,11 @@
 function! ale#fixers#biome#Fix(buffer) abort
     let l:executable = ale#handlers#biome#GetExecutable(a:buffer)
     let l:options = ale#Var(a:buffer, 'biome_options')
-    let l:apply = ale#Var(a:buffer, 'biome_fixer_apply_unsafe') ? '--write --unsafe' : '--write'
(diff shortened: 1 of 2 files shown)


hsanson:
This looks like a breaking change. What this --stdin-file-path flag added recently to biome? if so would be better to check the tool version and add this flag only if supported.

apeschar:
I don't think it is a breaking change. Both the `--write` and `--stdin-file-path` flags were added in the same version 1.8.0, so this change should be compatible with every version of Biome that ALE currently works with (ie, 1.8.0 and up).

~~~
dev % ./node_modules/.bin/biome --version
Version: 1.7.3

dev % ./node_modules/.bin/biome check --write --stdin-file-path test.js <<<"hey"
Error: --write is not expected in this context
dev % ./node_modules/.bin/biome check --write test.js
Error: --write is not expected in this context
~~~

~~~
dev % ./node_modules/.bin/biome --version
Version: 1.8.0

dev % ./node_modules/.bin/biome check --write --stdin-file-path test.js <<<"hey"
hey;
dev % ./node_modules/.bin/biome check --write test.js
Checked 1 file in 874µs. No fixes needed.
~~~

hsanson (approved):
Thanks.

PR #5053, opened 2025-10-07
fix phpstan: use configured level 0

### LINKED ISSUE #3139: phpstan is called with "-l 4" instead of 0 in certain situations
## Information

**VIM version**
NVIM v0.5.0-dev
Build type: RelWithDebInfo

Operating System: Linux, Ubuntu 18.04.4 LTS

## What went wrong

Ale always runs PHPStan with `-l 4` (level 4 of strictness) if `php_phpstan_level =0` and there's no `phpstan.neon` file.

## Reproducing the bug

1. Create PHP project without `phpstan.neon` file or with configuration in `phpstan.neon.dist` 
2. add `let g:ale_php_phpstan_level = 0` in your `~/.vimrc`
3. run ALE and check ALEInfo

The issue is caused by [these phpstan.vim lines](https://github.com/dense-analysis/ale/blob/master/ale_linters/php/phpstan.vim#L22):
```vim
let l:config_file_exists = ale#path#FindNearestFile(a:buffer, 'phpstan.neon')

if empty(l:level) && empty(l:config_file_exists)
    " if no configuration file is found, then use 4 as a default level
    let l:level = '4'
endif
```

If `l:level` is set to `0` above code `empty(l:level)` evaluates to `true`. If `l:config_file_exists` is empty as well the level value is always being set to `4`.

I know nothing about vimscript but my guess is `FindNearestFile` should be only called if there's no explicit configuration path in `g:ale_php_phpstan_configuration` or that file doesn't exist (maybe), and when it's called it should look for both `phpstan.neon` and `phpstan.neon.dist`. Also there should be specific check for `l:level == 0` because the result of `empty(0)` gives false positive.

### :ALEInfo
```
(finished - exit code 1) ['/bin/bash', '-c', '''bin/phpstan'' analyze --no-progress --error-format raw -c ''phpstan.neon.dist'' -l ''4'' ''/home/cg/Projects/mailservice/src/MailService/MailShipment/Domain/Event/ShipmentWasCreated.php'' < '
'/tmp/nvimczCUv3/5/ShipmentWasCreated.php''']
<<<OUTPUT STARTS>>>
/home/cg/Projects/mailservice/src/MailService/MailShipment/Domain/Event/ShipmentWasCreated.php:14:Method MailService\MailShipment\Domain\Event\ShipmentWasCreated::new() should return MailService\MailShipment\Domain\Event\ShipmentWasCreated
 but returns CG\EventSourcing\AggregateEvent.
<<<OUTPUT ENDS>>>
```

biinari:
Fixes #3139

Use `g:ale_php_phpstan_level` for `phpstan` `-l` (`--level`) command argument if it is configured to `0` (as a number rather than string).

`0` (number) is considered to be empty, but `'0'` (string) is not.

### DIFF
--- a/ale_linters/php/phpstan.vim
+++ b/ale_linters/php/phpstan.vim
@@ -27,6 +27,10 @@ function! ale_linters#php#phpstan#GetCommand(buffer, version) abort
 
     let l:level =  ale#Var(a:buffer, 'php_phpstan_level')
 
+    if type(l:level) is v:t_number
+        let l:level = string(l:level)
+    endif
+
     if empty(l:level) && empty(ale_linters#php#phpstan#FindConfigFile(a:buffer))
         " if no configuration file is found, then use 4 as a default level
         let l:level = '4'
--- a/test/linter/test_phpstan.vader
+++ b/test/linter/test_phpstan.vader
@@ -59,6 +59,13 @@ Execute(project with level set to 3):
   AssertLinter 'phpstan',
   \ ale#Escape('phpstan') . ' analyze --no-progress --errorFormat json -l ' . ale#Escape('3') . ' %s'
 
+Execute(project with level set to 0):
+  call ale#test#SetFilename('phpstan-test-files/foo/test.php')
+  let g:ale_php_phpstan_level = 0
+
+  AssertLinter 'phpstan',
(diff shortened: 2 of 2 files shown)


hsanson (approved):

biinari:
Thanks

PR #5057, opened 2025-10-27
fix s:TemporaryFilename in win11, & verilator in win32

congyong:
## Summary

**Operating System:** Windows 11  
**Language:** Verilog  
**Linter Tool:** Verilator   4.216 2021-12-05 rev v4.216-22-g141c5da3  &  5.026
**Tested Code:** [a.zip](https://github.com/user-attachments/files/23161048/a.zip)

---

## Motivation and Context

While testing the Verilator integration on **Windows 11**, I observed that the temporary file path generated by `s:TemporaryFilename` was inconsistent with the expected Windows path format.  
For example:

```
D:/Personal/Temp/V6UA83E.tmp\a.v
```

The existing implementation uses the following logic  (  function! s:TemporaryFilename(buffer) in `verilator.vim`) :

```vim
return ale#util#Tempname() . (has('win32') ? '\' : '/') . l:filename
```

This approach always appends a backslash (`\`) on Windows systems.  
However, on Windows 11, temporary paths returned by `ale#util#Tempname()` may contain either forward slashes (`/`) or backslashes (`\`), resulting in mixed and invalid paths.

---

## Description of Changes in ‘command.vim’

To ensure consistent and valid path construction across Windows versions while maintaining backward compatibility,  
the separator is now determined dynamically based on the actual temporary path:

```vim
if has('win32') && l:tmpname =~ '\\'
    let l:sep = '\'
else
    let l:sep = '/'
endif
```

This change ensures correct handling of paths on Windows 11 and remains compatible with older Windows systems.

---

## Changes in `verilator.vim`

Windows paths differ from Unix-style paths in that they include drive letters (e.g., `D:`) and additional colons.  
The parsing logic in `verilator.vim` has been updated to correctly handle such cases, preventing path-related errors when running Verilator on Windows.

```vim

if has('win32')
    " match filename until last colon before line number
    let l:pattern = '^%\(Warning\|Error\)[^:]*:\s*\(.\{-}\):\(\d\+\):\(\d\+\)\?:\? \(.\+\)$'
else
    let l:pattern = '^%\(Warning\|Error\)[^:]*:\s*\([^:]\+\):\(\d\+\):\(\d\+\)\?:\? \(.\+\)$'
endif

```

---

## Testing

The changes were tested locally on **Windows 11** with Verilator.  
After applying the modifications, the Verilog linter executed successfully, and temporary files were created using valid path formats.

---

## Additional Notes

These updates improve cross-platform consistency without affecting Unix-based environments.  
No regressions were observed in local testing.

### DIFF
--- a/autoload/ale/util.vim
+++ b/autoload/ale/util.vim
@@ -317,6 +317,25 @@ function! ale#util#Tempname() abort
     return l:name
 endfunction
 
+" Detect the best path separator for the current environment.
+function! ale#util#PathSeparator(...) abort
+    " 1. Get base dir from args OR  a temporary file name (reflects the real system path style)
+    let l:tmpname = a:0 ? a:1 : ale#util#Tempname()
+    " 2. On Windows, if the path contains '\', use it.
+    if has('win32')
+        if l:tmpname =~# '\\'
+            return '\\'
+        endif
+        " Otherwise, fallback to '/' (e.g., WSL, Git Bash, MSYS2)
+        return '/'
+    endif
+    " 3. On Unix-like systems, always use '/'
+    return '/'
+endfunction
+
+
+
+
 " Given a single line, or a List of lines, and a single pattern, or a List
 " of patterns, return all of the matches for the lines(s) from the given
 " patterns, using matchlist().
--- a/ale_linters/elixir/elixir_ls.vim
+++ b/ale_linters/elixir/elixir_ls.vim
@@ -6,7 +6,7 @@ call ale#Set('elixir_elixir_ls_config', {})
 
 function! ale_linters#elixir#elixir_ls#GetExecutable(buffer) abort
     let l:dir = ale#path#Simplify(ale#Var(a:buffer, 'elixir_elixir_ls_release'))
-    let l:cmd = has('win32') ? '\language_server.bat' : '/language_server.sh'
+    let l:cmd = ale#util#PathSeparator(l:dir) . (has('win32') ? 'language_server.bat' : 'language_server.sh')
 
     return l:dir . l:cmd
 endfunction
--- a/ale_linters/elm/make.vim
+++ b/ale_linters/elm/make.vim
@@ -175,7 +175,9 @@ function! ale_linters#elm#make#IsTest(buffer) abort
         return 0
     endif
 
-    let l:tests_dir = join([l:root_dir, 'tests', ''], has('win32') ? '\' : '/')
+    let l:sep = ale#util#PathSeparator(l:root_dir) 
+
+    let l:tests_dir = join([l:root_dir, 'tests', ''], l:sep)
 
     let l:buffer_path = fnamemodify(bufname(a:buffer), ':p')
 
--- a/ale_linters/java/javac.vim
+++ b/ale_linters/java/javac.vim
@@ -64,7 +64,7 @@ function! ale_linters#java#javac#GetCommand(buffer, import_paths, meta) abort
 
         " Automatically include the jaxb directory too, if it's there.
         let l:jaxb_dir = fnamemodify(l:src_dir, ':h:h')
-        \   . (has('win32') ? '\jaxb\' : '/jaxb/')
+        \   . (ale#util#PathSeparator(l:src_dir) == '\' ? '\jaxb\' : '/jaxb/') 
 
         if isdirectory(l:jaxb_dir)
             call add(l:sp_dirs, l:jaxb_dir)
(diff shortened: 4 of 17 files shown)


rymdbar:
Thanks for your neat and well-explained pull request, @congyong!

One worry I see is that the replaced pattern is still used elsewhere within the ale sources. Running `grep -R "has('win32') ? '.' : '/'" . | wc -l` shows 20 matches.

Fixing this instance is great. Fixing the others too would be even better. Not every occurrence of the pattern might be lucky enough to have access to a value from a freshly called `tempname()` to determine separator from. Thus as an alternative solution I suggest adding an `ale#util#PathSeparator()` returning either '/' or '\\'. Likely using the same logic as `tempname()` documents and using that new utility function instead.

Maybe something like this? (Not verified to work)

```vim
function! ale#util#Pathseparator() abort
    let l:separator = ''

    if has('win32') && exists('+shellslash')
        if &shellslash
            let l:separator = '/'
        else
            let l:separator = '\\'
        endif
    elseif has('win32')
        if &shellcmdflag =~ '^-' && &shell !~ 'powershell\|pwsh'
            let l:separator = '/'
        else
            let l:separator = '\\'
        endif
    else
        let l:separator = '/'
    endif

    return l:separator
endfunction
```

Checking whether the result of `tempname()` matches '\\' or '/' depending on that suggested `…#PathSeparator()` function returns would make a great (couple of) test case(s).

Would you agree addressing the pattern seems like a reasonable thing to do? Could you be up for doing it within the scope of this pull request?

congyong:
Hello, [rymdbar](https://github.com/rymdbar) 
 initially, I also intended to approach it this way. However, I was concerned about potential historical contexts or legacy considerations that I might not have fully understood. so I only fixed this specific instance. 
Going forward, I can allocate some time to review other sections of the code to assess whether they can also be repaired by replacing the relevant logic with SAME  function.

and There are too many corners in the forward combing, so consider other ways to implement the code.
maybe we can try this function


```vim
" Detect the best path separator for the current environment.
function! ale#util#PathSeparator(...) abort
    " 1. Get base dir from args OR  a temporary file name (reflects the real system path style)
    let l:tmpname = a:0 ? a:1 : ale#util#Tempname()
    " 2. On Windows, if the path contains '\', use it.
    if has('win32')
        if l:tmpname =~# '\\'
            return '\\'
        endif
        " Otherwise, fallback to '/' (e.g., WSL, Git Bash, MSYS2)
        return '/'
    endif
    " 3. On Unix-like systems, always use '/'
    return '/'
endfunction

```


~Yong

congyong:
Background of this function: I have reviewed the relevant code that uses Win32 for checking, and most of its calling scenarios contain directory information. Therefore, I added a judgment logic to the PathSep function. If the passed-in args include a directory, the format of this directory will be verified; if not, a temporary file (Tempfile) will be generated to parse the directory format.
There are many failed test cases (case failed). I need to check this issue when free ,  PAUSE 

In addition, the code style under ale_linter is not consistent, and each file needs to be reviewed one by one.

autoload/ale/util.vim
```diff
@@ -317,6 +317,25 @@ function! ale#util#Tempname() abort
...
+    if has('win32')
+        if l:tmpname =~# '\\'
+            return '\\'
+        endif
+        " Otherwise, fallback to '/' (e.g., WSL, Git Bash, MSYS2)
+        return '/'
+    endif
+    " 3. On Unix-like systems, always use '/'
+    return '/'
+endfunction
```
w0rp: Can we pre-compute the value to return here and store it in an `s:` variable with an `exists` check on the `s:` variable? That will avoid multiple calls to `Tempname()`. We can probably make directories consistent so they don't flip-flop between `\` and `/` on Windows.

congyong: sure,   
```vim
" Detect the best path separator for the current environment.
function! ale#util#PathSeparator(...) abort
    " If the cached value doesn't exist, compute and initialize it
    if !exists('s:SeparatorValue')
        " 1. Get base dir from args OR a temporary file name (reflects the real system path style)
        let l:tmpname = a:0 ? a:1 : ale#util#Tempname()
        
        " 2. Determine and cache the separator
        if has('win32')
            if l:tmpname =~# '\\'
                let s:SeparatorValue = '\\'
            else
                " Fallback to '/' (e.g., WSL, Git Bash, MSYS2)
                let s:SeparatorValue = '/'
            endif
        else
            " 3. On Unix-like systems, always use '/'
            let s:SeparatorValue = '/'
        endif
    endif

    " Return the cached value directly
    return s:SeparatorValue
endfunction
```

w0rp:

w0rp:
Thanks for the effort on this. I've taken an alternative approach to solving this by using `ale#path#GetAbsPath` on the tempname and filename so they are joined and then simplified to have the same slashes in a few places. Let me know if you have any issues, or open another PR if needed.

See: [https://github.com/dense-analysis/ale/commit/641294465e20750b64b6984433fc2e5229c68044](<https://github.com/dense-analysis/ale/commit/641294465e20750b64b6984433fc2e5229c68044>)

PR #5065, opened 2025-11-12
Fix j2lint linter

PsickOSSH:
This PR **adds tests** to verify that the option formats work correctly for **j2lint**.


hsanson (approved):
Thanks for the contribution.

PR #5055, opened 2025-10-20
Add SuperHTML linter for HTML

yoan667:
This pull request adds support for [superhtml](https://github.com/kristoff-it/superhtml) as a linter for html templates in ALE.

Fixes https://github.com/dense-analysis/ale/issues/4850

PR #5069, opened 2025-11-22
Finalize #5055, adding SuperHTML linter for HTML

rymdbar:
Remove bogus test of non-existing g:ale_html_superhtml_options.

As suggested and ignored in comment to #5055.

### DIFF
--- a/test/linter/test_superhtml.vader
+++ b/test/linter/test_superhtml.vader
@@ -15,12 +15,6 @@ Execute(The executable should be configurable):
   AssertLinter '/usr/local/bin/superhtml',
   \ ale#Escape('/usr/local/bin/superhtml') . ' check --stdin'
 
-Execute(The options should be configurable):
-  let g:ale_html_superhtml_options = '--some-option'
-
-  AssertLinter 'superhtml',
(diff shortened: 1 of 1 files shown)


rymdbar:
It seems it is not possible to add @yoan667 as a reviewer here, but a review from that account would be practically required.

yoan667:
Hello @rymdbar, sorry for the oversight. I had no intention of ignoring your feedback. I thought I had resolved the issue in the Vader test file, but I must have forgotten to push the change. I’ll be more careful next time, and again, I'm sorry.

rymdbar:
> Hello @rymdbar, sorry for the oversight.

It might actually be me who ought to apologize for having phrased myself unnecessarily harsh. Sorry about that, @yoan667!

I do get angry when these things happen, but that anger should be directed at Microsoft who provides this horrible platform. Wouldn't be great if ALE was hosted on some forge which helps with reviews rather than does it best of hiding and hindering clear communication‽ Even Atlassian gets that working better in Bitbucket for Christs sake, and that's not exactly an example of anything done right…

PR #5073, opened 2025-12-04
Fix linter `npm-groovy-lint`

Jorenar:
I wanted to use `npm-groovy-lint` to check some Gradle files, but I was not getting any diagnostics.

Changes:
* removed `filename` field from handler's output (it was pointing to temporary file),
* removed `--loglevel warning` from default options (so by default we see linter's defaults),
* added `--noserver` as a default option (to avoid lingering process after closing Vim),
* appending rule name to diagnostic message (so user knows which rules to toggle/modify in linter's config),
* adjusted tests,
* updated docs.


Jorenar:
Instead of `--noserver`, a potentially better option would be to call `npm-groovy-lint --killserver` on VimLeave

PR #5074, opened 2025-12-05
fix-proselint

### LINKED ISSUE #5042: Upcoming proselint release requires new "check" verb
See for reference:
- https://github.com/amperser/proselint/issues/1378#issuecomment-3253594866
- https://github.com/amperser/proselint/commit/01b507e3d6eb23f21a009913556468c974b14082
- https://github.com/amperser/proselint/pull/1400

cc @Nytelife26

I'm not sure what the release timeline is or what the best way is for Ale to support proselint: we could try to support multiple versions and insert the `check` verb as appropriate, or we could declare that only the next release (whatever it is) is supported.

It looks like there are lots of proselint linters (I count 14) that duplicate the call to `ale#linter#Define`, so a first step might be to refactor a common `ale#proselint#GetCommand` or similar.

PsickOSSH:
This PR fixes the problem with newer versions of Proselint by using the output of the `--version` command.

Fixes #5042

### DIFF
--- a/ale_linters/asciidoc/proselint.vim
+++ b/ale_linters/asciidoc/proselint.vim
@@ -1,9 +1,11 @@
 " Author: Daniel M. Capella https://github.com/polyzen
 " Description: proselint for AsciiDoc files
 
+call ale#Set('proselint_executable', 'proselint')
+
 call ale#linter#Define('asciidoc', {
 \   'name': 'proselint',
-\   'executable': 'proselint',
-\   'command': 'proselint %t',
+\   'executable': function('ale#proselint#GetExecutable'),
+\   'command': function('ale#proselint#GetCommandWithVersionCheck'),
 \   'callback': 'ale#handlers#unix#HandleAsWarning',
 \})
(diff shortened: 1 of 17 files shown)


lveneris:
Thank you for taking the time to update this. Following [a conversation](https://github.com/amperser/proselint/issues/1451#issuecomment-3555490886) in amperser/proselint#1451, we were [considering re-adding `--version` as an alias to `proselint version`](https://github.com/amperser/proselint/pull/1460) for backwards compatibility, but I realise this breaks the fundamental assumption upon which your PR rests. 

~~Since I am not familiar with ALE internals, I have some questions: How is this handled for other applications? Are there facilities for parsing the output of `--version` if we were to go ahead with putting it back? Would it be preferable to add a `version` key to the JSON output? Or is it ultimately easiest to just leave `--version` out at this point?~~

Update: I have a potential resolution to this. `--output-format json` did not exist prior to 0.16. Therefore, you can use `proselint version --output-format json` instead of `proselint --version`, and if that succeeds, you're on 0.16+.

PsickOSSH:
I’m not familiar with ALE’s internals either, but the solution you propose using `proselint version --output-format json` instead of `proselint --version` works for me, and I can adapt to that.

PR #5081, opened 2025-12-27
fix: replace deprecated client.request method call with colon syntax

ilovemikael:
I don't have much else to say; sorry if this is already being worked on elsewhere (as the deprecation warnings have been present for months now, not just an issue with this project either), but I couldn't find it being discussed anywhere.

### DIFF
--- a/lua/ale/lsp.lua
+++ b/lua/ale/lsp.lua
@@ -83,7 +83,7 @@ module.start = function(config)
         -- we can't get a client via get_client_by_id until after `on_init` is
         -- called. By deferring execution of calling the init callbacks we
         -- can only call them after the client becomes available, which
-        -- will make ications for configuration changes work, etc.
+        -- will make notifications for configuration changes work, etc.
         vim.defer_fn(function()
             vim.fn["ale#lsp#CallInitCallbacks"](config.name)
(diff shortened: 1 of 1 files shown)


w0rp:
Do we have a way of checking which versions of Neovim this will work with? I'd like to ensure we can be both backwards and forwards compatible.

ilovemikael:
W.r.t. forwards compatibility, I use nvim built against the latest git master branch and have been using my fork of this repo to remove the deprecation warnings; ale has still worked fine without a hitch for the last two weeks or so of doing that.

Sorry for not providing documentation referencing the deprecation, it can be found [here](https://neovim.io/doc/user/deprecated.html#_lsp) (scroll down to below deprecated in 0.11, it only lets you link to changes for 0.12 =( ). Since the methods were deprecated in 0.11, I'd say any version ≥0.11 would be supported.

There's a [PR](https://github.com/nvimdev/lspsaga.nvim/pull/1546) covering basically the same topic which you can look at for reference; the creator of that said colon syntax is supported since nvim 0.5 but later changed it to use the regular syntax for ≤0.10 to avoid some errors that I'm really not knowledgeable of, you should probably just review it yourself.

Issue #3814, opened 2021-07-11
Fixing rust files using `rustfmt` does not recognize cargo project settings


jmcantrell:
## Information

**VIM version**
NVIM v0.5.0
Build type: Release

Operating System: arch linux

## What went wrong

Since ale is simply calling the `rustfmt` executable with whatever options were configured, some options that are crucial to how `rustfmt` works are not being recognized. In particular, when creating a new cargo project, the default edition is 2018, whereas the default edition when running `rustfmt` (as opposed to `cargo fmt`) is 2015. As long as you're not using any language features that are 2018-only, everything appears to work fine.

One solution is to have ale always run `rustfmt` with the later edition (`let g:ale_rust_rustfmt_options = '--edition 2018'`), but that might not be appropriate if you're working on an older rust project.

For now, I'm going to work on making my configuration of ale a little smarter by trying to detect whether or not the context is a cargo project. I'll post any findings here, in case anyone else runs into this.

Even though I'm not convinced this is technically a bug anymore, it's definitely something that's unexpected, and is likely to affect many people, as most rust development is done in a cargo project, and the default edition for new cargo projects is 2018.

## Reproducing the bug

1. install the ale plugin (broken on master and v3.1.0)
2. have an ftplugin (`~/.config/nvim/ftplugin/rust.vim`) defining `b:ale_fixers = ['rustfmt']`.
3. start a new cargo project (specifically edition 2018): `cargo new --edition 2018 foo`
3. edit `src/main.rs`
4. make a change that is fixable in either edition 2015 or 2018 (like indent size)
5. verify that fixing works
4. make a change that is specific to edition 2018 (like define a function declared with the `async` keyword)
4. verify that fixing no longer works
5. add `let g:ale_rust_rustfmt_options = '--edition 2018'` to the ftplugin
6. verify that fixing works again

### :ALEInfo

```
 Current Filetype: rust
Available Linters: ['analyzer', 'cargo', 'rls', 'rustc']
  Enabled Linters: ['analyzer']
  Ignored Linters: []
 Suggested Fixers: 
  'remove_trailing_lines' - Remove all blank lines at the end of a file.
  'rustfmt' - Fix Rust files with Rustfmt.
  'trim_whitespace' - Remove all trailing whitespace characters at the end of every line.
 Linter Variables:

let g:ale_rust_analyzer_config = {}
let g:ale_rust_analyzer_executable = 'rust-analyzer'
 Global Variables:

[...]

PR #5083, opened 2025-12-29
Add cargo-fmt fixer for rust

Felixoid:
The issue is described in #3814; `rustfmt` produces different output without additional tweaks.

Here, the `cargo-fmt` fixer is added together with tests and docs.


Felixoid:
While working a little bit with it, I found it's a bad idea. `cargo-fmt` can't work in stdin mode; it always processes all files in a project.

PR #5084, opened 2026-01-08
Utilizing -fzf for tsserver responses along with regular lsp responses

stevematney:
I was really excited about the addition of #5018, but was a bit disappointed when it didn't work in my typescript project. This gets that working and unifies the display and formatting of reference responses down singular code paths, so all responses get processed if there are more output branches in the future.

### DIFF
--- a/autoload/ale/references.vim
+++ b/autoload/ale/references.vim
@@ -19,64 +19,39 @@ function! ale#references#ClearLSPData() abort
 endfunction
 
 function! ale#references#FormatResponseItem(response_item, options) abort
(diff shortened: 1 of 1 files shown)


stevematney:
I was thinking about the possibility of future `FindReferences` use cases, and realized that the previous code I'd written would not have leant itself at all well to the prospect of more use cases. That being the case, I reconfigured `FormatResponseItem` to take a `response_item` object that is expected to be sanitized when it gets passed in, and allowed each response case (tsserver or lsp) to handle its own configuration of that object. This allows for variation to happen at the point where it is most relevant, and opens up `FindReferences` to allow for any number of outliers.

PR #5094, opened 2026-02-10
Fix multiple lints appearing on the same line.

1wsx10:
The [ale demo video](https://github.com/dense-analysis/ale/assets/3518142/8f2ff760-6a87-4704-a11e-c109b8e9ec41) on the main page shows two lints on the same line briefly. the second lint appears on the line below the first one, instead of being concatenated.

vim9 virtualtext support was added to ale, but it was not made to put subsequent lints on following lines - instead they are concatenated to the end of the same line (and so get elided quickly).

this patch fixes it so that it matches the demo video.

### DIFF
--- a/autoload/ale/virtualtext.vim
+++ b/autoload/ale/virtualtext.vim
@@ -210,12 +210,18 @@ function! ale#virtualtext#ShowMessage(buffer, item) abort
             call add(s:hl_list, l:hl_group)
         endif
 
+        let l:align = "after"
+        if !empty(prop_list(l:line, { 'bufnr': a:buffer, 'types': s:hl_list }))
+            let l:align = "below"
+        endif
+
         " We ignore all errors from prop_add.
         silent! call prop_add(l:line, 0, {
         \   'type': l:hl_group,
         \   'text': ' ' . l:msg,
         \   'bufnr': a:buffer,
         \   'text_padding_left': l:col_pad,
+        \   'text_align': l:align,
         \})
     endif
 endfunction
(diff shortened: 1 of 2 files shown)


1wsx10:
The multiline option also does not sort by priority, even though the single line version does.
Untitled

this causes the highest priority item to appear at the end of the list (as it was previously sorted with `ale#util#LocItemCompare` which puts higher priorities later in the list.)
image

I'm not sure why the highest priority items should be at the end of the list, rather than the beginning. I assume its good for other code-paths and wasn't checked for multiline virtual text. maybe the multiline virtual text codepath should `reverse()` the list?

w0rp:
> I'm not sure why the highest priority items should be at the end of the list, rather than the beginning. I assume its good for other code-paths and wasn't checked for multiline virtual text. maybe the multiline virtual text codepath should `reverse()` the list?

It's been a while since I've thought about it, and I can't quite remember myself, but there's a reason for it. I think it has to do with how some code works out which message is the most important one to show by taking the last. That would makes sense when you have loops that are setting data and the later value naturally replaces the first one.

I prefer ALE's behavior the way it is now where messages are concatenated. I might update that demo video later. The default behavior these days is to show online one message per line.

1wsx10:
> > I'm not sure why the highest priority items should be at the end of the list, rather than the beginning. I assume its good for other code-paths and wasn't checked for multiline virtual text. maybe the multiline virtual text codepath should `reverse()` the list?
> 
> It's been a while since I've thought about it, and I can't quite remember myself, but there's a reason for it. I think it has to do with how some code works out which message is the most important one to show by taking the last. That would makes sense when you have loops that are setting data and the later value naturally replaces the first one.
> 
> I prefer ALE's behavior the way it is now where messages are concatenated. I might update that demo video later. The default behavior these days is to show online one message per line.

Could we have it as an option then?

w0rp:
> Could we have it as an option then?

Yeah, you can add it as an option you can turn on. Just remember to document it. I'll re-open the PR so you can work on it here.

w0rp:

PR #5113, opened 2026-04-03
fix expert lsp stdio and register capability

coriocactus:
tested with expert v0.1.0 installed via homebrew and ale at `ba8b9cba` (2026-04-01) installed via vim-plug on macos

expert's lsp server requires `--stdio` to use stdio transport (see [expert README](https://github.com/elixir-lang/expert/blob/9de53c22d2a829549356090451a122fa2d0cfd51/README.md?plain=1#L40) and `expert --help`), but the ale linter definition launched it without the flag so the server never entered lsp mode

separately, expert sends a `client/registerCapability` request after initialization to dynamically register capabilities, and ale had no handler for it, i.e., the server blocked waiting for a response and never processed `textDocument/didOpen` or any subsequent messages, making features like go-to-definition non-functional

the first commit adds `--stdio` to the expert launch command, the second adds a handler that acknowledges `client/registerCapability` requests with a null response, matching how ale already handles `workspace/configuration`

note: `client/registerCapability` is part of the lsp spec and other servers may also use it, so this fix is not expert-specific

### DIFF
--- a/ale_linters/elixir/expert.vim
+++ b/ale_linters/elixir/expert.vim
@@ -7,6 +7,6 @@ call ale#linter#Define('elixir', {
 \   'name': 'expert',
 \   'lsp': 'stdio',
 \   'executable': {b -> ale#Var(b, 'elixir_expert_executable')},
-\   'command': '%e',
+\   'command': '%e --stdio',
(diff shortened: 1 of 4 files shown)


hsanson (approved):
Thanks for the fix.

PR #5114, opened 2026-04-04
add stable keyboard scrolling to vim floating preview popups

coriocactus:
vim floating preview windows for hover and detail output had no keyboard scrolling, so long content required the mouse to read

this adds a popup filter for Vim popupwin previews that handles <C-n>/<C-p> for line scrolling, <C-d>/<C-u> for half-page scrolling, mouse wheel events, and <Esc> to dismiss; unhandled keys pass through normally and moved: 'any' still closes the popup on cursor movement; to prevent resize jitter while scrolling, ALE locks popup width to the maximum content display width and height to the rendered height when users have not set size limits themselves, and disables resize while those managed dimensions are active; scroll clamping at both boundaries prevents the viewport from moving into blank space, and the bottom boundary accounts for line wrapping as well as wrap: 0; users can override the filter or supply explicit sizing via g:ale_floating_preview_popup_opts, and neovim floating windows are unchanged

the filter runs in filtermode: 'n' so it does not interfere with insert-mode completion keys; mapping is left at the default so user normal-mode mappings are respected, which means custom mappings on <C-n> etc will take precedence over popup scrolling; the size-locking cleanup uses unlet! defensively so future refactors cannot hit a missing-key error on managed_size

### DIFF
--- a/autoload/ale/floating_preview.vim
+++ b/autoload/ale/floating_preview.vim
@@ -86,12 +86,15 @@ function! s:VimShow(lines, options) abort
         call s:VimCreate(a:options)
     endif
 
+    call s:ResetManagedPopupSize()
+
     " Execute commands in window context
     for l:command in get(a:options, 'commands', [])
         call win_execute(w:preview['id'], l:command)
     endfor
 
     call popup_settext(w:preview['id'], a:lines)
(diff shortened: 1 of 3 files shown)


w0rp:

PR #5066, opened 2025-11-15
Add markdownlint fixer for markdown

yoan667:
This pull request adds support for [markdownlint](https://github.com/igorshubovych/markdownlint-cli) as a fixer for markdown templates in ALE.

Issue #5064, opened 2025-11-10
Add support for rumdl


robinbowes:
**Name:** rumdl
**URL:** https://github.com/rvben/rumdl

rumdl is a high-performance Markdown linter and formatter that helps ensure consistency and best practices in your Markdown files. Inspired by [ruff](https://github.com/astral-sh/ruff) 's approach to Python linting, rumdl brings similar speed and developer experience improvements to the Markdown ecosystem.

PR #4565, opened 2023-07-28
feat(markdown): add marksman lsp

pbnj:
Todo:

- [x] add docs
- [x] add tests

PR #5115, opened 2026-04-04
feat(markdown): add rumdl server and fixer

vEnhance:
Implements #5064. Some notes:

- I basically copied #4565 and #5066 for the linter and fixer respectively.
- I have configuration options for both the linter and the fixer separately, but I could make them a single option if desired. Although in practice I expect the options list to usually be blank.
- For finding the project root, I look for both `.rumdl.toml` or `rumdl.toml` and take the shorter one if both are found. Otherwise it looks for `pyproject.toml`, else `.git`.
  - I assume it's overkill for ALE to read `pyproject.toml` looking to see if it has a `tool.rumdl` section.
- If no `.git` or `rumdl.toml`/`.rumdl.toml` is found I just set project root as the folder of the buffer.
- Besides the two vader files, I've put the linter and fixer on my own computer in dotfiles (here's [linter](https://github.com/vEnhance/dotfiles/blob/main/vim/ale_linters/markdown/rumdl.vim) and [fixer](https://github.com/vEnhance/dotfiles/blob/main/vim/autoload/ale/fixers/rumdl.vim)) to make sure they pass sanity check.

I hope this is all correct, but it'd be good if someone competent looked over this given I'm first-time contributing and had assistance from an LLM.

### DIFF
--- a/ale_linters/markdown/rumdl.vim
+++ b/ale_linters/markdown/rumdl.vim
@@ -6,33 +6,34 @@ call ale#Set('markdown_rumdl_executable', 'rumdl')
 call ale#Set('markdown_rumdl_options', '')
 
 function! ale_linters#markdown#rumdl#GetProjectRoot(buffer) abort
-  let l:dotconfig = ale#path#FindNearestFile(a:buffer, '.rumdl.toml')
-  let l:config = ale#path#FindNearestFile(a:buffer, 'rumdl.toml')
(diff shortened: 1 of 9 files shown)


vEnhance:
so sorry i apparently only did run-tests on the vader files, i should read `ale-development` more carefully.

`./run-tests --fast -q` should pass now on a8212ef

hsanson (approved):
Thanks, looks good to merge.

PR #5118, opened 2026-04-08
Use ale#Pad for option padding across the codebase

matren33:
Standardize option padding by replacing verbose ternary checks with ale#Pad()

Replace patterns like:
```vim
(empty(x) ? '' : '' . l:options)
```

with:
```vim
ale#Pad(l:options)
```

Similar to `af5a38c697e0ba062f04347d5165d97d7d1894a3`


hsanson (approved):
Thanks for the contribution.

PR #5125, opened 2026-05-11
Feat: replace em dash with --

micahcc:
LLMs like to use em dash for no reason. I don't like non-ascii character in my text files, so I made a fixer to convert them from em dash to regular old `--` which latex and markdown treat as EM dashes. Added tests. Yes some of this code was was partially created by claude. 

Naming is the hard part, definitely open to renames. `no_emdash` `ascii_emdash` `replace_emdash` `em_dash` or `emdash`? etc.

It might be argued there are quite a few annoying unicodes that could be replaced, who really wants a 🤗 in their C++ code?

### DIFF
--- a/autoload/ale/fix/registry.vim
+++ b/autoload/ale/fix/registry.vim
@@ -207,6 +207,11 @@ let s:default_registry = {
 \       'suggested_filetypes': ['puppet'],
 \       'description': 'Run puppet-lint -f on a file.',
 \   },
+\   'replace_emdash': {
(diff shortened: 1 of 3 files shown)


Spixmaster:
This has no ending. Why limit to only one character? If one inputs garbage, it is the own fault. I dislike this very special use case.

w0rp (approved):
Nice to see someone submitting a simple fixer function.

Cheers! :beers:

w0rp:
> I dislike this very special use case.

You can not use the function.

PR #5130, opened 2026-05-19
Fix infinite autocompletion loop

letorbi:
I stumbled upon an infinite loop while using "pure" omnicompletion with ALE's omnifunction and Vim's native autocompletion:

```
set completeopt=menu,menuone,noselect,noinsert
set autocomplete
autocmd FileType * setlocal omnifunc=ale#completion#OmniFunc
```

Apparently `ale#completion#OmniFunc` does not ever return, if the language server has no completion capabilities. I was able to track the cause down to the while-loop, which waits for the server result. This PR breaks the infinite-loop by introducing a timeout check in the while-loop. The timeout is 3 seconds by default, but can be configured via `g:ale_completion_timeout`, which takes a float value that is interpreted as seconds. 

Additionally, I was able to fix the underlying cause by returning an empty result, if the language server does not have completion capabilities. This fixes the problem, but IMHO the timeout should still be included in order to prevent locks due to other reasons.

A test has also been added, but I am not sure whether it should really be included in the test suite, because it might lock the whole suite, if the timeout does not work as expected.

### DIFF
--- a/autoload/ale/completion.vim
+++ b/autoload/ale/completion.vim
@@ -952,15 +952,14 @@ function! ale#completion#OmniFunc(findstart, base) abort
         let l:timeout_start = reltime()
 
         while l:result is v:null && !complete_check()
+            sleep 2ms
+            let l:result = ale#completion#GetCompletionResult()
+
             if reltimefloat(reltime(l:timeout_start)) > l:timeout
                 " no-custom-checks
                 echoerr 'no result within timeout (' . l:timeout . 's)'
                 break
             endif
-
-            sleep 2ms
-
-            let l:result = ale#completion#GetCompletionResult()
         endwhile
 
         return l:result isnot v:null ? l:result : []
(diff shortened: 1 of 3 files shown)


letorbi:
Sorry for adding additional commits, despite the PR was not marked as a draft. I am done now and the PR is ready for review.

w0rp (approved):
This looks sensible to me. 

Cheers! 🍻

w0rp:
There's probably some way to mitigate the test suite locking up, but if it works it'll work.

letorbi:
Great, thanks for merging :)

letorbi:
> There's probably some way to mitigate the test suite locking up, but if it works it'll work.

The only way I could think of, would be to run the test in a separate test-suite with a timeout. Something like:

```
$ timeout 5 /usr/bin/vim '+Vader!*'
```

But maybe setting up a separate test-suite just to mitigate a problem, which is unlikely to happen, is a bit over-engineered here ;)

PR #5138, opened 2026-06-02
Fix Vint version detection

letorbi:
This PR fixes the broken Vint version detection due to Python warnings that interfere with ALE's version detection algorithm.

I discovered that my Vint based Vim Script linting did not work anymore and was able to track it down to a new Python version (3.14) that has been installed on my system recently. The reason is the `pkg_resources` API, which is still used by the last stable Vint version (0.3.21), but has been declared as obsolete for a longer time now.

This causes `vint --version` to return the following:

```
/usr/lib/python3.14/site-packages/vint/linting/cli.py:4: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
0.3.21
```

ALE's semver module parses this as `[3, 14, 0]`, possibly due to `python3.14` in the first line. Since ALE assumes that this version of Vint has stdin support, it calls Vint with the `--stdin-display-name` option, which is not yet available in Vint 0.3.21.

This PR changes the call to print Vint's version from just `vint --version` to `PYTHONWARNINGS=ignore vint --version`. This suppresses all Python warnings and thus allows ALE's semver module to detect the correct version.

### DIFF
--- a/ale_linters/vim/vint.vim
+++ b/ale_linters/vim/vint.vim
@@ -58,7 +58,7 @@ call ale#linter#Define('vim', {
 \   'command': {buffer -> ale#semver#RunWithVersionCheck(
 \       buffer,
 \       ale#Var(buffer, 'vim_vint_executable'),
-\       '%e --version',
+\       'PYTHONWARNINGS=ignore %e --version',
 \       function('ale_linters#vim#vint#GetCommand'),
 \   )},
 \   'callback': 'ale_linters#vim#vint#Handle',
--- a/test/linter/test_vint.vader
+++ b/test/linter/test_vint.vader
@@ -10,7 +10,7 @@ After:
 
 Execute(The default command should be correct):
   AssertLinter 'vint', [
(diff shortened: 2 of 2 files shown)


hsanson (approved):
Thanks,

letorbi:
Thanks for merging, @hsanson.

I wonder whether it might be a good idea, if ALE's version detection would ignore lines with `warning` or `error` (case-insensitive) in general. This could avoid similar problems with other tools in the future, but might also have the side effect that version detection won't work, if the version number is part of an ignored line. On the other hand, it seems quite unlikely to me, that the version number and  `warning` or `error` appear in the same line.

I could create a PR for the line filtering, but I'd like to have your opinion on this (@w0rp's would be great as well), before I start coding.

PR #5146, opened 2026-07-13
chktex: always use the right format and grab the error code

### LINKED ISSUE #3856: Missing warning number for chktex
The warning for chktex is missing the warning number, e.g., "Warning 8," which is useful for suppressing warnings. I think that a fix may go in this code:

https://github.com/dense-analysis/ale/blob/f0887d3e6178482255f11aa378124aef3699245f/ale_linters/tex/chktex.vim#L29-L47

Steap:
Using -v0 works out of the box, but chktex grabs the "v0 format" from
its configuration file:

    OutFormat
    {
        # -v0; silent mode
        %f%b%l%b%c%b%n%b%m!n
        # -v1; normal mode
        ...
    }

Should the user modify the definition of v0, then the linter would no
longer work. We therefore use "--format" to force the output format.

Also grab the error code so it is correctly displayed.

Closes: #3856

### DIFF
--- a/ale_linters/tex/chktex.vim
+++ b/ale_linters/tex/chktex.vim
@@ -12,7 +12,9 @@ function! ale_linters#tex#chktex#GetCommand(buffer, version) abort
     let l:options = ''
 
     " Avoid bug when used without -p (last warning has gibberish for a filename)
-    let l:options .= ' -v0 -p stdin -q'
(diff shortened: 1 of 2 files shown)


hsanson (approved):
Thanks, forcing v0 makes sense since ALE expects it.

PR #5147, opened 2026-07-18
Handle nil result in pull diagnostic handler

pjcj:
## Summary

The `textDocument/diagnostic` handler assumed a non-nil `result` whenever no error was returned. A stale or cancelled pull diagnostic request can legitimately be answered with a nil result and no error under the LSP specification, so the handler indexed a nil value and threw `attempt to index local 'result' (a nil value)` inside a scheduled callback.

## Reproduction

- With a language server that supports pull diagnostics, open a file and let a request go stale (for example by reloading the buffer while a request is in flight). The server answers the now-stale request with a nil result and no error, and the handler crashes.

## Changes

- Guard the pull diagnostic handler so a nil result is ignored rather than indexed, matching the existing `err` check.
- Add a test covering a nil result with no error.

### DIFF
--- a/lua/ale/lsp.lua
+++ b/lua/ale/lsp.lua
@@ -77,7 +77,7 @@ module.start = function(config)
         end,
         -- Handle pull model diagnostic data.
         ["textDocument/diagnostic"] = function(err, result, request, _)
-            if err == nil then
+            if err == nil and result ~= nil then
                 local diagnostics
 
                 if result.kind == "unchanged" then
--- a/test/lua/ale_lsp_start_spec.lua
+++ b/test/lua/ale_lsp_start_spec.lua
@@ -444,4 +444,31 @@ describe("ale.lsp.start", function()
             },
         }, vim_fn_calls)
     end)
+
+    it("should ignore pull model diagnostics with a nil result", function()
+        lsp.start({name = "server:/code"})
+
+        eq(1, #start_calls)
+
+        local handlers = start_calls[1][1].handlers
+
+        eq("function", type(handlers["textDocument/diagnostic"]))
+
+        -- A stale request can be answered with a nil result and no error,
+        -- which is valid per the LSP specification. The handler should do
+        -- nothing rather than index the nil result.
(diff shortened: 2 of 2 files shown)


rymdbar:
For what it's worth, this looks good to me @pjcj. Thanks!

I'm no code owner and unable to approve pull-requests here, but any interested party can post their thoughts and questions, right?

The commit (and source code) comment claims a nil result (response?) for `textDocument/diagnostic` is valid per spec. I have no reason to distrust that, and it even seems very plausible. Yet the spec isn't always trivial to grasp. Does this follow from anything under [Pull Diagnostics][pd]? Could you perhaps spare a couple of minutes to explain how you read it?

Keeping things in the abstract shows how you are a great programmer, and that's defintely the right way to do things. Yet I can't help to wonder whether you happen to know any implementations in particular which ALE emits this broken behaviour with?

[pd]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_pullDiagnostics

pjcj:
Hello @rymdbar - thanks for checking this over!

Taking your second question first, I started looking into this because I was noticing the behaviour under the svelte language server.  I believe the exact code causing the problem is https://github.com/sveltejs/language-tools/blob/fbb12913ceab9dd3d33de531e0fbdbe432bc8e6b/packages/language-server/src/server.ts#L250-L253

Now, as to whether this is strictly standards compliant, it's possible that I may have overstated it a little in my wording.  I think it is allowed based on https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#responseMessage but the svelte language server should likely be handling this in another way, perhaps by returning ServerCancelled.

But regardless of the compliance it's probably a good idea to be defensive anyway.  This is the approach neovim seems to take: https://github.com/neovim/neovim/blob/v0.12.4/runtime/lua/vim/lsp/diagnostic.lua#L286-L288 and there's effectively no cost.

hsanson (approved):
Thanks @rymdbar for the review and @pjcj for the contribution.

PR #5150, opened 2026-07-22
Add ale_python_pyright_options variable

rymdbar:
Add ale_python_pyright_options variable, following the ALE typical configuration pattern.

Commit also replaces link to lsp config doc from its source code to the published rendered page version.


hsanson (approved):
Thanks for the contribution.

PR #5153, opened 2026-07-26
Add option to use human readable ruff errors

alexfikl:
`ruff` started adding some checks to push towards comments of the form `# ruff: ignore[some-rule-name]` ([RUF106](https://docs.astral.sh/ruff/rules/rule-codes-in-suppression-comments/)) instead of `# noqa: ignore[F401]` ([RUF105](https://docs.astral.sh/ruff/rules/noqa-comments/)). 

This adds an option `python_ruff_human_readable` to allow ale to show the human readable version if the project uses that instead of the error codes. With this enabled, the diagnostic looks something like
```
[ruff] `collections.abc.Sequence` imported but unused (unused-import) [Warning]
```


hsanson (approved):
Thanks for the contribution.

PR #5154, opened 2026-07-27
Only run actionlint on github workflow files

aminvakil:
This PR adds a test to show that current code runs `actionlint` on `.github/dependabot.yml` and it fails (as it's not a workflow).

And it changes `ale`, so it only looks for files in `.github/workflows` instead of `.github`.


hsanson (approved):
Thanks, looks good.

Issue #4519, opened 2023-05-09
Add support for OpenSCAD language server Leathong/openscad-LSP


pozix604:
OpenSCAD is 3D modelling software with scripting as its primary interface.

The most featureful LSP for for OpenSCAD is https://github.com/Leathong/openscad-LSP. Please add support for it.

Usage instructions very simple: https://github.com/Leathong/openscad-LSP#usage

nospam2998:
I configured my local ale to make use of Leathong's lsp recently. Adding ale support for language servers truly is trivial.

As this ticket has been filed, I understand the task of adding support has already been claimed, and wont step on any toes by posting a pull request.

```vim
function! LeathongGetProjectRoot(buffer) abort
    let l:potential_roots = [
    \   bufname(a:buffer),
    \ ]

    for l:root in l:potential_roots
        let l:project_root = ale#path#FindNearestFile(
        \   a:buffer,
        \   l:root,
        \ )

        if empty(l:project_root)
            let l:project_root = ale#path#FindNearestDirectory(
            \   a:buffer,
            \   l:root,
            \ )
        endif

        if !empty(l:project_root)
            " dir:p expands to /full/path/to/dir/ whereas
            " file:p expands to /full/path/to/file (no trailing slash)
            " Appending '/' ensures that :h:h removes the path's last segment
            " regardless of whether it is a directory or not.
            return fnamemodify(l:project_root . '/', ':p:h:h')
        endif
    endfor

    return ''
endfunction

call ale#linter#Define('openscad', {
\  'name': 'leathong',
\  'executable': 'openscad-lsp',
\  'lsp': 'stdio',
\  'command': '%e --stdio',
\  'project_root': function('LeathongGetProjectRoot'),
\})
```

Recognizing owner of this ticket as someone quite extroverted, I am leaving the above snippet here for others to use in their `.vimrc`. In case this is another one of those no-action-intended-magic-will-solve-it tickets.

PR #5156, opened 2026-08-02
feat: openscad-lsp support

jniklasson:
Just started out learning openscad and I found that openscad-lsp was missing. 

Based on the suggestion in #4519.

On the current version of ALE, autocompletion does not work due to ALE assuming no auto completion is supported if a server is returning `completionProvider: {}`. I've got a separate PR ready that solves this.


hsanson (approved):
Thanks for the work.

PR #5157, opened 2026-08-02
fix(lsp): check completionProvider key existence, not emptiness

jniklasson:
completionProvider: {} is valid per the LSP spec (means completion enabled with default options). The old !empty() check incorrectly rejected empty objects, causing ALE to silently disable completion for servers that send completionProvider: {}.

This change fixes auto completion for openscad-lsp added in #5156.


hsanson (approved):
Thanks for the correction.

PR #5159, opened 2026-08-04
Add support for fatou

jolars:
Add a linter that runs the fatou language server (`fatou lsp`) for Julia diagnostics, and a fixer that formats Julia with `fatou format`.

I've added several tests in `test/fixers/test_fatou_fixer_callback.vader` and `test/linter/test_julia_fatou.vader`. I ran the ALE suite in a local docker and it checks out.


hsanson (approved):
Thanks for the contribution.

PR #5160, opened 2026-08-06
Support the LSP built into Typescript 7 (typescript-go)

### LINKED ISSUE #5128: Add support for tsgo
**Name:** tsgo
**URL:** https://github.com/microsoft/typescript-go

The coming version of TypeScript is a rewrite into Go with a new language server. We should add support for it now, so we can be ready for the next version of TypeScript. Very interesting because `tsserver` predates LSP and now we're coming full circle.

_Suggested on Reddit here: https://www.reddit.com/r/neovim/comments/1sys7t0/comment/oj2zfkk/_

Jiggins:
Typescript moved to go for the 7.0 release. They also ditched `tsserver`
in favour of `tsc --lsp`. This commit implements an ALE LSP definition
for `tsc --lsp`.

From testing, `tsc --lsp` doesn't seem to be as good as the old
`tsserver` but hopefully it'll get better.

Fixes: #5128

### DIFF
--- a/ale_linters/typescript/tsc.vim
+++ b/ale_linters/typescript/tsc.vim
@@ -1,0 +1,10 @@
+call ale#Set('typescript_tsc_config_path', '')
+call ale#Set('typescript_tsc_use_global', get(g:, 'ale_use_global_executables', 0))
+
+call ale#linter#Define('typescript', {
+\   'name': 'tsc',
+\   'lsp': 'stdio',
+\   'executable': function('ale#handlers#typescript#GetExecutable'),
+\   'command': '%e --lsp --stdio',
+\   'project_root': function('ale#handlers#typescript#GetProjectRoot'),
+\})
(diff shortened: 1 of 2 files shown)


hsanson (changes requested):
Before merging this needs to include tests and documentation updates.  See this PR as reference:

- https://github.com/dense-analysis/ale/pull/5159/changes

In specific:

- New test `test/linter/test_typescript_tsc.vader`
- Entry in supported-tools.md
- Entry in txt/ale.txt
- Entry in doc/ale-supported-languages-and-tools.txt
- Help entry in doc/ale-typescript.txt

Important: entries in documentation must be added in alphabetical order to avoid linter errors.

wutje:
Note that this does not really work for `typescriptreact`; it does linting/checking, but it for .ts files and gives error when reaching a react component.
I duplicated the ale#linter#Define and changed to 'typescript' to 'tsx' to make it work for typescriptreact files.

It works, but I have know knowledge on ALE aliasing or how LSP work so perhaps this is not the best solution.

PR #5161, opened 2026-08-07
fix: linter fortitude: renamed location key from "row" to "line"

gomfol12:
Hi, 

fortitude has renamed the key `row` to `line` in the `location` and `end_location` field. They changed it in this commit PlasmaFAIR/fortitude@0eda11975096517b16bf30bf683011f83a24e55e.

This PR fixes this.


hsanson (changes requested):
This looks like a very recent breaking change on fortitude side. I recommend we keep support for the previous 'row' as removing it may break this linter for users that are stuck with previous versions of the linter.  Keep the tests using 'row' intact and add a new set of test for 'line'.

gomfol12:
Yes makes sense. I now test for `line` and fallback to `row`.

